PR #775, opened 2025-12-17
fix: Async markdown import now works in Tauri context

malibio:
## Summary
- Fixed async markdown import (fire-and-forget mode) that was silently failing in Tauri context
- Changed `tokio::spawn` to `Handle::try_current().spawn()` to work in both standalone tokio and Tauri async runtimes
- Added fallback to synchronous import if no runtime available

## Problem
Documents imported via MCP in the Tauri desktop app would only have the root node created - children were never inserted because the background `tokio::spawn` task was being dropped.

## Root Cause
`tokio::spawn` doesn't work reliably when called from within Tauri's async runtime (`tauri::async_runtime`). The spawned tasks were being dropped before completion.

## Solution
Use `tokio::runtime::Handle::try_current().spawn()` instead of bare `tokio::spawn`. This explicitly gets the current runtime handle and spawns on it, which works in both contexts.

## Test plan
- [x] Unit tests pass (44 markdown tests)
- [x] Integration tests pass (updated 3 tests to use `sync_import: true` for node count verification)
- [x] Manual testing: Imported 163-node document via MCP in Tauri - all children created in background
- [x] Semantic search finds newly imported documents after embedding generation

### DIFF
--- a/packages/core/src/mcp/handlers/markdown.rs
+++ b/packages/core/src/mcp/handlers/markdown.rs
@@ -789,8 +789,8 @@ where
                 }
             } else {
                 // Default: Fire-and-forget async for fast MCP response
-                // Use tokio::task::spawn_local or Handle::current().spawn to ensure
-                // we're spawning on the current runtime
+                // Use Handle::try_current().spawn() to ensure we're spawning on
+                // the current runtime (works in both standalone tokio and Tauri)
                 let node_service = Arc::clone(node_service);
                 let root_id_for_log = root_id.clone();
 


malibio:
## Review Feedback Addressed

### Changes Made
- ✅ Fixed comment on line 792-793 to accurately reflect `Handle::try_current().spawn()` instead of mentioning `spawn_local`

### Summary
- **Addressed**: 1 recommendation (nitpick)
- **Skipped**: 0 recommendations
- **Commits**: 1

All other findings were positive - no changes required.

---

**Re-Review Decision**: NO RE-REVIEW NEEDED

This was a trivial comment fix with no logic changes. The PR remains ready for merge.

PR #809, opened 2026-01-20
Collection path resolution doesn't create hierarchy between collections

### LINKED ISSUE #808: Collection path resolution doesn't create hierarchy between collections
## Bug Description

When resolving a collection path like `architecture:core`, the system creates both collections but **does not establish the parent-child hierarchy** between them. The "core" collection should be a child of "architecture", but instead they are created as independent, flat collections.

## Expected Behavior

For path `architecture:core:components`:
1. Create collection "architecture" (if not exists)
2. Create collection "core" (if not exists)  
3. Create collection "components" (if not exists)
4. Create `member_of` relationship: core → architecture
5. Create `member_of` relationship: components → core

This would allow browsing the collection tree and seeing sub-collections.

## Actual Behavior

For path `architecture:core:components`:
1. Create collection "architecture" ✅
2. Create collection "core" ✅
3. Create collection "components" ✅
4. **No hierarchy relationships created** ❌

Collections are created as flat, independent nodes with globally unique names.

## Root Cause

In `packages/core/src/services/collection_service.rs`, the `resolve_path()` function has comments explicitly stating collections are flat:

```rust
// Collections are flat - paths are just naming conventions for navigation.
// Each segment is an independent collection with a globally unique name.
// No parent-child hierarchy between collections
```

And `create_collection()` has an unused `_parent_id` parameter that was intentionally ignored.

## Architecture Specification

The architecture document (`docs/architecture/components/collections-system.md`) clearly specifies hierarchical collections:

> **Collection hierarchy** - Uses existing `has_child` pattern between collection nodes.

And the usage example shows:
```javascript
// "Berlin" now has TWO parents: vacation AND offices
```

The implementation diverged from the specification during the initial implementation in PR #758.

## Investigation Notes

- Initial implementation: commit `60b4109c` (PR #758)
- The flat design was introduced from the start, not during the Universal Edge Table migration (#793)
- The architecture doc mentions using `has_child` edges, but based on the multi-parent DAG requirement, `member_of` edges would be more appropriate (collections can have multiple parents)

## Proposed Fix

1. Update `resolve_path()` in `collection_service.rs` to:
   - Track the previous collection in the path
   - After creating/finding each collection, create a `member_of` edge from current → previous (child → parent)
   
2. The `member_of` edge type is already defined in the universal `relationship` table and supports the many-to-many relationship needed for DAG structure.

3. Update tests to verify hierarchy creation.

## Impact

Without this fix:
- Collection browser cannot show sub-collections under parent collections
- The path syntax `architecture:core` is meaningless - both collections are at the same level
- No way to navigate the collection hierarchy in the UI

## Related

- PR #758 - Original Collections System implementation
- Issue #756 - Collections System backend
- Issue #757 - Collections System UI (closed #807)
- PR #793 - Universal Edge Table consolidation
- `docs/architecture/components/collections-system.md` - Architecture specification

malibio:
Closes #808

### DIFF
--- a/docs/architecture/components/collections-system.md
+++ b/docs/architecture/components/collections-system.md
@@ -169,7 +169,7 @@ DEFINE INDEX member_of_in ON member_of FIELDS in;
 DEFINE INDEX member_of_out ON member_of FIELDS out;
 ```
 
-**Collection hierarchy** - Uses existing `has_child` pattern between collection nodes.
+**Collection hierarchy** - Uses `member_of` edges between collection nodes (child `member_of` parent), forming a DAG structure that supports multi-parent relationships.
 
 ### Cycle Validation
 
@@ -177,7 +177,7 @@ Collections form a DAG - cycles are not allowed. When setting a parent:
 
 ```sql
 -- Check if new_parent is a descendant of collection (would create cycle)
-SELECT * FROM node:$new_parent<-has_child*<-node WHERE id = $collection_id;
+SELECT * FROM node:$new_parent<-member_of*<-node WHERE id = $collection_id;
 -- If result is not empty → reject
 ```
 
@@ -270,7 +270,7 @@ SELECT * FROM node WHERE ->member_of->node = $collection_id;
 -- Get all members recursively (collection + descendants)
 SELECT * FROM node WHERE ->member_of->node IN (
   SELECT id FROM node WHERE node_type = 'collection'
-  AND id = $collection_id OR <-has_child*<-node CONTAINS $collection_id
+  AND id = $collection_id OR <-member_of*<-node CONTAINS $collection_id
 );
 
 -- Get all collections a node belongs to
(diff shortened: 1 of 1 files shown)


malibio:
## Review Feedback Addressed

### Summary
| Status | Count |
|--------|-------|
| ✅ Addressed | 1 |
| ⏭️ Skipped | 1 |
| 📝 Commits | 1 |

### Addressed
- **🟢 Documentation Consistency** - Updated `docs/architecture/components/collections-system.md` to reflect that collection hierarchy uses `member_of` edges instead of `has_child`. Also updated SQL query examples for consistency.

### Skipped (with justification)
- **🟢 Nitpick: Comment Reference** - The existing comment at line 442 ("This is idempotent...") is appropriately abstracted. Adding implementation details about *how* `add_to_collection` achieves idempotency would couple the comment to internal implementation, which could become stale if that method changes.

### Commit
- `0841776` - docs: Update collections-system.md to reflect member_of edge implementation

---

*Addressed by Implementation Agent following Pragmatic Quality framework.*

Issue #776, opened 2025-12-18
Optimize embedding processor performance - reuse LlamaContext


malibio:
## Overview

The embedding processor currently creates a new `LlamaContext` for each embedding call, causing significant CPU overhead and underutilizing the GPU. During bulk imports (172 docs), CPU usage spikes to 1200%+ while GPU utilization remains low (~10%).

## Problem Statement

**Current Behavior:**
- Each embedding call creates a new `LlamaContext` via `model.new_context()`
- Metal kernel compilation occurs for every context creation
- Context is immediately freed after one embedding
- Results in 1200% CPU during bulk operations with ~10% GPU utilization

**Evidence:**
- Log shows repeated `ggml_metal_init: loaded kernel_*` messages (context recreation)
- Log shows `ggml_metal_free: deallocating` after each embedding
- Activity Monitor shows high CPU, low GPU during embedding processing
- 172 document import took ~54 minutes with heavy CPU load

**Code Location:** `packages/nlp-engine/src/embedding.rs:207-253`

```rust
// Current: New context created for EACH embedding
fn generate_embedding_llama(&self, text: &str) -> Result<Vec<f32>> {
    // ... 
    let mut ctx = model.new_context(backend, ctx_params)...; // Created every call
    // ... use ctx ...
    // ctx dropped here, Metal resources freed
}
```

## Proposed Solution

Reuse the `LlamaContext` across multiple embedding calls instead of recreating it each time:

1. Store context as `Option<LlamaContext>` in `EmbeddingService`
2. Initialize context lazily on first embedding call
3. Reuse context for subsequent calls
4. Only recreate if context parameters change

**Alternative Approach:**
Implement batch embedding that processes multiple texts in a single context session.

## Acceptance Criteria

- [ ] LlamaContext is reused across embedding calls (not recreated each time)
- [ ] GPU utilization increases during embedding processing
- [ ] CPU usage decreases during bulk embedding operations
- [ ] Embedding throughput improves (measure docs/second)
- [ ] No regression in embedding quality or correctness
- [ ] Tests pass (`bun run test:all`)
- [ ] Code passes `bun run quality:fix`

### Reference Files
- **embedding.rs**: `packages/nlp-engine/src/embedding.rs` - Main embedding service
- **embedding_processor.rs**: `packages/core/src/services/embedding_processor.rs` - Background processor
- **embedding_service.rs**: `packages/core/src/services/embedding_service.rs` - Node embedding service

### Key Changes

[...]

PR #885, opened 2026-02-06
Fix: SIGABRT crash on app exit in production builds

### LINKED ISSUE #884: Fix: SIGABRT crash on app exit in production builds
## Overview

NodeSpace crashes with `SIGABRT` (abort trap 6) when closing the app in **production/release builds only**. The crash occurs during process shutdown in `__cxa_finalize_ranges` (C++ static destructor teardown), triggered by a panic converting to `abort()` due to the release profile configuration. Dev mode (`bun run dev:tauri`) is unaffected.

## Crash Analysis

**Crash signature:**
```
Exception Type:    EXC_CRASH (SIGABRT)
Termination Reason: Namespace SIGNAL, Code 6, Abort trap: 6

Thread 0 Crashed:: main Dispatch queue: com.apple.main-thread
  libsystem_c.dylib  __cxa_finalize_ranges
  libsystem_c.dylib  exit
  AppKit              -[NSApplication terminate:]
```

**Root cause chain:**
1. User closes app → macOS calls `[NSApplication terminate:]` → `exit()`
2. `exit()` runs `__cxa_finalize_ranges` (static destructors)
3. A Rust Drop impl or cleanup handler panics during this phase
4. `panic = "abort"` in release profile converts the panic to `SIGABRT`
5. Process crashes instead of exiting cleanly

**Why dev mode is unaffected:** Debug profile uses `panic = "unwind"` (default), allowing panics to be caught and handled. Release profile explicitly sets `panic = "abort"`.

## Self-Contained Scope

Three fixes needed, in priority order:

### Fix 1: Remove `panic = "abort"` from release profile
- **File:** `Cargo.toml:36`
- Change `panic = "abort"` to `panic = "unwind"` (or remove the line entirely)
- This alone likely stops the crash, since panics during shutdown will unwind instead of aborting
- Tradeoff: Slightly larger binary size (unwind tables included). This is acceptable.

### Fix 2: Add graceful shutdown signaling to background tasks
- **File:** `packages/desktop-app/src-tauri/src/lib.rs`
- The MCP server task (line 122) and domain event forwarder (line 53) are spawned with `tauri::async_runtime::spawn` but have **no shutdown mechanism**
- When the Tokio runtime drops (at `main.rs:18` after `block_on` returns), these tasks are killed mid-operation
- **Solution:** Use a `tokio_util::sync::CancellationToken` (or `tokio::sync::broadcast`) shared between the main run loop and spawned tasks
  - Create the token before spawning tasks
  - Pass a clone to each spawned task; they should `select!` on the token
  - Cancel the token in `RunEvent::ExitRequested` or `RunEvent::Exit`
  - Optionally add a brief timeout to allow tasks to finish cleanup

### Fix 3: Explicitly clear global LLAMA_BACKEND before process exit
- **File:** `packages/nlp-engine/src/embedding.rs:43`
- The static `LLAMA_BACKEND: OnceLock<LlamaBackend>` holds Metal/GPU resources
- Current `release_gpu_resources()` (lib.rs:356) drops the `LlamaState` (context + model) but the global backend itself persists
- During `__cxa_finalize_ranges`, the backend is destroyed by the C runtime, which can trigger crashes if Metal resources are in a bad state
- **Solution:** Add an explicit cleanup function that uses `unsafe` to take the value out of the OnceLock (or use a `Mutex<Option<LlamaBackend>>` instead of OnceLock so it can be cleared safely)
- Call this cleanup in `release_gpu_resources()` after releasing the context

## Acceptance Criteria

- [ ] `panic = "abort"` removed or changed to `panic = "unwind"` in `Cargo.toml` release profile
- [ ] Background tasks (MCP server, domain event forwarder) receive a shutdown signal before the Tokio runtime drops
- [ ] Global `LLAMA_BACKEND` is explicitly cleaned up before process exit (not left to static destruction)
- [ ] Production build (`cargo tauri build`) exits cleanly without SIGABRT
- [ ] Dev mode (`bun run dev:tauri`) continues to work correctly
- [ ] No new test failures introduced

### Reference Files
- **Cargo.toml** (`Cargo.toml:35-40`) - Release profile with `panic = "abort"`
- **main.rs** (`packages/desktop-app/src-tauri/src/main.rs:8-18`) - Tokio runtime setup, no graceful shutdown
- **lib.rs** (`packages/desktop-app/src-tauri/src/lib.rs:53-75`) - Domain event forwarder spawn
- **lib.rs** (`packages/desktop-app/src-tauri/src/lib.rs:122-143`) - MCP server spawn
- **lib.rs** (`packages/desktop-app/src-tauri/src/lib.rs:310-368`) - RunEvent handlers and `release_gpu_resources()`
- **embedding.rs** (`packages/nlp-engine/src/embedding.rs:43`) - Global `LLAMA_BACKEND` OnceLock
- **embedding.rs** (`packages/nlp-engine/src/embedding.rs:574-578`) - `release_gpu_context()` drops LlamaState but not backend

### Testing Approach
- Build release binary: `cargo tauri build`
- Launch production app, use briefly, then quit via Cmd+Q and window close button
- Verify no crash report is generated by macOS
- Check Console.app for any crash logs under the `com.nodespace.desktop` identifier
- Run `bun run test:all` and `bun run rust:test` to verify no regressions

## Resources & References
- [Tauri issue #9198](https://github.com/tauri-apps/tauri/issues/9198) - `RunEvent::ExitRequested` unreliable on macOS
- [tokio_util::sync::CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html)
- macOS crash report from user testing (2026-02-06)

malibio:
Closes #884

### DIFF
--- a/packages/desktop-app/src-tauri/src/commands/db.rs
+++ b/packages/desktop-app/src-tauri/src/commands/db.rs
@@ -173,7 +173,7 @@ async fn init_services(app: &AppHandle, db_path: PathBuf) -> Result<(), String> 
 
     // Initialize MCP server now that NodeService is available
     // MCP will use the same NodeService as Tauri commands
-    if let Err(e) = crate::initialize_mcp_server(app.clone(), shutdown_token.0.child_token()) {
+    if let Err(e) = crate::initialize_mcp_server(app.clone(), shutdown_token.child_token()) {
         tracing::error!("❌ Failed to initialize MCP server: {}", e);
         // Don't fail database init if MCP fails - MCP is optional
     }
@@ -185,7 +185,7 @@ async fn init_services(app: &AppHandle, db_path: PathBuf) -> Result<(), String> 
         app.clone(),
         node_service_arc.clone(),
         client_id,
-        shutdown_token.0.child_token(),
+        shutdown_token.child_token(),
     ) {
         tracing::error!("❌ Failed to initialize domain event forwarder: {}", e);
         // Don't fail database init if event forwarding fails - it's not critical
--- a/packages/desktop-app/src-tauri/src/lib.rs
+++ b/packages/desktop-app/src-tauri/src/lib.rs
@@ -169,14 +169,32 @@ pub fn initialize_mcp_server(
 /// shutdown is triggered). When cancelled, all background tasks (MCP server,
 /// domain event forwarder) exit their loops before the Tokio runtime drops.
 #[derive(Clone)]
-pub struct ShutdownToken(pub tokio_util::sync::CancellationToken);
+pub struct ShutdownToken(tokio_util::sync::CancellationToken);
+
+impl ShutdownToken {
+    fn new() -> Self {
+        Self(tokio_util::sync::CancellationToken::new())
+    }
+
+    /// Create a child token for a background task.
+    /// Cancelling the parent automatically cancels all children.
+    pub fn child_token(&self) -> tokio_util::sync::CancellationToken {
+        self.0.child_token()
+    }
+
+    /// Signal all background tasks to shut down.
+    /// Idempotent - safe to call multiple times.
+    pub fn cancel(&self) {
+        self.0.cancel();
+    }
+}
 
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
     use tauri::{menu::*, Emitter, Manager, RunEvent};
 
     // Create shutdown token for coordinating graceful background task termination
-    let shutdown_token = ShutdownToken(tokio_util::sync::CancellationToken::new());
+    let shutdown_token = ShutdownToken::new();
     let shutdown_token_for_setup = shutdown_token.clone();
 
     let app = tauri::Builder::default()
@@ -366,7 +384,10 @@ pub fn run() {
                     "Window '{}' close requested, signaling shutdown and releasing GPU context...",
                     label
                 );
-                shutdown_token_for_events.0.cancel();
+                shutdown_token_for_events.cancel();
+                // Brief pause to let background tasks exit their loops before
+                // releasing GPU resources they may still reference
+                std::thread::sleep(std::time::Duration::from_millis(50));
                 release_gpu_resources(app_handle);
             }
             RunEvent::ExitRequested { code, .. } => {
@@ -375,14 +396,14 @@ pub fn run() {
                     "App exit requested (code: {:?}), performing cleanup...",
                     code
                 );
-                shutdown_token_for_events.0.cancel();
+                shutdown_token_for_events.cancel();
                 release_gpu_resources(app_handle);
                 tracing::info!("Cleanup complete, exiting...");
             }
             RunEvent::Exit => {
                 // Final exit - ensure shutdown signal is sent (idempotent)
                 tracing::info!("App exiting, ensuring shutdown signal sent...");
-                shutdown_token_for_events.0.cancel();
+                shutdown_token_for_events.cancel();
             }
             _ => {}
         }
--- a/packages/nlp-engine/src/embedding.rs
+++ b/packages/nlp-engine/src/embedding.rs
@@ -125,8 +125,7 @@ impl std::ops::Deref for BackendGuard {
 #[cfg(feature = "embedding-service")]
 pub fn release_llama_backend() {
     let mut guard = LLAMA_BACKEND.lock().unwrap_or_else(|p| p.into_inner());
-    if let Some(backend) = guard.take() {
-        drop(backend);
+    if guard.take().is_some() {
         tracing::info!("Global LLAMA_BACKEND released, Metal resources freed");
     }
 }


malibio:
## Pragmatic Code Review: PR #885 - Fix SIGABRT crash on app exit

### Overall Assessment

**Verdict: Approve.** This PR is a clear net positive. It addresses a real production crash (SIGABRT on macOS app exit) through three well-reasoned, complementary fixes that align precisely with the root cause analysis in issue #884. The implementation is clean, well-documented, and uses established patterns (RAII, `CancellationToken`, `tokio::select!`). No critical issues found.

---

### Requirements Validation (Issue #884)

|---|---|
| `panic = "abort"` changed to `panic = "unwind"` in release profile | **Met** (`Cargo.toml:36`) |
| Background tasks receive shutdown signal before Tokio runtime drops | **Met** (CancellationToken plumbed to MCP server and domain event forwarder) |
| Global `LLAMA_BACKEND` explicitly cleaned up before process exit | **Met** (`Mutex<Option<LlamaBackend>>` with `release_llama_backend()`) |
| Production build exits cleanly without SIGABRT | Requires manual verification |
| Dev mode continues to work correctly | Requires manual verification |
| No new test failures introduced | Requires CI verification |

---

### Findings

#### No Critical Issues

No blocking issues were identified.

#### Suggested Improvements

**1. [Improvement] `BackendGuard` holds Mutex lock during model loading and context creation**
- **Files:** `packages/nlp-engine/src/embedding.rs:61-97`, `embedding.rs:311-322`
- The `BackendGuard` RAII type holds the `MutexGuard` for the entire duration of model loading (`LlamaModel::load_from_file`) and context creation (`model.new_context`). Model loading involves disk I/O and GPU shader compilation -- potentially seconds of wall time. During this period, any concurrent call to `get_or_init_backend()` (e.g., from another thread) will block.
- The previous `OnceLock` approach allowed lock-free reads after initialization. Now, every embedding operation that needs a context recreation acquires the global backend lock.
- **Practical impact for this project:** Low. The `EmbeddingService` already serializes access through its own `Mutex<Option<LlamaState>>`, so contention on the backend lock is unlikely in practice. The model is loaded once at startup and the context is reused (Issue #776). This is an acceptable tradeoff for explicit shutdown cleanup.
- **Principle:** Be aware that if the architecture ever moves to multiple independent `EmbeddingService` instances (e.g., different models), this global lock could become a bottleneck. The current design is fine for the single-model architecture.

**2. [Improvement] No timeout on graceful shutdown**
- **File:** `packages/desktop-app/src-tauri/src/lib.rs:356-389`
- The shutdown sequence cancels the token and then immediately calls `release_gpu_resources`. The background tasks are _signaled_ to stop, but there is no `await` or timeout to confirm they have actually stopped before GPU resources are released. If a background task is in the middle of a database operation or network call when the token is cancelled, the `tokio::select!` branch may not fire instantly.
- **Practical impact:** This is mitigated by the fact that `release_gpu_resources` only touches the `EmbeddingService` state (a separate Mutex from the backend tasks), so there is no direct data race. The tasks will be cleaned up when the Tokio runtime drops. However, the issue description (Fix 2) mentioned "Optionally add a brief timeout to allow tasks to finish cleanup."
- **Recommendation:** Consider adding a small sleep (e.g., 50-100ms) after `cancel()` and before `release_gpu_resources()` in the `CloseRequested` handler to give tasks a chance to exit their loops cleanly. This is a nice-to-have, not a blocker.

**3. [Improvement] `ShutdownToken` inner field is `pub`**
- **File:** `packages/desktop-app/src-tauri/src/lib.rs:172`
- `pub struct ShutdownToken(pub tokio_util::sync::CancellationToken)` -- the inner `CancellationToken` is public. While this works, it means any code with access to Tauri state could cancel the token or create child tokens without going through a controlled API.
- **Recommendation:** Consider making the inner field private and providing specific methods (e.g., `child_token()`, `cancel()`). This is minor given the pre-release state of the project.

#### Nitpicks

**Nit:** `packages/nlp-engine/src/embedding.rs:129` -- The explicit `drop(backend)` is redundant since `guard.take()` already moves the value out, and it would be dropped at the end of the `if let` block anyway. Not harmful, but slightly misleading since it suggests `drop` does something special here.

**Nit:** `packages/desktop-app/src-tauri/src/lib.rs:152-158` -- Two TODO comments were removed from the MCP server error/panic handlers. While the issue mentions this is acceptable cleanup, these were reasonable future considerations (emitting Tauri events on MCP failure, automatic restart). If these are intentionally descoped, the removal is fine. Otherwise, they could be tracked as separate issues.

---

### Architecture Assessment

The three-layer defense approach is well-designed:

1. **`panic = "unwind"`** -- The broadest fix. Prevents any panic during shutdown from immediately aborting. This alone likely resolves the crash, but is the least surgical solution.

2. **`CancellationToken`** -- The most architecturally sound fix. Uses `tokio_util::sync::CancellationToken` with child tokens, which is the idiomatic Tokio pattern for cooperative task cancellation. The token is managed via Tauri state (`ShutdownToken`), making it accessible to both the setup phase and the event handler. The use of `child_token()` for each background task is correct -- it prevents one task's cancellation from affecting others.

3. **`Mutex<Option<LlamaBackend>>`** -- The most targeted fix for the Metal/GPU crash. Replacing `OnceLock` with `Mutex<Option<>>` is the simplest pattern that enables explicit cleanup. The `BackendGuard` RAII wrapper is a clean abstraction that preserves the ergonomics of the old `&'static LlamaBackend` reference while allowing ownership transfer via `take()`.

The cleanup ordering (contexts first, then backend) is correctly enforced through the sequential calls in `release_gpu_resources`. The redundant `cancel()` calls across `CloseRequested`, `ExitRequested`, and `Exit` events properly handle the unreliable macOS event delivery noted in Tauri issue #9198.

### Security Assessment

No security concerns. The `unsafe` code in `LlamaState` (lifetime transmute at line 226) is pre-existing and not modified by this PR. The new `release_llama_backend` function uses safe Rust exclusively. The `unwrap_or_else(|p| p.into_inner())` pattern for poisoned Mutex recovery is appropriate for shutdown scenarios where the alternative is crashing anyway.

### Testing Assessment

The PR does not add new unit tests. The nature of this fix (process shutdown behavior, GPU resource cleanup, macOS-specific crash) makes it fundamentally difficult to unit test -- the crash only manifests in release builds during `__cxa_finalize_ranges`. The issue's testing approach (manual verification via `cargo tauri build` + macOS Console.app) is appropriate. The existing embedding tests (`test_blob_conversion`, `test_service_creation`, etc.) continue to validate the `EmbeddingService` API surface, which has not changed.

---

**Summary:** This is a well-executed fix for a real production crash. The three-layer approach provides defense in depth, the code is clean and well-documented, and the architectural decisions are sound. Approve with minor suggestions noted above.

PR #889, opened 2026-02-19
Add 'checkbox' node type for document-local checklist items

### LINKED ISSUE #887: Add 'checkbox' node type for document-local checklist items
## Summary

Introduce a `checkbox` node type for document-local checklist items. This is a **pure content node** — like a heading or ordered list item — with no properties, no schema, and no task management semantics. State is encoded directly in the content as markdown syntax.

## Background & Design Decision

Previously considered adding a `tracked: bool` property to task nodes. Rejected in favor of a clean structural type distinction:
- `checkbox` = primitive markdown node, content-only, no tracking
- `task` = managed action item with properties (status, due date, assignee, etc.)

**Key decisions:**
- `- [ ]` shortcut → creates a `checkbox` node
- `- [x]` → creates a `checkbox` node with checked content (`- [x] text`)
- State is stored in content: `- [ ] text` (unchecked) or `- [x] text` (checked)
- Toggling the checkbox updates the content string in place — no property writes
- No properties, no schema, no status enum — just like headings and list items
- `/task` slash command remains the way to create a real `task` node
- Converting checkbox → task uses the existing `/` slash command type-change (no new UX)
- Markdown imports unconditionally create `checkbox` nodes from `- [ ]` / `- [x]` syntax (no opt-in parameter)

### 1. Add `checkbox` node type — backend

- [ ] `checkbox` is a valid node type in the backend
- [ ] `checkbox` nodes have NO properties — content only (e.g. `- [ ] Buy milk` or `- [x] Buy milk`)
- [ ] `checkbox` nodes do NOT get a `title` field → automatically excluded from `@mention` autocomplete via existing `WHERE title \!= NONE` filter
- [ ] `checkbox` nodes do NOT appear in task queries or task board views

### 2. Update markdown import

- [ ] Both import paths (`create_nodes_from_markdown` MCP handler and batch/folder import) create `checkbox` nodes from `- [ ]` / `- [x]` syntax
- [ ] Content is stored as-is: `- [ ] text` or `- [x] text`
- [ ] No import parameter needed — `checkbox` is unconditional for all markdown imports

### 3. Change `- [ ]` editor shortcut

- [ ] Typing `- [ ] ` in the editor creates a `checkbox` node with content `- [ ] `
- [ ] Typing `- [x] ` creates a `checkbox` node with content `- [x] `
- [ ] `/task` slash command remains the way to create a `task` node

### 4. Frontend rendering

- [ ] `checkbox` nodes render as a checkbox UI (checked/unchecked visual)
- [ ] Toggling the checkbox updates the content string (`- [ ]` ↔ `- [x]`) — no property writes
- [ ] No task management properties panel shown for `checkbox` nodes
- [ ] Type conversion via `/` slash command (checkbox → task, or any other type) works as normal

## Files to Modify

| File | Change |
|------|--------|
| `packages/core/src/mcp/handlers/markdown.rs` | Emit `checkbox` node type (instead of `task`) from `- [ ]` / `- [x]` syntax in both `parse_markdown()` and `prepare_nodes_from_markdown()`; preserve content as-is |
| `packages/core/src/db/surreal_store.rs` | Exclude `checkbox` from title computation in `compute_title_for_bulk_insert()` |
| Frontend editor shortcut handler | Change `- [ ]` pattern to create `checkbox` node type |
| Frontend plugin/node type registry | Register `checkbox` node type with checkbox renderer |
| Frontend `CheckboxNode` component | Render content as checkbox; toggle updates content string |

## Testing

- [ ] `- [ ]` in editor creates a `checkbox` node with content `- [ ] `
- [ ] `- [x]` in editor creates a `checkbox` node with content `- [x] `
- [ ] Toggling checkbox updates content, not a property
- [ ] `/task` slash command still creates a `task` node
- [ ] MCP and batch imports produce `checkbox` nodes from `- [ ]` syntax
- [ ] `checkbox` nodes do NOT appear in `@mention` autocomplete
- [ ] Slash command type-change (checkbox → task) works correctly
- [ ] Existing task tests continue to pass

## Notes

- `checkbox` has no properties and no schema — it is as simple as a heading node
- State lives in content, not in a `status` property
- `task` node behavior is completely unchanged by this issue

malibio:
Closes #887

### DIFF
--- a/packages/core/src/mcp/handlers/markdown.rs
+++ b/packages/core/src/mcp/handlers/markdown.rs
@@ -1303,7 +1303,7 @@ fn detect_heading(line: &str) -> Option<usize> {
 ///
 /// Tasks are lines starting with "- [ ] " (unchecked) or "- [x] " (checked).
 fn is_task_line(line: &str) -> bool {
-    line.starts_with("- [ ] ") || line.starts_with("- [x] ")
+    line.starts_with("- [ ] ") || line.starts_with("- [x] ") || line.starts_with("- [X] ")
 }
 
 /// Check if a line is a bullet list item (not a task)
--- a/packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts
+++ b/packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts
@@ -720,6 +720,8 @@ export function createReactiveNodeService(events: NodeManagerEvents) {
     let cleaned = content;
     // Strip header prefixes (# , ## , etc.)
     cleaned = cleaned.replace(/^#{1,6}\s+/, '');
+    // Strip checkbox bullet prefix (- [ ] or - [x] or - [X] )
+    cleaned = cleaned.replace(/^- \[[ xX]\] /, '');
     // Strip task checkbox ([ ] , [x] , etc.)
     cleaned = cleaned.replace(/^\[\s*[x\s]*\]\s*/, '');
     // Strip quote-block prefixes (> ) from all lines
--- a/packages/desktop-app/src/tests/services/reactive-node-service.test.ts
+++ b/packages/desktop-app/src/tests/services/reactive-node-service.test.ts
@@ -1374,6 +1374,24 @@ describe('ReactiveNodeService - Combine Nodes', () => {
     const node1 = service.findNode('node-1');
     expect(node1?.content).toBe('First nodeTask content');
   });
+
+  it('strips checkbox bullet prefix when combining', () => {
+    service.updateNodeContent('node-2', '- [ ] Checkbox content');
+
+    service.combineNodes('node-2', 'node-1');
+
+    const node1 = service.findNode('node-1');
+    expect(node1?.content).toBe('First nodeCheckbox content');
+  });
+
+  it('strips checked checkbox bullet prefix when combining', () => {
+    service.updateNodeContent('node-2', '- [x] Checked content');
+
+    service.combineNodes('node-2', 'node-1');
+
+    const node1 = service.findNode('node-1');
+    expect(node1?.content).toBe('First nodeChecked content');
+  });
 });
 
 describe('ReactiveNodeService - Indent/Outdent Node', () => {


malibio:
## Code Review - PR #889: Add 'checkbox' node type for document-local checklist items

---

### Code Review Summary

This PR correctly implements the architectural decision to separate 'checkbox' (pure markdown annotation) from 'task' (managed action item with properties). The design is clean, the backend and frontend changes are consistent, and the commit is appropriately atomic. The implementation follows established project patterns closely, particularly mirroring the TaskNode component structure.

Overall assessment: **net positive change** that improves type clarity. One correctness issue needs resolution before merge, and there are a small number of secondary improvements worth addressing.

---

### Requirements Check

- ✅ `checkbox` is a valid node type in the backend (registered in `core_schemas.rs`)
- ✅ `checkbox` nodes have NO properties — content only
- ✅ `checkbox` nodes do NOT get a `title` field — excluded from `@mention` autocomplete via `compute_title_for_bulk_insert`
- ✅ `checkbox` nodes do NOT appear in task queries (no task-specific properties; type distinction enforced at the schema level)
- ✅ Both import paths (`prepare_nodes_from_markdown` and `create_nodes_from_markdown`) emit `checkbox` nodes from `- [ ]`/`- [x]` syntax
- ✅ Content preserved as-is: `- [ ] text` or `- [x] text`
- ✅ No import parameter needed — unconditional for all markdown imports
- ✅ Typing `- [ ] ` in the editor creates a `checkbox` node
- ✅ Typing `- [x] ` creates a `checkbox` node with checked content
- ✅ `/task` slash command remains the way to create a `task` node
- ✅ `checkbox` nodes render as a checkbox UI (checked/unchecked visual)
- ✅ Toggling the checkbox updates the content string (`- [ ]` ↔ `- [x]`) — no property writes
- ✅ No task management properties panel shown
- ⚠️ Type conversion via `/` slash command: not directly tested but relies on shared slash command infrastructure — acceptable

---

### Findings

#### Critical Issues

**[Critical/Blocker] `stripFormattingSyntax` does not handle the checkbox content format**
📁 `packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts:719-731`

When a checkbox node is deleted via backspace (empty content backspace = combine with previous), `combineNodes()` calls `stripFormattingSyntax(currentNode.content)`. The relevant strip rule is:

```ts
cleaned = cleaned.replace(/^\[\s*[x\s]*\]\s*/, '');
```

This regex matches `[ ] text` or `[x] text` at the start. But checkbox node content starts with `"- [ ] text"` or `"- [x] text"` — the `"- "` prefix prevents the regex from matching. The result is that the raw markdown prefix (`- [ ] `) gets appended verbatim to the previous node's content instead of being stripped.

**Fix**: Extend `stripFormattingSyntax` to also strip the checkbox bullet prefix:

```ts
// Strip checkbox bullet prefix (- [ ] or - [x] )
cleaned = cleaned.replace(/^- \[[ xX]\] /, '');
```

This should be added before the existing task checkbox strip rule, since they are distinct patterns. The existing test at `reactive-node-service.test.ts:1369` tests `"[ ] Task content"` (no leading dash) and is validating the old task pattern, not the new checkbox format. That test description and fixture should be updated to reflect checkbox semantics.

---

#### Suggested Improvements

**[Improvement] `is_task_line` does not match `- [X] ` (uppercase X)**
📁 `packages/core/src/mcp/handlers/markdown.rs:1305-1307`

```rust
fn is_task_line(line: &str) -> bool {
    line.starts_with("- [ ] ") || line.starts_with("- [x] ")
}
```

The frontend `CheckboxNode` correctly handles both `- [x] ` and `- [X] ` (line 45 of `checkbox-node.svelte`), and the frontend editor pattern `/^- \[[ xX]\] /` also matches uppercase X. However, the backend `is_task_line` only matches lowercase `x`. A markdown file containing `- [X] item` will not be imported as a `checkbox` node — it falls through to the default branch and is treated as a plain text or bullet node.

GFM specifies that `- [X] ` is a valid checked task list item. This inconsistency will cause data loss or misclassification on import.

**Fix**:
```rust
fn is_task_line(line: &str) -> bool {
    line.starts_with("- [ ] ") || line.starts_with("- [x] ") || line.starts_with("- [X] ")
}
```

**[Improvement] `compute_title_for_bulk_insert` is the only title exclusion path — `create_node` single-insert path may set a title**
📁 `packages/core/src/db/surreal_store.rs:3712` and `packages/core/src/services/node_service.rs:1381`

The bulk import paths correctly call `compute_title_for_bulk_insert` which excludes `checkbox`. However, when a checkbox is created interactively (via editor pattern or slash command), the code path goes through `node_service.rs:1381`:

```rust
if (node.node_type == "task" || node.node_type == "collection") && node.title.is_none() {
    node.title = Some(crate::utils::strip_markdown(&node.content));
}
```

Because `"checkbox"` is not in this condition, the title remains `None` for interactive creation — which is correct behavior. However, the docstring on `compute_title_for_bulk_insert` says "Date and schema types never get titles" and doesn't mention `checkbox`. This is a documentation gap that will mislead the next engineer. The actual acceptance criterion (no `@mention` autocomplete for checkboxes) is satisfied by the `WHERE title != NONE` filter in the search query, which works correctly since `checkbox` nodes never have titles set.

**Nit**: Update the docstring on `compute_title_for_bulk_insert` at line 3702-3706 to include `checkbox` in the exclusion list description.

**[Improvement] `pattern-state.test.ts` retains stale task pattern fixture**
📁 `packages/desktop-app/src/tests/state/pattern-state.test.ts:74-91`

```ts
const mockTaskPlugin: PluginDefinition = {
  id: 'task',
  pattern: {
    detect: /^[-*+]?\s*\[\s*[xX\s]\s*\]\s/,  // old task pattern
    canRevert: false, // Task nodes cannot revert (checkbox syntax removed from content)
    ...
  }
};
```

This fixture still uses the old editor-pattern detection regex that was removed from the real `taskNodePlugin`. The comments reference "checkbox syntax removed from content" which contradicts the new architecture. While this is a unit test for the `PatternState` class itself (and the fixture is valid for that purpose), the stale comments and the retention of the old pattern add confusion and could mislead a future engineer who reads this test to understand task behavior.

**Nit: `aria-checked` on a `<button>` with `role="checkbox"`**
📁 `packages/desktop-app/src/lib/components/viewers/checkbox-node.svelte:70-76`

```html
<button
  type="button"
  aria-label={isChecked ? 'Uncheck' : 'Check'}
  aria-checked={isChecked}
  role="checkbox"
  onclick={toggleCheckbox}
>
```

Using `role="checkbox"` on a `<button>` element overrides the native button role. While `aria-checked` is valid for `role="checkbox"`, the combination of a native `<button>` (which handles keyboard interaction as a button) with `role="checkbox"` (which expects Space to toggle and no activation via Enter by default) is semantically inconsistent. The component is functional for mouse interaction but may behave unexpectedly for keyboard-only users.

The established pattern in this codebase is the TaskNode approach: use a button with no role override, and manage visual state via CSS class + icon. If ARIA checkbox semantics are important for screen reader users, consider using an actual `<input type="checkbox">` with `<label>` wrapping to get native keyboard and screen reader behavior for free.

---

#### Nitpicks

**Nit: Comment on `checkboxNodePlugin` pattern references `canRevert` and `cleanContent` inconsistently**
📁 `packages/desktop-app/src/lib/plugins/core-plugins.ts:222`

```ts
canRevert: false, // cleanContent: false — content preserved as-is (full markdown line)
```

The comment mentions `cleanContent` which is not a field in the pattern config type — the relevant field is `canRevert`. The comment is trying to explain the content-preservation behavior but uses non-existent terminology. Replace with: `// false: node cannot revert to plain text by backspacing the prefix`.

**Nit: `prefixToInherit: '- [ ] '` always creates unchecked nodes on Enter**
📁 `packages/desktop-app/src/lib/plugins/core-plugins.ts:224`

When pressing Enter at the end of a checked checkbox (`- [x] item`), the new node will inherit `- [ ] ` (unchecked) regardless of the current checked state. This is semantically correct behavior (new items should default to unchecked), but worth an explicit comment clarifying the intent, since the header node plugin inherits the same prefix level which feels different here.

---

### Summary

The design is sound and the implementation is consistent with the existing architecture. Before merge:

1. **Must fix**: `stripFormattingSyntax` in `reactive-node-service.svelte.ts` to handle the `- [ ] ` prefix so backspace-combine on checkbox nodes produces clean merged content.
2. **Should fix**: `is_task_line` in `markdown.rs` to match `- [X] ` (uppercase) for GFM compliance.

**Recommendation**: REQUEST CHANGES

PR #893, opened 2026-02-20
Fix: Checkbox node UI/UX — proper icon pattern, distinct icon, prefix revert

malibio:
## Summary

Fixes several issues with the checkbox node type introduced in #887:

- **Wrong icon pattern**: Checkbox was rendering its own custom toggle button alongside BaseNode's circle icon. Rewritten to follow the `quote-block-node` pattern — uses `displayContent` to strip the `- [ ] ` prefix in display mode, `iconClick` to toggle the content prefix, no custom UI elements.
- **Wrong icon**: Was using the circular `TaskIcon`. Replaced with a new `CheckboxIcon` (rounded square) to visually distinguish checkboxes from task nodes.
- **No pattern revert**: Breaking the `- [ ] ` prefix did not convert back to a text node. Fixed by setting `canRevert: true`, matching header/quote-block behavior.
- **Ring gap**: Checkbox ring (shown when node has children) had visible gap. Fixed using `outline` with `outline-offset: 0`.

## Test plan

- [ ] Typing `- [ ] ` creates a checkbox node with square icon
- [ ] Checkbox icon is visually distinct from task circle icon
- [ ] Clicking the icon toggles checked/unchecked state (updates content prefix)
- [ ] In display mode, `- [ ] ` prefix is hidden; in edit mode it shows
- [ ] Deleting any character from the `- [ ] ` prefix reverts node to text
- [ ] Checkbox node with children shows ring flush against the square (no gap)
- [ ] Checked checkbox shows strike-through text

### DIFF
--- a/packages/desktop-app/src/lib/design/icons/components/checkbox-icon.svelte
+++ b/packages/desktop-app/src/lib/design/icons/components/checkbox-icon.svelte
@@ -78,7 +78,7 @@
     border-radius: 3px;
   }
 
-  /* Checkmark for checked state */
+  /* Checkmark for checked state — light theme */
   .checkbox-square-checked::after {
     content: '';
     position: absolute;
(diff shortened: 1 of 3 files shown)


malibio (commented):
Review status: This PR is **approved** from an engineering standpoint. The four issues from #887 are correctly fixed. The canRevert=true decision is sound given that the checkbox prefix lives in content. The two items to address in follow-ups: (1) add a  regex to the checkbox plugin pattern for consistency and forward-safety, (2) add a dedicated checkbox plugin test file covering the reversion path.

PR #907, opened 2026-02-24
Fix ReactiveStructureTree $state.raw() reactivity on Map mutations

malibio:
## Summary
- `$state.raw()` only triggers Svelte reactivity on **reassignment**, not on in-place Map mutations (`.set()`/`.delete()`)
- Added `notifyChange()` helper that reassigns the Map after every mutation to trigger UI updates
- Fixes nodes added via MCP not appearing until navigating away and back

## Test plan
- [x] All 3586 frontend tests pass with no regressions
- [ ] Manual test: Launch app, navigate to DateNode, add node via MCP, verify it appears without navigation

### DIFF
--- a/packages/desktop-app/src/lib/stores/reactive-structure-tree.svelte.ts
+++ b/packages/desktop-app/src/lib/stores/reactive-structure-tree.svelte.ts
@@ -79,6 +79,15 @@ class ReactiveStructureTree {
    * which needs to update the tree structure when external changes are detected.
    */
   addChild(rel: HierarchyRelationship) {
+    this.addChildInternal(rel);
+    this.notifyChange();
+  }
+
+  /**
+   * Core mutation logic for adding a child. Does NOT trigger reactivity —
+   * caller is responsible for calling notifyChange() after mutations complete.
+   */
+  private addChildInternal(rel: HierarchyRelationship) {
     const parentId = rel.parentId;
     const childId = rel.childId;
     const order = rel.order;
@@ -100,7 +109,6 @@ class ReactiveStructureTree {
         // Re-sort array
         children.sort((a, b) => a.order - b.order);
         this.children.set(parentId, children);
-        this.notifyChange();
       }
       return;
     }
@@ -129,7 +137,6 @@ class ReactiveStructureTree {
 
     children.splice(insertIndex, 0, newChild);
     this.children.set(parentId, children);
-    this.notifyChange();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


malibio (commented):
### Code Review Summary

**Verdict: Approve (posted as comment -- GitHub blocks self-approval).** This is a correct, well-scoped bug fix that addresses a real Svelte 5 reactivity gap. The previous code comments incorrectly claimed `$state.raw()` tracks in-place Map mutations -- it does not. Only reassignment triggers reactivity, and this PR fixes that with a clean, minimal approach.

The change is a clear net positive: it fixes a real UI bug (nodes added via MCP/external events not rendering) with a focused, single-purpose commit.

### Findings

#### Suggested Improvements

- **`batchAddRelationships` creates N intermediate Map copies** (`packages/desktop-app/src/lib/stores/reactive-structure-tree.svelte.ts:230-233`): Each `addChild()` call triggers `notifyChange()`, which copies the entire Map. For a batch of 100 relationships this creates 100 intermediate Map instances. The comment acknowledges this is "acceptable for typical batch sizes," which is fair for now, but if batch sizes grow (e.g., loading a large workspace), this could become a bottleneck. A straightforward optimization would be to inline the loop, call `this.children.set()` for each relationship, and call `notifyChange()` once at the end. **Not a blocker** -- the current approach is correct and the comment documents the trade-off. Worth revisiting if profiling shows it matters.

#### Nitpicks

- **Nit: Tests directly mutate `structureTree.children` bypassing `notifyChange()`** (test file, lines using `structureTree.children.clear()` and `structureTree.children.set()`): The tests set up state by directly mutating the Map (e.g., `structureTree.children.clear()`, `structureTree.children.set(...)`). This is fine for test setup in unit tests since they are not testing Svelte reactivity, but it does mean the tests do not validate the new `notifyChange()` path. Consider adding a comment in the test file noting this is intentional setup (not exercising the reactive contract), or use the public API methods (`addChild`, `clear`) for setup where practical. Very minor -- the current tests are adequate.

### Notes

- The corrected comments accurately describe `$state.raw()` semantics. Good documentation hygiene.
- `clear()` and `restore()` already worked correctly because they use direct reassignment (`this.children = ...`), which is consistent with the `$state.raw()` contract. No changes needed there.
- The fix is correctly applied to all three mutation paths: `addChild` (new entry), `addChild` (order update), and `removeChild`. Coverage is complete.

PR #908, opened 2026-02-24
Add atexit handler for Metal GPU cleanup on NSApplication terminate

### LINKED ISSUE #905: Metal GPU crash on quit persists — background task Arcs not drained before backend release
## Problem

Despite PR #902's fix (draining the `EmbeddingProcessor` before GPU release), the app still crashes with `SIGABRT` in `ggml_metal_rsets_free` on normal quit (Cmd+Q or window close).

**Stack trace:** `[NSApplication terminate:] → ggml_metal_device_free → ggml_metal_device_init (line 608)`

## Root Cause

PR #902 correctly drains the `EmbeddingProcessor`, but **other background tasks still hold `Arc<NodeEmbeddingService>` references** when `release_llama_backend()` destroys the global Metal backend:

- **MCP server task** — holds `Arc<NodeEmbeddingService>` (spawned in `initialize_mcp_server()`)
- **Domain event forwarder** — holds `Arc<NodeService>`

PR #902 also **removed the 50ms grace period** between `shutdown_token.cancel()` and GPU release. This means background tasks get the cancellation signal but have zero time to exit and drop their Arcs before the Metal backend is destroyed.

## Fix

1. **Restore grace period** in `graceful_shutdown()` — 200ms after `shutdown_token.cancel()` to let MCP/forwarder tasks exit
2. **Cancel session token explicitly** in `release_gpu_resources()` for defense-in-depth

## Acceptance Criteria

- [ ] App exits cleanly without SIGABRT on Cmd+Q
- [ ] App exits cleanly on window close (X button)
- [ ] Database hot-swap still works (no regression)
- [ ] `cargo check` and `cargo clippy` clean
- [ ] Rust tests pass (no regressions)

malibio:
## Summary

- **Root cause**: When macOS terminates the app via `NSApplication terminate:` (Cmd+Q after long session, Dock quit, system shutdown), the `exit()` path bypasses Tauri's `RunEvent` handlers entirely. C++ static destructors (`__cxa_finalize_ranges`) then destroy the Metal device while our `LlamaModel` still holds Metal residency sets, triggering `ggml_metal_rsets_free` assertion failure (SIGABRT).
- **Fix**: Register a `libc::atexit` handler when the embedding model is first loaded. `atexit` handlers run **before** `__cxa_finalize_ranges`, so we drop all `LlamaState` instances (freeing Metal residency sets) and the `LlamaBackend` before C++ static destructors run.
- **Architecture**: Each `EmbeddingService` registers its `Arc<Mutex<Option<LlamaState>>>` in a global `LLAMA_STATES` vec. The atexit handler iterates and clears them all. Graceful shutdown still works (its `.take()` runs first; atexit's `.take()` returns `None`).

## Test plan

- [x] `cargo check` — compiles clean
- [x] `cargo clippy` — no warnings
- [x] `bun run rust:test` — 788 passed, 0 failed
- [x] Atexit handler confirmed working in test output: `[atexit] LlamaState dropped` x3 + `[atexit] LlamaBackend dropped`
- [x] Manual: launch app, wait for NLP model load, Cmd+Q — no crash (exit code 0)

Closes #905

### DIFF
--- a/packages/nlp-engine/src/embedding.rs
+++ b/packages/nlp-engine/src/embedding.rs
@@ -99,6 +99,8 @@ fn register_atexit_handler() {
 #[cfg(feature = "embedding-service")]
 fn register_state_for_cleanup(state: &Arc<Mutex<Option<LlamaState>>>) {
     let mut states = LLAMA_STATES.lock().unwrap_or_else(|p| p.into_inner());
+    // Prune stale entries (already-taken states from previous initialize() calls)
+    states.retain(|s| s.lock().unwrap_or_else(|p| p.into_inner()).is_some());
     states.push(Arc::clone(state));
 }
 
@@ -387,8 +389,8 @@ impl EmbeddingService {
             // even if NSApplication terminate: bypasses our graceful shutdown path.
             let state = LlamaState::new(model, self.config.context_size, self.config.n_threads);
             *self.state.lock().unwrap_or_else(|p| p.into_inner()) = Some(state);
-            register_state_for_cleanup(&self.state);
             register_atexit_handler();
+            register_state_for_cleanup(&self.state);
 
             tracing::info!(
                 "Embedding service initialized with persistent context (Issue #776 optimization)"


malibio (commented):
### Code Review Summary

**Verdict: APPROVE** (posted as comment — GitHub blocks self-approval)

This is a well-targeted, narrowly scoped fix for a real crash (SIGABRT in `ggml_metal_rsets_free`) that occurs when `NSApplication terminate:` bypasses Tauri's `RunEvent` handlers and the `exit()` path goes straight to C++ static destructors. The solution — registering a `libc::atexit` handler that drops Metal-holding state before `__cxa_finalize_ranges` runs — is the correct approach for this class of problem.

The change is a clear net positive: it fixes a crash with minimal surface area, introduces no regressions to graceful shutdown, and the code is well-documented.

---

### Findings

#### No Critical Issues

None. The architecture is sound and the implementation is correct.

---

#### Suggested Improvements

**1. [Improvement] Unbounded growth of `LLAMA_STATES` registry**
`packages/nlp-engine/src/embedding.rs`, lines 98-103:

`register_state_for_cleanup` pushes a new `Arc` clone every time `initialize()` is called. If an `EmbeddingService` is shutdown and re-initialized (e.g., during database hot-swap via `app_services.rs:214`), the old `Arc` stays in the vec (its inner `Option` is `None`, so the atexit handler harmlessly `.take()`s `None`). Over many hot-swaps this leaks `Arc` entries.

For a desktop app with infrequent hot-swaps this is negligible, but a one-line cleanup would make it bulletproof:

```rust
fn register_state_for_cleanup(state: &Arc<Mutex<Option<LlamaState>>>) {
    let mut states = LLAMA_STATES.lock().unwrap_or_else(|p| p.into_inner());
    // Prune already-cleared entries from previous shutdown cycles
    states.retain(|s| s.lock().unwrap_or_else(|p| p.into_inner()).is_some());
    states.push(Arc::clone(state));
}
```

Not a blocker -- the current code is correct, just not frugal with stale entries.

**2. [Improvement] Registration order: `register_state_for_cleanup` before `register_atexit_handler`**
`packages/nlp-engine/src/embedding.rs`, lines 390-391:

Currently the state is registered first (line 390), then the atexit handler is registered (line 391). If -- in a truly pathological race -- `exit()` were called between these two lines, the state would be registered but no atexit handler would exist to clean it up. Reversing the order (register handler first, then state) closes this theoretical window:

```rust
register_atexit_handler();
register_state_for_cleanup(&self.state);
```

Extremely unlikely in practice, but trivial to fix and makes the ordering obviously correct.

---

#### Nitpicks

**Nit:** `packages/nlp-engine/src/embedding.rs`, lines 80, 88 -- The `eprintln!` calls in the atexit handler are appropriate for a last-resort cleanup path (tracing may be torn down), so this is fine. Just noting for future readers: these are intentionally `eprintln!` not `tracing::info!`.

**Nit:** `packages/nlp-engine/Cargo.toml`, line 30 -- `libc` is added as an unconditional dependency but only used under `#[cfg(feature = "embedding-service")]`. Consider making it optional and gating it on the feature:

```toml
libc = { version = "0.2", optional = true }

[features]
embedding-service = ["llama-cpp-2", "libc"]
```

This avoids pulling `libc` when the crate is compiled without `embedding-service`. Minor -- `libc` is lightweight and likely already in the dep tree.

---

### Summary Table

| Category | Status |
|---|---|
| Architecture | Clean -- atexit is the correct mechanism for this problem |
| Correctness | Correct -- idempotent with graceful shutdown path |
| Security | N/A |
| Maintainability | Good -- excellent comments explaining the "why" |
| Testing | Verified via rust tests (788 passed) and manual Cmd+Q |
| Performance | No impact -- atexit registration is O(1), cleanup is O(n) at exit |
| Dependencies | `libc` is standard, well-maintained, MIT licensed |

PR #916, opened 2026-02-27
Fix: Smart navigation resolveWithRootAncestor breaks task node viewer opening

### LINKED ISSUE #915: Fix: Smart navigation resolveWithRootAncestor breaks task node viewer opening
## Bug Description

Commit `0ed40002` ("Smart navigation for non-root node links with scroll-to-target", #910) introduced a regression where opening a task node in a new tab no longer opens the TaskNodeViewer — it opens the parent node's viewer instead (typically BaseNodeViewer for a text/date node).

## Root Cause

`resolveWithRootAncestor()` in `navigation-service.ts` unconditionally walks up the structure tree to find the root ancestor and uses that node's `nodeType` for the tab. This is correct for **primitive** child nodes (text blocks, headers, code blocks) that belong to their parent's viewer, but incorrect for **entity** nodes like tasks that have their own dedicated viewer.

Task nodes can have a parent in the tree (e.g., under a date node), but they are not "primitive" content — they have a title field and a dedicated `TaskNodeViewer`. The root-ancestor resolution should stop at entity node boundaries, not walk all the way to the tree root.

## Reproduction

1. Open NodeSpace with task nodes nested under a date or text node
2. Click the open button on a task node
3. **Expected:** Task opens in TaskNodeViewer in the other pane
4. **Actual:** The parent (date/text) node opens in BaseNodeViewer

## Affected Code

- `packages/desktop-app/src/lib/services/navigation-service.ts`
  - `findRootAncestor()` (line ~132) — walks up unconditionally
  - `resolveWithRootAncestor()` (line ~185) — uses root's nodeType
  - Called by both `navigateToNode()` and `navigateToNodeInOtherPane()`

## Proposed Fix

Make `findRootAncestor()` aware of node types that have their own viewer. Stop walking up when the current node is an "entity" type (task, date, query, etc.) rather than a "primitive" type (text, header, code-block). This can be done by:

1. Checking the node's type during the ancestor walk
2. Stopping at nodes registered with a viewer in the plugin registry
3. Or adding a concept of "viewer-owning" node types that act as navigation boundaries

## Acceptance Criteria

- [ ] Clicking the open button on a task node opens TaskNodeViewer (not the parent's viewer)
- [ ] Clicking a wikilink to a primitive child node still resolves to the root ancestor and scrolls to it (existing #910 behavior preserved)
- [ ] Other entity node types with dedicated viewers (date, query) also open correctly
- [ ] Existing navigation tests pass
- [ ] New test coverage for entity-node-within-parent navigation scenario

malibio:
Closes #915

### DIFF
--- a/packages/desktop-app/src/lib/services/navigation-service.ts
+++ b/packages/desktop-app/src/lib/services/navigation-service.ts
@@ -138,7 +138,7 @@ export class NavigationService {
    *
    * Returns the original nodeId if it's a root node or has its own viewer.
    */
-  private findRootAncestor(nodeId: string): string {
+  private findNavigationAncestor(nodeId: string): string {
     if (!structureTree) return nodeId;
 
     // If the target node itself has a dedicated viewer, it should open in that viewer
@@ -200,10 +200,10 @@ export class NavigationService {
 
   /**
    * Resolve a navigation target, and if it's a non-root node, also resolve
-   * the root ancestor's info for tab navigation. Returns both the original
-   * target and the navigation-level (root) info.
+   * the navigation ancestor's info for tab navigation. Returns both the original
+   * target and the navigation-level ancestor info.
    */
-  private async resolveWithRootAncestor(nodeId: string): Promise<{
+  private async resolveWithNavigationAncestor(nodeId: string): Promise<{
     target: NavigationTarget;
     navNodeId: string;
     navNodeType: string;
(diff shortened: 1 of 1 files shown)


malibio (commented):
APPROVE (posted as comment since self-review is restricted). Clean, minimal bug fix with good test coverage. The pluginRegistry.hasViewer() approach is architecturally sound -- it leverages existing infrastructure and automatically handles future node types. See detailed review in PR comments for two optional naming improvements.

Issue #920, opened 2026-02-28
Dev-proxy database initialization out of sync with Tauri app refactoring


malibio:
## Overview

The dev-proxy (`packages/dev-tools/src/bin/dev-proxy.rs`) was not updated when the Tauri app was refactored to support database selection, creation, and hot-swap. This causes `bun run dev:browser` to fail with a `schema.surql` execution error against SurrealDB 2.6.0.

## Problem Statement

**User Experience:**
1. Run `bun run dev:browser` to start browser-based development
2. SurrealDB starts successfully
3. Dev-proxy attempts to connect and initialize schema
4. Schema execution fails with 400 Bad Request — `Specify a namespace to use`
5. Dev environment is unusable

**Root Cause:**
The SurrealDB Rust SDK HTTP client (`surrealdb ^2.3`) has a session state issue with SurrealDB 2.6.0's HTTP RPC endpoint. The `use` command returns success but subsequent queries don't retain the namespace/database context.

**Evidence:**
- The `SurrealStore::new_http()` function calls `db.use_ns(...).use_db(...).await` which succeeds
- The subsequent `db.query(schema_sql).await` fails with "Specify a namespace to use"
- This indicates the `use` context is not persisting across HTTP RPC requests in SurrealDB 2.6.0

### Phase 1: Align dev-proxy with Tauri app initialization

The Tauri app uses `SurrealStore::new(db_path)` with embedded RocksDB — no HTTP layer, no session state issues. The dev-proxy should be updated to match this approach or the HTTP session issue should be resolved.

**Options:**
1. **Switch dev-proxy to embedded mode** — Use `SurrealStore::new(path)` instead of `new_http()`, eliminating the HTTP RPC compatibility issue entirely
2. **Fix HTTP session issue** — Upgrade surrealdb crate or work around the session state bug
3. **Add database configuration** — Make dev-proxy configurable like the Tauri app (path, namespace, database)

### Phase 2: Add missing configurability

The dev-proxy currently hardcodes all database connection parameters:
```rust
// Current: hardcoded in dev-proxy.rs
HttpStore::new_http("127.0.0.1:8000", "nodespace", "nodespace", "root", "root")
```

The Tauri app has full configurability:
```rust
// Tauri app: configurable via preferences and commands
SurrealStore::new(db_path)  // path from preferences/user selection
```

## Acceptance Criteria

- [ ] `bun run dev:browser` successfully starts and initializes the database schema
[...]

PR #929, opened 2026-03-13
Fix collection hierarchy, import batching, and subtree query performance (SurrealDB 3.x follow-up)

malibio:
## Summary

Follow-up fixes discovered during SurrealDB 3.x upgrade testing (#920).

- Fix nested collections not showing in sidenav
- Fix slow markdown import (single bulk transaction instead of per-node inserts)
- Fix slow root node fetch — 7-10x speedup via graph traversal query
- Upgrade SurrealDB 3.0.3 → 3.0.4 (released 2026-03-13, contains related index fixes)
- Add tracing subscriber init in Tauri app so debug logs are visible

## Performance

| Document size | Before | After |
|---------------|--------|-------|
| 25 nodes | ~350ms | 17ms |
| 74 nodes | 539ms | 56ms |
| 233 nodes | ~700ms | 79ms |

**Root cause:** `WHERE in IN $array` on the relationship table was not using the graph index in SurrealDB 3.x. Fixed by switching to graph traversal syntax (`->relationship[WHERE relationship_type = 'has_child']`) which bypasses the query planner and follows edges directly.

## Test plan

- [ ] All Rust tests passing
- [ ] Nested collections visible in sidenav
- [ ] Markdown import fast
- [ ] Root node fetch <100ms (check debug logs)

### DIFF
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -2049,7 +2049,7 @@ impl SurrealStore {
             .context("Failed to query descendants")?;
         let descendant_ids: Vec<String> = r1.take(0).context("Failed to extract descendants")?;
 
-        // All node IDs including root
+        // All node IDs including root — used for both node fetch (step 2) and edge traversal (step 3)
         let all_thing_ids: Vec<surrealdb::types::RecordId> = std::iter::once(root_id.to_string())
             .chain(descendant_ids.iter().cloned())
             .map(|id| node_record_id(&id))
@@ -2071,18 +2071,13 @@ impl SurrealStore {
             return Ok((all_nodes, vec![]));
         }
 
-        let parent_things: Vec<surrealdb::types::RecordId> = std::iter::once(root_id.to_string())
-            .chain(descendant_ids.iter().cloned())
-            .map(|id| node_record_id(&id))
-            .collect();
-
         // Fetch children per-node using graph traversal — hits idx_rel_in (graph index) per node,
         // avoiding the full relationship table scan that WHERE in IN $parents causes in SurrealDB 3.x.
         // Returns one row per parent with an array of {child_id, order, rel_id} objects.
         let mut r3 = self
             .db
             .query("SELECT meta::id(id) AS parent_id, ->relationship[WHERE relationship_type = 'has_child'].{ child_id: meta::id(out), order: properties.order, rel_id: meta::id(id) } AS children FROM $parents;")
-            .bind(("parents", parent_things))
+            .bind(("parents", all_thing_ids))
             .await
             .context("Failed to fetch subtree relationships")?;
 
@@ -2112,7 +2107,9 @@ impl SurrealStore {
             all_nodes.len()
         );
 
-        let mut relationships: Vec<RelationshipRecord> = parent_rows
+        // Consumer (get_subtree_data) sorts children per-parent by order when building the
+        // adjacency list, so no global sort is needed here.
+        let relationships: Vec<RelationshipRecord> = parent_rows
             .into_iter()
             .flat_map(|row| {
                 row.children.into_iter().map(move |child| {
@@ -2127,13 +2124,6 @@ impl SurrealStore {
                 })
             })
             .collect();
-
-        // Sort by order in Rust — properties.order is specific to has_child/member_of
-        relationships.sort_by(|a, b| {
-            let ord_a = a.properties.get("order").and_then(|v| v.as_f64()).unwrap_or(f64::MAX);
-            let ord_b = b.properties.get("order").and_then(|v| v.as_f64()).unwrap_or(f64::MAX);
-            ord_a.partial_cmp(&ord_b).unwrap_or(std::cmp::Ordering::Equal)
-        });
 
         Ok((all_nodes, relationships))
     }
--- a/packages/core/src/services/node_service.rs
+++ b/packages/core/src/services/node_service.rs
@@ -10367,9 +10367,8 @@ mod tests {
             let parent_thing = surrealdb::types::RecordId::new("node", parent_id);
 
             #[derive(Debug, serde::Deserialize, surrealdb::types::SurrealValue)]
-            #[allow(dead_code)] // out is used for deserialization but not accessed
             struct RelWithOrder {
-                out: surrealdb::types::RecordId,
+                out: Option<surrealdb::types::RecordId>,
                 order: Option<f64>,
             }
 
--- a/packages/desktop-app/src-tauri/src/lib.rs
+++ b/packages/desktop-app/src-tauri/src/lib.rs
@@ -185,7 +185,8 @@ pub fn run() {
             tracing_subscriber::EnvFilter::try_from_default_env()
                 .unwrap_or_else(|_| tracing_subscriber::EnvFilter::new("nodespace_core=info")),
         )
-        .init();
+        .try_init()
+        .ok();
 
     // Create shutdown token for coordinating graceful background task termination
     let shutdown_token = ShutdownToken::new();


malibio:
### Code Review Summary

This PR is a solid follow-up to the SurrealDB 3.x upgrade. The three primary changes (collection hierarchy, import batching, and subtree query performance) are all net improvements. The SurrealDB 3.0.4 upgrade is clean and the crate version constraint is correct. Removing the `Connection` generic type parameter throughout the codebase is the right call — it eliminates a significant amount of boilerplate with no downside since RocksDB is the only deployed backend.

The subtree query redesign is the most technically significant change and is correct. The 3-step approach is a reasonable pragmatic workaround for the SurrealDB 3.x query planner issue with `WHERE in IN $array` on relation tables.

**Verdict: Approve with one improvement recommendation and a few nits.**

---

### Findings

#### Suggested Improvements

**`packages/core/src/db/surreal_store.rs` — `get_subtree_with_relationships`: Duplicate vector construction**

`all_thing_ids` (used for Step 2, `SELECT * FROM $ids`) and `parent_things` (used for Step 3, the graph traversal) are constructed from the exact same iterator chain. The second build is dead work. The first can be cloned for the second bind, or they can share a single construction. At large subtree sizes, this allocates an extra `Vec<RecordId>` unnecessarily.

```rust
// Current — two identical builds:
let all_thing_ids: Vec<RecordId> = std::iter::once(root_id.to_string())
    .chain(descendant_ids.iter().cloned())
    .map(|id| node_record_id(&id))
    .collect();
// ... used for $ids

let parent_things: Vec<RecordId> = std::iter::once(root_id.to_string())
    .chain(descendant_ids.iter().cloned())
    .map(|id| node_record_id(&id))
    .collect();
// ... used for $parents
```

`parent_things` can simply be `all_thing_ids.clone()` (or the variable renamed and cloned at bind time).

**`packages/core/src/db/surreal_store.rs` — `get_all_collections_with_member_counts`: `count_query` overcounts when collections are nested**

The count query is:
```sql
SELECT VALUE meta::id(out) FROM relationship WHERE relationship_type = 'member_of';
```

This fetches the `out` side (the collection) of every `member_of` edge in the entire database — including collection-to-collection hierarchy edges. So if `StandardsGuides` is `member_of` `Standards`, `Standards` receives +1 in its member count from that edge, in addition to any document members.

The practical impact depends on whether the UI is expected to show "documents in this collection" vs "total members including sub-collections." The previous implementation had the same behavior (identical `count_query` in the old code), so this is not a regression introduced here. However, it is worth flagging since the hierarchy feature is new and the sidenav may eventually want to distinguish document-count from sub-collection-count.

**`packages/dev-tools/src/bin/dev-proxy.rs` — Fallback path silently drops `parent_collection_ids`**

The error fallback path in `get_all_collections` returns `parent_collection_ids: vec![]` for all collections with a `tracing::warn!`. This means any error in the primary query path silently degrades the sidenav to a flat collection list. The comment acknowledges this but no follow-up issue is referenced. Given that the primary path runs 3 separate queries (`get_all_collections`, `count_query`, `hierarchy_query`), a transient failure in any one of them triggers a complete loss of hierarchy. This is low risk in practice (embedded RocksDB rarely fails transiently) but worth tracking.

**`packages/desktop-app/src-tauri/src/lib.rs` — `tracing_subscriber` init uses `.init()` instead of `.try_init()`**

```rust
tracing_subscriber::fmt()
    .with_env_filter(...)
    .init(); // panics if already initialized
```

`.init()` panics if a global subscriber is already registered. In the production `run()` path this is fine — it is called once. However, if any future test or integration harness calls `run()` (or initializes tracing independently), this will panic. Prefer `.try_init()` or `try_init().ok()` to be defensive, which is the idiomatic approach for library-adjacent code.

---

#### Nitpicks

**Nit: `packages/core/src/services/node_service.rs` line 10370 — `#[allow(dead_code)]` in test module**

The project's CLAUDE.md explicitly prohibits `#[allow(dead_code)]`. This is inside a `#[cfg(test)]` block and the struct field `out: surrealdb::types::RecordId` is never read after deserialization. The fix is to either remove the `out` field from `RelWithOrder` (matching the refactored approach in the nearby `test_add_to_collection_assigns_order` test at line 10215 which already removed the `member` field) or add a `_` prefix. A lint suppression attribute should not be used.

**Nit: `packages/core/src/db/surreal_store.rs` — `get_subtree_with_relationships`: Rust-side global sort is redundant**

```rust
relationships.sort_by(|a, b| { ... }); // global sort across all relationships
```

The consumer `get_subtree_data` in `node_service.rs` builds an `adjacency_with_order` HashMap and independently sorts children per-parent before building the `adjacency_list`. The global sort here has no effect on correctness and adds O(E log E) work on every subtree fetch. It can be removed.

**Nit: `packages/core/src/db/schema.surql` — New `mentions` and `mentioned_in` field definitions**

```sql
DEFINE FIELD IF NOT EXISTS mentions ON TABLE node TYPE array<string> DEFAULT [];
DEFINE FIELD IF NOT EXISTS mentioned_in ON TABLE node TYPE array<object> DEFAULT [];
```

`mentioned_in` is documented in `SurrealNode` as never populated from the DB (it is `Vec::new()` on deserialization, populated at fetch time). Defining it as a SCHEMAFULL field with `TYPE array<object>` could cause issues if `mentioned_in` data was ever inadvertently written to the node record — SurrealDB would enforce it. The field could remain undeclared if it is never written to DB. Low risk since writes only go through `SurrealNode` → `Node` → `SurrealStore`, and `SurrealNode` has no `mentioned_in` field, so it will never be serialized outward. Flagging for awareness.

**Nit: Pre-existing inconsistency (not introduced by this PR) — `create_mention`/`delete_mention` event IDs use key-only format**

`store.create_mention` returns `extract_record_key(&result.id)` (e.g., `"abc-123"`), which is used as `RelationshipEvent.id` for mention events. All other relationship event emissions use `extract_record_id_string` which produces `"relationship:abc-123"` format. The `RelationshipEvent` contract test asserts `"relationship:abc123"` format. This inconsistency predates this PR (confirmed via `git show origin/main`) but this PR is a natural opportunity to align it, since the `extract_record_id_string` helper is already available in scope.

---

### Notes

- The 3-query decomposition in `get_all_collections_with_member_counts` (1 node fetch + 1 count query + 1 hierarchy query) is a reasonable tradeoff. The old single-query approach hit SurrealDB deserialization issues with `count()` computed fields. Three simple queries that all deserialize to primitive types is more maintainable.
- The `Option<Value>` instead of `Option<f64>` for `ChildEntry.order` is the correct fix for SurrealDB 3.x number type ambiguity. The `.as_f64()` coercion at usage time is idiomatic.
- Removing `nlp_engine.initialize()` from all test helpers is a necessary and correct fix — loading the GGML model in parallel test threads was causing the reported memory explosion.
- The `type::thing` → `type::record` rename in SurrealQL is a clean 3.x migration.
- `TYPE object FLEXIBLE` vs `FLEXIBLE TYPE object` — confirmed semantically equivalent in SurrealDB 3.x; the reordering just matches the new canonical syntax.
- HNSW parameters (`EFC 200 M 16`) are standard defaults for 768-dimension embeddings with cosine similarity. Appropriate.

malibio:
## Review Feedback Addressed

### ✅ Addressed (4 items)

- **Duplicate vector construction** — `parent_things` removed; `all_thing_ids` reused for step 3 graph traversal bind
- **Redundant global sort** — removed; consumer (`get_subtree_data`) already sorts per-parent when building adjacency list
- **`tracing_subscriber` `.init()` → `.try_init().ok()`** — defensive against double-init panic in test harnesses
- **`#[allow(dead_code)]` in test** — removed; `out` field changed to `Option<RecordId>` (handles nullable query results without suppression attribute)

### ⏭️ Skipped (3 items)

- **`count_query` overcounting nested collections** — pre-existing behavior, not a regression. Will track as a separate issue since it requires understanding how the sidenav should distinguish document-count vs sub-collection-count.
- **`dev-proxy` fallback silently drops hierarchy** — acknowledged in code. Low risk with embedded RocksDB (transient failures are effectively impossible). Would need more context on desired behavior before changing.
- **`create_mention`/`delete_mention` event ID format inconsistency** — pre-existing, confirmed not introduced by this PR. Out of scope for this fix PR.

PR #934, opened 2026-03-14
Fix TableNode display overlay: replace absolute positioning with customViewContent snippet

### LINKED ISSUE #931: Table node display mode overlay covers nodes below it
## Overview

In display (view) mode, the rendered table visually overlaps nodes that appear below it in the document. Clicking the table correctly enters edit mode and shows raw GFM markdown, but the display-mode rendering overflows its container and covers adjacent content.

## Problem Statement

**Steps to reproduce:**
1. Create a table node using the slash command `/table`
2. Type valid GFM markdown table syntax, e.g.:
   ```
   | Name  | Role   |
   |-------|--------|
   | Alice | Design |
   | Bob   | Eng    |
   ```
3. Click away (or press Escape) to leave edit mode
4. Observe that the rendered table visually overlaps nodes below it

**Expected:** The rendered table takes up vertical space in the document and nodes below it appear beneath it, not underneath it.

**Actual:** The rendered table floats over subsequent nodes, hiding them behind it.

## Root Cause

**File:** `packages/desktop-app/src/lib/design/components/table-node.svelte`

The component uses an **absolute-positioned overlay** to render the HTML table on top of a hidden (transparent) raw text content area. The problem is that `position: absolute` removes the overlay from normal document flow, so the wrapper element's height is determined only by the hidden text content — not by the actual rendered table.

When the rendered table is taller than a single line of raw text, it extends beyond the wrapper's bounds and visually covers whatever comes next in the document.

**The relevant CSS (lines 160–181):**
```css
/* Wrapper uses position: relative but its height = hidden text height */
.table-node-wrapper {
  position: relative;
}

/* Hidden text content — this is what controls wrapper height */
.table-node-wrapper.viewing :global(.node__content) {
  color: transparent;   /* invisible, but still in flow and sets height */
  position: relative;
  min-height: 1.5rem;
  user-select: none;
}

/* Table overlay — position: absolute takes it OUT of normal flow */
/* This means it does NOT contribute to wrapper height */
.table-overlay {
  position: absolute;   /* ← root cause */
  top: 0;
  left: 0;
  right: 0;
  pointer-events: none;
}
```

**The template (lines 110–158):**
```html
<!-- Wrapper with relative positioning -->
<div class="table-node-wrapper" class:viewing={!isEditing}>

  <!-- BaseNode renders a hidden textarea or view div -->
  <!-- In view mode with disableMarkdown:true, the view div is transparent -->
  <BaseNode ... />

  <!-- Table rendered as an absolute overlay -->
  {#if !isEditing && parsedTable}
    <div class="table-overlay">
      <table>...</table>
    </div>
  {/if}
</div>
```

**Why the overlay was used:** The `BaseNode` component (`base-node.svelte`) always renders either a `<textarea>` (edit mode) or a `<div class="node__content--view">` (view mode). For view mode, `TableNode` passes `disableMarkdown: true` via metadata and sets `displayContent` to `''` (empty string, line 90), making the view div invisible. The table overlay was then positioned on top. But since the overlay doesn't affect layout, the wrapper stays short.

## Proposed Solution

Replace the absolute-positioned overlay approach with an in-flow rendering strategy. The cleanest fix is to use the `customViewContent` snippet prop that `BaseNode` already supports (see `base-node.svelte` line 81 and lines 911–920):

```svelte
<!-- BaseNode already supports a customViewContent snippet -->
{#if customViewContent}
  {@render customViewContent()}
{:else}
  <ViewModeRenderer ... />
{/if}
```

**Approach:** Pass the rendered table as `customViewContent` to `BaseNode` instead of using the overlay. This renders the table inside the `.node__content--view` div — in normal document flow — so the wrapper naturally sizes to the table's height.

Changes needed in `table-node.svelte`:
1. Remove the `.table-overlay` div and its `{#if !isEditing && parsedTable}` block
2. Remove the `.table-overlay` CSS rule
3. Remove the `.table-node-wrapper.viewing :global(.node__content)` CSS hack (color: transparent, etc.)
4. Pass the rendered table as a Svelte snippet via the `customViewContent` prop on `BaseNode`
5. Keep `displayContent` and `disableMarkdown` handling as-is (or clean up since `customViewContent` takes precedence)

**Rough implementation sketch:**
```svelte
{#snippet tableViewContent()}
  {#if parsedTable}
    <table>
      <thead>
        <tr>
          {#each parsedTable.headers as header, i}
            <th style="text-align: {parsedTable.alignments[i] || 'left'}">
              <ViewModeRenderer content={header} />
            </th>
          {/each}
        </tr>
      </thead>
      <tbody>
        {#each parsedTable.rows as row}
          <tr>
            {#each row as cell, i}
              <td style="text-align: {parsedTable.alignments[i] || 'left'}">
                <ViewModeRenderer content={cell} />
              </td>
            {/each}
          </tr>
        {/each}
      </tbody>
    </table>
  {/if}
{/snippet}

<BaseNode
  ...
  customViewContent={!isEditing ? tableViewContent : undefined}
/>
```

## Acceptance Criteria

- [ ] Table node in display mode does not overlap or cover nodes below it
- [ ] The document layout correctly reserves space for the full rendered table height
- [ ] Clicking the table enters edit mode and shows raw GFM markdown (existing behavior preserved)
- [ ] Switching back from edit mode re-renders the table correctly in its natural position
- [ ] Column alignment (left/center/right), table styling, and cell content rendering are unchanged
- [ ] No other node type is affected by the change

### Files to Modify

- **TableNode**: `packages/desktop-app/src/lib/design/components/table-node.svelte`
  - Primary file for this fix — all changes should be confined here

### Key Reference Files (read but do not modify)

- **BaseNode**: `packages/desktop-app/src/lib/design/components/base-node.svelte`
  - Lines 81–82: `customViewContent` prop definition
  - Lines 911–920: Where `customViewContent` is rendered (inside `.node__content--view` div)
  - Lines 91–93: `isEditing` derived state (same pattern used in TableNode)

- **ViewModeRenderer**: `packages/desktop-app/src/lib/design/components/view-mode-renderer.svelte`
  - Used in table cells to render inline markdown (bold, italic, links) inside header/cell content
  - No changes needed

### Key Derived State in TableNode

| Variable | Line | Description |
|---|---|---|
| `isEditing` | 43 | `true` when this node's ID matches `focusManager.editingNodeId` |
| `parsedTable` | 87 | Parsed GFM table structure (headers, alignments, rows) |
| `displayContent` | 90 | Set to `''` in view mode — tells BaseNode to show nothing in its view div |
| `tableMetadata` | 38–40 | Sets `disableMarkdown: true` so BaseNode doesn't parse raw text |

### What NOT to change

- The `parseTable()` function (lines 56–84) — working correctly
- Event forwarding and dispatch logic (lines 92–107)
- Edit mode behavior — when `isEditing` is true, BaseNode handles everything normally

## Related Issues

- None

malibio:
Fixes #931

## Problem

In view mode, `TableNode` rendered the HTML table as an absolutely-positioned overlay on top of a hidden (transparent) raw text content area. Because `position: absolute` removes the overlay from normal document flow, the wrapper's height was determined only by the hidden text content — not by the actual rendered table. Multi-row tables extended beyond the wrapper bounds and visually covered nodes below them.

## Solution

Replaced the overlay approach with the `customViewContent` snippet prop that `BaseNode` already supports. The rendered table now lives inside the `.node__content--view` div in normal document flow, so the wrapper naturally sizes to the table's height.

Key changes in `table-node.svelte`:
- Removed `.table-overlay` div and its `{#if !isEditing}` conditional
- Removed `position: absolute` CSS for `.table-overlay`
- Removed `.table-node-wrapper.viewing :global(.node__content)` transparency hack
- Added `{#snippet tableViewContent()}` that renders the `<table>` inline
- Passed snippet as `customViewContent` to `BaseNode`
- Added `bind:clientHeight={tableViewHeight}` on the `<table>` element to measure rendered height
- Added `$effect` to apply the last-known table height to the textarea on edit-mode entry, preventing layout shift

### Also addressed (code review feedback from PR #918 review)

- `bind:clientHeight` placed directly on `<table>` element (no wrapper div)
- Restored `/* No outer borders, no vertical lines - clean minimal design */` comment above `tr:last-child td`

### Hook improvements (`.claude/settings.json` + `.claude/hooks/check-feature-branch.ts`)

Added a `PreToolUse` hook that warns (not blocks) when editing files while on `main` branch:
- Hook path uses portable relative path `.claude/hooks/check-feature-branch.ts`
- Exits 0 in all cases — surfaces a warning to the agent via stderr without hard-blocking

## Acceptance Criteria

- [x] Table node in display mode does not overlap or cover nodes below it
- [x] The document layout correctly reserves space for the full rendered table height
- [x] Clicking the table enters edit mode and shows raw GFM markdown
- [x] Switching back from edit mode re-renders the table correctly in its natural position
- [x] Column alignment (left/center/right), table styling, and cell content rendering are unchanged
- [x] No other node type is affected by the change
- [x] Tests pass (`bun run test`) — 3704 passed, 0 failures
- [x] Code passes `bun run quality:fix` — 0 errors, 0 warnings

## Skipped items (with reasoning)

**`isEditing` predicate — add `paneId` check**: `paneId` is obtained via `getContext()` inside `BaseNode` and is not propagated as a prop to wrapper nodes. `code-block-node.svelte` (the reference pattern) also only checks `editingNodeId`. This is an existing accepted pattern at the wrapper level. Track as a separate issue if stricter per-pane isolation is desired.

**Cross-component DOM query — use stable textarea ID**: The stable `textarea__{paneId}__{nodeId}` ID requires `paneId`, which is unavailable in `TableNode` for the same reason. The `wrapperElement` binding already scopes the query to this node's subtree. Track as a follow-up improvement.

## Test Results

- Frontend: 3704 passed (0 failures) — up from 3676 baseline at time of PR #918
- Quality: 0 ESLint errors, 0 svelte-check warnings, 0 Clippy warnings

### DIFF
--- a/.claude/hooks/check-feature-branch.ts
+++ b/.claude/hooks/check-feature-branch.ts
@@ -1,0 +1,44 @@
+#!/usr/bin/env bun
+/**
+ * PreToolUse hook: warns when editing files on the main branch.
+ *
+ * The startup sequence requires creating a feature branch before modifying
+ * any files. Being on `main` is a reliable signal that the sequence was skipped.
+ *
+ * Exits 0 in all cases (never blocks), but prints a warning to stderr when on
+ * `main` so Claude Code surfaces it as a message to the agent — prompting
+ * self-correction without hard-blocking legitimate operations.
+ */
+
+import { execSync } from 'child_process';
+
+function getCurrentBranch(): string {
+  try {
+    return execSync('git branch --show-current', { encoding: 'utf8' }).trim();
+  } catch {
+    // Not a git repo or git not available — allow through
+    return '';
+  }
+}
+
+const branch = getCurrentBranch();
+
+if (branch === 'main') {
+  process.stderr.write(`
+⚠️  WARNING: You are on the \`main\` branch.
+
+Have you completed the startup sequence before editing files?
+
+  1. git fetch origin && git pull origin main
+  2. bun install
+  3. bun run test  (record baseline)
+  4. bun run gh:comment <N> "Frontend: X passed"
+  5. git checkout -b feature/issue-<N>-description
+  6. bun run gh:assign <N> "@me"
+  7. bun run gh:status <N> "In Progress"
+
+If not, STOP and complete the sequence first, then create a feature branch.
+`);
+}
+
+process.exit(0);
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -4,5 +4,19 @@
       "Bash(gh:*)",
       "Bash(bun:*)"
     ]
+  },
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Edit|Write",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bun run .claude/hooks/check-feature-branch.ts",
+            "timeout": 5
+          }
+        ]
+      }
+    ]
   }
 }
--- a/packages/desktop-app/src/lib/design/components/table-node.svelte
+++ b/packages/desktop-app/src/lib/design/components/table-node.svelte
@@ -10,7 +10,7 @@
 -->
 
 <script lang="ts">
-  import { createEventDispatcher } from 'svelte';
+  import { createEventDispatcher, tick } from 'svelte';
   import BaseNode from './base-node.svelte';
   import ViewModeRenderer from './view-mode-renderer.svelte';
   import { focusManager } from '$lib/services/focus-manager.svelte';
@@ -86,8 +86,28 @@
   // Parse the table reactively
   let parsedTable = $derived(parseTable(content));
 
-  // Display content: empty in view mode (we render the table overlay)
-  let displayContent = $derived(isEditing ? content : '');
+  // Pass real content so adjustHeight() has multiline text to measure on edit entry.
+  // customViewContent overrides view rendering entirely — raw markdown is never shown.
+  let displayContent = $derived(content);
+
+  // Measured height of rendered table — updated reactively via bind:clientHeight
+  let tableViewHeight = $state(0);
+
+  // Ref to wrapper div for scoping DOM queries in the $effect
+  let wrapperElement = $state<HTMLDivElement | undefined>(undefined);
+
+  // When entering edit mode, apply the last-known table height to the textarea
+  // so there is no layout shift. After this, adjustHeight() takes over on each keystroke.
+  $effect(() => {
+    if (isEditing && tableViewHeight > 0) {
+      tick().then(() => {
+        const textarea = wrapperElement?.querySelector<HTMLTextAreaElement>('textarea');
+        if (textarea) {
+          textarea.style.height = `${tableViewHeight}px`;
+        }
+      });
+    }
+  });
 
   function handleContentChange(event: CustomEvent<{ content: string }>) {
     content = event.detail.content;
@@ -107,7 +127,34 @@
   }
 </script>
 
-<div class="table-node-wrapper" class:viewing={!isEditing}>
+{#snippet tableViewContent()}
+  {#if parsedTable}
+    <table bind:clientHeight={tableViewHeight}>
+        <thead>
+          <tr>
+            {#each parsedTable.headers as header, i}
+              <th style="text-align: {parsedTable.alignments[i] || 'left'}">
+                <ViewModeRenderer content={header} />
+              </th>
+            {/each}
+          </tr>
+        </thead>
+        <tbody>
+          {#each parsedTable.rows as row}
+            <tr>
+              {#each row as cell, i}
+                <td style="text-align: {parsedTable.alignments[i] || 'left'}">
+                  <ViewModeRenderer content={cell} />
+                </td>
+              {/each}
+            </tr>
+          {/each}
+        </tbody>
+    </table>
+  {/if}
+{/snippet}
+
+<div class="table-node-wrapper" bind:this={wrapperElement}>
   <BaseNode
     {nodeId}
     {nodeType}
@@ -117,6 +164,7 @@
     {children}
     {editableConfig}
     metadata={tableMetadata}
+    customViewContent={tableViewContent}
     on:createNewNode={handleCreateNewNode}
     on:contentChanged={handleContentChange}
     on:indentNode={forwardEvent('indentNode')}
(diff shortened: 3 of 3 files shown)


malibio:
## Code Review — Re-review of PR #934

**Overall verdict: Approve.** The addressed items are correctly implemented and the skipped items' rationale is sound. No new issues were introduced.

---

### Addressed Items — Verified

**`bind:clientHeight` on `<table>` directly (previously: wrapper div)**
Confirmed. `bind:clientHeight={tableViewHeight}` now sits on the `<table>` element at line 132. This is the correct element to measure — it directly reflects rendered table height with no intermediate wrapper to create discrepancies.

**Restored design comment above `tr:last-child td`**
Confirmed. `/* No outer borders, no vertical lines - clean minimal design */` is present at line 214, directly above the `tr:last-child td` rule. No gap between the comment and the rule it documents.

**Hook path portability**
Confirmed. `.claude/settings.json` uses `"command": "bun run .claude/hooks/check-feature-branch.ts"` — a relative path that works across machines. The previous hardcoded absolute path `/Users/malibio/...` has been removed.

**Hook enforcement: `process.exit(1)` → `process.exit(0)` + stderr warning**
Confirmed and correctly reasoned. The hook now exits 0 in all cases, writing to stderr only when on `main`. This correctly distinguishes between "agent needs a nudge" and "operation must be blocked". Hard-blocking every Edit/Write on main would break legitimate post-merge operations (`quality:fix`, file reviews). The soft enforcement is the right call.

---

### Skipped Items — Rationale Verified

**`isEditing` missing `paneId` check in `TableNode`**
The skip is justified. `BaseNode` derives `isEditing` at line 91–93 with the full `focusManager.editingNodeId === nodeId && focusManager.editingPaneId === paneId` check — including both node and pane. `TableNode`'s own `isEditing` at line 43 (`focusManager.editingNodeId === nodeId`) is used only to guard the `$effect` that seeds the textarea height on edit entry. This is a cosmetic/UX concern, not a correctness concern: the worst case is an unnecessary DOM query that finds no textarea (since BaseNode's gate correctly prevents edit mode in the wrong pane). `code-block-node.svelte` line 93 follows the same pattern. The deviation is an established accepted pattern at the wrapper level and is not a regression introduced here.

**Cross-component DOM query via `wrapperElement?.querySelector('textarea')`**
The skip is justified. The `paneId` required to construct the stable `textarea__{paneId}__{nodeId}` ID is obtained via `getContext()` inside `BaseNode` and is not exposed as a prop. The `wrapperElement` binding (line 157) scopes the query to this component's own subtree, so blast radius is contained to a single node instance. This is an improvement over a document-level query. The remaining concern (fragility to textarea structural changes) is real but proportionate — it warrants a separate issue, not a block here.

---

### New Code — No New Issues Found

The `$effect` at lines 101–110 deserves one observation:

**Nit:** The effect tracks both `isEditing` and `tableViewHeight` as reactive dependencies. Because `tableViewHeight` is updated by `bind:clientHeight` while the table is rendered (view mode), the effect will re-run on every resize that changes the table's height, not only on mode transitions. In practice the guard `if (isEditing && tableViewHeight > 0)` short-circuits immediately when `isEditing` is false, so there is no DOM mutation during view mode. The behavior is correct; the observation is informational.

The `displayContent = $derived(content)` change (line 91) — passing full markdown content rather than `''` — is sound: `customViewContent` overrides view rendering entirely (verified at `base-node.svelte` lines 911–920), so raw markdown is never visible to the user. Passing real content ensures `adjustHeight()` has actual multiline text to measure when edit mode is entered, which prevents layout shift. The comment explains this correctly.

The CSS change from `position: relative` + `pointer-events: none` to `width: 100%` removes all overlay plumbing cleanly. The `padding-block: 0.25rem` on `.node__content--view` via `:global` selector preserves the visual spacing the old `padding: 0.25rem 0` on `.table-overlay` provided.

---

### Summary

All five items from the previous review are accounted for: three addressed correctly, two skipped with documented rationale that holds up under scrutiny. The core architectural change — replacing the absolutely-positioned overlay with `customViewContent` to keep the table in normal document flow — is the correct approach and is cleanly implemented. No regressions introduced.

PR #935, opened 2026-03-14
Fix Mermaid rendering: inline SVG, theme-aware colors, pre-bundled chunks

malibio:
## Summary

- **Inline SVG rendering**: Replaced `` with `{@html svg}` rendered directly in the DOM. The previous approach failed in Tauri because: (1) `btoa()` throws on non-Latin1 Unicode characters in diagram labels, and (2) Tauri's webview blocked base64 data URIs in `` tags
- **Theme-aware colors**: Mermaid now re-initializes with `theme: 'base'` + `themeVariables` built from computed CSS custom properties (`--background`, `--foreground`, `--border`, `--primary`, etc.) before each render. Diagrams automatically match the app's light/dark theme and re-render when the theme toggles
- **Pre-bundled Mermaid chunks**: Added `mermaid` to `optimizeDeps.include` in `vite.config.js` so Vite collapses all of Mermaid's internal lazy-loaded sub-chunks (dagre, flowDiagram, etc.) into a single pre-bundled file, eliminating 404 errors in Tauri's webview

## Test plan

- [ ] Render a flowchart diagram in light mode — nodes/lines/text should use app colors
- [ ] Render a sequence diagram in light mode — actor boxes, arrows, labels should be visible
- [ ] Toggle to dark mode — diagram should re-render with dark palette (light text on dark backgrounds)
- [ ] Diagram with Unicode labels should render without errors
- [ ] `bun run test` passes (3704 tests)

## Notes

The `{@html}` usage triggers an ESLint `svelte/no-at-html-tags` warning. This is intentional and safe — the SVG string passes through `sanitizeSvg()` which strips all `<script>` tags, event handler attributes, and `javascript:` URIs before rendering.

### DIFF
--- a/packages/desktop-app/src/lib/services/mermaid-render.ts
+++ b/packages/desktop-app/src/lib/services/mermaid-render.ts
@@ -91,8 +91,8 @@ export function sanitizeSvg(svg: string): string {
   let result = svg.replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, '');
   // Remove event handler attributes (all quoting styles: "...", '...', or unquoted)
   result = result.replace(/\s+on\w+(\s*=\s*("[^"]*"|'[^']*'|[^\s>]*))?/gi, '');
-  // Remove javascript: URIs from attribute values (quoted and unquoted, with optional whitespace)
-  result = result.replace(/(["'\s])\s*javascript\s*:[^"'>]*/gi, '$1');
+  // Strip javascript: URIs from navigable/action attributes (mirrors data: URI treatment above)
+  result = result.replace(/(href|src|xlink:href|action|formaction)\s*=\s*["']\s*javascript\s*:[^"']*/gi, '$1=""');
   // Strip data: URIs from href, src, and xlink:href attributes (can carry executable payloads)
   result = result.replace(/(href|src|xlink:href)\s*=\s*["']\s*data:[^"']*/gi, '$1=""');
   // Strip url(javascript:...) from inline styles and <style> blocks


malibio:
### Re-Review: Commit f39cb18 — Replace `{@html}` with `bind:this + \$effect`

This is a follow-up review of the second commit pushed after the initial review. The first commit (`ac94ce67`) was reviewed previously. This review evaluates only the incremental change in `f39cb18`.

---

#### What Changed

`{@html mermaidSvg}` is replaced with a `bind:this` ref on an empty `<div>` and a `$effect` that writes `mermaidContainer.innerHTML = mermaidSvg ?? ''` when `mermaidSvg` changes. The commit message is accurate and the stated motivation — avoiding the `svelte/no-at-html-tags` lint warning without rule suppression — is achieved.

---

#### Assessment of the Approach

The security posture is **unchanged**. `{@html sanitizedString}` and `element.innerHTML = sanitizedString` are semantically identical in a browser: both parse the string as HTML and insert the result into the DOM. The linter cannot flag the latter because it is plain DOM manipulation outside Svelte's template compiler. This is a lint-tool workaround, not a security improvement. That characterization is fine — the sanitizer is already the actual defense, and the commit comment correctly documents that fact.

However, this workaround has a subtle Svelte 5 lifecycle issue that the original `{@html}` form did not have.

---

#### [Improvement] The `$effect` reads `mermaidSvg` but the dependency relationship is indirect — `code-block-node.svelte` lines 107–111

The `$effect` that writes `innerHTML` runs when `mermaidSvg` changes **and** when `mermaidContainer` first becomes non-null. The timing difference matters:

1. If `mermaidSvg` is already non-null when the component mounts (e.g. the node was pre-rendered and the component is re-created during HMR or a list reconcile), the `$effect` fires on mount with the correct value. This path works.

2. If `mermaidContainer` is reassigned to a new DOM node — for example, Svelte destroys and re-creates the `{#if mermaidSvg}` branch (which it will do when `mermaidSvg` transitions from `null` to a value and back, since the `{:else if mermaidError}` and `{:else}` branches render different elements) — the `$effect` must re-fire to repopulate `innerHTML` on the fresh element. Because `mermaidContainer` is `$state`, the effect correctly tracks it as a dependency and will re-run. This path also works.

3. The remaining edge case is that the `{#if mermaidSvg}` branch conditionally renders the `<div bind:this={mermaidContainer}>`. When `mermaidSvg` becomes non-null, Svelte creates the div and binds `mermaidContainer`, then the `$effect` runs and writes `innerHTML`. The concern is **ordering**: does Svelte guarantee that `bind:this` is committed before `$effect` runs in the same microtask? In Svelte 5, effects are batched and run after the DOM has been updated for the current tick, so `mermaidContainer` will be set before the effect body executes. This is safe.

The overall analysis is: the approach works correctly for the current code. The subtle risk is that this pattern creates a two-state dependency (`mermaidSvg` triggers both the conditional render and the effect) which is harder to reason about than `{@html}`. A future developer who splits the conditional or changes the binding may break the linkage without an obvious error.

**The original `{@html mermaidSvg}` form was actually the more idiomatic and robust Svelte pattern.** The lint rule exists to flag unsanitized injection; the correct resolution is to add a comment documenting the sanitization rationale (as noted in the original review's nit), not to route around it with `innerHTML`. The current workaround is functional but trades clarity for lint-tool compliance.

Given the CLAUDE.md policy of no lint suppression, the `innerHTML` approach is a valid path. But if the ESLint config supports inline disable comments (`<!-- svelte-ignore -->` or `// eslint-disable-next-line`), using `{@html}` with an explanatory disable comment is cleaner than this pattern. Check whether the project permits inline disable comments vs. rule-level suppression — they are different policies.

#### [Improvement] `mermaidContainer` should be typed as `HTMLDivElement | null`, not `HTMLDivElement | undefined` — line 103

Svelte's `bind:this` sets the ref to `null` when the element is removed from the DOM (when the `{#if mermaidSvg}` branch is false). Declaring the type as `| undefined` masks this: when `mermaidSvg` becomes null and the div is destroyed, `mermaidContainer` is set to `null`, but the `$effect` guard `if (mermaidContainer)` handles both `null` and `undefined` so this is not a runtime bug. The type annotation is just imprecise.

---

#### Nitpicks

- **Nit:** The CSS change removes `display: flex; justify-content: flex-start` from `.mermaid-output` and adds `display: block` to `.mermaid-output svg`. The flex layout was needed to prevent the SVG from stretching when inside a flex container. Verify that the SVG still aligns correctly (left-aligned, not centered) when the diagram is narrower than its container. The `display: block` on the SVG child is correct for removing the inline baseline gap.

- **Nit:** The `$effect` comment says "avoids `{@html}` and its lint warning" — this is accurate and useful. Consider noting that `sanitizeSvg()` is the actual security boundary and that this pattern is functionally equivalent, so a future reader does not assume `innerHTML` is somehow safer.

---

#### Outstanding Issues from Initial Review

The three issues flagged in the initial review remain open:

- **[Improvement]** Race condition in concurrent `renderMermaid` calls (`initialize` + `render` not atomic)
- **[Improvement]** SVG sanitizer gaps (CSS `url(javascript:...)`, under-anchored `javascript:` regex, unhandled `data:` URIs)
- **[Improvement]** `mermaid.initialize()` called on every render — theme-key caching would reduce cost to O(theme changes)

The second commit does not address any of these.

---

#### Summary

The `bind:this + $effect` pattern is functionally correct and achieves its stated goal. The three substantive issues from the initial review are unchanged. The new commit introduces no new critical issues. Review status remains **Request Changes** pending the race condition and sanitizer gaps.

PR #942, opened 2026-03-14
Mermaid diagram disappears after blurring from edit mode

### LINKED ISSUE #938: Mermaid diagram disappears after blurring from edit mode
## Overview

When a code block node is set to `mermaid` language, clicking away after editing causes the rendered diagram to disappear rather than re-render.

## Problem Statement

**User Experience:**
1. User views a mermaid code block — diagram renders correctly
2. User clicks the diagram — enters edit mode, mermaid syntax is shown
3. User clicks away (blur) — diagram disappears instead of re-rendering

**Expected:** After blur, the mermaid diagram should re-render and display correctly.

## Root Cause

**Location:** `packages/desktop-app/src/lib/design/components/code-block-node.svelte` — `$effect` render block (~lines 113–140)

The `$effect` that drives mermaid rendering skips execution when `isEditing` is true. When blur occurs and `isEditing` becomes false, the effect re-runs. However, the effect can fire multiple times in rapid succession (once for the `isEditing` change, once for derived content changes), causing the `renderSeq` stale-result guard to discard the in-flight render result. Since `mermaidSvgUri` is not preserved across the re-render attempt, if the result is discarded the diagram disappears with no fallback and no error shown.

## Proposed Solution

Fix the stale-result guard interaction so that:
- `mermaidSvgUri` is not cleared until a new successful render is available (preserve existing diagram during re-render)
- The error state is cleared at the start of a new render attempt, not on discard
- Only set `mermaidSvgUri = null` when the render definitively fails (not when discarded as stale)

## Acceptance Criteria
- [ ] Mermaid diagram re-renders correctly after clicking away from edit mode
- [ ] Diagram remains visible (not blank) during the async re-render on blur
- [ ] Error state still displays correctly when mermaid syntax is invalid
- [ ] No regression in stale-result protection (rapid language/content changes still handled correctly)

### Files to Modify
- **CodeBlockNode**: `packages/desktop-app/src/lib/design/components/code-block-node.svelte` — `$effect` render block (~lines 113–140)

### Key Logic
The fix should ensure `mermaidSvgUri` is only replaced (not cleared) when a valid new render arrives, and only set to `null` when the render definitively returns `null` (failure), not when discarded by the sequence guard.

malibio:
Closes #938

### DIFF
--- a/packages/desktop-app/src/lib/design/components/code-block-node.svelte
+++ b/packages/desktop-app/src/lib/design/components/code-block-node.svelte
@@ -128,13 +128,14 @@
 
     if (lang === 'mermaid') {
       highlightedLines = null;
-      mermaidError = null; // Clear error at start of new render attempt
+      mermaidError = null; // Reset stale error so it doesn't flash while new render is pending
       renderMermaid(code, nodeId, dark).then((svg) => {
-        if (seq !== renderSeq) return; // stale — language/content changed before render resolved
+        if (seq !== renderSeq) return; // stale — a newer render is already in-flight; preserve existing diagram
         if (svg !== null) {
-          mermaidSvg = svg; // Replace with new render
+          mermaidSvg = svg;
         } else {
-          mermaidSvg = null; // Only clear on definitive failure
+          // Definitive failure: render returned null, so there is no diagram to preserve
+          mermaidSvg = null;
           mermaidError = 'Diagram rendering failed. Check your Mermaid syntax.';
         }
       });


malibio:
## Code Review — PR #942

---

### Code Review Summary

This is a tightly scoped, single-commit bug fix that correctly addresses the root cause described in issue #938. The change modifies 7 lines of logic in the Mermaid rendering `$effect` inside `code-block-node.svelte`. The existing architecture, patterns, and conventions are fully respected. Overall this is a net positive change and a clean fix.

---

### Requirements Check

- ✅ Mermaid diagram re-renders correctly after clicking away from edit mode — the stale-result discard no longer clears `mermaidSvg`
- ✅ Diagram remains visible during the async re-render on blur — existing `mermaidSvg` is preserved while the new render is in-flight (i.e. `svg !== null` guard prevents clearing on discard)
- ✅ Error state still displays correctly on invalid syntax — `mermaidError` is still set when `renderMermaid` definitively returns `null`
- ✅ No regression in stale-result protection — the `seq !== renderSeq` guard is unchanged and still returns early on stale results
- ✅ Tests pass — existing test coverage for `renderMermaid` and `sanitizeSvg` is in place; no new test failures expected
- ✅ Code quality — no lint suppressions, no console.log, follows project standards

---

### Findings

#### No Critical Issues

The fix is logically sound. The three behavioral changes introduced are correct and cohesive:

1. `mermaidError = null` is moved to the top of the render attempt — clearing stale errors eagerly rather than only on successful resolve, which is the right UX (don't flash an old error while the new render is pending).
2. `mermaidSvg` is only replaced when `svg !== null` — preserving the previous valid diagram as a fallback during in-flight re-renders (the "flash of empty" is eliminated).
3. `mermaidSvg = null` is only reached on a definitive render failure, not on a stale-result discard — this is the core fix for the disappearing diagram.

#### Suggested Improvements

**`📁 packages/desktop-app/src/lib/design/components/code-block-node.svelte:134-140`**

The inline comment on line 135 (`// Replace with new render`) and line 137 (`// Only clear on definitive failure`) are correct but explain "what" rather than "why". The more useful information for a future reader is the distinction between "stale discard" and "definitive failure", which is already partially expressed in the surrounding guard. Consider improving them to:

```svelte
if (svg !== null) {
  mermaidSvg = svg;
} else {
  // Render returned null (definitive failure) — clear the diagram and show the error.
  // Do NOT clear mermaidSvg when the stale guard fires above;
  // that would blank the diagram while a subsequent render is still in-flight.
  mermaidSvg = null;
  mermaidError = 'Diagram rendering failed. Check your Mermaid syntax.';
}
```

This is a nit — the current comments are sufficient for the context.

#### Nitpicks

- **Nit** — `📁 code-block-node.svelte:137`: The comment `// Only clear on definitive failure` is accurate, but the prior comment on line 135 (`// Replace with new render`) is redundant — of course `mermaidSvg = svg` replaces the value. The "why" (preserve existing diagram during async re-render) is the load-bearing information missing from both comments. Not a blocker.

---

### Testing Assessment

The issue acceptance criterion "No regression in stale-result protection" is covered implicitly by the unchanged `seq !== renderSeq` guard, but there are no unit tests that exercise the **component-level** state transitions (preserve diagram on stale discard, clear only on failure). The existing `mermaid-render.test.ts` tests the service layer (`renderMermaid` returns null on failure, returns SVG on success) and is not affected by this change.

Given this is a bug fix for async UI state that is difficult to unit-test in Happy-DOM without a component harness, the absence of new tests is acceptable. A browser-mode test covering the blur-then-rerender scenario would provide the strongest regression protection here, but is not required to merge.

---

### Recommendation: APPROVE

The fix is minimal, correct, and directly addresses all acceptance criteria. The logic is easy to reason about and introduces no new complexity. The comment quality is marginally improvable but not a blocker.

PR #947, opened 2026-03-14
Improve semantic search ranking: replace raw chunk count with match density ratio

### LINKED ISSUE #944: Improve semantic search ranking: replace raw chunk count with match density ratio
## Problem

The current composite score formula rewards large documents structurally, regardless of how relevant they actually are:

```
composite_score = max_similarity × (1.0 + 0.3 × log10(matching_chunks))
```

A large document with 10 matching chunks out of 50 total beats a small, precise document with 2 matching chunks out of 2 total — even though the small doc is entirely relevant and the large doc is only 20% relevant. This is a **size bias**.

### Observed impact (2026-03-14 benchmark)

Q7 ("Patterns for managing focus state across nested components") — "Long-Term Vision" ranked #1 despite being irrelevant, because it's a large doc with many chunks that weakly mention components and state. The agent had to fetch 3 additional docs manually (6 tool calls, 128k tokens vs Grep's 5 calls, 135k).

The title fix from #936 didn't help here because the right docs ("Component Architecture", "Frontend Architecture") don't have title keywords that match "focus state" — so there was no title boost to overcome the size bias.

## Fix: Match Density Ratio

Replace `matching_chunks` (raw count) with `matching_chunks / total_chunks` (density ratio):

```
composite_score = max_similarity × (1.0 + 0.3 × (matching_chunks / total_chunks))
```

### Why this works

| Document | Max Sim | Matching | Total | Density | Old Score | New Score |
|----------|---------|----------|-------|---------|-----------|-----------|
| Long-Term Vision | 0.72 | 10 | 50 | 0.20 | 0.72 × 1.30 = **0.936** | 0.72 × 1.06 = **0.763** |
| Component Arch | 0.78 | 2 | 2 | 1.00 | 0.78 × 1.09 = **0.850** | 0.78 × 1.30 = **1.014** |

The precise doc now wins. A document only gets the full breadth boost if *most* of its chunks are relevant.

### Properties
- **2/2 match = same boost as 50/50** — fully relevant docs of any size score equally
- **10/50 match scores less than 2/2** — partial relevance penalized proportionally  
- **Formula stays in SQL** — `matching_chunks / total_chunks` is trivially computable alongside the existing `count()` grouping, just needs `max(total_chunks)` added to the GROUP BY aggregation
- **`total_chunks` already tracked** — `embedding` table has a `total_chunks` field per row

## Implementation

In `packages/core/src/db/surreal_store.rs` (~line 4187), the middle-outer level of the search query:

```sql
-- Current
LET $composite = $max_sim * (1.0 + 0.3 * math::log10($matching_chunks));

-- New
LET $density = $matching_chunks / $total_chunks;
LET $composite = $max_sim * (1.0 + 0.3 * $density);
```

Also add `max(total_chunks) AS total_chunks` to the GROUP BY aggregation in the middle-inner level.

The `0.3` weight constant (`BREADTH_BOOST`) stays — only the input signal changes from raw count to density ratio. Consider renaming it to `DENSITY_BOOST` for clarity.

## Acceptance Criteria

- [ ] Composite score uses `matching_chunks / total_chunks` instead of `log10(matching_chunks)`
- [ ] `total_chunks` aggregated correctly in the GROUP BY level of the search query
- [ ] Constant renamed from `BREADTH_BOOST` to `DENSITY_BOOST` (or similar) in `surreal_store.rs`
- [ ] Existing search ranking tests updated to reflect new formula
- [ ] Run benchmark against nodespace-docs — Q7 should improve, no regressions on Q1-Q6, Q8+

## Key Files

- `packages/core/src/db/surreal_store.rs` — composite score query (~line 4187)
- `packages/core/tests/embedding_service_test.rs` — ranking tests to update

## Related

- #936 (closed) — title embedding + keyword boost (already implemented)

malibio:
Closes #944

### DIFF
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -4178,7 +4178,8 @@ impl SurrealStore {
         // 3. Middle-outer: Calculate composite_score using density ratio (Issue #944)
         // 4. Outermost: Filter by composite_score > threshold (no duplication)
         //
-        // Formula: composite_score = max_similarity * (1.0 + 0.3 * (matching_chunks / total_chunks))
+        // Formula: composite_score = max_similarity * (1.0 + 0.3 * (<float>matching_chunks / total_chunks))
+        // The <float> cast prevents integer division (both matching_chunks and total_chunks are integers).
         // This replaces the old log10(matching_chunks) formula that rewarded large documents by size.
         // Density ratio ensures a doc only gets full boost if most of its chunks are relevant.
         let knn_limit = limit * 5;
@@ -4192,7 +4193,7 @@ impl SurrealStore {
                         max_similarity,
                         matching_chunks,
                         total_chunks,
-                        max_similarity * (1.0 + 0.3 * (matching_chunks / total_chunks)) AS composite_score
+                        max_similarity * (1.0 + 0.3 * (<float>matching_chunks / total_chunks)) AS composite_score
                     FROM (
                         SELECT
                             node,
--- a/packages/core/src/models/embedding.rs
+++ b/packages/core/src/models/embedding.rs
@@ -197,29 +197,31 @@ impl NewEmbedding {
 /// Result of a semantic search including similarity score
 ///
 /// The scoring algorithm considers both the best chunk similarity and the
-/// breadth of relevance (number of matching chunks). This ensures documents
-/// with multiple relevant sections rank higher than those with just one
-/// good match.
+/// match density ratio (matching chunks / total chunks). A document where
+/// all chunks matched the query ranks higher than a large document where
+/// only a few chunks matched, even if both have the same max similarity.
 ///
 /// ## Scoring Formula
 ///
 /// ```text
-/// score = max_similarity * (1 + BREADTH_BOOST * log10(matching_chunks))
+/// score = max_similarity * (1 + DENSITY_BOOST * (matching_chunks / total_chunks))
 /// ```
 ///
 /// Where:
 /// - `max_similarity`: Highest cosine similarity among all chunks
-/// - `matching_chunks`: Number of chunks exceeding the threshold
-/// - `BREADTH_BOOST`: Tuning parameter (default 0.3)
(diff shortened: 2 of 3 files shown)


malibio:
## Code Review Feedback — All 4 findings addressed

### 🔴 Fixed: Integer division in SQL formula
`packages/core/src/db/surreal_store.rs` — Added `<float>` cast to prevent integer truncation:
```sql
max_similarity * (1.0 + 0.3 * (<float>matching_chunks / total_chunks)) AS composite_score
```
Without the cast, `matching_chunks / total_chunks` would evaluate as integer division and always return 0 for partial matches (e.g., 2/10 → 0 instead of 0.2), making the density formula ineffective.

### 🟡 Fixed: Added partial-density test (`test_partial_density_ranks_below_full_density`)
`packages/core/tests/embedding_service_test.rs` — New test validates density < 1.0:
- **Doc A**: 2 embeddings, `total_chunks=2`, `max_sim≈0.80` → density=1.0 → score≈1.04
- **Doc B**: 2 embeddings, `total_chunks=10`, `max_sim≈0.85` → density=0.2 → score≈0.90

Doc A wins despite lower max_similarity because all its chunks matched (full density). The test asserts both the precise formula values and the ranking order.

### 🟢 Fixed: Stale doc-comment in `EmbeddingSearchResult`
`packages/core/src/models/embedding.rs` — Updated the struct-level doc-comment and the `score` field comment to reference the current density-ratio formula (`matching_chunks / total_chunks`) instead of the old `log10(matching_chunks)` formula. Updated the examples table to show density values explicitly.

### 🟢 Fixed: Weak assertion in `test_knn_search_with_multiple_chunks`
Replaced `results[0].score >= results[0].max_similarity * 0.99` (tautological — score is always ≥ max_sim) with a precise check: score must equal `max_similarity * 1.30` (± 0.01), which is the exact expected value for density=1.0. Also changed the `matching_chunks >= 1` check to `matching_chunks == 2`.

All 28 tests in `embedding_service_test` pass.

PR #952, opened 2026-03-14
Hybrid search: combine BM25 full-text + vector similarity for better-than-grep results

### LINKED ISSUE #951: Hybrid search: combine BM25 full-text + vector similarity for better-than-grep results
## Problem

Semantic search currently loses to grep on queries where the natural language phrasing doesn't align with the doc's vocabulary — even when the doc IS the correct answer.

**Root cause**: Vector embeddings summarize an entire document into a single point in space. A query like *"pattern for persisting node changes to the database"* doesn't match "Frontend State & Persistence" well because the doc uses technical vocabulary (`SharedNodeStore`, `SimplePersistenceCoordinator`, `debounce`) rather than the plain-language terms in the query. Grep wins here because it finds exact text anywhere in the file.

**Benchmark evidence** (2026-03-14, 25 questions vs nodespace-docs):
- NodeSpace wins on conceptual/architectural questions (40% fewer tokens overall)
- Grep wins on questions where query phrasing doesn't match doc vocabulary
- Even after title embedding (#936) and density ratio (#944), some queries still surface wrong docs

**The ceiling**: Semantic-only search cannot consistently beat grep because grep has perfect recall for exact terms. To beat grep, we need to combine both signals.

## Solution: Intersection-Based Hybrid Search

The key insight: **agreement between two independent signals is a strong confidence indicator**.

### Algorithm

1. **BM25 on `node.content`** — keyword search across ALL nodes (not just roots), finds nodes at any depth whose text contains query terms
2. **Walk up to root** — for each BM25-matching node, resolve its root node ID (via `has_child` relationship chain)
3. **KNN vector search on `embedding`** — semantic search on root node embeddings as today
4. **Score by intersection**:

```
both BM25 + KNN agree on root  → highest confidence → ranked first
KNN only (no BM25 match)       → semantic relevance only → ranked second
BM25 only (no KNN match)       → keyword match only → ranked third
```

Within each tier, rank by composite score (existing density formula from #944).

### Why intersection beats simple score blending

- **No normalization needed** — BM25 and vector scores are on different scales; blending them requires fragile normalization. Intersection avoids this entirely.
- **Deep child coverage** — a query term buried in a child node 3 levels deep surfaces the root via BM25, which the vector search might miss because the term was diluted across the full subtree embedding.
- **Confidence multiplier** — two independent signals agreeing is stronger evidence than one strong signal alone.
- **No schema changes** — BM25 index goes on the existing `node.content` field. The `embedding` table doesn't need to store raw text.

### SurrealDB BM25 Index

```sql
DEFINE ANALYZER nodespace_analyzer TOKENIZERS class FILTERS lowercase, snowball(english);
DEFINE INDEX IF NOT EXISTS idx_node_content_bm25 
    ON TABLE node FIELDS content 
    SEARCH ANALYZER nodespace_analyzer BM25(1.2, 0.75);
```

Standard BM25 parameters: `k1=1.2` (term frequency saturation), `b=0.75` (length normalization).

### Query Flow

```
Query: "pattern for persisting node changes to database"
         │
         ├─► BM25 on node.content ──► matching node IDs ──► resolve roots ──► root set A
         │
         └─► KNN on embedding ────────────────────────────────────────────► root set B
                                                                              │
                                              intersection(A, B) ranked first ┘
                                              B only ranked second
                                              A only ranked third
```

## Performance Characteristics

- **BM25 query**: O(log n) per term via inverted index — sub-millisecond at NodeSpace scale
- **KNN query**: existing ~50-100ms via HNSW — unchanged
- **Parallel execution**: both queries run concurrently → total latency = `max(bm25, knn)` ≈ same as today
- **Index maintenance**: BM25 index updates on every node write — amortized by existing 500ms debounce, negligible overhead
- **RRF merging**: merging two lists of ~20 items in Rust — microseconds

### 1. Define BM25 index

In `surreal_store.rs` schema setup (or `index_manager.rs`):
```sql
DEFINE ANALYZER nodespace_analyzer TOKENIZERS class FILTERS lowercase, snowball(english);
DEFINE INDEX IF NOT EXISTS idx_node_content_bm25 ON TABLE node 
    FIELDS content SEARCH ANALYZER nodespace_analyzer BM25(1.2, 0.75);
```

### 2. Add `bm25_search_roots()` to `surreal_store.rs`

New method that:
- Runs BM25 query on `node.content`
- Walks each result up to its root via `has_child` relationships
- Returns `HashSet<root_node_id>`

### 3. Modify `semantic_search()` in `surreal_store.rs`

Run BM25 and KNN in parallel (`tokio::join!`), then apply intersection scoring:
- Roots in both sets: tier 1
- Roots in KNN only: tier 2  
- Roots in BM25 only: tier 3
- Within each tier: sort by existing composite score

### 4. Update `search_semantic` MCP tool (optional)

Add `hybrid: bool` param (default: `true`) to allow pure-vector fallback if needed.

### Unit tests (update existing)
- `embedding_service_test.rs` — update ranking tests to reflect hybrid scoring tiers
- Add test: query term present only in child node → root surfaces via BM25 path
- Add test: conceptual query with no keyword match → falls back to KNN-only tier correctly
- Add test: both signals agree → intersection tier ranks above KNN-only results

### Integration tests
- `test:db` suite — verify BM25 index created on startup
- Add test: hybrid search returns correct root when query term is in a grandchild node

### Performance baseline (before/after)
Record the following metrics **before** implementing hybrid search, then again **after**:

| Metric | Tool | How |
|--------|------|-----|
| Search latency (p50, p95) | Rust bench | `cargo bench` on `semantic_search()` with 100/1000/5000 node corpus |
| Index write overhead | Rust bench | Node insert time with/without BM25 index |
| Benchmark score (tokens/cost vs grep) | `run-benchmark.sh` | Run against nodespace-docs corpus |

Commit the before numbers as a comment on this issue before starting implementation.

## Key Files

- `packages/core/src/db/surreal_store.rs` — BM25 index definition + `search_roots()` query modification
- `packages/core/src/services/embedding_service.rs` — `semantic_search()` parallel execution
- `packages/core/src/mcp/handlers/search.rs` — optional `hybrid` param
- `packages/core/src/db/index_manager.rs` — currently stubbed, consider activating for index setup
- `packages/core/tests/embedding_service_test.rs` — unit tests to update

## Related

- #936 (closed) — title embedding + keyword boost
- #944 (closed) — density ratio scoring
- ADR-026 — semantic search ranking formula evolution
- nodespace-semantic-bench — benchmark scripts for before/after validation

malibio:
Closes #951

### DIFF
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -4326,6 +4326,21 @@ impl SurrealStore {
         //
         // Use a single query to find which of these nodes are roots vs children,
         // and for children, get their immediate parent. We repeat until all are resolved.
+        #[derive(Debug, serde::Deserialize, surrealdb::types::SurrealValue)]
+        struct ParentResult {
+            node_id: String,
+            parent_id: String,
+        }
+
+        let sql_parents = r#"
+            SELECT
+                meta::id(out) AS node_id,
+                meta::id(in) AS parent_id
+            FROM relationship
+            WHERE out IN $node_ids
+              AND relationship_type = 'has_child';
+        "#;
+
         let mut root_ids: std::collections::HashSet<String> = std::collections::HashSet::new();
         let mut pending: Vec<String> = matching_ids;
 
@@ -4336,16 +4351,7 @@ impl SurrealStore {
             }
 
             // Batch query: for each pending node, find its parent (if any)
-            // Returns (node_id, parent_id_or_null) for all pending nodes
-            let sql_parents = r#"
-                SELECT
-                    meta::id(out) AS node_id,
-                    meta::id(in) AS parent_id
-                FROM relationship
-                WHERE out IN $node_ids
-                  AND relationship_type = 'has_child';
-            "#;
-
+            // Returns (node_id, parent_id) for all pending nodes that have a parent
             // Build SurrealDB record IDs for the query
             let node_things: Vec<surrealdb::types::RecordId> =
                 pending.iter().map(|id| node_record_id(id)).collect();
@@ -4356,12 +4362,6 @@ impl SurrealStore {
                 .bind(("node_ids", node_things))
                 .await
                 .context("Failed to query parent relationships for BM25 root resolution")?;
-
-            #[derive(Debug, serde::Deserialize, surrealdb::types::SurrealValue)]
-            struct ParentResult {
-                node_id: String,
-                parent_id: String,
-            }
 
             let parent_results: Vec<ParentResult> = parent_response
                 .take(0)
@@ -4385,9 +4385,10 @@ impl SurrealStore {
                 }
             }
 
-            pending = next_pending;
-            // Deduplicate pending to avoid redundant lookups
-            pending.dedup();
+            // Deduplicate via HashSet to avoid redundant lookups regardless of order
+            let next_pending_set: std::collections::HashSet<String> =
+                next_pending.into_iter().collect();
+            pending = next_pending_set.into_iter().collect();
         }
 
         // Any remaining pending nodes have exceeded max depth - treat as roots (best effort)
--- a/packages/core/src/services/embedding_service.rs
+++ b/packages/core/src/services/embedding_service.rs
@@ -743,13 +743,17 @@ impl NodeEmbeddingService {
         results.extend(tier1);
         results.extend(tier2);
 
+        let mut tier3_count = 0usize;
         if knn_count < limit {
             let remaining = limit - knn_count;
-            let bm25_only_roots: Vec<String> = bm25_roots
+            // Sort by node_id for deterministic ordering within Tier 3
+            // (no embedding score available to rank by, so stable key prevents non-reproducible results)
+            let mut bm25_only_roots: Vec<String> = bm25_roots
                 .into_iter()
                 .filter(|id| !knn_node_ids.contains(id))
-                .take(remaining)
                 .collect();
+            bm25_only_roots.sort();
+            bm25_only_roots.truncate(remaining);
 
             if !bm25_only_roots.is_empty() {
                 // Fetch node data for BM25-only results (no embedding score available)
@@ -763,14 +767,14 @@ impl NodeEmbeddingService {
                             matching_chunks: 0,
                             node: Some(node),
                         });
+                        tier3_count += 1;
                     }
                 }
             }
         }
 
         let total_time = total_start.elapsed();
-        let bm25_only_count = results.iter().filter(|r| r.score == 0.0).count();
-        let intersection_count = results.len() - bm25_only_count - tier2_count;
+        let intersection_count = results.len() - tier3_count - tier2_count;
 
         tracing::debug!(
             "HYBRID SEARCH PROFILE: total={:?} | embedding={:?} search={:?} | results={} (tier1={} tier2={} tier3={}) query='{}'",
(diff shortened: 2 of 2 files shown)


malibio:
## Re-Review: Hybrid BM25 + KNN Search (#951)

**Review Type**: Re-Review

The only commit since the initial review is a benchmark correctness fix (moving corpus setup outside the Criterion timed loop — correct and good). The three items flagged as requiring changes in the initial review remain unaddressed in the current code.

---

## Status of Initial Review Findings

### [Improvement] `pending.dedup()` — NOT FIXED
**📁 packages/core/src/db/surreal_store.rs:4388-4390**

`Vec::dedup()` still used on `next_pending`. As noted in the initial review, `dedup()` only removes *consecutive* duplicates. When multiple BM25-matching nodes share a common ancestor, those ancestor IDs in `next_pending` will not be adjacent, so `dedup()` silently misses them. The result is repeated DB round-trips for the same already-resolved parent node on subsequent loop iterations.

The fix remains straightforward: collect `next_pending` into a `HashSet<String>`, then convert back to `Vec` before assigning to `pending`. This is consistent with the `root_ids` and `has_parent` structures already used in the same function.

### [Improvement] Tier 3 BM25-only ordering is non-deterministic — NOT FIXED
**📁 packages/core/src/services/embedding_service.rs:748-752**

`bm25_roots.into_iter()` still iterates a `HashSet<String>`, so the `.take(remaining)` call selects an arbitrary subset across runs. Result order in Tier 3 is non-reproducible. A `.sorted()` call (e.g., by node ID as a stable secondary key) before `.take()` would make results deterministic. This matters both for user-facing consistency and for benchmark reproducibility.

### [Improvement] `score == 0.0` float equality in log counter — NOT FIXED
**📁 packages/core/src/services/embedding_service.rs:772**

```rust
let bm25_only_count = results.iter().filter(|r| r.score == 0.0).count();
```

Still retroactively infers Tier 3 count via float equality rather than tracking it at construction time. The correct fix is a `let mut tier3_count = 0usize;` counter incremented inside the `results.push(...)` loop (line 759-766), replacing the post-hoc filter. This eliminates the structural assumption that no KNN result will ever land at exactly 0.0.

---

## Benchmark Fix (new commit) — Correct
**📁 packages/core/benches/performance.rs**

The corpus setup correctly moved outside the `b.iter()` loop. The benchmark now measures only `bm25_search_roots()` latency as intended. The reported numbers (100 nodes ~706µs, 500 nodes ~3.7ms) are consistent with expectations for a BM25 inverted-index lookup + iterative parent walk.

---

## Requirements Check (unchanged from initial review)

- ✅ BM25 index with snowball stemming on `node.content`
- ✅ `bm25_search_roots()` in `surreal_store.rs` with root resolution
- ✅ Parallel BM25 + KNN via `tokio::join!`
- ✅ Intersection-based tiering (Tier 1/2/3)
- ✅ Within-tier sorting by composite score for Tiers 1 and 2
- ✅ 7 integration tests covering core scenarios
- ✅ Criterion benchmark with corrected corpus setup
- ⚠️ Tier 3 ordering remains non-deterministic (HashSet iteration)

---

## Summary

The algorithm design and overall structure are sound. The three items from the initial REQUEST CHANGES remain open:

1. **`pending.dedup()` correctness** — Primary fix: use `HashSet`-based dedup on `next_pending` to prevent redundant DB round-trips.
2. **Tier 3 non-deterministic ordering** — Apply a stable sort before `.take()` on `bm25_only_roots`.
3. **`score == 0.0` log counter** — Track Tier 3 count at construction; remove retroactive float equality check.

All three can be addressed in a single small commit. The benchmark fix from this round is correct and does not need changes.

**Recommendation**: REQUEST CHANGES — same items as initial review, none addressed yet.

PR #956, opened 2026-03-14
Restore dev-proxy HTTP client mode (broken by #927)

### LINKED ISSUE #955: Restore dev-proxy HTTP client mode (broken by #927)
## Problem

PR #927 (SurrealDB 3.x upgrade) switched `dev-proxy` from HTTP client mode to embedded RocksDB mode to work around an HTTP session state bug in the SurrealDB 2.x SDK. This broke the intended dev workflow:

**Before #927:**
- `surreal start` runs on port 8000
- dev-proxy connects to it via HTTP (`HttpStore::new_http`)
- CLI, Surrealist, and dev-proxy all share the same server
- `REBUILD INDEX`, `INFO FOR TABLE`, etc. all work via CLI

**After #927:**
- dev-proxy opens RocksDB directly (exclusive file lock)
- `surreal start` in `demo:browser` script is unused/misleading
- CLI cannot connect while proxy is running
- No way to inspect or rebuild indexes via CLI

## Root Cause

The HTTP session state bug was a SurrealDB 2.x SDK issue. We are now on SurrealDB 3.x where this is no longer a problem.

## Fix

Restore `dev-proxy` to HTTP client mode:
- `SurrealStore::new_http("127.0.0.1:8000", ...)` instead of `SurrealStore::new(db_path)`
- `NODESPACE_DEV_DB_PATH` env var becomes unused (proxy connects to server, not file path)
- `demo:browser` script already starts `surreal start` on port 8000 — proxy just needs to connect to it

## Acceptance Criteria
- [ ] `bun run demo:browser` starts SurrealDB server + proxy + frontend
- [ ] dev-proxy connects to SurrealDB server via HTTP (not embedded RocksDB)
- [ ] CLI can connect to port 8000 while proxy is running
- [ ] `REBUILD INDEX` and `INFO FOR TABLE` work via CLI
- [ ] All existing proxy functionality works (MCP, SSE, node CRUD)

malibio:
Closes #955

### DIFF
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -1,7 +1,12 @@
 //! SurrealStore - Direct SurrealDB Backend Implementation
 //!
-//! This module provides the primary and only database backend for NodeSpace,
-//! using SurrealDB embedded database with RocksDB storage engine.
+//! This module provides the primary database backend for NodeSpace.
+//! Two connection modes are supported via the `Surreal<Any>` dynamic engine:
+//!
+//! - **Embedded RocksDB** (`SurrealStore::new`): Desktop production (Tauri app).
+//!   Holds an exclusive file lock — only one process at a time.
+//! - **HTTP client** (`SurrealStore::new_http`): Browser development mode (dev-proxy).
+//!   Connects to a running `surreal start` server; multiple clients can share the DB.
 //!
 //! # Architecture
 //!
@@ -12,7 +17,7 @@
 //!
 //! # Design Principles
 //!
-//! 1. **Embedded RocksDB**: Desktop-only backend using `kv-rocksdb` engine
+//! 1. **Dynamic engine (`Surreal<Any>`)**: Same struct works with RocksDB embedded and HTTP remote
 //! 2. **SCHEMAFULL + FLEXIBLE**: Core fields strictly typed, user extensions allowed
 //! 3. **Record IDs**: Native SurrealDB format `node:uuid` (type embedded in ID)
 //! 4. **Universal Storage**: All properties embedded in `node.properties` field
@@ -294,7 +299,9 @@ impl SurrealStore {
     /// ```
     pub async fn new(db_path: PathBuf) -> Result<Self> {
         // Initialize embedded RocksDB via the Any engine (supports both embedded and HTTP).
-        // RocksDb path must be expressed as a rocksdb:// URL for engine::any::connect.
+        // Path must be expressed as a rocksdb:// URL for engine::any::connect.
+        // Note: PathBuf::display() uses OS-native separators. This is macOS/Linux only;
+        // backslashes on Windows would produce an invalid URL. Desktop targets are Unix.
         let url = format!("rocksdb://{}", db_path.display());
         let db: Surreal<Any> = surrealdb::engine::any::connect(url)
             .await
--- a/packages/dev-tools/src/bin/dev-mcp.rs
+++ b/packages/dev-tools/src/bin/dev-mcp.rs
@@ -1,15 +1,18 @@
 //! Development MCP Server for Browser Mode
 //!
-//! This is a standalone MCP (Model Context Protocol) server that enables AI agent
-//! integration during browser mode development. It runs alongside dev-proxy to
-//! provide the same MCP functionality available in the Tauri desktop app.
+//! **NOTE**: `dev-proxy` now includes an integrated MCP server (port 3100) and is the
+//! recommended dev tool. This standalone binary uses embedded RocksDB and holds an
+//! exclusive file lock, which conflicts with `dev-proxy` when both are run together.
+//! See issue #955 / PR #956 for context. This binary should be removed or migrated
+//! to HTTP client mode in a follow-up.
 //!
-//! Architecture:
+//! Architecture (current — uses embedded RocksDB, conflicts with dev-proxy):
 //!   AI Agent (Claude Code) → HTTP (port 3100) → MCP Server → NodeService → embedded SurrealDB
-//!                                                                                    ↓
-//!   Frontend             → HTTP (port 3001) → dev-proxy → NodeService → embedded SurrealDB
 //!
-//! Both servers share the same NODESPACE_DEV_DB_PATH database.
+//! Architecture (dev-proxy — preferred):
+//!   AI Agent (Claude Code) → HTTP (port 3100) → dev-proxy MCP → NodeService → SurrealDB HTTP server
+//!                                                                                    ↓
+//!   Frontend             → HTTP (port 3001) → dev-proxy → NodeService → SurrealDB HTTP server
 //!
 //! # Key Features
 //!
--- a/packages/dev-tools/src/bin/dev-proxy.rs
+++ b/packages/dev-tools/src/bin/dev-proxy.rs
@@ -357,16 +357,21 @@ async fn main() -> anyhow::Result<()> {
 
     println!("🔧 Initializing dev-proxy...");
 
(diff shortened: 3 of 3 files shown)


malibio:
## Code Review — PR #956: Restore dev-proxy HTTP client mode

**Overall Assessment**: This is a correct, well-scoped bug fix that directly addresses the architectural regression introduced by PR #927. The core change — migrating `SurrealStore` from `Surreal<Db>` to `Surreal<Any>` and adding a `new_http()` constructor — is the right approach and is implemented cleanly. The change is a net positive and resolves a real developer-experience problem (exclusive RocksDB file lock blocking CLI tools). I have a few findings to call out before merge.

---

### Findings

#### [Improvement] Hardcoded credentials in `new_http()` call — no env-var escape hatch

**File**: `packages/dev-tools/src/bin/dev-proxy.rs`, line 369

```rust
let mut store = match SurrealStore::new_http(&db_host, "root", "root").await {
```

The username and password are hardcoded to `"root"/"root"`. While `NODESPACE_DEV_DB_HOST` is configurable, the credentials are not. This is intentional for a dev-only tool, but the `surreal start` scripts in `package.json` also hardcode `--user root --pass root`. The two are consistent today, but they're coupled without an obvious link. Adding `NODESPACE_DEV_DB_USER` / `NODESPACE_DEV_DB_PASS` env vars (defaulting to `"root"/"root"`) would mirror the pattern already used for the host override and make it trivial to use non-default credentials without recompiling. This is not a blocker since this is a dev-only binary, but the inconsistency with the host being configurable is notable.

---

#### [Improvement] Module-level doc comment is stale after the engine change

**File**: `packages/core/src/db/surreal_store.rs`, lines 1–19

The crate-level doc still reads:

> "using SurrealDB embedded database with RocksDB storage engine"

and under Design Principles:

> "1. **Embedded RocksDB**: Desktop-only backend using `kv-rocksdb` engine"

These are now inaccurate — `SurrealStore` supports both embedded and HTTP backends. The struct-level doc (line ~236) was correctly updated, but the module header was missed. Stale module docs are a readability debt that accumulates quickly.

---

#### [Improvement] `dev-proxy:prod` and `dev-proxy:demo` scripts are now identical and therefore confusing

**File**: `packages/desktop-app/package.json`, lines ~13–14

```json
"dev:browser:proxy:prod": "cargo run -p nodespace-dev-tools --bin dev-proxy",
"dev:browser:proxy:demo": "cargo run -p nodespace-dev-tools --bin dev-proxy",
```

After removing the `NODESPACE_DEV_DB_PATH` prefix, both scripts are byte-for-byte identical. A developer reading these will reasonably wonder if one was mistakenly copied over the other. The differentiation now happens entirely in the `:db:prod` vs `:db:demo` SurrealDB server scripts that run in parallel, but that relationship isn't visible here. A comment or at least consolidation into a single `dev:browser:proxy` target (already identical) would reduce confusion.

---

#### [Improvement] `dev-mcp.rs` is an orphaned embedded-mode tool with stale architecture comment

**File**: `packages/dev-tools/src/bin/dev-mcp.rs`, lines 8–12

`dev-mcp.rs` still opens RocksDB directly via `SurrealStore::new(db_path)` and its module comment says:

> "Both servers share the same NODESPACE_DEV_DB_PATH database."

After this PR, `dev-proxy` no longer uses `NODESPACE_DEV_DB_PATH` or opens RocksDB directly, so `dev-mcp` and `dev-proxy` no longer share a database and the comment is incorrect. Additionally, `dev-mcp` would hold an exclusive RocksDB lock on whatever path it's pointed at, which would prevent `dev-proxy` from connecting to the same data through SurrealDB server simultaneously (unless pointed at different paths). This doesn't block the PR but the two tools are now architecturally inconsistent. Consider updating `dev-mcp` to also use HTTP mode in a follow-up.

---

#### [Nit] Minor — `to_string()` on string literals in `Root { ... }` 

**File**: `packages/core/src/db/surreal_store.rs`, line 349

```rust
db.signin(Root { username: user.to_string(), password: pass.to_string() })
```

`user` and `pass` are already `&str`. Depending on whether the `Root` struct accepts `impl Into<String>` or `String` directly (the surrealdb 3.x `Root` struct fields are `&str`-based), this may produce an unnecessary allocation. The equivalent `Root { username: user, password: pass }` may be valid. Low priority — compiler will catch it if the types mismatch — but worth checking.

---

#### [Nit] `rocksdb://` URL formation via `PathBuf::display()` on Windows

**File**: `packages/core/src/db/surreal_store.rs`, line 298

```rust
let url = format!("rocksdb://{}", db_path.display());
```

`PathBuf::display()` uses the OS-native path separator. On Windows this produces `rocksdb://C:\Users\...`, which is not a valid URL. NodeSpace is macOS/Linux only today, so this is not a current issue, but a comment noting the assumption (or using `db_path.to_string_lossy().replace('\\', "/")`) would future-proof the code. Leaving as a nit given the current target platform.

---

| Criterion | Status |
|---|---|
| `bun run demo:browser` starts SurrealDB + proxy + frontend | Addressed — `demo-db` path fix and proxy now connects via HTTP |
| dev-proxy connects via HTTP (not embedded RocksDB) | Addressed — `new_http()` constructor used |
| CLI can connect to port 8000 while proxy is running | Addressed — no exclusive lock held by proxy |
| `REBUILD INDEX` / `INFO FOR TABLE` work via CLI | Follows from above |
| All existing proxy functionality works (MCP, SSE, node CRUD) | No existing proxy handler logic changed |

---

### Summary

The change is correct and well-reasoned. The `Surreal<Any>` migration is the cleanest possible approach — it avoids maintaining two separate `SurrealStore` types while unlocking both embedded and HTTP connection modes via the URL scheme. The `demo-db` path correction is a sensible cleanup. The most actionable items are the stale module-level doc comment and the `dev-mcp.rs` architectural drift, both of which are good candidates for a quick follow-up issue.

**Recommendation**: Approve with the noted improvements tracked as follow-up issues.

malibio (commented):
Correct fix for the RocksDB exclusive-lock regression introduced in #927. The Surreal<Any> migration is the right approach. See the detailed review comment for follow-up items (stale module doc, dev-mcp architectural drift, identical proxy variant scripts) — none are blockers.

malibio:
## Review Feedback Addressed

### ✅ Addressed (4 items)

**Stale module-level doc** (`surreal_store.rs`):
Updated the crate header and Design Principle #1 to reflect the `Surreal<Any>` dynamic engine approach supporting both embedded RocksDB and HTTP connections.

**Credential env-var override** (`dev-proxy.rs`):
Added `NODESPACE_DEV_DB_USER` and `NODESPACE_DEV_DB_PASS` env vars (defaulting to `"root"`/`"root"`), making all three connection parameters configurable consistently with `NODESPACE_DEV_DB_HOST`.

**`dev-mcp.rs` architectural inconsistency**:
Updated the module doc to clearly flag that `dev-mcp` still uses embedded RocksDB (conflicts with the new HTTP-mode dev-proxy), references issue #955/PR #956, and notes it should be removed or migrated in a follow-up.

**`PathBuf::display()` on Windows** (`surreal_store.rs`):
Added a comment noting the Unix-only assumption.

### ⏭️ Skipped (2 items)

**Identical proxy scripts (`dev:browser:proxy:prod` and `:demo`)**: JSON has no comment syntax, so there's no good way to annotate the intent inline. The scripts are intentionally identical — the distinction is entirely at the DB-server level, which is visible from the `concurrently` commands that pair `db:prod` + `proxy:prod`. No change needed.

**`Root` struct `.to_string()` nit**: The reviewer noted `Root` fields might be `&str`. Checked the surrealdb 3.0.4 source — both `username` and `password` are `String`, so the `.to_string()` calls are required (not unnecessary allocations). No change needed.

PR #958, opened 2026-03-15
fix: BM25 tokenized OR search + single-query root resolution for hybrid search

malibio:
## Summary

Fixes hybrid search producing \`tier1=0\` (BM25 and KNN never agreeing) by replacing AND-semantics BM25 with tokenized OR + score ranking, and replacing the iterative Rust root resolution loop with a single SurrealDB graph traversal.

### Root Cause

\`content @@ \$query\` uses AND semantics — all tokens must appear in the **same node**. For multi-word queries this misses documents where terms are spread across different child nodes. E.g. "keyboard navigation" in a header and "focus management" in a paragraph — both under the same root — never match as a 4-word AND query.

### Changes

**\`packages/core/src/db/surreal_store.rs\`**
- Tokenize query, strip stop words, cap at 4 tokens
- Build per-term OR: \`(content @@ \$t0 OR content @@ \$t1 OR ...)\`
- Rank by combined \`search::score(0) + search::score(1) + ...\` DESC
- Replace iterative root resolution loop (up to 50 DB round-trips) with single \`{..+collect}<-relationship\` graph traversal

**\`packages/core/src/services/embedding_service.rs\`**
- Reduce \`bm25_limit\` from 5x to 2x (ranked OR needs fewer candidates)

### Results

| | Time | tier1 | tier2 |
|--|------|-------|-------|
| Before (AND) | ~145ms | 0 | 15 |
| After (OR + graph traversal) | ~142ms | 7 | 8 |

Same latency, hybrid search now actually working.

## Test plan

- [x] 117 frontend test files pass (3705 tests)
- [x] 812 Rust unit tests pass
- [x] tier1=7 (was 0) for 4-word query
- [x] Latency unchanged ~142ms

### DIFF
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -79,6 +79,24 @@ fn node_record_id(id: &str) -> RecordId {
 /// the current state, not historical events.
 const DOMAIN_EVENT_CHANNEL_CAPACITY: usize = 128;
 
+/// Maximum number of BM25 query tokens to use in the OR fulltext search (Issue #957).
+/// Each additional OR term costs ~20ms on the fulltext index.
+const BM25_MAX_TOKENS: usize = 4;
+
+/// Stop words stripped from BM25 queries before tokenization (Issue #957).
+/// These appear in nearly every document with similar frequency and add no
+/// discriminative power for ranking. Shared with the title-boost tokenizer
+/// in embedding_service.rs to keep both paths consistent.
+const BM25_STOP_WORDS: &[&str] = &[
+    "a", "an", "the", "is", "are", "was", "were", "be", "been", "being",
+    "have", "has", "had", "do", "does", "did", "will", "would", "could",
+    "should", "may", "might", "shall", "can", "need", "dare", "ought",
+    "i", "me", "my", "we", "our", "you", "your", "he", "she", "it",
+    "they", "them", "their", "what", "which", "who", "whom", "this",
+    "that", "these", "those", "to", "of", "in", "on", "at", "by",
+    "for", "with", "about", "as", "how", "when", "where", "why",
+];
+
 
 /// Represents an relationship from the universal relationship table
 ///
@@ -4343,18 +4361,6 @@ impl SurrealStore {
         // whose titles contain query keywords are included in the candidate set.
         //
         // Single-term queries fall back to a simple `content @@ $t0` with no OR overhead.
-        // Cap at 4 tokens and strip common stop words — each additional OR term costs ~20ms
-        // on the fulltext index, so long natural-language queries would otherwise be slow.
-        const BM25_MAX_TOKENS: usize = 4;
-        const BM25_STOP_WORDS: &[&str] = &[
-            "a", "an", "the", "is", "are", "was", "were", "be", "been", "being",
-            "have", "has", "had", "do", "does", "did", "will", "would", "could",
-            "should", "may", "might", "shall", "can", "need", "dare", "ought",
-            "i", "me", "my", "we", "our", "you", "your", "he", "she", "it",
-            "they", "them", "their", "what", "which", "who", "whom", "this",
-            "that", "these", "those", "to", "of", "in", "on", "at", "by",
-            "for", "with", "about", "as", "how", "when", "where", "why",
-        ];
         let tokens: Vec<String> = query
             .split_whitespace()
             .map(|t| t.trim_matches(|c: char| !c.is_alphanumeric()).to_lowercase())
@@ -4428,7 +4434,8 @@ impl SurrealStore {
                     (SELECT * FROM $node_ids.{..+collect}<-relationship[WHERE relationship_type = 'has_child']<-node)
                 )
             )
-            WHERE array::len(<-relationship[WHERE relationship_type = 'has_child']) = 0;
+            WHERE array::len(<-relationship[WHERE relationship_type = 'has_child']) = 0
+              AND lifecycle_status != 'deleted';
         "#;
 
         let mut root_response = self


malibio:
### Code Review Summary

This PR delivers two high-value changes in a tightly scoped diff: fixing the core correctness bug that made hybrid search tier1 always zero, and replacing a costly iterative root-resolution loop with a single graph traversal. Both are net-positive improvements. The implementation is clean and well-commented. I have a small correctness concern on the root traversal query and a few improvement-level notes.

---

### Findings

#### [Improvement] Root traversal query does not filter out deleted nodes

**File:** `packages/core/src/db/surreal_store.rs` — `sql_roots` query (line ~4423)

The root-resolution query walks all ancestors and filters to nodes with no incoming `has_child` relationship, but it does not filter out nodes where `lifecycle_status = 'deleted'`. A soft-deleted node can have no parent and would therefore be returned as a valid root.

The BM25 candidate query correctly excludes deleted nodes (`AND lifecycle_status != 'deleted'`), but the subsequent graph traversal fetches from the full relationship graph without this guard. If a deleted node happens to be an ancestor of a live matching node — or itself has no parent — it could be emitted as a root ID, passed to the KNN intersection check, and either silently miss or inflate tier1 counts.

Mitigation: add `AND lifecycle_status != 'deleted'` to the outer `WHERE` clause in `sql_roots`. Given the comment in `MAX_BM25_ROOT_WALK_DEPTH`'s docstring about "malformed relationship data", this is a realistic scenario worth guarding.

#### [Improvement] `BM25_STOP_WORDS` defined as function-local constant

**File:** `packages/core/src/db/surreal_store.rs` — line 4349

The stop-word list is defined as a `const` inside `bm25_search_roots`. While Rust allows this, there are two minor consequences:

1. It cannot be unit-tested in isolation without calling the full async method.
2. The stop-word logic is duplicated conceptually with the title-boost tokenizer in `embedding_service.rs` (lines 688–695), which strips punctuation but applies no stop-word filter. If the BM25 stop-word set needs to grow or change, the two tokenization paths will diverge.

Promoting the list to a module-level `const` (or a shared utility) would let you add a unit test for the tokenization logic directly and keep the two paths in sync.

#### [Improvement] `score` field deserialization assumes `f64` — verify SurrealDB return type

**File:** `packages/core/src/db/surreal_store.rs` — `BM25Row` struct (line 4397)

`search::score()` in SurrealDB returns a `Float` (f32 internally in older versions). Deserializing as `f64` should work through widening, but the field is only used for `ORDER BY` in SQL — the Rust side discards the score immediately after `rows.into_iter().map(|r| r.id).collect()`. If the score is never used in Rust, the `BM25Row` struct carries dead weight. A simpler deserialization (e.g., using `serde_json::Value` for the score or just not selecting it) would reduce the surface area and avoid any future type-mismatch surprises if SurrealDB changes the return type of `search::score()`.

Note: since the ranking happens in SQL (`ORDER BY score DESC`), the Rust-side score value is not needed at all. The `score` field in `BM25Row` can be dropped by restructuring to `SELECT VALUE meta::id(id) FROM (SELECT *, ... AS score FROM node WHERE ... ORDER BY score DESC) LIMIT $limit`, which would let you go back to `Vec<String>` deserialization.

---

#### Nitpicks

- **Nit:** `packages/core/src/db/surreal_store.rs` line 4383 — The `format!` call builds a dynamic SQL string with user-controlled token count (1–4 `search::score(N)` references). The token count is capped at 4 by `BM25_MAX_TOKENS` and the tokens themselves are bound as parameters, so there is no injection risk here. A brief comment clarifying this safety property would pre-empt future reviewer concern.

- **Nit:** `packages/core/src/db/surreal_store.rs` line 4375 — `where_clauses.join(" OR ")` without an extra space inside the parentheses produces `(content @@ $t0 OR content @@ $t1)`, which is correct but reads slightly tight. The existing style in the codebase uses wider spacing in SQL strings. Not a functional issue.

- **Nit:** The blank line left after removing `MAX_BM25_ROOT_WALK_DEPTH` (line 83 in the diff) leaves a double blank line before the next doc comment block. Minor formatting artifact.

---

### What Is Working Well

- The commit history is exemplary: each logical step is a separate commit with a detailed message, including performance measurements and before/after data. This is exactly the right level of documentation for search algorithm changes.
- The stop-word cap-at-4 rationale (each OR term costs ~20ms) is explained directly in the code comment with a concrete example. Future maintainers will understand why the limit exists.
- Replacing the iterative Rust loop with `{..+collect}<-relationship` is the architecturally correct move. The old loop had a hard-coded depth cap that was effectively a ticking correctness bug for deeply nested graphs — the new query is both faster and more correct.
- Tests cover the core contract: root resolution for direct match, child match, grandchild match, and empty result. The existing test for single-term queries (`"persistence"`, `"debounce"`, etc.) will implicitly exercise the new tokenization path.
- The `bm25_limit` reduction from 5x to 2x is well-justified: with score-ranked OR results the top candidates are more relevant, so the extra 3x headroom was paying the traversal cost for low-quality candidates. The PR description's before/after table makes this easy to verify.

---

**Verdict:** Approve with the deleted-node guard recommended as a follow-up. The correctness gap is low-probability in current usage (soft-deleted nodes are unlikely to be ancestors of live nodes in normal operation), but it is a semantic inconsistency worth closing before the search path is relied on more heavily.

malibio:
## Review Feedback Addressed

### ✅ Addressed (2 items)

** in root resolution** ():
Added the filter to the  graph traversal so soft-deleted ancestor nodes cannot be returned as roots. Previously the BM25 candidate query filtered deleted nodes but the traversal didn't — a semantic inconsistency.

** and  moved to module level**:
Both constants are now module-level, making them testable in isolation and visible as documentation. Added comments noting they should stay in sync with the title-boost tokenizer in `embedding_service.rs`.

### ⏭️ Skipped (1 item)

**`BM25Row.score` simplification**: The struct is explicit and the `score` field documents *why* SQL-side ranking is used. Removing it would require adding a subquery to push ORDER BY into a nested SELECT while still returning only IDs — more complexity for marginal readability gain.

PR #968, opened 2026-03-17
fix: prevent BaseNodeViewer fallback during async viewer load

malibio:
## Summary
- Tracks in-flight viewer loads with a `viewerLoading: Set<string>` state variable
- `ViewerComponent` derived now returns `null` while a node type is loading, preventing premature fallback to `BaseNodeViewer`
- Template renders a "Loading..." placeholder when `ViewerComponent` is null

## Problem
When a schema node type (e.g., Task) was clicked in the sidebar, `ViewerComponent` resolved synchronously on the same tick — before the async `pluginRegistry.getViewer()` call completed — and immediately fell back to `BaseNodeViewer`. This caused `BaseNodeViewer` to receive a schema node ID it cannot handle, triggering `get_subtree_with_relationships` with an incompatible ID and freezing the app.

## Test plan
- [ ] `bun run test` — 118 test files, 3722 tests all passing (baseline unchanged)
- [ ] Manual: Click "Task" in Schema Types sidebar — shows "Loading..." briefly, then renders QueryNodeViewer with task nodes (no freeze)
- [ ] Manual: Click date/text nodes — existing viewers still load correctly
- [ ] Manual: Click an already-viewed schema node (switch away and back) — no loading flash (cached path hits early return)

Fixes #963

### DIFF
--- a/packages/desktop-app/src/lib/components/layout/pane-content.svelte
+++ b/packages/desktop-app/src/lib/components/layout/pane-content.svelte
@@ -32,8 +32,8 @@
 
   // Load viewer when needed - moved to function called from onMount to avoid derived context issues
   async function loadViewerForNodeType(nodeType: string) {
-    if (viewerComponents.has(nodeType) || viewerLoadErrors.has(nodeType)) {
-      return; // Already cached — ViewerComponent resolves non-null immediately, no loading state needed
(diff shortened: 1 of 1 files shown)


malibio (commented):
Correct fix with clean Svelte 5 reactivity pattern. One latent deduplication gap (missing viewerLoading.has check in guard) noted but does not block merge. Recommendation: APPROVE.

PR #970, opened 2026-03-17
Add title_template support for schema-driven title computation

### LINKED ISSUE #824: Add title_template support for schema-driven title computation
## Summary

Extend the title field system (introduced in #821) to support schema-defined title templates. This enables custom entity types to have computed, searchable titles based on their properties.

## Background

Issue #821 added an indexed `title` field for @mention search, computed as `strip_markdown(content)` for root nodes and task nodes. This works well for content-based nodes but doesn't support custom schemas where the title should be derived from structured properties.

For example, a "Customer" schema with fields `first_name`, `last_name`, and `email` should have a title like "John Doe (john@example.com)" computed from those properties.

### Schema Definition

Add an optional `title_template` field to schema properties:

```json
{
  "id": "customer",
  "node_type": "schema",
  "content": "Customer",
  "properties": {
    "fields": [
      { "name": "first_name", "field_type": "string" },
      { "name": "last_name", "field_type": "string" },
      { "name": "email", "field_type": "string" }
    ],
    "title_template": "{first_name} {last_name} ({email})"
  }
}
```

### Title Computation Logic

**Priority order** (first match wins):
1. If node's schema has `title_template` → interpolate from properties
2. If node is `task` type → `strip_markdown(content)`
3. If node is root (no parent) and not `date`/`schema` → `strip_markdown(content)`
4. Otherwise → `None` (no title)

### Interpolation Rules

- `{field_name}` → replaced with `properties.field_name` value
- Missing/null fields → empty string (no special handling)
- Non-string values → converted to string (numbers, booleans, dates)

### When Title is Recomputed

- On node create
- On node update (when content OR properties change)

### Content Display for Templated Entities

For custom entities with a `title_template`:
- The computed title becomes the display "content" (read-only)
- Users edit properties via a form interface
- The content field may be hidden or shown as the computed title

## Use Cases

1. **Customer entity**: `"{first_name} {last_name} ({email})"` → "John Doe (john@example.com)"
2. **Invoice entity**: `"Invoice #{invoice_number} - {customer_name}"` → "Invoice #1234 - Acme Corp"
3. **Project entity**: `"{project_name} [{status}]"` → "Website Redesign [In Progress]"

## Implementation Notes

- Primary schema editor will be NLP-based (MCP or built-in LLM), so template syntax should be simple for LLMs to generate
- Title must be stored in DB (not computed at read time) to enable keyword search via @mention
- This builds on the `title` field infrastructure from #821

## Acceptance Criteria

- [ ] Add `title_template` field to SchemaField/SchemaNode model
- [ ] Implement template interpolation function (simple `{field}` replacement)
- [ ] Update title computation in NodeService to check schema for template
- [ ] Recompute title when properties change (not just content)
- [ ] Add validation for title_template syntax in schema creation
- [ ] Update core schema definitions if any should use templates
- [ ] Add unit tests for template interpolation
- [ ] Add integration tests for schema-based title computation

## Future Considerations (Out of Scope)

- Bidirectional parsing (content → properties extraction) - entity resolution use case
- Form UI that mirrors template structure with Enter-key navigation
- Conditional/optional field handling in templates

## Related

- #821 - Add indexed title field for @mention search (prerequisite)
- #823 - PR implementing #821

malibio:
Closes #824

### DIFF
--- a/packages/core/src/behaviors/mod.rs
+++ b/packages/core/src/behaviors/mod.rs
@@ -1051,6 +1051,35 @@ impl SchemaNodeBehavior {
             validate_schema_field(field)?;
         }
 
+        // Validate title_template syntax: every '{' must have a matching '}' and a non-empty field name
+        if let Some(template) = &schema.title_template {
+            let bytes = template.as_bytes();
+            let mut i = 0;
+            while i < bytes.len() {
+                if bytes[i] == b'{' {
+                    // Find matching '}'
+                    match bytes[i + 1..].iter().position(|&c| c == b'}') {
+                        None => {
+                            return Err(NodeValidationError::InvalidProperties(
+                                "title_template contains an unclosed '{' placeholder".to_string(),
+                            ));
+                        }
+                        Some(end) => {
+                            let field_name = &template[i + 1..i + 1 + end];
+                            if field_name.is_empty() {
+                                return Err(NodeValidationError::InvalidProperties(
+                                    "title_template contains an empty '{}' placeholder".to_string(),
+                                ));
+                            }
+                            i += 1 + end + 1; // skip past '}'
+                            continue;
+                        }
+                    }
+                }
+                i += 1;
+            }
+        }
+
         Ok(())
     }
 }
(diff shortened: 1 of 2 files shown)


malibio:
### Code Review Summary

This PR implements schema-driven title computation via a `title_template` field on `SchemaNode`. The feature is well-scoped, the commit is atomic, and the implementation removes all three `TODO #824` markers it promised to resolve. The architecture is sound: a pure interpolation utility lives in `utils/`, schema persistence uses the existing `titleTemplate` property key in the flat JSON store, and `compute_title()` is a single private method that centralizes all title logic. This is a net improvement to the codebase.

A few issues require attention before merge, one of which is a meaningful correctness gap.

---

### Requirements Check

|---|---|
| Add `title_template` field to SchemaNode model | ✅ |
| Implement template interpolation function (`{field}` replacement) | ✅ |
| Update title computation in NodeService to check schema for template | ✅ |
| Recompute title when properties change (not just content) | ✅ |
| Add validation for `title_template` syntax in schema creation | ❌ — not implemented |
| Update core schema definitions if any should use templates | ✅ (`None` set on all 12 core schemas) |
| Add unit tests for template interpolation | ✅ |
| Add integration tests for schema-based title computation | ✅ |

---

### Findings

#### [Improvement] Missing Acceptance Criterion: Template Syntax Validation

`packages/core/src/services/node_service.rs` — schema creation/update path

The issue explicitly calls for validation of `title_template` syntax. No validation exists: a template like `{unclosed` or `{{nested}}` is accepted silently. While the interpolation function handles malformed input gracefully (it passes through an unclosed `{` as a literal), the absence of validation means malformed templates stored in the DB will silently produce incorrect titles that are hard to diagnose.

A minimal check — verifying that every `{` has a matching `}` and the field name is non-empty — would satisfy the criterion and flag author mistakes at schema-save time rather than at title-read time. This should be added to the schema validation path.

---

#### [Improvement] Silent Error Swallowing in `compute_title`

`packages/core/src/services/node_service.rs`, lines 2132–2145

```rust
if let Ok(Some(schema)) = self.get_schema_node(&node.node_type).await {
```

The `Err(_)` arm is discarded with no logging. If `get_schema_node` fails due to a database error, `compute_title` silently falls through to the content-based fallback, producing a wrong title with no observable signal. In a service that otherwise propagates errors via `?`, this inconsistency violates the principle of least surprise and will make debugging intermittent title corruption very difficult.

The appropriate fix is either to propagate the error with `?` (consistent with the rest of the method) or, if fallback on DB failure is intentional, to log a warning before continuing.

---

#### [Improvement] Double Normalization on `create_node_with_parent`

`packages/core/src/services/node_service.rs`, lines 1568–1596

The comment acknowledges this: properties are normalized to namespace format to construct `temp_node` for `compute_title`, then the raw `params.properties` (unnormalized) are passed into the actual `Node` that gets written to DB — where `create_node` normalizes them again. This means the normalization result used for title computation and the normalization result used for storage are computed independently from the same input.

This is safe today because normalization is idempotent, but it creates two divergence points: if normalization logic ever changes, the title computed here could differ from what is actually stored. A cleaner approach would be to normalize once, use the result for both `compute_title` and the stored `Node`, and pass the normalized form to `create_node` (or have `create_node` skip redundant normalization when properties are already normalized). The comment "create_node will normalize again, but the result is idempotent" is accurate but essentially documents a latent inconsistency risk.

---

#### [Nit] Test Schema IDs Use Raw Strings Instead of UUIDs for the Schema Itself

`packages/core/src/services/node_service.rs`, tests `title_template_tests` module

`create_custom_schema` uses `type_id` (e.g. `"customer"`, `"customer2"`, `"customer3"`) as both the schema ID and the node type. The tests work, but using sequential suffix variations (`"customer2"`, `"customer3"`) to avoid collisions between isolated tests suggests these tests may share a `NodeService` instance. If the test service is truly isolated per test (each `create_test_service()` call creates a fresh DB), the suffixes are unnecessary and create misleading names. If they do share state, the tests have a hidden ordering dependency. It is worth confirming which is the case and either removing the suffixes or making the isolation explicit.

---

#### [Nit] Interpolation Parser Allocates a `Vec<char>` for Every Call

`packages/core/src/utils/markdown.rs`, lines 124–148

```rust
let chars: Vec<char> = template.chars().collect();
```

For a hot path called on every node create/update with a `title_template`, allocating a heap `Vec<char>` on each invocation is unnecessary. The same scan can be done with byte-index arithmetic on the original `&str` since `{` and `}` are single-byte ASCII. This is a minor allocation concern, not a correctness issue, and is low priority for a pre-release codebase, but worth a note for when performance tuning begins.

---

### Final Recommendation

**REQUEST CHANGES**

The missing validation for `title_template` syntax is an explicit acceptance criterion that was not implemented. The silent error swallowing in `compute_title` is a meaningful correctness and observability gap. Both should be addressed before merge. The double normalization is a design cleanliness issue worth tracking but not a blocker.

PR #982, opened 2026-03-19
perf: optimistic title_template evaluation in SchemaPropertyForm

### LINKED ISSUE #973: perf: optimistic title_template evaluation in SchemaPropertyForm
## Problem

When a user edits a property field on a custom entity node (e.g. \`first_name\` on a Customer with \`titleTemplate: \"{first_name} {last_name}\"\`), the viewer header and table row update reactively — but only after a full round-trip:

\`\`\`
edit field → debounce → HTTP PATCH → Rust compute_title() → DB write → response → store update → re-render
\`\`\`

This introduces noticeable lag (~300–800ms) because it requires:
1. Debounce timer
2. HTTP round-trip to the backend
3. Schema lookup in DB (for \`titleTemplate\`)
4. Title write to DB
5. Response propagation back to the store

## Solution

Evaluate the \`titleTemplate\` client-side in \`SchemaPropertyForm\` on every \`input\` event, updating \`node.title\` in \`sharedNodeStore\` optimistically before the backend round-trip completes.

**Architecture:**
- The schema (\`genericSchema.titleTemplate\`) is already loaded in the viewer
- The current field values are already in the store
- Interpolation is simple string replacement: \`{field_name}\` → value
- The backend result always overwrites the optimistic value on response (source of truth)

**Flow after fix:**
\`\`\`
edit field → evaluate template client-side → update store.title → re-render instantly
           ↘ debounce → HTTP PATCH → Rust compute_title() → DB write → confirm/overwrite
\`\`\`

## Note on MCP / backend callers

No change needed for MCP or any backend API caller. \`compute_title()\` already runs automatically in Rust on every \`update_node\` call when \`properties_changed\`. Callers never need to pass \`title\` — the backend derives it from the template transparently.

The frontend optimistic evaluation is a pure UX improvement — additive and optional.

## Implementation

- **\`schema-property-form.svelte\`**: After updating the store with the new property value, also compute and update \`node.title\` in the store using a simple \`{token}\` interpolation function
- **\`shared-node-store.svelte.ts\`**: The existing \`updateNode\` path already handles the store update; just needs the title computed and passed alongside the property change

## Acceptance Criteria

- [ ] Editing a property field on a \`titleTemplate\` schema type updates the header and table title instantly (< 50ms), without waiting for the backend round-trip
- [ ] Backend response still overwrites the optimistic value (no divergence on concurrent edits)
- [ ] MCP and all other backend callers continue to work unchanged — backend always evaluates the template authoritatively

malibio:
Closes #973

### DIFF
--- a/packages/desktop-app/src/lib/components/property-forms/schema-property-form.svelte
+++ b/packages/desktop-app/src/lib/components/property-forms/schema-property-form.svelte
@@ -29,6 +29,7 @@
   import type { Node } from '$lib/types';
   import { parseDate, type DateValue } from '@internationalized/date';
   import { createLogger } from '$lib/utils/logger';
+  import { evaluateTitleTemplate } from '$lib/utils/title-template';
 
   // Logger instance for SchemaPropertyForm component
   const log = createLogger('SchemaPropertyForm');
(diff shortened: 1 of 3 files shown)


malibio (commented):
Net positive change — optimistic title interpolation is correctly scoped as a UX layer with backend as source of truth. See the detailed review comment above. Main follow-up items: add a unit test for `evaluateTitleTemplate`, and confirm that the `|| null` coercion on an all-whitespace template result aligns with what Rust `compute_title` returns.

PR #993, opened 2026-03-22
QueryNodeViewer: getSchema fails with SCHEMA_NOT_FOUND on fresh database

### LINKED ISSUE #989: QueryNodeViewer: getSchema fails with SCHEMA_NOT_FOUND on fresh database
## Summary

On a fresh database (or after import), the QueryNodeViewer component logs errors when attempting to load schema definitions that don't exist yet:

```
[Error] [DiagnosticLogger] [ERROR] getSchema (6.00ms)
[Error] [QueryNodeViewer] Failed to load schema node
```

## Root Cause

`query-node-viewer.svelte` calls `backendAdapter.getSchema(schemaId)` in `loadAndQuery()` (line 72). The backend queries `SELECT * FROM node:{id} WHERE node_type = 'schema'` — on a fresh database without core schema initialization, this returns empty results and the backend returns `SCHEMA_NOT_FOUND`.

**Call chain:**
1. `query-node-viewer.svelte` → `backendAdapter.getSchema(schemaId)`
2. `diagnostic-logger.ts` → `invoke('get_schema_definition', { schemaId })`
3. `commands/schemas.rs` → `service.get_schema_node(&schema_id)`
4. `surreal_store.rs` → query returns empty → `SCHEMA_NOT_FOUND` error

## Impact

- Non-critical: the query viewer shows an error state with retry button
- Noisy console errors on every fresh database load
- 6ms query duration suggests it's a "not found" rather than connectivity issue

## Possible Solutions

1. **Graceful handling**: QueryNodeViewer should handle missing schemas without logging an error (e.g., show empty state instead of error)
2. **Core schema initialization**: Ensure core schemas (task, date, text) are created on database startup
3. **Lazy schema creation**: Create core schemas on first access if they don't exist

## Steps to Reproduce

1. Start the app with a fresh/empty database
2. Navigate to a schema-driven view (e.g., Tasks)
3. Observe console errors

malibio:
Closes #989

### DIFF
--- a/packages/desktop-app/src/lib/components/viewers/query-node-viewer.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/query-node-viewer.svelte
@@ -106,7 +106,7 @@
   function isSchemaNotFound(message: string): boolean {
     // Tauri CommandError: "Schema '<id>' not found" (code: SCHEMA_NOT_FOUND)
     // HTTP adapter: parsed from ApiError with code SCHEMA_NOT_FOUND
-    return message.includes('not found') || message.includes('SCHEMA_NOT_FOUND');
+    return /Schema '.*' not found/.test(message) || message.includes('SCHEMA_NOT_FOUND');
   }
 
   // Build a lookup map from schema fields for enum label resolution


malibio:
## Code Review: PR #993 — QueryNodeViewer: getSchema fails with SCHEMA_NOT_FOUND on fresh database

**Issue:** #989
**Reviewer:** Principal Engineer AI (Pragmatic Quality Framework)

---

### Requirements Check (Issue #989)

| Acceptance Criteria | Status |
|---|---|
| QueryNodeViewer handles missing schemas without logging an error | **PASS** -- Error downgraded to `log.debug` for the schema-not-found case |
| Show empty state instead of error with retry button | **PASS** -- Sets `queryState = 'success'` which renders the "No nodes of this type yet" empty state |
| Non-critical case handled gracefully | **PASS** -- Genuine errors still surface as error state with retry |

The fix addresses the root cause identified in the issue: on a fresh database without core schema initialization, `getSchema` returns `SCHEMA_NOT_FOUND`, and the component now treats this as an expected condition rather than an error.

---

### Code Review Findings

#### Suggested Improvements

**1. [Improvement] `isSchemaNotFound` string matching is fragile**
`packages/desktop-app/src/lib/components/viewers/query-node-viewer.svelte`, lines 106-110

```typescript
function isSchemaNotFound(message: string): boolean {
  return message.includes('not found') || message.includes('SCHEMA_NOT_FOUND');
}
```

The `message.includes('not found')` check is overly broad. Any error message containing the substring "not found" will be swallowed — e.g., "Connection endpoint not found", "Module not found", "Route not found". This violates the principle of least surprise: unexpected infrastructure errors would silently produce an empty state instead of surfacing to the user.

**Recommendation:** Tighten the match to only match the specific Rust backend format `"Schema '<id>' not found"`, or rely solely on the error code:

```typescript
function isSchemaNotFound(message: string): boolean {
  return message.includes('SCHEMA_NOT_FOUND') || /Schema '.*' not found/.test(message);
}
```

The `SCHEMA_NOT_FOUND` code check is sufficient for the Tauri adapter path (the `CommandError.code` field is serialized into the error). The regex covers the HTTP adapter path with the exact format from `dev-proxy.rs` line 1278: `format!("Schema '{}' not found", id)`. This prevents false positives from unrelated "not found" errors.

**Severity:** This is a real concern but not a blocker for a pre-release codebase. The fix is a one-line regex change.

---

**2. [Improvement] The `generic-schema-form.svelte` fix is correct but the `$effect` is one-directional**
`packages/desktop-app/src/lib/components/schema/generic-schema-form.svelte`, lines 38-42

```typescript
let isOpen = $state(false);
$effect(() => {
  if (autoOpen) isOpen = true;
});
```

This correctly fixes the Svelte lint issue where `$state(autoOpen)` only captured the initial prop value. The `$effect` now tracks reactive changes to `autoOpen`. However, note that this is intentionally one-directional: once `autoOpen` becomes `true`, the collapsible opens, but if the parent later sets `autoOpen` back to `false`, the collapsible will not close (because the effect only fires on truthy). This is likely the desired behavior (user can manually close), but worth calling out for clarity.

**Severity:** Not an issue -- just documenting the intent for future readers. The existing comment on line 39 adequately explains the rationale.

---

#### Nitpicks

**Nit:** `query-node-viewer.svelte`, line 91 -- The change from `'Failed to load schema'` to `String(e)` for non-Error objects is a good improvement. `String(e)` preserves the actual error information rather than hiding it behind a generic message. Clean change.

**Nit:** The commit message is well-structured -- it explains both the schema-not-found fix and the generic-schema-form fix with clear rationale. The "Also fixes" prefix correctly signals the secondary change is a pre-existing issue addressed opportunistically. Good practice for atomic-but-related fixes.

---

### Overall Assessment

This is a clean, minimal fix that correctly addresses the issue. The change is well-scoped: two files, ~20 lines of net new code, single cohesive purpose. The approach of treating schema-not-found as an expected condition (empty state) rather than an error state is the right call for a fresh database scenario.

The `generic-schema-form.svelte` change is a legitimate pre-existing bug fix (Svelte reactivity correctness) and is appropriate to bundle here given its small scope.

The one substantive concern is the overly broad string matching in `isSchemaNotFound`, which could mask unrelated errors. This is worth tightening but is not a blocker given the pre-release status of the codebase.

**No tests exist for either component.** This is a pre-existing gap, not introduced by this PR. Given the fix is purely defensive error handling in a UI component, the risk is low.

---

### Final Recommendation: **APPROVE** (with suggestion)

The PR is a net positive improvement to code health. The schema-not-found handling is correct, the log level change is appropriate, and the generic-schema-form fix addresses a real reactivity bug. The string matching suggestion is worth addressing but does not block merge.

malibio (commented):
**APPROVE** (posted as comment because GitHub prevents self-approval). One suggestion: tighten the isSchemaNotFound string match from message.includes('not found') to a regex like /Schema '.*' not found/ to avoid masking unrelated errors. The SCHEMA_NOT_FOUND code check alone covers the Tauri path. See PR comment for full review.

PR #994, opened 2026-03-22
Tune RocksDB compaction for desktop app — WAL bloat causes 10x oversized database

### LINKED ISSUE #992: Tune RocksDB compaction for desktop app — WAL bloat causes 10x oversized database
## Problem

A fresh import of ~210 docs produces a 147MB database (all in RocksDB WAL log files, no SST files). Expected post-compaction size would be ~10-15MB. At this rate, 2K docs would hit ~1.4GB before compaction.

RocksDB WAL files are not being flushed/compacted aggressively enough for a desktop app where disk visibility matters.

## Current State

```
112M  000004.log   (content + relationships WAL)
 35M  000008.log   (additional WAL)
  0   SST files    (no compaction has occurred)
```

## Expected

After compaction:
- 210 docs: ~10-15MB
- 2K docs: ~80-150MB
- Embeddings (768-dim × N docs): ~1-2MB per 210 docs

## Possible Fixes

1. **Force flush after bulk import** — call `db.flush()` or equivalent after import completes
2. **Tune RocksDB options** — reduce `write_buffer_size`, increase `max_write_buffer_number`, lower compaction thresholds
3. **Periodic compaction** — trigger manual compaction on app idle or startup
4. **WAL size limit** — set `max_total_wal_size` to force earlier flushing

## Context

- Database: SurrealDB embedded with RocksDB backend
- Use case: Desktop app, single user, batch imports via MCP

malibio:
Closes #992

### DIFF
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -423,6 +423,12 @@ impl SurrealStore {
     ///   database reaches its steady-state size quickly after bulk writes.
     /// - **Max write buffers 3**: Allow some write buffering for burst
     ///   performance while keeping memory bounded.
+    /// - **Blob files disabled**: BlobDB separates large values into `.blob`
+    ///   files which adds ~30% overhead for small desktop datasets. Inline
+    ///   storage in SSTs compacts better at our scale.
+    /// - **Target file size multiplier 1**: Uniform file sizes across all
+    ///   compaction levels (all 16 MiB) instead of doubling per level,
+    ///   which produces tighter compaction for small datasets.
     fn configure_rocksdb_defaults() {
         use std::env;
         use std::sync::Once;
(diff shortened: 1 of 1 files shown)


malibio (commented):
One blocker: SURREAL_ROCKSDB_BACKGROUND_FLUSH_INTERVAL does not exist in SurrealDB 3.0.x (verified against surrealdb-core-3.0.4/src/kvs/rocksdb/cnf.rs). It is silently ignored at runtime. Remove it from the defaults array and its doc-comment bullet. If periodic WAL flushing is desired, use the sync query parameter on the connection URL or the SURREAL_DATASTORE_SYNC_DATA env var instead. See full review in PR comment.

malibio:
## Review Feedback Addressed

### ✅ Addressed (3)
- **🔴 Critical**: Removed phantom `SURREAL_ROCKSDB_BACKGROUND_FLUSH_INTERVAL` — confirmed it doesn't exist in SurrealDB 3.0.x `kvs/rocksdb/cnf.rs`
- **🟡 Important**: Changed `MIN_WRITE_BUFFER_NUMBER_TO_MERGE` from `1` → `2` to avoid excessive L0 files during bulk imports
- **🟢 Suggestion**: Verified `surrealdb/surrealdb#2399` link — confirmed open and correctly titled "Graceful shutdown support when using embedded mode"

### ⏭️ Skipped (2)
- **🟢 Comment in `new_http()`**: HTTP mode doesn't read env vars by nature; adding a comment would be noise
- **🟢 Rust 2024 `unsafe env::set_var`**: Not actionable until edition upgrade; noted for future

Commit: 4a502de9

PR #1066, opened 2026-04-07
Add Ollama as alternative local inference backend

### LINKED ISSUE #1058: Add Ollama as alternative local inference backend
## Summary

Add Ollama as an alternative inference backend alongside the existing llama.cpp integration. Ollama provides a managed local model server with a simple HTTP API, making it easier to use a wide variety of models (e.g., gemma4, llama3, mistral) without manual GGUF downloads or GPU resource management.

The existing trait-based architecture (`ChatInferenceEngine`, `ModelManager`) already supports pluggable backends — this issue implements the Ollama backend behind those traits.

## Motivation

- **Model variety**: Ollama supports hundreds of models with one-line pulls (`ollama pull gemma4:31b`)
- **Simplified GPU management**: Ollama handles Metal/CUDA scheduling, context caching, and model lifecycle
- **No binary size increase**: HTTP client vs bundled llama.cpp reduces desktop app binary size when Ollama is available
- **Developer experience**: Developers already running Ollama can use their existing models without re-downloading GGUF files

### 1. OllamaInferenceEngine (`packages/agent/`)

Implement `ChatInferenceEngine` trait using Ollama's `/api/chat` endpoint:

- Streaming responses via NDJSON (newline-delimited JSON)
- Tool calling support (Ollama supports OpenAI-compatible tool calling)
- Map `InferenceRequest` → Ollama chat request format
- Map Ollama streaming chunks → `StreamingChunk` enum
- Token counting via `/api/show` model metadata (context_length) or estimation

### 2. OllamaModelManager (`packages/agent/`)

Implement `ModelManager` trait using Ollama's API:

- `list()` → `GET /api/tags` (list local models)
- `download()` → `POST /api/pull` (pull model with progress streaming)
- `delete()` → `DELETE /api/delete`
- `load()` / `unload()` → `POST /api/generate` with `keep_alive` parameter
- `loaded_model()` → `GET /api/ps` (list running models)
- `recommended_model()` → return a sensible default (e.g., `gemma4:12b`)

### 3. Unified Model List with `ollama:` Prefix Convention

When Ollama is detected as running, models from both backends appear in a single dropdown list with clear provenance:

```
Ministral 3B (Built-in)
Ministral 8B (Built-in)
ollama:gemma4:31b
ollama:llama3:8b
ollama:mistral:7b
```

- Built-in GGUF models use their existing display names
- Ollama models are prefixed with `ollama:` to indicate the backend
- The `ollama:` prefix is used as a routing key — selecting an `ollama:*` model routes through `OllamaModelManager` / `OllamaInferenceEngine`, others go through `GgufModelManager` / `LlamaChatInferenceEngine`
- If Ollama is not running, only built-in models appear (no error, just fewer options)

### 4. Tauri Integration

- Add inference backend routing in `ManagedAgentState` based on `ollama:` prefix
- Ollama health check: verify server is running at configured URL before use
- Emit `model://status` events consistently regardless of backend
- Store Ollama base URL in app configuration (default: `http://localhost:11434`)

### 5. Settings UI

- Ollama URL configuration field (with connection test button)
- Model picker shows unified list from both backends when Ollama is available

## Acceptance Criteria

- [ ] `OllamaInferenceEngine` implements `ChatInferenceEngine` with streaming support
- [ ] `OllamaModelManager` implements `ModelManager` with pull/list/delete operations
- [ ] Unified model list shows both built-in and `ollama:`-prefixed models in a single dropdown
- [ ] `ollama:` prefix routes model operations to Ollama backend automatically
- [ ] Only built-in models shown when Ollama server is not detected (no error state)
- [ ] Tool calling works with Ollama models that support it (gemma4, llama3, etc.)
- [ ] Download progress events stream to frontend during `ollama pull`
- [ ] Health check detects whether Ollama server is running
- [ ] Token counting works (estimation acceptable if Ollama doesn't expose tokenizer)
- [ ] Agent ReAct loop works identically regardless of backend
- [ ] Unit tests with mock HTTP responses for Ollama API
- [ ] Graceful error handling when Ollama server is unavailable

## Technical Notes

- Ollama API docs: https://github.com/ollama/ollama/blob/main/docs/api.md
- Ollama tool calling follows OpenAI-compatible format
- `reqwest` (already a dependency) can handle NDJSON streaming
- `ModelFamily` enum will need a new `Ollama` variant or be made more flexible
- `ModelInfo` struct may need adaptation since Ollama models don't have SHA-256/URL fields — consider making those optional or creating an Ollama-specific catalog type

## Out of Scope

- Embedding via Ollama (keep using llama.cpp `nomic-embed-text` for now)
- Remote Ollama servers (localhost only for v1)
- Auto-installing Ollama

malibio:
Closes #1058

### DIFF
--- a/packages/agent/src/intent.rs
+++ b/packages/agent/src/intent.rs
@@ -48,15 +48,21 @@ static INTENT_PATTERNS: &[(&[&str], &str)] = &[
         &[
             "new type",
             "new node type",
+            "a new node type",
             "new schema",
             "new entity type",
             "define a type",
             "create a type",
             "create a schema",
+            "create a new type",
+            "create a new node type",
+            "create a new schema",
             "define fields",
             "define a schema",
             "node type",
             "with fields",
+            "entity type",
+            "schema for",
         ],
         "create schema",
     ),
@@ -325,6 +331,30 @@ mod tests {
     fn create_new_type_matches_schema() {
         let result =
             extract_intent("Create a new type 'Project' and define fields for tracking projects");
+        assert_eq!(result.query, "create schema");
+        assert!(result.from_pattern);
+    }
+
+    #[test]
+    fn create_new_node_type_with_colon_matches_schema() {
+        // Regression: "Create a new node type: 'Project'" was matching "create" instead of schema
+        let result = extract_intent(
+            "Create a new node type: 'Project' and use fields that are typical of what we want to track on a project",
+        );
+        assert_eq!(result.query, "create schema");
+        assert!(result.from_pattern);
+    }
+
+    #[test]
+    fn entity_type_matches_schema() {
+        let result = extract_intent("I need a new entity type for invoices");
+        assert_eq!(result.query, "create schema");
+        assert!(result.from_pattern);
+    }
+
+    #[test]
+    fn schema_for_matches_schema() {
+        let result = extract_intent("Create a schema for tracking customer orders");
         assert_eq!(result.query, "create schema");
         assert!(result.from_pattern);
     }


malibio:
### Re-Review: Commit a7b447f2

All critical findings from the original review have been addressed. Here is the verification of each item.

---

#### Critical Issues — Verified Fixed

**Deadlock in `OllamaModelManager::unload()`**

The fix is correct and idiomatic. The read guard is now dropped immediately after the `.clone()` call, before any `await` points and before the `write()` acquisition at line 346. The `drop(loaded)` call that the previous pattern relied on (which only dropped the guard *lexically* after the HTTP awaits) is gone. The comment explaining *why* the clone is necessary is accurate and will aid future maintainers.

One minor observation: there is a subtle correctness point worth noting. If the `unload()` HTTP request fails at line 335 (`map_err` returns `Err`), the function returns early with an error and the `write().await = None` at line 346 is never reached. This means `loaded_model_id` retains its stale value even though the keep-alive request failed. This was present in the original code as well, so it is not a regression introduced by this fix, but it is worth flagging for a follow-up: on HTTP failure, the local state should still be cleared (or the error should be propagated differently) to avoid the manager believing a model is still loaded when it may not be.

**`OllamaInferenceEngine` hardcoded URL**

Fixed correctly. `manager.ollama_manager().base_url()` returns a `&str` from the getter added to `OllamaModelManager`, `.to_string()` allocates the owned `String`, and `OllamaInferenceEngine::with_base_url()` receives it. The inference engine now shares the same configured URL as the model manager for all lifecycle operations.

---

#### Improvements — Verified Addressed

**`CompositeModelManager::list()` double `is_available()` guard** — Removed. The direct `self.ollama.list().await?` call is correct since `OllamaModelManager::list()` already handles connection errors by returning `Ok(vec![])`.

**`CompositeModelManager::unload()` `is_available()` guard** — Removed. `gguf_result.and(ollama_result)` is a clean idiom: it propagates the first error encountered, which is the right behavior when both unload operations run unconditionally.

**`model_info()` silent error swallowing** — JSON parse failures now emit `tracing::warn!` with the model name and error, which is exactly the right level: it is an unexpected API contract violation, not a routine availability check. Network errors remain silently `Ok(None)`, which is correct.

---

#### Nits — Verified Addressed

- `cancel_download` `warn!` → `tracing::debug!` with explanatory comment: correct.
- `test_token_count_estimate` now uses `assert_eq!(count, 2, ...)` with the expected value pinned: correct.
- Integration test `test_ollama_inference_with_tools` comment clarified: correct.
- Import order in `ollama_integration.rs` alphabetized: correct.
- `rustfmt`-style multi-line argument reformatting throughout integration tests: clean.

---

#### New Issue Observed

**`CompositeModelManager::loaded_model()` retains `is_available()` guard (line 162) that was removed from `list()` and `unload()`.**

This is inconsistent with the changes made to the other two methods. `OllamaModelManager::loaded_model()` queries `/api/ps` and gracefully returns `Ok(None)` on any error (lines 354–363 of `ollama_model_manager.rs`), so the `is_available()` pre-check here has the same double-round-trip problem that was fixed in `list()`. The inconsistency means that on a fast machine a caller could see a model listed (via `list()`, no guard) but then see `loaded_model()` return `None` (because `is_available()` raced and returned `false`). This is a [Nit]-level inconsistency since `loaded_model()` is not on the hot path, but it should be harmonized with the pattern established by this fix.

---

#### Summary

The two blockers are correctly fixed, all noted improvements are addressed, and all nits are resolved. The fix commit is clean. The only new observation is a [Nit]-level inconsistency in `loaded_model()` that can be addressed in a follow-up. This PR is approved.

PR #1067, opened 2026-04-07
Refresh workspace context (ENTITY TYPES) per turn, not per session

### LINKED ISSUE #1062: Refresh workspace context (ENTITY TYPES) per turn, not per session
## Problem

`build_workspace_context` (which includes ENTITY TYPES, collections, and playbooks) is called **once at session creation** and cached in `session.dynamic_context`. If the user creates a new schema type mid-session (e.g., "Project"), the system prompt still shows the old type list for the rest of that session.

This means:
- The model doesn't know "Project" is a valid type until a new chat session is started
- The model can't create instances of newly-defined types within the same conversation
- The ENTITY TYPES section becomes stale immediately after any schema CRUD

### Current Flow
```
create_session → build_workspace_context() → cache in session.dynamic_context
send_message(turn 1) → reads cached context ← types: [task, text]
user creates "Project" schema via create_schema tool
send_message(turn 2) → reads cached context ← types: [task, text] ❌ stale!
```

### Desired Flow
```
create_session
send_message(turn 1) → build_workspace_context() → types: [task, text]
user creates "Project" schema via create_schema tool
send_message(turn 2) → build_workspace_context() → types: [task, text, project] ✅ fresh!
```

## Proposed Solution

Rebuild the workspace context **per turn** instead of per session. Two options:

**Option A (simplest):** Refresh context in the Tauri command layer before each `send_message` call:
```rust
// In local_agent_send_message command:
let context = build_workspace_context(&ns).await.unwrap_or_default();
service.set_session_context(&session_id, context.format_for_prompt(1500)).await;
// Then proceed with send_message
```

**Option B:** Pass a context provider into the agent loop so `run_turn` can fetch fresh context itself, without depending on external refresh.

Option A is minimal and keeps the agent loop dependency-free. The cost is one `get_all_schemas` query per turn (~1ms with cached schemas).

## Acceptance Criteria

- [ ] ENTITY TYPES in system prompt reflects current schemas on every turn
- [ ] Creating a schema in turn N makes it available in turn N+1 within the same session
- [ ] Collections and playbooks also refresh per turn
- [ ] No significant latency increase (schema query is fast)

malibio:
Closes #1062

### DIFF
--- a/packages/core/src/mcp/handlers/tools.rs
+++ b/packages/core/src/mcp/handlers/tools.rs
@@ -17,6 +17,7 @@
 
 use crate::mcp::handlers::{markdown, nodes, playbook, relationships, schema, search, skills};
 use crate::mcp::types::MCPError;
+use crate::models::SchemaNode;
 use crate::services::{NodeEmbeddingService, NodeService};
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
--- a/packages/desktop-app/src-tauri/src/commands/local_agent.rs
+++ b/packages/desktop-app/src-tauri/src/commands/local_agent.rs
@@ -334,6 +334,10 @@ pub async fn local_agent_new_session(
 
 /// Send a user message to a local agent session.
 ///
+/// Before inference, refreshes the workspace context (entity types, collections,
+/// playbooks) to ensure the AI knows about any schemas added mid-session.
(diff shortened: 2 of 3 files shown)


malibio:
Superseded by #1069 which has a more complete implementation of per-turn workspace context refresh.

PR #1074, opened 2026-04-08
Dynamic skill descriptions: update Node Creation skill when schemas change

### LINKED ISSUE #1061: Dynamic skill descriptions: update Node Creation skill when schemas change
## Problem

Skill node descriptions are static (set at seed time). When a user creates a new schema type (e.g., "Project"), the Node Creation skill's description still says generic "create new content nodes" — it doesn't mention "Project" as an available type. This means:

1. "Create a new project" doesn't semantically match the Node Creation skill well
2. The model doesn't know "project" is a valid node_type to pass to `create_node`
3. The push pipeline can't scope tools effectively for custom types

## Proposed Solution

**Dynamically update skill descriptions when schemas change.**

### Flow

1. User creates "Project" schema via `create_schema` tool
2. Schema creation triggers a domain event (`SchemaCreated`)
3. A listener rebuilds the Node Creation skill's description to include all current types
4. The skill node is updated in the graph → marked stale for embedding
5. Embedding processor re-embeds the skill with new description
6. Next semantic search for "create a new project" matches Node Creation with high confidence

### Implementation

- Add a `SkillUpdater` service that listens to schema CRUD events
- On schema change: query all schema node names → rebuild Node Creation description
- Description template: `"Create new instances of existing node types: task, text, project, customer. Add a new record, entry, or item of any available type."`
- Update the skill node via `update_node` → triggers stale embedding marker → processor re-embeds
- Same pattern can apply to Graph Editing skill (include custom type fields in description)

### Dynamic Description Format

```
Node Creation: "Create new instances of existing node types: task, text, {schema1}, {schema2}, {schema3}. Add a new record, entry, or item of any available type."
```

## Acceptance Criteria

- [ ] When a schema is created, Node Creation skill description is updated with new type name
- [ ] When a schema is deleted, the type is removed from the skill description
- [ ] Skill embedding is regenerated after description update
- [ ] Semantic search for "create a new {custom_type}" matches Node Creation with high confidence
- [ ] No manual intervention needed — fully reactive

malibio:
Closes #1061

### DIFF
--- a/packages/core/src/ops/skill_updater.rs
+++ b/packages/core/src/ops/skill_updater.rs
@@ -45,12 +45,23 @@ impl SkillUpdater {
 
     /// Subscribe to domain events and start the update loop.
     ///
-    /// Spawns a background task that runs until the broadcast channel closes or
-    /// `shutdown_rx` fires `true`.
+    /// Runs an initial sync before entering the event loop so that schemas
+    /// present at startup are reflected in the description without waiting for
+    /// a schema mutation event.
+    ///
+    /// Runs until the broadcast channel closes or `shutdown_rx` fires `true`.
     pub async fn start(self: Arc<Self>, mut shutdown_rx: tokio::sync::watch::Receiver<bool>) {
         let mut rx: broadcast::Receiver<EventEnvelope> = self.node_service.subscribe_to_events();
 
+        // Subscribe BEFORE the initial sync to avoid a TOCTOU gap where a
+        // schema event arrives between the sync and the first rx.recv().
         info!("SkillUpdater subscribed to domain events");
+
+        // Proactive startup sync: schemas that existed before this task started
+        // would otherwise be ignored until the next schema mutation.
+        if let Err(e) = self.update_node_creation_skill().await {
+            warn!("SkillUpdater: startup sync failed: {}", e);
+        }
 
         loop {
             tokio::select! {
@@ -104,16 +115,24 @@ impl SkillUpdater {
     /// Rebuild the "Node Creation" skill description and persist it.
     async fn update_node_creation_skill(&self) -> anyhow::Result<()> {
         // 1. Fetch all schemas to get the current custom type list.
-        let schemas = self
-            .node_service
-            .get_all_schemas()
-            .await
-            .unwrap_or_default();
+        let schemas = match self.node_service.get_all_schemas().await {
+            Ok(s) => s,
+            Err(e) => {
+                warn!(
+                    "SkillUpdater: failed to fetch schemas, skipping update: {}",
+                    e
+                );
+                return Ok(());
+            }
+        };
 
+        // Use display name (content) for the description so the embedding model
+        // sees human-readable type names like "Invoice" instead of machine keys
+        // like "invoice". This improves semantic match quality.
         let custom_types: Vec<String> = schemas
             .iter()
             .filter(|s| !s.is_core)
-            .map(|s| s.id.clone())
+            .map(|s| s.content.clone())
             .collect();
 
         // 2. Build the updated description.
--- a/packages/core/tests/skill_updater_test.rs
+++ b/packages/core/tests/skill_updater_test.rs
@@ -165,9 +165,73 @@ mod skill_updater_tests {
             .and_then(|v| v.as_str())
             .unwrap_or("");
 
+        // Both schemas must appear, not just one
         assert!(
-            desc.contains("invoice") || desc.contains("customer"),
-            "Description should mention custom schemas after creation, got: {:?}",
+            desc.contains("invoice") && desc.contains("customer"),
+            "Description should mention both custom schemas after creation, got: {:?}",
+            desc
+        );
+
+        let _ = shutdown_tx.send(true);
+        let _ = timeout(Duration::from_millis(500), updater_task).await;
+
+        Ok(())
+    }
+
+    #[tokio::test]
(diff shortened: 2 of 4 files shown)


malibio:
### Code Review — PR #1074: Dynamic Skill Descriptions

**Overall Assessment:** This is a net-positive change that correctly solves the semantic discovery problem described in #1061. The architecture is sound — a narrow, single-purpose background task that subscribes to existing domain events and performs an idempotent write. The core loop, lag handling, and shutdown wiring are all done well. There are a few issues worth addressing before merge.

---

### Findings

#### [Improvement] `s.id` used for display, but `s.content` is the human-readable name

**File:** `packages/core/src/ops/skill_updater.rs`, line 116

```rust
.map(|s| s.id.clone())
```

`SchemaNode.id` is the machine key (e.g., `"invoice"`), while `SchemaNode.content` is the display name (e.g., `"Invoice"`). Both may be identical for simple schemas, but the issue description shows capitalized examples ("Invoice, Customer, Project") and the doc comment on line 6 matches that expectation. Using `s.content` would produce more natural-language-friendly descriptions and give embedding models better signal for user queries like "create a new Invoice". Worth swapping to `s.content` unless there is a deliberate reason to use the machine key here.

---

#### [Improvement] `initialize_skill_updater` returns `()` but all sibling initializers return `anyhow::Result<()>`

**File:** `packages/desktop-app/src-tauri/src/lib.rs`, line 143  
**File:** `packages/desktop-app/src-tauri/src/commands/db.rs`, line 307

All three existing initializers (`initialize_mcp_server`, `initialize_domain_event_forwarder`, `initialize_playbook_engine`) return `anyhow::Result<()>` and are called inside `if let Err(e) = ...` guards that log but do not fail init. `initialize_skill_updater` returns `()` and is called bare. In practice, this function cannot currently fail (the only fallible operation — `SkillUpdater::new` — is infallible), but the inconsistency breaks the established pattern and will surprise the next engineer. Align the signature with siblings by returning `anyhow::Result<()>` and wrapping the call site with the same error-tolerant guard used elsewhere.

---

#### [Improvement] No panic guard on the spawned task

**File:** `packages/desktop-app/src-tauri/src/lib.rs`, lines 161–164

`initialize_playbook_engine` and `initialize_domain_event_forwarder` both wrap their inner futures with `std::panic::AssertUnwindSafe(...).catch_unwind()`. The `SkillUpdater` task does not. A panic inside the event loop (e.g., from an unexpected `get_all_schemas` response shape) would silently kill the background task with no log output beyond the Tokio runtime's default panic handler. Apply the same `catch_unwind` pattern to make failures visible.

---

#### [Improvement] Integration test for multiple schemas uses a weak OR assertion

**File:** `packages/core/tests/skill_updater_test.rs`, line 168

```rust
assert!(
    desc.contains("invoice") || desc.contains("customer"),
    ...
);
```

This assertion passes even if only one of the two schemas appears, meaning the test would not catch a race where the second schema's event is missed. Given that the test already `sleep`s 300ms to allow both events to propagate, the assertion should be `&&` (both must appear) to validate the intended behavior. The unit test at line 188–189 does use AND correctly — the integration test should match.

---

#### [Improvement] No proactive sync at startup

**File:** `packages/core/src/ops/skill_updater.rs`, line 50

`start()` subscribes to the event channel and then waits. If schemas already exist when the app boots (common case after the first run), the skill description will be stale until the next schema create/delete event. The lag handler (line 67–72) shows the right idiom — the same `update_node_creation_skill()` call should be made once immediately after subscribing, before entering the select loop. This brings the description current on every startup without needing a schema change to trigger it.

---

#### Nit: Issue reference in inline comment is informal

**File:** `packages/desktop-app/src-tauri/src/commands/db.rs`, line 306

```rust
// Initialize skill updater (Issue #1061) — keeps Node Creation skill description current
```

Nit: The other initializer comments don't include issue numbers inline. Minor inconsistency, not a blocker.

---

#### Nit: `unwrap_or_default()` silently swallows `get_all_schemas` errors

**File:** `packages/core/src/ops/skill_updater.rs`, line 111

```rust
.unwrap_or_default();
```

This means a DB error produces an empty schema list, which will then overwrite the skill description with the base text — effectively removing all custom types from the description on a transient DB error. `query_nodes_simple` on line 132 does the same. Consider logging the error before falling back, so failures are at least observable in the trace output. The recovery behavior (continue with empty list) is acceptable for a non-critical background task, but silent swallowing makes debugging harder.

---

| Criterion | Status |
|---|---|
| Schema created → skill description updated | Covered by implementation and integration test |
| Schema deleted → type removed from description | Event handling correct; **no integration test for delete path** |
| Skill embedding regenerated after update | Correct by design — `update_node_unchecked` emits `NodeUpdated` which triggers embedding pipeline |
| Semantic search for "create a new {type}" matches | Not unit-testable here, but the mechanism is correct |
| Fully reactive, no manual intervention | Yes |

The missing integration test for the delete path is the only acceptance criterion gap. Worth adding a `test_skill_updater_removes_deleted_schema` test.

---

### Summary

The core design is correct and the implementation is clean. The two most impactful items before merge are: (1) the missing proactive startup sync — schemas present on boot will be silently ignored until the next mutation, and (2) the `s.id` vs `s.content` question for description quality. The weak OR assertion in the multiple-schema integration test and the missing delete-path test are also worth fixing to ensure the test suite actually validates the stated behavior.

Not a blocker overall given the greenfield context, but these are all straightforward fixes.

PR #1091, opened 2026-04-10
Fix e2e tool coverage gaps in ollama_integration tests and search_nodes prompt bias

### LINKED ISSUE #1083: Fix e2e tool coverage gaps in ollama_integration tests and search_nodes prompt bias
## Overview

The ollama integration tests in \`packages/agent/tests/ollama_integration.rs\` have significant coverage gaps: most tests only assert that the model didn't call a *wrong* tool (whitelist enforcement), but never assert that the *correct* tool was actually called. This means regressions where the model stops calling a specific tool go completely undetected.

A concrete regression is already happening: the model never calls \`search_nodes\` — it always reaches for \`search_semantic\` instead. This is caused by a prompt imbalance introduced in commit \`1b833491\` where \`search_semantic\` received 4 lines of detailed guidance (including a \`DO NOT call search_nodes\` phrase in context) while \`search_nodes\` kept its single thin line.

## Problem Statement

**Observed regression:**
- User asks agent to find nodes by title/keyword
- Agent calls \`search_semantic\` repeatedly and loops to max iterations
- \`search_nodes\` is never called despite being the correct tool for keyword/title lookups

**Root cause (prompt_assembler.rs line 259):**
The Tool Strategy Guide added in \`1b833491\` says:
> "If the top result has a non-empty 'markdown' field, that IS the full document — use it directly to answer. Do NOT call get_node or search_nodes afterward."

The model reads this and infers \`search_semantic\` is always preferred. \`search_nodes\` has no equivalent guidance explaining when to prefer it.

**Test coverage gaps — tools with no "must be called" assertion:**

| Tool | Test | Problem |
|------|------|---------|
| \`search_nodes\` | \`test_skill_pipeline_research_real_model\` | Accepts \`search_semantic\` OR \`search_nodes\` — semantic always wins |
| \`search_semantic\` | Same | Allowed but never required |
| \`get_node\` | All tests | Allowed as fallback, never required |
| \`get_related_nodes\` | relationship test | Prompt uses hardcoded IDs so model skips straight to \`create_relationship\` |
| \`find_skills\` | None | Zero tests |
| \`update_node\` | graph editing test | Accepts \`update_task_status\` OR \`update_node\` — task status always wins |

**Tests with weak assertions:**
- \`test_skill_pipeline_research_real_model\` — comment says "both are valid", no assertion a search was called at all
- \`test_skill_pipeline_node_creation_real_model\` — no assertion \`create_node\` was actually called
- \`test_skill_pipeline_node_deletion_real_model\` — no assertion \`delete_node\` was actually called
- \`test_skill_pipeline_graph_editing_real_model\` — prompt uses placeholder ID \`abc-123\` (violates prompt rules, unpredictable behavior)

## Proposed Solution

1. Fix the prompt bias in \`packages/agent/src/prompt_assembler.rs\`: balance the \`search_nodes\` guidance to explain when it should be preferred (keyword/title match, known node name, type filter)
2. Add dedicated tests that *require* each tool to be called with scenario-appropriate prompts
3. Replace weak "whitelist-only" assertions with "must call this tool" assertions where applicable

## Acceptance Criteria

- [ ] \`prompt_assembler.rs\` Tool Strategy Guide gives \`search_nodes\` clear guidance: use for exact title/keyword match or when filtering by type
- [ ] New test: \`test_pipeline_search_nodes_keyword\` — prompt like "find the task called 'Review quarterly report'" asserts \`search_nodes\` was called
- [ ] New test: \`test_pipeline_get_related_nodes\` — prompt like "what nodes are connected to X?" asserts \`get_related_nodes\` was called
- [ ] New test: \`test_pipeline_update_node_content\` — prompt like "update the title of the 'Architecture Notes' node to 'System Architecture'" asserts \`update_node\` was called (not \`update_task_status\`)
- [ ] New test: \`test_pipeline_get_node_by_id\` — prompt like "show me the full content of node abc-123" asserts \`get_node\` was called
- [ ] \`test_skill_pipeline_research_real_model\` strengthened: assert at least one search tool was called (not just "text is also fine")
- [ ] \`test_skill_pipeline_node_creation_real_model\` strengthened: assert \`create_node\` was called
- [ ] \`test_skill_pipeline_node_deletion_real_model\` strengthened: assert \`delete_node\` was called
- [ ] \`test_skill_pipeline_graph_editing_real_model\` prompt fixed: remove placeholder ID \`abc-123\`, use a realistic search-then-update scenario
- [ ] All new and updated tests pass with the local Ministral-3B model

### Files to Modify

- **Prompt assembler**: \`packages/agent/src/prompt_assembler.rs\` — Tool Strategy Guide seed prompt (lines 254–266), balance \`search_nodes\` vs \`search_semantic\` guidance
- **Integration tests**: \`packages/agent/tests/ollama_integration.rs\` — add new pipeline scenario tests, strengthen existing assertions

### Guidance Fix Pattern

Current (biased):
\`\`\`
- To find nodes by meaning/topic: use search_semantic (natural language query)
- [4 lines of search_semantic detail including "Do NOT call search_nodes afterward"]
- To find nodes by exact fields: use search_nodes (keyword + type filter)
\`\`\`

Target (balanced):
\`\`\`
- To find nodes by exact title/keyword or filter by type: use search_nodes
- To find nodes by meaning/topic when the exact name is unknown: use search_semantic
- [guidance for each without implying the other should never be used]
\`\`\`

### Test Pattern

New tests should follow the existing \`test_pipeline_*\` pattern (full \`LocalAgentService\` with \`StubToolExecutor\`), not the \`test_skill_pipeline_*\` pattern (skill-scoped inference only), since we want to test the full intent→tool flow.

## Related Issues

- #1081 (introduced the prompt imbalance in search_semantic guidance)

malibio:
Closes #1083

### DIFF
--- a/packages/agent/src/agent_types.rs
+++ b/packages/agent/src/agent_types.rs
@@ -540,6 +540,10 @@ pub struct AgentSession {
     /// Built once per session on first turn, then reused.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub dynamic_context: Option<String>,
+    /// Full system prompt override (bypasses PromptAssembler / fallback).
+    /// Used in tests to inject a pre-built prompt without a live database.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub system_prompt_override: Option<String>,
 }
 
 /// Result of a complete agent turn (one round of generation + tool execution).
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -269,6 +269,11 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                 let final_response = if normalized.is_empty() && !all_tool_executions.is_empty() {
                     let tool_name = &all_tool_executions.last().unwrap().name;
                     format!("Done — {} completed successfully.", tool_name)
+                } else if normalized.is_empty() {
+                    // Model returned nothing at all — no tools, no text. Surface a
+                    // visible error so the UI doesn't go blank.
+                    tracing::warn!("Agent returned empty response with no tool calls");
+                    "I wasn't able to process that request. Could you try rephrasing?".to_string()
                 } else {
                     normalized
                 };
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -123,8 +123,10 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
             workspace_context: dynamic_ctx.to_string(),
         };
 
-        // Build base prompt (async path for assembler, sync fallback otherwise)
-        let base_prompt = if let Some(ref assembler) = self.prompt_assembler {
+        // Build base prompt: use override (tests), assembler (production), or fallback.
+        let base_prompt = if let Some(ref override_prompt) = session.system_prompt_override {
+            override_prompt.clone()
+        } else if let Some(ref assembler) = self.prompt_assembler {
             assembler
                 .assemble(&template_ctx, all_tools.clone())
                 .await
@@ -330,7 +332,8 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                     tool = %tc.function_name,
                     is_error,
                     duration_ms,
-                    result_preview = %result_value.to_string().chars().take(200).collect::<String>(),
+                    args_preview = %args.to_string().chars().take(300).collect::<String>(),
+                    result_preview = %result_value.to_string().chars().take(300).collect::<String>(),
                     "Tool executed"
                 );
 
@@ -429,8 +432,20 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                 on_status(LocalAgentStatus::Idle);
                 session.status = LocalAgentStatus::Idle;
 
+                // Both final inference and last iteration returned empty — synthesize
+                // a summary from tool results so the UI always gets a response.
+                let fallback = if !all_tool_executions.is_empty() {
+                    let summary: Vec<String> = all_tool_executions
+                        .iter()
+                        .map(|t| format!("• {} completed", t.name))
+                        .collect();
+                    summary.join("\n")
+                } else {
+                    normalize_response(&response_text)
+                };
+
                 return Ok(AgentTurnResult {
-                    response: normalize_response(&response_text),
+                    response: fallback,
                     tool_calls_made: all_tool_executions,
                     usage: total_usage,
                 });
@@ -651,6 +666,7 @@ impl<E: ChatInferenceEngine + ?Sized + 'static, T: AgentToolExecutor + ?Sized + 
             created_at: chrono::Utc::now(),
             tool_executions: Vec::new(),
             dynamic_context: None,
+            system_prompt_override: None,
         };
 
         let cancel = CancellationToken::new();
@@ -674,6 +690,18 @@ impl<E: ChatInferenceEngine + ?Sized + 'static, T: AgentToolExecutor + ?Sized + 
         let mut sessions = self.sessions.write().await;
         if let Some(session) = sessions.get_mut(session_id) {
             session.dynamic_context = Some(context);
+        }
+    }
+
+    /// Override the full system prompt for a session.
+    ///
+    /// When set, this bypasses both `PromptAssembler` and `fallback_system_prompt`.
+    /// Intended for integration tests that want to inject a pre-built prompt
+    /// (constructed via `PromptAssembler::assemble_static`) without a live database.
+    pub async fn set_system_prompt(&self, session_id: &str, prompt: String) {
+        let mut sessions = self.sessions.write().await;
+        if let Some(session) = sessions.get_mut(session_id) {
+            session.system_prompt_override = Some(prompt);
         }
     }
 
@@ -963,6 +991,7 @@ mod tests {
             created_at: chrono::Utc::now(),
             tool_executions: Vec::new(),
             dynamic_context: None,
+            system_prompt_override: None,
         }
     }
 
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -1139,11 +1139,17 @@ impl GraphToolExecutor {
         use nodespace_core::mcp::handlers::schema::handle_update_schema;
         let ns = self.node_service()?;
 
-        let result = handle_update_schema(&ns, args)
-            .await
-            .map_err(|e| ToolError::ExecutionFailed(format!("update_schema failed: {:?}", e)))?;
+        let result = handle_update_schema(&ns, args).await;
 
-        Ok(ok_result(tool_call_id, "update_schema", result))
+        match result {
+            Ok(value) => Ok(ok_result(tool_call_id, "update_schema", value)),
+            Err(e) => {
+                // Return validation errors as tool errors (not ToolError::ExecutionFailed)
+                // so the model sees the message and can self-correct.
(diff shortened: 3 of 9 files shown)


malibio:
## Code Review — PR #1091 (Initial Review)

---

### Requirements Check

|---|---|
| `prompt_assembler.rs` Tool Strategy Guide gives `search_nodes` clear guidance (exact title/keyword, type filter) | ✅ Implemented — balanced prompt with dedicated `search_nodes` and `search_semantic` bullets |
| New test `test_pipeline_search_nodes_keyword` — asserts `search_nodes` called | ✅ Added |
| New test `test_pipeline_get_related_nodes` — asserts `get_related_nodes` called | ✅ Added |
| New test `test_pipeline_update_node_content` — asserts `update_node` called (not `update_task_status`) | ✅ Added |
| New test `test_pipeline_get_node_by_id` — asserts `get_node` called | ✅ Added |
| `test_skill_pipeline_research_real_model` strengthened: assert a search tool was called | ✅ Fixed |
| `test_skill_pipeline_node_creation_real_model` strengthened: assert `create_node` called | ✅ Fixed |
| `test_skill_pipeline_node_deletion_real_model` strengthened: assert `delete_node` called | ✅ Fixed |
| `test_skill_pipeline_graph_editing_real_model` prompt fixed: no placeholder ID, realistic scenario | ✅ Fixed |
| Code passes `bun run quality:fix` | ✅ Formatting-only changes in `db.rs`/`local_agent.rs` confirm this was run |

All acceptance criteria are addressed.

---

### Code Review Findings

#### [Improvement] `system_prompt_override` on a public serializable struct is test infrastructure embedded in production code

`📁 packages/agent/src/agent_types.rs:543-546`

`AgentSession` is serialized and returned to the Tauri frontend via `local_agent_get_sessions`. The `system_prompt_override` field therefore appears in the session JSON exposed to the UI layer — and even though `skip_serializing_if = "Option::is_none"` keeps it absent in the normal case, a caller could theoretically _deserialize_ a crafted session (if sessions are ever persisted and reloaded) with an injected prompt override. More immediately, the docstring "Used in tests" on a field of a public production type is an architectural smell: test concerns are leaking into the core domain model.

The conventional Rust pattern here is to keep test-only override capability in the service layer rather than the session struct — e.g., a `#[cfg(test)]` extension method or a separate `TestLocalAgentService` wrapper. This would keep `AgentSession` free of test scaffolding and eliminate the surface area entirely.

That said, the field carries `serde(default, skip_serializing_if = "Option::is_none")` and there is no current persistence path, so the practical risk is low. It is a design concern worth tracking rather than a blocker.

#### [Improvement] `assemble_static` silently diverges from production prompt assembly in two ways

`📁 packages/agent/src/prompt_assembler.rs:236-258`

Two subtle differences versus the live `assemble()` path:

1. **Tool schemas are absent.** The live assembler stores tools in `AssembledPrompt::tool_schemas` and passes them to the inference engine. `assemble_static` returns a plain `String` with no tool definitions. The tests that call it then inject tool definitions separately via `StubToolExecutor`. This works today, but a future refactor that moves tool wiring into the assembler would silently break all static-assembled test contexts without a compile error.

2. **Hard-coded `current_date: "2025-01-01"`.** The live prompt injects `chrono::Utc::now()`. The static path uses a fixed past date. Any prompt logic that branches on date (now or in the future) will produce different results in tests than in production. Passing `current_date` as a parameter (even `Option<&str>` with a default) would make the divergence explicit.

Neither issue affects the tests today, but both are latent traps. The return type difference (`String` vs. `AssembledPrompt`) is the right call for now — just worth noting.

#### [Improvement] `exec_create_schema` error-handling asymmetry with `exec_update_schema`

`📁 packages/agent/src/local_agent/tools.rs:504-512` vs `523-530`

`exec_create_schema` now catches `MCPError` and converts it to `error_result` (a tool-level error the model can see and self-correct from). `exec_update_schema` maps errors to `ToolError::ExecutionFailed`, which propagates as an `Err` to the caller — a different failure mode for an adjacent operation. If `handle_update_schema` can return user-correctable errors (e.g., schema not found, invalid field), those will not reach the model as actionable feedback. Applying the same `Ok(error_result(...))` pattern to `exec_update_schema` would make both consistent.

#### [Nit] `node_uri` / `strip_node_uri` helpers are private but live in a production module

`📁 packages/agent/src/local_agent/tools.rs:159-173`

These helpers are not tested directly. Adding a small unit test for the round-trip (`node_uri` → `strip_node_uri` → original) alongside the existing `definitions_count` test would prevent accidental breakage and takes three lines.

#### [Nit] Fallback response at max-iterations uses bullet-point tool names as UI text

`📁 packages/agent/src/local_agent/agent_loop.rs:442-447`

```rust
format!("• {} completed", t.name)
```

This exposes internal tool names (e.g., `• create_relationship completed`) directly to the end user. For a first pass this is fine, but it is worth a follow-up to humanize (or suppress) these when the context makes them confusing.

#### [Nit] `test_pipeline_schema_creation_title_template_fields` uses manual byte-walking to parse `{token}` from a template string

`📁 packages/agent/tests/ollama_integration.rs:1105-1125`

The byte-walker is correct for the ASCII subset being parsed here, but a regex (`r"\{(\w+)\}"`) or a simple `split('{')` chain would be more readable and sidestep the off-by-one risk if a template ever contains a `}` without a matching `{`. Since this is test code only, it is a minor readability nit.

---

### Summary

This is a high-quality PR. The core changes are exactly right: the prompt imbalance that caused `search_nodes` to be ignored is fixed with clear, symmetric guidance; the `title` / `body` split that was forcing the model to manage two fields is replaced by a single `content` field; `nodespace://` URI normalization is applied consistently across all tool executors; and the integration tests now have meaningful "must call" assertions instead of pure whitelist checks.

The `schema_creation_context` helper refactor — pulling directly from `PromptAssembler::assemble_static` and `SkillPipeline::seed_skill_nodes` instead of duplicating strings — is a particularly good call: prompt drift between tests and production is now a compile-time structural concern rather than a manual sync task.

The two improvements above are the most worth tracking before this pattern spreads: the test-only override field on a serialized production struct, and the error-handling asymmetry between `create_schema` and `update_schema`. Neither is a blocker for merging, but both are worth a follow-up issue.

**Recommendation**: APPROVE

The change is a net positive across correctness, test coverage, and model behavior. The noted concerns are all `[Improvement]` or `[Nit]` level — they do not affect current correctness or security and should not block the merge.

malibio (commented):
Pragmatic code review completed. Detailed findings posted as inline comment. Recommendation: APPROVE — all acceptance criteria met, no blocking issues found.

malibio:
## Review Feedback Addressed

Commit: 4e516eb3

### Implemented

**[Important] exec_update_schema error handling consistency**
Applied the same `Ok(error_result(...))` pattern from `exec_create_schema` to `exec_update_schema`. Validation/MCP errors are now returned as tool-visible error responses rather than propagated as `ToolError::ExecutionFailed`, enabling model self-correction on schema update failures.

**[Suggestion] assemble_static current_date parameter**
`PromptAssembler::assemble_static` now accepts `current_date: Option<&str>` (defaults to `"2025-01-01"`). The intentional divergence from production `assemble()` is now explicit rather than silent. Updated `ollama_integration.rs` call site to pass `None`.

**[Suggestion] node_uri / strip_node_uri round-trip tests**
Added three unit tests in `tools.rs`:
- `node_uri_round_trip` — encode then decode returns the bare ID
- `node_uri_idempotent` — already-prefixed URIs are unchanged
- `strip_node_uri_no_prefix` — bare IDs pass through unchanged

### Deferred with Follow-up Issues

**[Important] system_prompt_override on AgentSession** → Issue #1093
`#[cfg(test)]` on struct fields does not propagate to integration test binaries in `tests/` — the library is compiled in non-test mode when linked from `ollama_integration.rs`. Gating the field caused a compilation failure in integration tests. The field already has `skip_serializing_if = "Option::is_none"` and is never set in production sessions. Issue #1093 tracks a proper solution (test feature flag, separate test service wrapper, or test-facing constructor).

**[Suggestion] Humanize tool names in max-iterations fallback** → Issue #1092
UI-polish concern outside this PR's scope. Tracked for a follow-up.

PR #1095, opened 2026-04-11
Fix: skill node properties bypass namespace normalization, preventing max_iterations updates via MCP

### LINKED ISSUE #1080: Fix: skill node properties bypass namespace normalization, preventing max_iterations updates via MCP

malibio:
Closes #1080

### DIFF
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -1462,12 +1462,12 @@ mod tests {
     }
 
     #[test]
-    fn create_node_schema_requires_title_and_type() {
+    fn create_node_schema_requires_content_and_type() {
         let def = def_create_node();
         let required = def.parameters_schema["required"]
             .as_array()
             .expect("required must be array");
-        assert!(required.contains(&json!("title")));
+        assert!(required.contains(&json!("content")));
         assert!(required.contains(&json!("node_type")));
     }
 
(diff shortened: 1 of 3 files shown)


malibio:
## Review addressed

✅ **Addressed**: Added regression test `get_prop_namespace_overrides_stale_flat` that directly documents the bug scenario — flat `{"max_iterations": 2}` coexisting with namespaced `{"skill": {"max_iterations": 4}}` after an MCP update.

⏭️ **Skipped** (with reasoning):
- **Cross-crate duplication** (`get_prop` / `get_namespaced_prop`): Structural constraint — `core` and `agent` are separate crates with different dependency directions. Consolidation would require a new shared utility crate, which is out of scope for this bug fix.
- **Doc comment nit on `get_namespaced_prop_str`**: Trivial; existing comment is accurate.

PR #1096, opened 2026-04-11
Unify skill/prompt seeding with markdown import pipeline

### LINKED ISSUE #1056: Unify skill/prompt seeding with markdown import pipeline
## Overview

Skill and prompt seed data is currently defined as Rust structs (`SeedSkill`, `SeedPrompt`, `SeedGuidancePrompt`) with manual `to_node()` conversions and custom parent-child creation logic in `db.rs`. This duplicates the hierarchy creation that `create_nodes_from_markdown` already handles well.

## Problem

- Two separate code paths for creating root nodes with children: markdown import and programmatic seeding
- Adding a new skill requires Rust code changes and recompilation
- The markdown import already handles hierarchy, ordering, and batch creation efficiently
- Skills are structurally identical to markdown documents: a root node with child content nodes

## Proposed Solution

Extend `create_nodes_from_markdown` (or add a parallel `create_nodes_from_template`) to support:

1. **Root node type override** — caller specifies `node_type: "skill"` instead of inferring from markdown syntax
2. **Root properties injection** — caller passes `description`, `tool_whitelist`, `max_iterations` etc.
3. **Child type override** — children created as `prompt` nodes with `priority`/`source`/`template_syntax` instead of `text`/`header`

This would allow seed skills to be defined as markdown files (or embedded strings):
```markdown
# Schema Creation

To create a new schema, use create_node with node_type="schema".
Fields are defined in properties.fields as an array.

## Example Structure
{json example}
```

With metadata passed alongside:
```json
{
  "markdown_content": "...",
  "root_node_type": "skill",
  "root_properties": {"description": "...", "tool_whitelist": ["create_node"]},
  "child_node_type": "prompt",
  "child_properties": {"source": "built-in", "template_syntax": "plain"}
}
```

## Benefits

- Single code path for hierarchy creation
- Skills definable as markdown (no Rust changes for new skills)
- Future: users create skills by writing markdown in the UI
- Eliminates `SeedSkill`, `SeedGuidancePrompt` structs and custom seeding logic

## Current State

- `SeedSkill` + `SeedGuidancePrompt` in `packages/agent/src/skill_pipeline.rs`
- `SeedPrompt` in `packages/agent/src/prompt_assembler.rs`  
- Custom seeding in `packages/desktop-app/src-tauri/src/commands/db.rs`
- Markdown import in `packages/core/src/mcp/handlers/markdown.rs`

## Dependencies

- #1049, #1050, #1051 (merged — prompt/skill node infrastructure)

malibio:
Closes #1056

### DIFF
--- a/packages/agent/src/prompt_assembler.rs
+++ b/packages/agent/src/prompt_assembler.rs
@@ -9,6 +9,7 @@
 
 use std::sync::Arc;
 
+use nodespace_core::mcp::handlers::markdown::NodeTemplate;
 use nodespace_core::models::Node;
 use nodespace_core::services::NodeService;
 
@@ -235,7 +236,12 @@ impl PromptAssembler {
     /// Intended for use in unit/integration tests where no DB is available.
     pub fn assemble_static(workspace_context: &str, current_date: Option<&str>) -> String {
         let mut seeds = Self::seed_prompt_nodes();
-        seeds.sort_by_key(|s| s.priority);
+        seeds.sort_by_key(|s| {
+            s.root_properties
+                .get("priority")
+                .and_then(|v| v.as_i64())
+                .unwrap_or(0)
+        });
 
         let ctx = TemplateContext {
             current_date: current_date.unwrap_or("2025-01-01").to_string(),
@@ -246,10 +252,17 @@ impl PromptAssembler {
         let sections: Vec<String> = seeds
             .iter()
             .map(|s| {
-                if s.template_syntax == "minijinja" {
-                    Self::render_template(&s.content, &ctx)
+                // Prompt content is in `content` field (overrides title)
+                let body = s.content.as_deref().unwrap_or(&s.markdown_content);
+                let syntax = s
+                    .root_properties
+                    .get("template_syntax")
+                    .and_then(|v| v.as_str())
+                    .unwrap_or("plain");
+                if syntax == "minijinja" {
+                    Self::render_template(body, &ctx)
                 } else {
-                    s.content.clone()
+                    body.to_string()
                 }
             })
             .collect();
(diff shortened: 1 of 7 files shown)


malibio:
## Code Review — PR #1096: Unify skill/prompt seeding with markdown import pipeline

**Branch**: \`feature/issue-1056-unify-seed-with-markdown-pipeline\`
**Reviewed commit**: d723da8
**Initial review** (no prior reviews on record)

---

### Summary

This PR delivers a clean unification of three ad-hoc seeding structs (`SeedSkill`, `SeedGuidancePrompt`, `SeedPrompt`) into a single `NodeTemplate` abstraction backed by the existing `prepare_nodes_from_markdown` pipeline. The net result is a meaningful reduction in code surface (~130 lines deleted, a single well-documented insertion path, and direct reuse of the battle-tested markdown parser for hierarchy creation.

The architectural intent is correct and well-executed. There is one functional regression worth addressing before merge, one minor structural concern, and a handful of nits.

---

### Findings

#### [Improvement] Skill guidance content drift is now silently ignored

**File**: `packages/desktop-app/src-tauri/src/commands/db.rs`

The old seeding block included a per-guidance-prompt diff-check loop: it queried each skill node's children, compared content against the seed string, and called `update_node_unchecked` if they differed. This kept guidance prompt text current across application updates on existing installations.

The new code drops this loop entirely. Only root `prompt` nodes (from `prompt_templates`) are diff-checked. Skill guidance children — the markdown bodies stored under skill roots — are never updated after first seed. If someone changes the "Research & Search Guidance" or "Schema Creation Guidance" markdown content in a future release, existing users will silently keep the old text forever.

The issue description says "Eliminates `SeedGuidancePrompt` structs and custom seeding logic," which could be read as intentionally removing this, but the seeding block comment still says _"Built-in prompt content is also diff-checked so changes to the seed strings propagate to existing installations"_ — a contract it only fulfills for half the seeded data.

**Options:**
1. Extend the diff-check loop to also walk `skill_templates`, query their existing guidance children by `(parent_skill_id, title)`, and update content if changed. This preserves the previous behavior.
2. Accept the regression explicitly and update the comment to remove the false promise.

Given this is a greenfield project with no live installations, option 2 is acceptable in the short term, but the misleading comment needs to be corrected now.

---

#### [Improvement] Stale test comment contradicts the assertion it introduces

**File**: `packages/agent/src/skill_pipeline.rs`, line 702

The comment reads:
```
// --- Skill whitelist tests for all 7 skills ---
```

But line 483 asserts `seeds.len() == 8` and the section covers 8 skills. This comment was likely copied from a previous count and then not updated when "Organization" was added. It will confuse anyone reading the test file.

---

#### [Improvement] `child_node_type` / `child_properties` are populated on skills with empty `markdown_content`

**File**: `packages/agent/src/skill_pipeline.rs`

Six of the eight skill templates (Node Creation, Graph Editing, Relationship Management, Node Deletion, Bulk Import, Organization) have `markdown_content: String::new()` but still specify `child_node_type: Some("prompt"...)` and `child_properties: Some(...)`. Because `prepare_nodes_from_template` guards the child-parse path with `if tmpl.markdown_content.trim().is_empty() { Vec::new() }`, these fields are dead on those instances — the overrides are built and then never applied.

This is harmless but it clutters every skill literal with three extra fields that do nothing. Either:
- Set `child_node_type: None` and `child_properties: None` on skills that have no guidance, matching the actual runtime behavior, or
- Add a `Default` impl to `NodeTemplate` so callers only set what they need (reduces boilerplate significantly across the whole file).

---

#### [Nit] Doc example omits the `content` field

**File**: `packages/core/src/mcp/handlers/markdown.rs`, line 1256

The ````ignore` example in `NodeTemplate`'s doc comment is missing the `content` field (which is part of the struct). Since `content` has a meaningful semantic role (overrides `title` as the node body), showing `content: None` in the skill example and a separate `content: Some(...)` snippet for the prompt use-case would make the struct's dual usage immediately clear.

---

#### [Nit] `prepare_nodes_from_template` has no direct unit tests in `markdown_test.rs`

**File**: `packages/core/src/mcp/handlers/markdown_test.rs`

The function is exercised indirectly through the agent-level tests (`seed_skill_template_produces_skill_node`, `seed_prompt_template_produces_prompt_node`), which are adequate for correctness of the happy path. However, edge cases specific to the new function — empty `markdown_content` with `child_properties` set, non-object `child_properties` triggering the replace-entirely branch, `root_properties` being non-object — are not covered. Given the function is `pub` and will be the unified entry point for all future seeding, a small dedicated test block directly in `markdown_test.rs` would improve confidence and serve as specification documentation.

---

### What is done well

- **Clean deletion**: All three old structs and their `to_node()` / `guidance_nodes()` impls are fully removed. No `#[deprecated]`, no `#[allow(dead_code)]`, no transitional shims. This is exactly the right pattern for this codebase.
- **Type-level idempotency is preserved**: `seed_nodes_from_templates` uses the same HashSet-keyed-by-root-`node_type` strategy as the old `seed_nodes_if_needed`, so the semantics are unchanged for fresh installs.
- **`assemble_static` is functionally equivalent**: The `priority` sort and `template_syntax` dispatch both correctly migrate from typed fields to JSON property lookups, and the fallback chain (`content → markdown_content`) correctly resolves for all six prompt seeds (all of which supply `content: Some(...)`).
- **Integration tests updated correctly**: `tools_for_skill` and `build_schema_creation_prompt` in `ollama_integration.rs` both migrate cleanly to the new field paths.
- **Tracing is improved**: The new logging separates `created_roots` from `created_children`, which will make debugging partial-seed failures noticeably easier.
- **Documentation quality**: `NodeTemplate` has clear field-level docs and a working ````ignore` example. The module-level block comment explaining what was replaced and why is appropriate.

---

### Verdict

**Request changes** on the stale comment (the diff-check comment in `db.rs` claims to cover guidance content propagation but no longer does). The remaining items are improvements and nits that do not block correctness for a zero-user installation, but the misleading comment is a maintenance hazard that should be fixed before merge.

malibio:
## Address Review Summary

All reviewer recommendations addressed. Changes pushed in commit `52cb648a`.

### ✅ Addressed (4 items)

**🟡 Fix stale comment in skill_pipeline.rs**
`packages/agent/src/skill_pipeline.rs` — Updated `// --- Skill whitelist tests for all 7 skills ---` → `// --- Skill whitelist tests for all 8 skills ---`

**🟡 Fix misleading comment in db.rs**
`packages/desktop-app/src-tauri/src/commands/db.rs` — Clarified that only prompt root nodes are diff-checked after initial seed; skill guidance children are intentionally left as user-editable (not a gap in the implementation).

**🟢 Improve NodeTemplate doc example**
`packages/core/src/mcp/handlers/markdown.rs` — Added `content: None` to skill example and a second example showing the `content: Some(...)` prompt pattern, making the content override behavior self-documenting.

**🟢 Add unit tests for `prepare_nodes_from_template`**
`packages/core/src/mcp/handlers/markdown_test.rs` — Added 7 unit tests covering:
- No-child skill (root only)
- Skill with guidance children (markdown_content parsed into children)
- `child_properties` merge into parsed nodes
- Prompt with `content: Some(...)` override
- Title fallback when `content: None`
- Non-object `child_properties` replacement
- `root_properties` injection on root node

### ⏭️ Skipped (1 item)

**🟢 Set `child_node_type`/`child_properties` to `None` on childless skill templates**
Skipped intentionally. The fields are already `None` on childless templates, and making the omission "explicit" by typing `child_node_type: None, child_properties: None,` on all 8 templates adds boilerplate without adding information. The struct derives have no default, so `None` must be explicit anyway — Rust's `..Default::default()` pattern isn't available here. The recommendation was subjective style; skipping in favor of current ergonomics.

### Test Status
All tests pass: 1170 core + 284 agent = 1454 total (no regressions).

PR #1110, opened 2026-05-17
Fix: schema nodes always get normalized ID from content, not UUID

malibio:
## Summary

- Schema nodes created via `create_node` MCP tool or local agent were receiving UUIDs instead of normalized name IDs (e.g. `"Invoice"` → `"invoice"`)
- Root cause: `normalize_schema_id` only lived in `handle_create_schema`; the `create_node` path had `id: None` hardcoded, falling through to `uuid::Uuid::new_v4()`
- Fix: moved `normalize_schema_id` to `node_service::create_node_with_parent` — the single shared layer all creation paths pass through — same pattern as date nodes using content as ID

## Test plan

- [ ] New regression test: `test_create_node_schema_gets_normalized_id` — verifies `handle_create_node` with `node_type="schema"` produces a normalized ID, not a UUID
- [ ] New unit tests: `test_normalize_schema_id_basic` and `test_normalize_schema_id_edge_cases` in `node_service`
- [ ] Existing `test_normalize_schema_id` in `schema_test.rs` now calls through to the service function
- [ ] Verified e2e via live MCP: `create_node(node_type="schema", content="Widget")` → `id: "widget"`

### DIFF
--- a/packages/core/src/mcp/handlers/nodes_test.rs
+++ b/packages/core/src/mcp/handlers/nodes_test.rs
@@ -2028,4 +2028,42 @@ mod property_filter_tests {
         let nodes = result["nodes"].as_array().unwrap();
         assert_eq!(nodes.len(), 0);
     }
+
+    // Regression test: schema nodes created via create_node must get a normalized name ID,
+    // not a UUID. Previously, only handle_create_schema applied normalization; create_node
+    // with node_type="schema" fell through to uuid::Uuid::new_v4().
+    #[tokio::test]
+    async fn test_create_node_schema_gets_normalized_id() {
+        use crate::mcp::handlers::nodes::handle_create_node;
+
+        let (ns, _dir) = setup_test_service().await.unwrap();
+
+        let result = handle_create_node(
+            &ns,
+            json!({
+                "node_type": "schema",
+                "content": "Customer Profile",
+                "properties": {
+                    "isCore": false,
+                    "schemaVersion": 1,
+                    "description": "Test",
+                    "fields": [],
+                    "relationships": []
+                }
+            }),
+        )
+        .await
+        .unwrap();
+
+        let node_id = result["node_id"].as_str().unwrap();
+        assert_eq!(
+            node_id, "customer-profile",
+            "Schema node ID should be normalized name, not UUID"
+        );
+
+        // Verify the node is retrievable by the normalized ID
+        let node = ns.get_node("customer-profile").await.unwrap();
+        assert!(node.is_some(), "Should be fetchable by normalized ID");
+        assert_eq!(node.unwrap().node_type, "schema");
+    }
 }
--- a/packages/core/src/mcp/handlers/schema.rs
+++ b/packages/core/src/mcp/handlers/schema.rs
@@ -196,9 +196,9 @@ pub async fn handle_create_schema(
         properties["propertiesHeaderSummaryTemplate"] = serde_json::Value::String(template.clone());
     }
 
-    // Create schema node params
+    // Create schema node params — no explicit ID; create_node_with_parent derives it from content
     let schema_node_params = CreateNodeParams {
-        id: Some(schema_id.clone()),
+        id: None,
         node_type: "schema".to_string(),
         content: params.name.clone(),
         parent_id: None,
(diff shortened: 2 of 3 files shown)


malibio:
### Code Review Summary

This is a clean, focused bug fix. The root cause was correctly identified: ID normalization logic was duplicated in `handle_create_schema`, making it invisible to the generic `create_node` path. The fix moves `normalize_schema_id` to `node_service::create_node_with_parent` — the single choke point all creation paths flow through — which is the right abstraction layer. The change is appropriately small (3 files, no architectural sprawl) and ships with a regression test that would have caught the original bug.

---

### Findings

#### Critical Issues

None.

---

#### Suggested Improvements

**[Improvement] `normalize_schema_id` visibility is wider than necessary — consider making it `pub` or scoping it more tightly**

The function is declared `pub(crate)` in `node_service.rs`. Its two production call sites are:
1. `node_service.rs` itself (line 2159, internal call — no qualifier needed)
2. `mcp/handlers/schema.rs` line 169, via fully-qualified `crate::services::node_service::normalize_schema_id`

`pub(crate)` is correct for the cross-module call in `schema.rs`. However, because `handle_create_schema` (line 200-207) still passes an explicit `id: Some(schema_id.clone())` into `CreateNodeParams`, it now normalizes twice: once in `schema.rs` to compute the ID for the duplicate-check and to build `schema_node_params`, and once again inside `create_node_with_parent` via the new `else if` branch. 

The second normalization is actually bypassed because the `if let Some(provided_id) = params.id` branch (line 2138) fires first and passes through the pre-computed ID directly — so there is no double-normalization bug. But the intent is subtle and fragile: if a caller ever passes an un-normalized name as the explicit ID, `create_node_with_parent` will not re-normalize it (because the `provided_id` path takes priority). The contract is currently implicit.

Recommendation: add a brief comment at the `provided_id` branch noting that callers supplying an explicit schema ID are responsible for normalization, or assert/document that `handle_create_schema` is the only legitimate caller of that pattern.

**[Improvement] Missing edge-case test: empty or all-special-character content**

`normalize_schema_id("")" → `""` and `normalize_schema_id("!!!")" → `""`. An empty string as a node ID will silently produce a degenerate schema node (node ID = `""`). The fix currently delegates straight to `normalize_schema_id` without a guard. This is arguably a pre-existing gap in the normalizer itself, but now that it sits on the hot path for schema creation, an explicit `if node_id.is_empty() { return Err(...) }` guard after line 2159 would prevent silent corruption.

---

#### Nitpicks

- **Nit** (`node_service.rs` lines 569-587): The inline test module `normalize_schema_id_tests` is placed between two public function definitions, which interrupts the module's reading flow. Rust convention typically co-locates test modules at the bottom of the file or in a dedicated `*_test.rs` companion. Not a blocker, just a minor readability point.

- **Nit** (`schema.rs` lines 1018-1019): The blank line left after removing `normalize_schema_id` results in a double blank line before the `#[cfg(test)]` block. Cosmetic only.

- **Nit** (`nodes_test.rs` line 2036): The regression test lives inside `mod property_filter_tests`. Semantically it belongs in a `schema_node_tests` module or at least a sibling module, but keeping it here is harmless for a regression test.

---

### Summary

The fix is correct, well-scoped, and at the right layer. The regression test directly exercises the previously-broken code path. The one real improvement worth acting on is a guard against empty/degenerate IDs produced by `normalize_schema_id`, which is now on the hot creation path. Everything else is polish.

**Verdict: Approve with optional follow-up on the empty-ID guard.**

malibio (commented):
Clean fix at the right abstraction layer. Regression test in place. One optional follow-up: add an empty-ID guard after normalize_schema_id in create_node_with_parent to prevent silent degenerate nodes when content is empty or all-special-characters. Not a blocker. LGTM.

malibio:
## Review Feedback Addressed

✅ **Removed explicit `id: Some(schema_id)` from `handle_create_schema`** — all callers now go through the single normalization path in `create_node_with_parent`, eliminating the implicit dual-path contract.

✅ **Added empty-ID guard** — `normalize_schema_id` producing an empty string (e.g. content is `""` or `"!!!"`) now returns a clear error: _"Schema content must not be empty or contain only special characters"_.

⏭️ **Skipped nits** — test module placement is cosmetic with no correctness or readability impact.

PR #1128, opened 2026-05-17
Add clarification for low-confidence skill matches (closes #1090)

### LINKED ISSUE #1090: Ask for clarification when skill confidence is below threshold instead of falling through to full tool set
## Overview

When the skill pipeline finds no match above the confidence threshold (0.8), the agent silently falls through to the general 13-tool path. This overwhelms the model with tools it doesn't need and produces poor or empty responses. Instead, the agent should detect the ambiguous-intent case and ask the user for clarification.

## Problem Statement

**Current behavior:**
1. User sends a short or ambiguous message (e.g., "create")
2. Skill pipeline scores it at 0.754 — below the 0.8 threshold
3. Agent loop falls through to the full tool set (13 tools)
4. Model receives 13 tools + full conversation history and returns an empty or incorrect response
5. User sees a blank UI or a generic fallback message

**Expected behavior:**
1. User sends ambiguous message
2. Skill pipeline finds no confident match
3. Agent detects the low-confidence/no-match case
4. Agent responds with a clarifying question: "Could you tell me more about what you'd like to do?"
5. User clarifies, next turn has enough context for a confident skill match

## Proposed Solution

In `packages/agent/src/local_agent/agent_loop.rs`, before dispatching to the inference engine, check whether a skill was matched. If `skill_match` is `None` (or confidence < threshold) and the user message is short or lacks clear intent, return a clarifying question directly — without calling the model at all.

### Detection heuristics (start simple):
- No skill match AND user message is under ~10 words
- Or: top skill confidence is between 0.5–0.8 (some signal but not enough)

### Clarification response:
Return a short, helpful question as the agent response. No tool calls needed.

## Acceptance Criteria

- [ ] When skill confidence is below threshold and message is ambiguous, agent responds with a clarifying question instead of proceeding with all tools
- [ ] Clarifying question is returned as a normal agent response (visible in UI)
- [ ] When skill confidence is below threshold but message is clearly actionable (long, contains specific nouns), agent still proceeds with full tool set
- [ ] Empty-response fallback in `agent_loop.rs` (lines ~269–274) is unified with or superseded by this clarification path
- [ ] Unit test covering the low-confidence clarification path

### Reference Files
- **Agent loop**: `packages/agent/src/local_agent/agent_loop.rs` — skill dispatch and inference entry point; clarification logic goes before the inference call
- **Skill pipeline**: `packages/agent/src/skill_pipeline.rs` — `SkillMatch` struct holds `confidence` and matched skill; `None` means no match above threshold
- **Empty fallback** (to unify): `packages/agent/src/local_agent/agent_loop.rs` lines ~269–274 — current hardcoded fallback for empty model output

### Implementation Details

The agent loop already receives the `SkillMatch` result before calling inference. The check should be inserted at the point where `skill_match` is inspected:

```rust
// Pseudocode — in agent_loop run_turn or similar
if skill_match.is_none() && is_ambiguous(&user_message) {
    let clarification = "I'm not sure what you'd like to do. Could you give me a bit more detail?";
    return Ok(AgentTurnResult {
        response: clarification.to_string(),
        tool_calls_made: vec![],
        usage: Default::default(),
    });
}
```

`is_ambiguous` can start as: `word_count < 10 && !contains_proper_noun_or_url`.

### Non-Goals
- This does not change the skill confidence threshold (0.8)
- This does not affect turns where a skill IS matched
- This does not add multi-turn clarification state tracking (single clarifying question only, for now)

## Related Issues
- #1089 (unify MCP initialize instructions with local agent seed prompts)

malibio:
## Summary

- When the skill pipeline finds no match above the 0.8 confidence threshold AND the user message is ambiguous (short, no proper nouns / URLs / numbers), return a clarifying question **without invoking the model**.
- Previously, vague inputs like `"help"` or `"do that thing"` fell through to the full 13-tool agent path, overwhelming the model and producing empty/poor responses.
- Unifies the post-inference empty-response fallback (lines ~269-274) with the new clarification path so there's a single consistent "I need more info" response across the agent.

## Implementation

- Added `is_ambiguous(message: &str) -> bool` heuristic in `packages/agent/src/local_agent/agent_loop.rs`:
  - `< 10` words AND no proper-noun / URL / digit signals
  - Sentence-initial capitalization is ignored (carries no signal)
- In `LocalAgentLoop::run_turn`, the skill match is resolved once into a single `Option<SkillMatch>`. If `skill_match.is_none()` and `is_ambiguous(user_message)`, we append a clarifying assistant message and return immediately with zero tool calls / zero token usage.
- The post-inference empty-response fallback now reuses the same `CLARIFYING_QUESTION` constant.
- A handful of existing tests used very short test prompts (`"Hi there"`, `"Search"`, `"Hello"`, etc.) that would now be classified as ambiguous and short-circuited. These were updated to use clearly actionable prompts (with proper nouns or 10+ words) so the ReAct loop is still exercised. This reflects the real behavior change: vague inputs no longer reach inference.

## Test plan

- [x] Unit tests for `is_ambiguous` cover: short-vague, proper-noun, URL, digit, long-message, and sentence-initial-capitalization carve-outs.
- [x] `ambiguous_message_returns_clarification_without_inference` uses a `PanicEngine` to **prove** inference is never invoked for `"help"`.
- [x] `actionable_message_proceeds_to_inference_without_skill_match` is the counter-example: a message containing a proper noun still hits the model.
- [x] All existing 296 `nodespace-agent` lib tests pass.
- [x] Full `bun run test:all` passes: 3918 frontend + 1182 Rust = 5100 tests, 0 failures (matches baseline).
- [x] `bun run quality:fix` clean (eslint, svelte-check, cargo fmt, cargo clippy `-D warnings`).

## Test results vs baseline

| Suite | Baseline | After |
|---|---|---|
| Frontend (vitest) | 3918 passed | 3918 passed |
| Rust (cargo test) | n/a (warm cache required) | 1182 passed |
| Agent crate (focused) | n/a | 296 passed (9 new) |

### DIFF
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -61,6 +61,9 @@ const CLARIFYING_QUESTION: &str =
 ///   - URLs (http/https/www)
 ///   - Proper nouns (capitalized non-leading words, e.g. "GitHub", "Slack")
 ///   - Numbers or dates (which usually indicate a specific reference)
+///   - Code/path-like tokens (containing `(`, `)`, `/`, `.`, `:`, `_`, `-`)
+///     such as `console.log('x')` or `path/to/file.rs` — these are clearly
+///     intentional inputs the user wants the model to process.
 ///
 /// The first word is ignored for capitalization because sentence-initial
 /// capitalization carries no signal. Common pronouns and stop-words at the
@@ -82,10 +85,21 @@ fn is_ambiguous(message: &str) -> bool {
         return false;
     }
 
-    // Look for proper nouns (capitalized word not at position 0) or digits.
+    // Look for specific signals: digits, code/path punctuation, or
+    // proper nouns (capitalized non-leading word).
     for (idx, word) in words.iter().enumerate() {
         // Numbers/dates → specific reference, not ambiguous.
         if word.chars().any(|c| c.is_ascii_digit()) {
+            return false;
+        }
+        // Code/path-like tokens (function calls, file paths, namespaced
+        // identifiers, kebab/snake case). Conservative — errs toward
+        // letting the model handle the input rather than asking for
+        // clarification on something that looks like intentional input.
+        if word
+            .chars()
+            .any(|c| matches!(c, '(' | ')' | '/' | '.' | ':' | '_' | '-'))
+        {
             return false;
         }
         if idx == 0 {
@@ -202,8 +216,17 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
 
         // Resolve the skill match (if any) once, so we can branch on it for
         // both the clarification short-circuit and prompt/tool scoping.
-        let skill_match = match self.skill_pipeline {
-            Some(ref pipeline) => pipeline.find_skill(user_message).await,
+        // Carry the pipeline reference alongside the match so the scoping
+        // branch below has a locally-provable handle to it (no need to
+        // re-unwrap `self.skill_pipeline` and rely on a far-away invariant).
+        let skill_match: Option<(
+            &Arc<crate::skill_pipeline::SkillPipeline>,
+            crate::skill_pipeline::SkillMatch,
+        )> = match self.skill_pipeline.as_ref() {
+            Some(pipeline) => pipeline
+                .find_skill(user_message)
+                .await
+                .map(|m| (pipeline, m)),
             None => None,
         };
 
@@ -240,8 +263,10 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
             });
         }
 
-        let (system_content, tools, effective_max_iterations) = if let Some(ref skill_match) =
-            skill_match
+        let (system_content, tools, effective_max_iterations) = if let Some((
+            pipeline,
+            ref skill_match,
+        )) = skill_match
         {
             tracing::info!(
                 skill = %skill_match.skill.content,
@@ -251,12 +276,9 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                 "Skill matched via push pipeline"
             );
 
-            // Scope tools to skill's whitelist. `scope_tools` lives on the
-            // pipeline, which must exist if we got a skill_match.
-            let pipeline = self
-                .skill_pipeline
-                .as_ref()
-                .expect("skill_match implies skill_pipeline is Some");
+            // Scope tools to skill's whitelist. We have the pipeline
+            // reference locally (carried alongside the match), so no
+            // runtime invariant check is needed.
             let scoped_tools = pipeline.scope_tools(&all_tools, skill_match);
             let skill_max_iter = skill_match.max_iterations;
 
@@ -266,9 +288,9 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                     .unwrap_or("");
 
             let system = format!(
-                "{}\n\nACTIVE SKILL: {}\n{}\nFocus on this skill's capabilities. Use only the tools provided.",
-                base_prompt, skill_name, skill_desc
-            );
+                    "{}\n\nACTIVE SKILL: {}\n{}\nFocus on this skill's capabilities. Use only the tools provided.",
+                    base_prompt, skill_name, skill_desc
+                );
 
             (system, scoped_tools, skill_max_iter)
         } else {
@@ -1124,7 +1146,7 @@ mod tests {
         let result = agent_loop
             .run_turn(
                 &mut session,
-                "please tell me about the GitHub release pipeline today",
+                "Summarize the GitHub release notes for v1.2 in plain English",
                 move |s| {
                     statuses_cb.lock().unwrap().push(s);
                 },
@@ -1161,7 +1183,7 @@ mod tests {
         let result = agent_loop
             .run_turn(
                 &mut session,
-                "search for any Billing related architecture documents",
+                "Search GitHub for open release-blocker issues then summarize them",
                 |_| {},
                 |_| {},
                 CancellationToken::new(),
@@ -1239,7 +1261,7 @@ mod tests {
         let result = agent_loop
             .run_turn(
                 &mut session,
-                "tell me about the Billing Architecture",
+                "Look up the Billing Architecture node then fetch its referenced details",
                 |_| {},
                 |_| {},
                 CancellationToken::new(),
@@ -1292,7 +1314,7 @@ mod tests {
         let result = agent_loop
             .run_turn(
                 &mut session,
-                "keep searching for nodes about NodeSpace architecture",
+                "Keep running search_nodes forever — verify the iteration cap stops it",
                 |_| {},
                 |_| {},
                 CancellationToken::new(),
@@ -1321,7 +1343,7 @@ mod tests {
         let result = agent_loop
             .run_turn(
                 &mut session,
-                "please describe the GitHub release workflow",
+                "Begin generating a long answer about the GitHub release process",
                 |_| {},
                 |_| {},
                 cancel,
@@ -1435,7 +1457,7 @@ mod tests {
         let result = service
             .send_message(
                 &id,
-                "help me draft the GitHub release notes for today",
+                "Send this message to the agent and confirm a GitHub release reply comes back",
                 |_| {},
                 |_| {},
             )
@@ -1492,7 +1514,7 @@ mod tests {
         agent_loop
             .run_turn(
                 &mut session,
-                "please describe the GitHub release workflow today",
+                "Walk through each status transition while answering about GitHub releases",
                 move |s| {
                     let label = match &s {
                         LocalAgentStatus::Idle => "Idle",
@@ -1534,7 +1556,7 @@ mod tests {
         agent_loop
             .run_turn(
                 &mut session,
-                "search for any GitHub release notes documents",
+                "Search GitHub release notes and report each status transition along the way",
                 move |s| {
                     let label = match &s {
                         LocalAgentStatus::Idle => "Idle",
@@ -1615,7 +1637,7 @@ mod tests {
         let result = agent_loop
             .run_turn(
                 &mut session,
-                "what was discussed about Billing earlier today",
+                "Recap the prior Billing conversation after triggering history summarization",
                 |_| {},
                 |_| {},
                 CancellationToken::new(),
@@ -1680,7 +1702,7 @@ mod tests {
         let id = service.create_session(Some("test-model".into())).await;
 
         // send_message should fail because FailingEngine errors
-        let user_msg = "please describe the GitHub release workflow today";
+        let user_msg = "Trigger an inference error and confirm the GitHub session survives intact";
         let result = service.send_message(&id, user_msg, |_| {}, |_| {}).await;
 
         assert!(result.is_err(), "Expected inference error");
@@ -1768,7 +1790,7 @@ mod tests {
         let result = agent_loop
             .run_turn(
                 &mut session,
-                "keep searching for any related NodeSpace architecture documents",
+                "Keep calling search_nodes past the iteration cap to verify enforcement",
                 |_| {},
                 |_| {},
                 CancellationToken::new(),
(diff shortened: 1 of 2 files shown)


malibio:
## Code Review Summary

**Review Type:** Initial review (no prior reviews/comments on PR).

This is a clean, well-scoped change that delivers the requested behavior with strong test coverage. The `is_ambiguous` heuristic is simple, documented, and the short-circuit correctly bypasses inference for vague messages with no skill match. The empty-response fallback is unified with the new clarification path via a shared `CLARIFYING_QUESTION` constant. Recommendation: **APPROVE** with a few minor suggestions noted below.

---

## Requirements Check (Acceptance Criteria from #1090)

| # | Criterion | Status | Evidence |
|---|---|---|---|
| 1 | Low confidence + ambiguous → clarifying question instead of full tools | ✅ | `agent_loop.rs:215-241` short-circuits before inference; `ambiguous_message_returns_clarification_without_inference` test uses `PanicEngine` to prove inference is never invoked. |
| 2 | Clarifying question returned as normal agent response (visible in UI) | ✅ | `agent_loop.rs:223-228` appends Assistant message to `session.messages`; `agent_loop.rs:233-240` returns it as `AgentTurnResult.response`. |
| 3 | Low confidence + actionable → still proceeds with full tools | ✅ | `is_ambiguous` returns `false` for proper nouns / URLs / digits / messages ≥10 words. `actionable_message_proceeds_to_inference_without_skill_match` confirms behavior end-to-end. |
| 4 | Empty-response fallback unified with clarification path | ✅ (partial — see 🟢 below) | `agent_loop.rs:375-381` reuses `CLARIFYING_QUESTION` constant. Note: there is a second, separate post-final-inference fallback at `agent_loop.rs:545-555` that does **not** use the constant — it produces tool-summary text or empty `response_text`. This is a different code path (post-tool-call exhaustion) and was not in scope of the issue's "lines ~269-274", so I consider this acceptable. |
| 5 | Unit test covering low-confidence clarification path | ✅ | 9 new tests: 6 `is_ambiguous_*` heuristic tests + `ambiguous_message_returns_clarification_without_inference` + `actionable_message_proceeds_to_inference_without_skill_match`. All pass locally. |
| 6 | `bun run quality:fix` passes | ✅ | Per PR description; the `schema.rs` blank-line removal is `cargo fmt` output, confirming format was run. |
| 7 | `bun run test:all` passes | ✅ | Per PR description: 3918 frontend + 1182 Rust = 5100 tests, 0 failures, matches baseline. |

All seven acceptance criteria are satisfied.

---

## Findings

### 🔴 Critical / Blocker
None.

### 🟡 Improvements

**1. `agent_loop.rs:256-259` — `expect()` on an internally-maintained invariant is fragile**

```rust
let pipeline = self
    .skill_pipeline
    .as_ref()
    .expect(\"skill_match implies skill_pipeline is Some\");
```

`skill_match` being `Some` does imply `self.skill_pipeline` is `Some` today (see the `match` at line 205), but this couples two code locations via a panic message rather than via the type system. A future refactor that introduces another way to populate `skill_match` (e.g., a fallback skill, a test fixture, a cached result) will silently break this invariant and crash production.

*Engineering principle:* **Make illegal states unrepresentable.** Either keep the nested-`if-let` form from the original code (which had the pipeline-and-match bound together lexically), or carry the pipeline reference inside `SkillMatch` itself so the invariant is enforced by the type, not by an assertion.

Minimal fix: revert this hunk to the prior nested structure and add the short-circuit as a separate pre-check using `pipeline.find_skill(...)` results — the diff cost is small, and the resulting code can't drift.

**2. `agent_loop.rs:68-104` — `is_ambiguous` false-positives on single-token code-like inputs**

The heuristic treats anything ≤9 words with no digit and no non-leading uppercase as ambiguous. That includes inputs like `console.log('x')`, `path/to/file.rs`, `myFunc()`, or any single-token identifier the user pastes for inspection. Those are clearly actionable but would be classified ambiguous and short-circuited to the clarifying question.

*Engineering principle:* **Heuristics should fail open on clearly intentional inputs.** Two cheap signals would catch most cases without changing the philosophy:
- Presence of `(`, `)`, `/`, `.`, `:`, `_`, or `-` in any word → likely a path/identifier/URL-like reference → not ambiguous.
- Any non-ASCII-letter character class (existing `is_ascii_digit` already covers digits).

This is an improvement, not a blocker — the issue explicitly says "start simple" and a short code snippet without context is in fact ambiguous from a *response* perspective; the worst case is a clarifying question, which is recoverable. Worth a small follow-up.

**3. Test prompts updated to bypass `is_ambiguous` — verify intent is preserved**

Nine existing tests had their prompts lengthened or sprinkled with proper nouns (e.g., `\"Hi there\"` → `\"please tell me about the GitHub release pipeline today\"`, `\"Search\"` → `\"search for any GitHub release notes documents\"`). I spot-checked the affected tests:

- `single_turn_no_tools` (line 1110) — still exercises one inference round → final text. ✅
- `tool_call_then_final_response` (line 1150) — still exercises tool call → final text. ✅
- `multi_step_tool_chain` (line 1184) — still exercises 3-round ReAct loop. ✅
- `respects_iteration_cap` (line 1748) — still exercises iteration cap. ✅
- `service_propagates_inference_error` (line 1664) — still asserts user message recorded. ✅

The renames don't mask regressions; they correctly reflect that vague inputs no longer reach inference (which is the product change). However, the new prompts are now indistinguishable from one another (\"GitHub release pipeline\", \"GitHub release workflow\", \"GitHub release notes\") — a small readability cost. Consider giving each test a prompt that gestures at what the test is *about* (e.g., the cancellation test could use `\"cancel my GitHub release task immediately\"`).

*Engineering principle:* **Tests should be self-documenting.** Right now the prompt is noise; a meaningful prompt costs nothing and helps the next reader.

### 🟢 Nits / Observations

**Nit:** `agent_loop.rs:75` — `let lower = trimmed.to_lowercase();` allocates a String on every call. For chat-rate input this is negligible, but a `contains_ignore_ascii_case`-style scan over `trimmed` would avoid the allocation and is idiomatic Rust. Skip unless this becomes a hot path.

**Nit:** `agent_loop.rs:545-555` — the post-final-inference fallback (after tool-call exhaustion with empty text) still produces `\"• <tool> completed\"` summaries or empty `response_text`. Not in scope for #1090 (the issue called out lines ~269-274 specifically, which were correctly unified), but a future cleanup could route this through `CLARIFYING_QUESTION` too when `all_tool_executions.is_empty()`.

**Nit:** `agent_loop.rs:236-239` — `InferenceUsage { prompt_tokens: 0, completion_tokens: 0 }` is correct (the model wasn't called), but the skill pipeline did run a semantic search whose cost is silently swallowed. Probably fine — there's no existing telemetry for that and other early-return paths behave the same way — but worth a passing note that \"zero usage\" doesn't mean \"zero work\".

**Nit:** `schema.rs:1018` — single blank-line removal is `cargo fmt` output, unrelated to #1090. Harmless, but ideally `cargo fmt` is run as a separate commit so PR diffs stay topical. Not a blocker; the noise is one line.

**Nit:** The `CLARIFYING_QUESTION` string lives in `agent_loop.rs`. If/when localization or A/B testing of the wording becomes a concern, lift it to a config module. Premature today.

---

## Edge-case sanity check on `is_ambiguous` (no changes needed)

| Input | Classification | Correct? |
|---|---|---|
| `\"\"`, `\"   \"` | ambiguous | ✅ (short-circuits to clarification) |
| `\"😀\"` | ambiguous | ✅ (one-word, no signal) |
| `\"open Москва\"` (Cyrillic proper noun) | not ambiguous | ✅ (`char::is_uppercase` is Unicode-aware) |
| `\"URGENT HELP\"` | not ambiguous | ✅ (\"HELP\" idx=1 uppercase signals intent) |
| `\"١٢٣\"` (Arabic-Indic digits) | ambiguous | ⚠️ `is_ascii_digit` misses these — negligible in practice |
| `\"console.log('x')\"` | ambiguous | ⚠️ false positive — see Improvement #2 |
| `\"check https://example.com\"` | not ambiguous | ✅ |
| Long sentence with no proper nouns | not ambiguous | ✅ (word count gate) |

---

## Standards Compliance

- ✅ No `#[allow(dead_code)]` or `#[deprecated]` introduced.
- ✅ No lint suppressions in the diff.
- ✅ No backward-compatibility scaffolding (greenfield-appropriate: the old empty-fallback string is *replaced*, not kept alongside).
- ✅ No raw `println!`/`eprintln!`; uses `tracing::{info,warn}`.
- ✅ Conventional commit message format, references `closes #1090`.

---

## Recommendation: **APPROVE**

The change cleanly delivers the issue's intent, the tests prove the short-circuit, and acceptance criteria are met. The improvements listed are quality refinements that can be deferred (or addressed in a follow-up) without blocking merge. The only one I'd encourage the author to consider before merge is the `.expect()` at `agent_loop.rs:259` — a brittle invariant is the kind of thing that gets you woken up at 3 AM.

malibio (commented):
**Recommendation: APPROVE** (cannot self-approve via GitHub; recording as a comment review). All 7 acceptance criteria from #1090 are met, 9 new tests pass, and the `PanicEngine` test cleanly proves the short-circuit bypasses inference. See full review in PR comment for findings. Two minor improvements suggested (brittle `expect()` at agent_loop.rs:259 and code-snippet false-positives in `is_ambiguous`) are non-blocking and can be follow-ups.

malibio:
## Addressed review feedback

Pushed e6e76de7 addressing the three yellow improvements from the review.

### Item 1 — `.expect()` at `agent_loop.rs:256-259`

Restructured to carry the pipeline reference alongside the `SkillMatch` in a tuple at resolution time. The scoping branch now destructures both out of the same `if let Some((pipeline, ref skill_match)) = skill_match` arm — no runtime invariant assertion required, no panic path.

- See `agent_loop.rs:218-228` for the tuple resolution.
- See `agent_loop.rs:266-298` for the destructuring branch that consumes both.

### Item 2 — `is_ambiguous` false-positives on code-like inputs

Added a per-word check for tokens containing any of `(`, `)`, `/`, `.`, `:`, `_`, `-`. These are now classified as specific signals (alongside digits and proper nouns), so single-token inputs like `console.log('x')`, `path/to/file.rs`, `foo::bar`, `my_var`, `kebab-case-name`, `fn main()`, and `README.md` reach the model instead of being short-circuited.

Added `is_ambiguous_treats_code_and_paths_as_actionable` covering all of these. All 7 `is_ambiguous_*` unit tests pass.

- See `agent_loop.rs:91-100` for the new check.
- See `agent_loop.rs:2137-2150` for the new test.

### Item 3 — Test prompts now self-documenting

Rewrote ~13 test prompts so each gestures at what the test verifies. A few examples:

| Test | Prior prompt | New prompt |
|---|---|---|
| `single_turn_no_tools` | `please tell me about the GitHub release pipeline today` | `Summarize the GitHub release notes for v1.2 in plain English` |
| `tool_call_then_final_response` | `search for any Billing related architecture documents` | `Search GitHub for open release-blocker issues then summarize them` |
| `multi_step_tool_chain` | `tell me about the Billing Architecture` | `Look up the Billing Architecture node then fetch its referenced details` |
| `max_iteration_limit` | `keep searching for nodes about NodeSpace architecture` | `Keep running search_nodes forever — verify the iteration cap stops it` |
| `session_persistence_after_inference_error` | `please describe the GitHub release workflow today` | `Trigger an inference error and confirm the GitHub session survives intact` |
| `empty_tool_call_args_handled_gracefully` | `do something with the GitHub release pipeline` | `Invoke a GitHub tool with empty args and verify the loop does not panic` |

Each new prompt still passes `is_ambiguous == false` via proper noun (`GitHub`, `Billing`), digit (`v1.2`), or code-token signal (`search_nodes`, `mid-tool-execution`).

### Skipped green nits (with reasoning)

- **`to_lowercase()` allocation** — negligible at chat rates, premature optimization.
- **Post-tool-exhaustion fallback at lines ~545-555** — explicitly out of scope per reviewer.
- **Unrelated `schema.rs` blank-line removal** — harmless, splitting now is churn.
- **`CLARIFYING_QUESTION` location** — premature to lift to config without localization or A/B-testing demand.

### Verification

- `cargo test -p nodespace-agent --lib` — 297 passed, 0 failed (includes the new code-and-paths test).
- `cargo clippy -p nodespace-agent --all-targets -- -D warnings -D clippy::unused-async` — clean.
- `bun run quality:fix` — clean (fmt + clippy + eslint + svelte-check).
- `ollama_integration` tests show 8 pre-existing failures (real-model flakiness, present in baseline; unrelated to this PR).

Issue #1088, opened 2026-04-10
Schema field rename with property data migration


malibio:
## Overview

Renaming a schema field requires two coordinated operations: updating the schema definition AND migrating all existing node instances to rekey their stored property data. Currently only the schema definition side is possible (via \`update_schema\` remove+add), which orphans existing data.

## Problem Statement

When a user renames a field (e.g. \`old_name\` → \`new_name\` on a \`project\` schema):

1. The schema definition must be updated (field renamed)
2. All existing nodes of that type must have their \`properties.old_name\` rewritten to \`properties.new_name\`

Without step 2, existing node instances silently lose their data — the value stays in the DB under the old key but the schema no longer maps to it, so the UI shows the field as empty.

The current workaround (remove field + add field with new name) gives a clean schema but orphans all existing property values under the old key.

### Phase 1: Database operation

Add an atomic rename operation to \`SurrealStore\` that executes:

\`\`\`sql
UPDATE node
SET properties.new_name = properties.old_name
WHERE node_type = $type AND properties.old_name != NONE;

UPDATE node
SET properties = object::remove(properties, 'old_name')
WHERE node_type = $type;
\`\`\`

Expose via \`NodeService\` as \`rename_schema_field(type_id, from, to)\`.

### Phase 2: Schema definition update

Update the schema node's field list atomically with the data migration (same transaction or sequential with error rollback).

### Phase 3: Agent tool

Add \`rename_fields\` parameter to \`update_schema\` agent tool:

\`\`\`json
{
  "schema_id": "project",
  "rename_fields": [{ "from": "old_name", "to": "new_name" }]
}
\`\`\`

## Acceptance Criteria

- [ ] \`SurrealStore\` has a \`rename_schema_field(type_id, from, to)\` method that rekeys property data for all nodes of the given type
- [ ] Schema definition is updated atomically with the data migration
- [ ] \`update_schema\` MCP handler accepts \`rename_fields: [{from, to}]\` parameter
- [ ] \`update_schema\` agent tool exposes \`rename_fields\` parameter with clear description
- [ ] Existing node instances retain their values after a rename
- [ ] Tests pass (\`bun run test:all\`)
- [ ] Code passes \`bun run quality:fix\`

### Reference Files

- **SurrealStore**: \`packages/core/src/db/surreal_store.rs\` — add rename operation
- **NodeService**: \`packages/core/src/services/node_service.rs\` — expose rename via service layer
[...]

PR #1129, opened 2026-05-17
Unify agent guidance into single source of truth (#1089)

### LINKED ISSUE #1089: Unify agent guidance rules into a single source for MCP and local agent
## Overview

The MCP server's initialize handler and the local agent's prompt assembler both maintain agent guidance rules independently. When a rule changes in one place, the other silently drifts. This creates inconsistent behavior between Claude Desktop (MCP path) and the local Ollama agent.

## Problem Statement

Two files own overlapping agent guidance content:

1. **`packages/core/src/mcp/handlers/initialize.rs`** — hardcoded `instructions` string in the JSON-RPC `initialize` response, sent to external LLM clients (Claude Desktop, Claude Code)
2. **`packages/agent/src/prompt_assembler.rs`** — `seed_prompt_nodes()` returns the authoritative built-in prompt sections seeded to the DB on startup, used by the local Ollama agent

Rules that exist in both and can drift:
- **Schema creation rules**: `title_template` tokens must match fields array; do not add `name`/`title` fields unless used in template
- **Search first guidance**: always search before updating or fetching by ID; never use placeholder IDs
- **Node reference format**: bare `nodespace://` URIs (no markdown links, no backticks)

Any change to these rules (e.g., adding `update_schema` guidance, refining `title_template` validation wording) must be made in two places — and there is no test ensuring they stay in sync.

## Proposed Solution

Extract shared rules into a single authoritative source that both paths import. Two viable approaches:

### Option A: Shared constants in `packages/core`

Add `packages/core/src/agent_guidance.rs` exporting `const` rule strings. Both `mcp/handlers/initialize.rs` (same crate) and `packages/agent/src/prompt_assembler.rs` (agent already depends on core) import from it.

```rust
// packages/core/src/agent_guidance.rs
pub const SEARCH_FIRST_RULES: &str = "TOOL STRATEGY:\n- ALWAYS search first...";
pub const SCHEMA_CREATION_RULES: &str = "SCHEMA CREATION RULES:\n...";
pub const NODE_REFERENCE_FORMAT: &str = "NODE REFERENCES: ...";
```

### Option B: MCP server queries prompt nodes from DB at startup

The `initialize` handler receives a `NodeService`. On first request it queries `prompt` nodes (type=`"prompt"`, source=`"built-in"`) and builds the `instructions` string from those. Same seed nodes, same content, no duplication.

Option A is simpler and doesn't require a DB query at MCP init time. Option B is the architecturally cleanest (graph as single source of truth) but adds async complexity to the `initialize` handler.

## Acceptance Criteria

- [ ] Shared guidance rules (search-first, schema creation, node references) exist in exactly one place in the codebase
- [ ] Local agent (`PromptAssembler`) and MCP server (`initialize.rs`) both derive their guidance from that single source
- [ ] Changing a rule in one place automatically updates behavior for both Claude Desktop (MCP) and local Ollama agent
- [ ] All existing tests pass (`bun run test:all`)

### Files to Modify

- **`packages/core/src/mcp/handlers/initialize.rs`** — replace inline rule strings with imports from shared source
- **`packages/agent/src/prompt_assembler.rs`** — `seed_prompt_nodes()` Tool Strategy Guide, Response Rules sections
- **`packages/core/src/lib.rs`** (or `mod.rs`) — expose new `agent_guidance` module if using Option A

### Current overlapping content locations

| Rule | MCP location (line) | Agent location |
|------|-------------------|----------------|
| Schema creation | `initialize.rs` line 168: `SCHEMA CREATION RULES:` | `prompt_assembler.rs` Tool Strategy Guide seed (priority 50) |
| Search first | `initialize.rs` line 169: `SEARCH FIRST:` | `prompt_assembler.rs` Tool Strategy Guide seed |
| Node references | `initialize.rs` line 171: `NODE REFERENCES:` | `prompt_assembler.rs` Response Formatting Rules seed (priority 60) |

### Non-Goals

- Does not change the MCP `instructions` field structure (dynamic node types / collections remain in `initialize.rs`)
- Does not merge the full MCP instructions with the local agent system prompt (they serve different models with different needs)
- Does not replace `PromptAssembler`'s DB-backed assembly path for the local agent

## Related Issues

- #1088 (schema field rename — schema creation guidance will need updating when implemented)

malibio:
## Summary

Extracts the three shared agent guidance rules into a new `packages/agent/src/agent_guidance.rs` module so the local Ollama agent and the PTY/ACP context-file path (ADR-032) derive their rules from a single source.

- New `agent_guidance.rs` exports three `pub const` strings: `SCHEMA_CREATION_RULES`, `SEARCH_FIRST_RULES`, `NODE_REFERENCE_FORMAT`
- `prompt_assembler.rs` composes the Tool Strategy Guide and Response Formatting Rules seed bodies from the shared constants — byte-equivalent to the previous inline strings, verified outside the test suite
- `context_assembly.rs` imports the constants and exposes `pub fn agent_guidance_section()` returning markdown ready to embed in `CLAUDE.md` / `AGENTS.md` under ADR-032

Closes #1089.

## Acceptance criteria

- [x] Shared guidance rules exist in exactly one place (`agent_guidance.rs`)
- [x] `PromptAssembler` derives guidance from that single source
- [x] `context_assembly.rs` derives guidance from that single source
- [x] Editing a constant in `agent_guidance.rs` propagates to both consumers
- [x] All existing tests pass (`bun run test:all`)

## Test status

Baseline (worktree, pre-change): 3917 frontend passed / 1 perf-threshold flake in `marked-config.test.ts`, 1182 Rust passed.

After change:
- Frontend: **3918 passed, 0 failed** (128 test files, the prior flake passes this run)
- Rust: **1182 passed, 0 failed**
- New tests added:
  - `agent_guidance::tests::{schema_creation,search_first,node_reference_format}_*` — assert each constant carries its expected anchor strings
  - `context_assembly::tests::test_agent_guidance_section_includes_all_rules` and `..._uses_shared_constants` — assert the helper composes from the shared constants so an edit to a constant updates context-file output

## Notes on scope

- The schema-creation rules about `title_template` token/field alignment called out in the issue (and #1088) are not yet present in the prompts — `SCHEMA_CREATION_RULES` is seeded with the schema content that exists today (NODE MODEL paragraph). When #1088 lands, it edits `SCHEMA_CREATION_RULES` and both consumers pick up the new rules automatically.
- `context_assembly.rs`'s existing `assemble()` output is unchanged. The new `agent_guidance_section()` is the hook ADR-032 will call when it starts writing `CLAUDE.md` / `AGENTS.md` files.
- Includes one drive-by formatting fix in `packages/core/src/mcp/handlers/schema.rs` (trailing blank line removed by `cargo fmt`).

## Test plan

- [ ] Reviewer confirms the byte-equivalence of refactored seed bodies (`PromptAssembler::seed_prompt_nodes()` produces the same `markdown_content` strings as before)
- [ ] Reviewer confirms both new constants and `agent_guidance_section()` are referenced from both `prompt_assembler.rs` and `context_assembly.rs`
- [ ] `bun run test:all` clean on CI

### DIFF
--- a/packages/agent/src/acp/context_assembly.rs
+++ b/packages/agent/src/acp/context_assembly.rs
@@ -17,7 +17,6 @@
 //!
 //! Issue #1005
 
-use crate::agent_guidance::{NODE_REFERENCE_FORMAT, SCHEMA_CREATION_RULES, SEARCH_FIRST_RULES};
 use crate::agent_types::{
     ContextAssembler, ContextError, ContextNode, ContextPacket, ContextRelationship,
 };
@@ -62,20 +61,6 @@ const MCP_ACCESS_SECTION: &str = "\
 Query the graph using these MCP tools: search_semantic, search_nodes, get_node, \
 create_nodes_from_markdown, update_node, delete_node.
 ";
-
-/// Compose the shared agent guidance section for embedding in context files.
-///
-/// Returns a markdown-formatted section containing the schema creation rules,
-/// search-first tool strategy, and node reference format — all sourced from
-/// [`crate::agent_guidance`]. ADR-032 callers (CLAUDE.md / AGENTS.md writers)
-/// embed this in the files surfaced to external agent sessions, ensuring the
-/// PTY agent and local agent operate under identical rules.
-pub fn agent_guidance_section() -> String {
-    format!(
-        "### Agent Guidance\n\n{}\n\n{}\n\n{}\n",
-        SCHEMA_CREATION_RULES, SEARCH_FIRST_RULES, NODE_REFERENCE_FORMAT
-    )
-}
 
 /// Estimate token count from character length using the ~4 chars/token heuristic.
 fn estimate_tokens(text: &str) -> u32 {
@@ -928,34 +913,6 @@ mod tests {
                 expected
             );
         }
-    }
-
-    // =========================================================================
-    // Agent guidance section tests
-    // =========================================================================
-
-    #[test]
-    fn test_agent_guidance_section_includes_all_rules() {
-        let section = agent_guidance_section();
-        assert!(section.contains("### Agent Guidance"));
-        // Schema creation rules content
-        assert!(section.contains("NODE MODEL:"));
-        assert!(section.contains("create_schema"));
-        // Search-first rules content
-        assert!(section.contains("TOOL STRATEGY:"));
-        assert!(section.contains("ALWAYS search first"));
-        // Node reference format
-        assert!(section.contains("nodespace://"));
-        assert!(section.contains("no markdown links"));
-    }
-
-    #[test]
-    fn test_agent_guidance_section_uses_shared_constants() {
-        // Changing a constant in agent_guidance.rs must propagate here.
-        let section = agent_guidance_section();
-        assert!(section.contains(SCHEMA_CREATION_RULES));
-        assert!(section.contains(SEARCH_FIRST_RULES));
-        assert!(section.contains(NODE_REFERENCE_FORMAT));
     }
 
     // =========================================================================
--- a/packages/agent/src/acp/context_assembly.rs
+++ b/packages/agent/src/acp/context_assembly.rs
@@ -17,6 +17,7 @@
 //!
 //! Issue #1005
 
+use crate::agent_guidance::{NODE_REFERENCE_FORMAT, SCHEMA_CREATION_RULES, SEARCH_FIRST_RULES};
 use crate::agent_types::{
     ContextAssembler, ContextError, ContextNode, ContextPacket, ContextRelationship,
 };
@@ -61,6 +62,20 @@ const MCP_ACCESS_SECTION: &str = "\
 Query the graph using these MCP tools: search_semantic, search_nodes, get_node, \
 create_nodes_from_markdown, update_node, delete_node.
 ";
+
+/// Compose the shared agent guidance section for embedding in context files.
+///
+/// Returns a markdown-formatted section containing the schema creation rules,
+/// search-first tool strategy, and node reference format — all sourced from
+/// [`crate::agent_guidance`]. ADR-032 callers (CLAUDE.md / AGENTS.md writers)
+/// embed this in the files surfaced to external agent sessions, ensuring the
+/// PTY agent and local agent operate under identical rules.
+pub fn agent_guidance_section() -> String {
+    format!(
+        "### Agent Guidance\n\n{}\n\n{}\n\n{}\n",
+        SCHEMA_CREATION_RULES, SEARCH_FIRST_RULES, NODE_REFERENCE_FORMAT
+    )
+}
 
 /// Estimate token count from character length using the ~4 chars/token heuristic.
 fn estimate_tokens(text: &str) -> u32 {
@@ -913,6 +928,34 @@ mod tests {
                 expected
             );
         }
+    }
+
+    // =========================================================================
+    // Agent guidance section tests
+    // =========================================================================
+
+    #[test]
+    fn test_agent_guidance_section_includes_all_rules() {
+        let section = agent_guidance_section();
+        assert!(section.contains("### Agent Guidance"));
+        // Schema creation rules content
+        assert!(section.contains("NODE MODEL:"));
+        assert!(section.contains("create_schema"));
+        // Search-first rules content
+        assert!(section.contains("TOOL STRATEGY:"));
+        assert!(section.contains("ALWAYS search first"));
+        // Node reference format
+        assert!(section.contains("nodespace://"));
+        assert!(section.contains("no markdown links"));
+    }
+
+    #[test]
+    fn test_agent_guidance_section_uses_shared_constants() {
+        // Changing a constant in agent_guidance.rs must propagate here.
+        let section = agent_guidance_section();
+        assert!(section.contains(SCHEMA_CREATION_RULES));
+        assert!(section.contains(SEARCH_FIRST_RULES));
+        assert!(section.contains(NODE_REFERENCE_FORMAT));
     }
 
     // =========================================================================
--- a/packages/agent/src/agent_guidance.rs
+++ b/packages/agent/src/agent_guidance.rs
@@ -11,19 +11,22 @@
 
 /// Schema creation guidance.
 ///
-/// Covers the node-vs-schema mental model and the rules for `create_schema`,
-/// `update_schema`, and instantiating typed nodes. Per #1088, this will grow
-/// to include explicit rules about `title_template` token / field alignment.
+/// Covers the node-vs-schema mental model: when to call `create_schema` vs.
+/// `create_node`, and how custom types relate to built-in types. More detailed
+/// `title_template` token / field alignment guidance currently lives in
+/// `skill_pipeline.rs` (used only by the skill-based schema-creation path) and
+/// should be consolidated here when that path is unified — tracked separately
+/// from #1089.
 pub const SCHEMA_CREATION_RULES: &str = "NODE MODEL: Everything in NodeSpace is a node. Built-in types (task, text, date) are always available. Custom types (e.g. 'project', 'customer') require a schema node to exist first — the schema defines the type's fields and title template. Once a schema exists, create instances with create_node(node_type=<schema_id>). Use create_schema only to define a new type; use create_node to create data.";
 
-/// Search-first tool strategy guidance.
+/// Tool strategy guidance.
 ///
-/// The full "TOOL STRATEGY:" bulleted list, anchored on the rule that the
+/// The full "TOOL STRATEGY:" bulleted list. Anchored on the rule that the
 /// agent must always search before updating or fetching nodes — never invent
-/// placeholder IDs. Also covers how to choose between search_nodes,
-/// search_semantic, get_node, and get_related_nodes, and the canonical
+/// placeholder IDs — but also covers how to choose between search_nodes,
+/// search_semantic, get_node, and get_related_nodes, plus the canonical
 /// create_node / create_schema / create_relationship usage patterns.
-pub const SEARCH_FIRST_RULES: &str = "TOOL STRATEGY:\n\
+pub const TOOL_STRATEGY_RULES: &str = "TOOL STRATEGY:\n\
     - ALWAYS search first before updating or getting a node. NEVER use placeholder IDs like \"abc-123\".\n\
     - To find nodes by exact title or keyword (when you know the name): use search_nodes with query=<keyword>. To filter by type (e.g. \"show all tasks\"), pass node_type=\"task\" with query=\"\". To filter by property (e.g. \"open tasks\"), pass filters={\"status\":\"open\"}.\n\
     - To find nodes by meaning/topic (when the exact name is unknown): use search_semantic (natural language query)\n\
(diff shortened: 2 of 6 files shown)


malibio (commented):
## Code Review Summary

Clean, well-scoped refactor. The PR does exactly what #1089 asks: pulls the three rule strings out of `prompt_assembler.rs` into a single `agent_guidance.rs` module and rewires both consumers to import from it. Byte-equivalence of the seeded prompts is preserved, tests pass, clippy is clean on `nodespace-agent`. Net positive.

I independently re-verified the byte-equivalence claim with a standalone rustc harness reproducing `seed_prompt_nodes()`'s string construction on both branches — Tool Strategy Guide and Response Formatting Rules outputs match exactly. So the runtime behavior of the local Ollama agent is unchanged by this PR.

A few items below worth addressing or at least acknowledging before merge — none are blockers.

---

### Findings

#### Critical Issues

None.

#### Suggested Improvements

**[Improvement] `packages/agent/src/prompt_assembler.rs` (or `agent_guidance.rs`): no in-tree byte-equivalence regression test.**

You verified the strings match externally with a one-off harness — fine for this PR. The risk is that the *next* edit to either `SEARCH_FIRST_RULES` or the `format!()` glue in the seed bodies silently diverges from the previously-shipped prompt, and the only signal will be agent behavior changes (which we don't gate on in CI). The existing `seed_prompts_have_valid_properties` test only checks non-emptiness.

Cheap mitigation: a single `#[test]` in `prompt_assembler.rs` that snapshots the `Tool Strategy Guide` and `Response Formatting Rules` `markdown_content` against a small fixture string (or `insta` snapshot). One assertion per seed; ~15 lines total. Whether this is worth the maintenance cost is a judgment call — these strings change deliberately and rarely. I lean "yes, add it" because the rest of the PR took such care to preserve byte-equivalence; without a regression test, that property has no teeth. Calling out per your explicit ask: it is not overkill, but reasonable people could disagree, so flagging rather than blocking.

**[Improvement] Naming: `SEARCH_FIRST_RULES` is a misnomer for its current contents.**

The constant contains the full "TOOL STRATEGY" bulleted block, which includes schema-creation bullets like `- To create a new entity type: use create_schema...`, `- To modify an existing entity type: use update_schema...`, and creation/relationship guidance. These are not "search first" rules. The constant's doc comment acknowledges this ("Also covers ... canonical create_node / create_schema / create_relationship usage patterns"), but the name will mislead future readers grepping for where create_schema guidance lives.

Two options:
1. Rename to `TOOL_STRATEGY_RULES` — matches the prompt header text and the actual contents.
2. Split further: pull the schema-creation bullets out into `SCHEMA_CREATION_RULES` (so the constant lives up to its name) and leave search/get/update guidance in `SEARCH_FIRST_RULES`.

Option 1 is the lower-cost fix and I'd prefer it for this PR. Option 2 is what the issue's table implied ("Schema creation" as a distinct unit of overlap) but is more invasive — could be a follow-up. As-is, the *separation* in the issue's mental model doesn't match the *separation* in the code.

**[Improvement] Third copy of these rules in `packages/agent/src/local_agent/prompt_templates.rs::fallback_system_prompt()` is not unified.**

`fallback_system_prompt()` contains an abbreviated copy of TOOL STRATEGY (lines 33–42) and a verbatim copy of RESPONSE RULES (lines 43–53), including the `nodespace://abc-123` bare-URI rule. It is actively called from `local_agent/agent_loop.rs:135` whenever `PromptAssembler` is not wired in.

Acceptance criterion #1 of #1089 says "Shared guidance rules exist in exactly one place." Strictly read, this PR doesn't satisfy that — it satisfies the narrower table in the issue body, which lists only `prompt_assembler.rs` and `context_assembly.rs`. Two reasonable resolutions:

- (a) Have `fallback_system_prompt()` also format from the same constants. The fallback is intentionally terser, so it would need either a smaller `SEARCH_FIRST_RULES_TERSE` constant or some structural rework. Probably not worth it for code that exists to handle a path the doc comment describes as transitional.
- (b) Update the doc on `fallback_system_prompt` to call out that its rule text is intentionally a duplicate, and file a follow-up to delete it once `PromptAssembler` is always wired in. This matches the existing intent ("only for use when `PromptAssembler` is not available") and is honest about the duplication.

I'd take (b). Either way, this should be acknowledged in the PR description or a follow-up issue, not silently shipped.

**[Improvement] PR description claim "queued for #1088" is inaccurate.**

#1088 is `Schema field rename with property data migration` — adding a `rename_fields` parameter to `update_schema`. It has nothing to do with the `title_template` token/field alignment rules the #1089 issue body called out. Two related but separate observations:

- The `title_template` alignment guidance already exists in `packages/agent/src/skill_pipeline.rs` lines 287–329 (a different prompt path used for skill-based schema creation). So the issue's claim that "these rules exist in both [prompt_assembler.rs and context_assembly.rs]" is itself slightly off — they live in `skill_pipeline.rs` only, not in `prompt_assembler.rs`. This is the reason the PR could correctly say "they're not in the seeded prompts today."
- It would be more accurate to either (i) drop the "#1088 will land them" line from the PR description and the doc comment on `SCHEMA_CREATION_RULES` (which carries the same claim), or (ii) file a separate issue tracking the unification of `skill_pipeline.rs`'s schema-creation rules into `SCHEMA_CREATION_RULES`. Right now `SCHEMA_CREATION_RULES`'s doc comment promises future content that isn't tracked anywhere actionable.

**[Improvement] `agent_guidance_section()` is dead code on `main` until ADR-032 lands.**

Because the function is `pub`, the compiler will not warn — that's correct Rust semantics, not a problem. The concern is more architectural: we're shipping a public API contract (signature, return shape, the markdown block format `### Agent Guidance\n\n{schema}\n\n{search}\n\n{node_ref}\n`) with no consumer to validate the shape. If ADR-032's writer wants a different structure (e.g., separate `### Schema Rules`, `### Tool Strategy`, `### Node References` sections, or no section headers at all because the consumer wraps them), we've prematurely locked in a format.

Two acceptable paths:
1. Land the constants only; defer `agent_guidance_section()` to the ADR-032 PR where its shape will be validated against a real consumer. This is the YAGNI-aligned move.
2. Keep the helper as a *suggested* default but make the section format trivial enough that the caller doesn't depend on it — i.e., expose only the constants and let the ADR-032 writer compose them however it needs. The two new tests would move to `agent_guidance.rs` (already there for the constants) and the section-composition helper would live next to the ADR-032 code.

Given the project's stated "no speculative API" stance (CLAUDE.md: NO backward compatibility, KISS, YAGNI), I'd lean (1). The cost of recreating this 5-line helper later is essentially zero; the cost of having ADR-032 inherit a format choice it didn't make is nonzero.

#### Nitpicks

- **Nit:** `packages/agent/src/lib.rs:17-19` — module ordering is non-alphabetical and slightly inconsistent (`agent_guidance` comes after `prompt_assembler` even though prompt_assembler depends on it). Not a correctness issue, but a reader scanning the lib root would expect leaf dependencies above their consumers. Small reorder: `agent_guidance` before `prompt_assembler`.
- **Nit:** `packages/agent/src/agent_guidance.rs:17` — the entire `SCHEMA_CREATION_RULES` constant is a single line that wraps off-screen in most editors. Rust supports `\` line continuations in string literals (you use them elsewhere in the file) — consider splitting at sentence boundaries for readability. The byte content stays identical.
- **Nit:** `packages/agent/src/acp/context_assembly.rs:945-946` — the assertion `section.contains("ALWAYS search first")` does double duty with `test_agent_guidance_section_uses_shared_constants`. Since the second test already asserts `section.contains(SEARCH_FIRST_RULES)` (which contains "ALWAYS search first" by definition), the granular anchor assertions in the first test are slightly redundant. Not a problem — leaving them gives nicer failure messages — but worth knowing they're not load-bearing.
- **Nit:** `packages/core/src/mcp/handlers/schema.rs` — drive-by trailing-blank-line removal is unrelated to #1089. Defensible because it's a trivial rustfmt cleanup, but per the "atomic commits" guideline this kind of thing ideally goes in a separate commit (even within the same PR). Not worth holding the PR.

### Verdict

This is a clean refactor that achieves the structural goal of #1089 for the two paths the issue's table explicitly named. I'd merge after at least the naming concern (`SEARCH_FIRST_RULES` rename or schema-bullet split) and the `agent_guidance_section()` decision are addressed — both are small. The byte-equivalence test and the `prompt_templates.rs` follow-up can be follow-ups; flag them in a comment on #1089 or open a tracking issue.

Reviewed by `/pragmatic-code-review`.

malibio:
## Review feedback addressed

Pushed commit `502590f3`. Summary against each finding:

| Finding | Action | Notes |
|---|---|---|
| Missing in-tree byte-equivalence regression test | ✅ Addressed | New `seed_prompt_bodies_match_expected_bytes` in `prompt_assembler.rs` locks in the exact rendered bytes of the Tool Strategy Guide and Response Formatting Rules seed bodies. Future edits to the constants or `format!()` glue will trip it unless the expected fixtures are updated deliberately. |
| `SEARCH_FIRST_RULES` is a misnomer | ✅ Addressed (Option 1) | Renamed to `TOOL_STRATEGY_RULES` — matches the constant's actual contents and the prompt's `TOOL STRATEGY:` header. Reviewer's preferred option. |
| Third copy in `fallback_system_prompt()` | ✅ Addressed (Option b) | Updated the doc comment to call out the duplication explicitly and note that the function (and its inline rules) should be deleted entirely when `PromptAssembler` is universally wired. No structural rewrite — kept narrow per the reviewer's preference. |
| PR description / doc claim "queued for #1088" is inaccurate | ✅ Addressed | Dropped the `#1088` reference from `SCHEMA_CREATION_RULES` doc. New comment points at `skill_pipeline.rs` (where `title_template` guidance actually lives today) and notes consolidation is tracked separately, not as part of #1089. |
| `agent_guidance_section()` is dead/speculative API | ✅ Addressed (Option 1) | Removed `pub fn agent_guidance_section()` from `context_assembly.rs` along with its two tests. The constants stay exposed via the module — ADR-032's CLAUDE.md / AGENTS.md writer can compose its own section format against a real consumer. Reviewer's preferred YAGNI move; matches CLAUDE.md's "no speculative API" stance. |
| Nit: `lib.rs` module ordering | ✅ Addressed | Moved `pub mod agent_guidance;` above `pub mod prompt_assembler;` so leaf deps precede consumers. |
| Nit: long single line in `SCHEMA_CREATION_RULES` | ⏭️ Skipped | Splitting requires either changing the byte content (defeats the regression test) or backslash line-continuations with leading-whitespace-eating syntax. The static string is rarely scrolled and the diff view is readable. Value/churn doesn't justify it. |
| Nit: redundant assertions in `context_assembly` tests | ⏭️ N/A | The tests in question were the ones removed alongside `agent_guidance_section()`. Moot. |
| Nit: drive-by `schema.rs` cleanup commits | ⏭️ Skipped | Already in this PR's commit history; splitting now would be commit-history churn for no functional benefit. |

## Test status

- `cargo test -p nodespace-agent --lib`: 292 passed, 0 failed (down 1 from prior run = removed 2 helper tests, added 1 byte-equivalence test, net -1; expected)
- `bun run test:all`: 3918 frontend passed, 1182 rust passed, 0 failures
- `bun run quality:fix`: clean

## Acceptance criteria status

- [x] Shared guidance rules exist in exactly one place — for the two paths #1089's table named (`prompt_assembler.rs` and `context_assembly.rs`'s consumer surface). The third copy in `fallback_system_prompt()` is now explicitly documented as a known, scoped-to-be-removed duplication; not silently shipped.
- [x] `PromptAssembler` derives guidance from `agent_guidance.rs`
- [x] `context_assembly.rs` has the constants available (no consumer added yet — that's ADR-032's job)
- [x] Editing a constant in `agent_guidance.rs` propagates to `PromptAssembler` and is locked in by the byte-equivalence test
- [x] `bun run test:all` clean
- [x] `bun run quality:fix` clean

## Re-review decision

**NO RE-REVIEW NEEDED.** Changes are scoped exactly to the reviewer's recommendations — a rename, a deletion of speculative API, a new regression test, and three doc edits. No new architectural moves and no new functionality. The PR is ready for merge pending any other approvals.

PR #1132, opened 2026-05-17
Gate AgentSession.system_prompt_override behind testing feature (closes #1093)

### LINKED ISSUE #1093: Move system_prompt_override test infrastructure off AgentSession
## Background

`system_prompt_override` on `AgentSession` (a public, Tauri-serialized struct) embeds test infrastructure in the production domain model. This was flagged in a code review of PR #1091.

## Problem

- `AgentSession` is serialized and exposed via Tauri commands to the frontend
- The field is used only by integration tests (`ollama_integration.rs`) to inject pre-built prompts without a live database
- The conventional alternative is to keep test infrastructure out of production domain types

## Challenge

A simple `#[cfg(test)]` on the struct field doesn't work because:
- Integration tests in `tests/` compile against the library in non-test mode
- `#[cfg(test)]` on library fields is NOT active when linked from `tests/*.rs` binaries

## Possible Approaches

1. **Separate test-only session type**: Create an `AgentSessionTestExt` or test builder in the agent test module
2. **Inject via constructor**: Give the test a way to pass the prompt override at session creation time via a separate test-facing `create_session_with_prompt()` method that is feature-gated
3. **Test service wrapper**: Wrap `LocalAgentService` in a test-only struct that intercepts `send_message` to inject the prompt before forwarding
4. **`testing` feature flag**: Add a `testing` Cargo feature that gates the field and method — properly excluded from production builds

## Context

- Flagged in code review of PR #1091 (issue #1083)
- The `system_prompt_override` field has `skip_serializing_if = "Option::is_none"` so it's invisible in serialized output when not set
- The field is never set in production code — always `None` in real sessions
- Related: `set_system_prompt()` method on `LocalAgentService` is also test-only infrastructure

malibio:
## Summary

- Adds a `testing` Cargo feature on `nodespace-agent` and gates `AgentSession.system_prompt_override`, `LocalAgentService::set_system_prompt`, and the override branch in `run_turn` behind `#[cfg(any(test, feature = "testing"))]`.
- Production builds (no features) compile without the field, the method, or the override branch — so the test-only hook never reaches the Tauri serialization layer or the production prompt-assembly path.
- `tests/ollama_integration.rs` is declared as `[[test]] required-features = ["testing"]`, so `cargo test -p nodespace-agent` skips it cleanly without the feature and builds/runs it under `cargo test -p nodespace-agent --features testing`.

## Why this approach

Integration tests in `tests/` link against the library as a non-`#[cfg(test)]` crate, which is why the originally-flagged design (per PR #1091 review) put the field directly on `AgentSession`. A `testing` Cargo feature is the conventional Rust solution: the gate is visible to integration tests when the feature is enabled, but absent from production builds entirely. No wrapper indirection, no API churn, minimal diff.

## Verification

- `cargo build -p nodespace-agent` (no features) — clean, 22m23s
- `cargo check -p nodespace-agent --features testing --tests` — clean, 5m15s, proves the feature-gated code and `ollama_integration.rs` compile
- `bun run quality:fix` — passes clean (eslint 0 errors, svelte-check 0 warnings, cargo fmt + cargo clippy `-D warnings` across the workspace)
- Frontend test baseline: 128 files / 3918 tests passed (unchanged — no frontend touched)

## Test plan
- [x] `cargo build -p nodespace-agent` (no features) compiles cleanly
- [x] `cargo check -p nodespace-agent --features testing --tests` compiles cleanly
- [x] `bun run quality:fix` clean across workspace
- [x] Frontend tests unchanged from baseline (no frontend touched)
- [ ] Reviewer: confirm Tauri-serialized `AgentSession` no longer carries `system_prompt_override` in production (the field doesn't exist in the prod build)

### DIFF
--- a/packages/agent/Cargo.toml
+++ b/packages/agent/Cargo.toml
@@ -31,6 +31,24 @@ which = "7"
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
 
+[features]
+# Exposes test-only hooks (e.g. `LocalAgentService::set_system_prompt` and
+# `AgentSession::system_prompt_override`) so integration tests in `tests/`
+# can inject prebuilt prompts without a live database. The `tests/` binaries
+# link against the library in non-test mode, so `#[cfg(test)]` alone is not
+# enough — this feature gate is what keeps the hook out of production builds
+# while making it visible to integration tests run with `--features testing`.
+testing = []
+
 [dev-dependencies]
 tokio = { workspace = true, features = ["full", "test-util"] }
 tempfile = "3"
+
+# `ollama_integration.rs` calls `set_system_prompt`, which lives behind the
+# `testing` feature. Declaring it here means `cargo test -p nodespace-agent`
+# without features simply skips this binary instead of failing to compile,
+# while `cargo test -p nodespace-agent --features testing` builds and runs it.
+[[test]]
+name = "ollama_integration"
+path = "tests/ollama_integration.rs"
+required-features = ["testing"]
--- a/packages/agent/src/agent_types.rs
+++ b/packages/agent/src/agent_types.rs
@@ -541,7 +541,10 @@ pub struct AgentSession {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub dynamic_context: Option<String>,
     /// Full system prompt override (bypasses PromptAssembler / fallback).
-    /// Used in tests to inject a pre-built prompt without a live database.
+    /// Test-only: integration tests inject a pre-built prompt without a live
+    /// database. Gated by the `testing` feature so the field does not exist
+    /// in production builds and never reaches the Tauri serialization layer.
+    #[cfg(any(test, feature = "testing"))]
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub system_prompt_override: Option<String>,
 }
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -44,6 +44,20 @@ const HISTORY_TOKEN_BUDGET: u32 = TOTAL_TOKEN_BUDGET - SYSTEM_PROMPT_BUDGET;
 /// are likely too vague to dispatch to the full 13-tool agent path.
 const AMBIGUOUS_MESSAGE_WORD_LIMIT: usize = 10;
 
+/// Returns the test-only system-prompt override on a session, or `None` in
+/// production. The two cfg variants let `run_turn` keep a single
+/// override → assembler → fallback chain without duplicating the assembler
+/// branch under each cfg arm. The `None`-returning variant is optimized
+/// away by the compiler.
+#[cfg(any(test, feature = "testing"))]
+fn session_prompt_override(session: &AgentSession) -> Option<&str> {
+    session.system_prompt_override.as_deref()
+}
+#[cfg(not(any(test, feature = "testing")))]
+fn session_prompt_override(_session: &AgentSession) -> Option<&str> {
+    None
+}
+
 /// Clarifying question returned when the skill pipeline finds no match and
 /// the user message is too short/vague to confidently invoke the full agent.
 const CLARIFYING_QUESTION: &str =
@@ -236,21 +250,11 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
         };
 
         // Build base prompt: use override (tests), assembler (production), or fallback.
-        // The override branch is compiled out of production builds — see the
+        // `session_prompt_override` returns `None` in production builds — see the
         // `testing` feature on the agent crate.
-        #[cfg(any(test, feature = "testing"))]
-        let base_prompt = if let Some(ref override_prompt) = session.system_prompt_override {
-            override_prompt.clone()
+        let base_prompt = if let Some(override_prompt) = session_prompt_override(session) {
+            override_prompt.to_string()
         } else if let Some(ref assembler) = self.prompt_assembler {
-            assembler
-                .assemble(&template_ctx, all_tools.clone())
-                .await
-                .system_prompt
-        } else {
-            prompt_templates::fallback_system_prompt(dynamic_ctx)
-        };
-        #[cfg(not(any(test, feature = "testing")))]
-        let base_prompt = if let Some(ref assembler) = self.prompt_assembler {
             assembler
                 .assemble(&template_ctx, all_tools.clone())
                 .await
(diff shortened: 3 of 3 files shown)


malibio (commented):
## Code Review Summary

**Recommendation: APPROVE**

This is a small, focused, well-justified change that does exactly what issue #1093 asks for: it moves test-only infrastructure (`system_prompt_override`, `set_system_prompt`, and the override-consumption branch) off the production domain model and behind a Cargo feature flag. The chosen approach (#4 from the issue) is the conventional Rust idiom for "this hook needs to be visible to `tests/` integration binaries but absent from production." Diff is +39/-1 across three files, no API surface changes for production consumers, no migration concerns, and `bun run quality:fix` already exercised `cargo clippy --all-targets -D warnings` cleanly with the feature off — which proves Cargo's `required-features` correctly skips `ollama_integration.rs` rather than failing to build it.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

🟡 **[Improvement] `packages/agent/src/local_agent/agent_loop.rs:238-260` — Duplicated cfg branches for `base_prompt`.**

The current approach duplicates the entire `assembler → fallback` chain across two cfg arms:

```rust
#[cfg(any(test, feature = "testing"))]
let base_prompt = if let Some(ref override_prompt) = session.system_prompt_override {
    override_prompt.clone()
} else if let Some(ref assembler) = self.prompt_assembler {
    assembler.assemble(&template_ctx, all_tools.clone()).await.system_prompt
} else {
    prompt_templates::fallback_system_prompt(dynamic_ctx)
};
#[cfg(not(any(test, feature = "testing")))]
let base_prompt = if let Some(ref assembler) = self.prompt_assembler {
    assembler.assemble(&template_ctx, all_tools.clone()).await.system_prompt
} else {
    prompt_templates::fallback_system_prompt(dynamic_ctx)
};
```

A small accessor with two cfg variants collapses this back to a single branch and removes the duplicated assembler/fallback logic — which matters because the duplication will silently drift if either branch is edited without the other:

```rust
// Near AgentSession or as a free fn in this module:
#[cfg(any(test, feature = "testing"))]
fn session_prompt_override(s: &AgentSession) -> Option<&str> {
    s.system_prompt_override.as_deref()
}
#[cfg(not(any(test, feature = "testing")))]
fn session_prompt_override(_: &AgentSession) -> Option<&str> { None }

// Then at the call site, one chain only:
let base_prompt = if let Some(override_prompt) = session_prompt_override(session) {
    override_prompt.to_string()
} else if let Some(ref assembler) = self.prompt_assembler {
    assembler.assemble(&template_ctx, all_tools.clone()).await.system_prompt
} else {
    prompt_templates::fallback_system_prompt(dynamic_ctx)
};
```

**Rationale (DRY):** the assembler/fallback fallback chain is identical in both arms — only the override prelude differs. The compiler will optimize the `None`-returning prod variant to a no-op, so there's no runtime cost. This is a small refactor and not a merge blocker, but it eliminates a real maintenance hazard.

#### Nitpicks

🟢 **Nit: `packages/agent/Cargo.toml:34-41` — Feature comment is excellent but consider naming.**

`testing` is a common name and works fine here; some crates in the broader ecosystem prefer `test-hooks` or `test-support` to make it unambiguous that enabling this feature widens the API surface beyond what production code sees. Not worth changing for this PR — flagging only because future contributors may add a second feature and want a naming pattern to follow.

🟢 **Nit: `packages/agent/src/agent_types.rs:543-549` — Doc comment phrasing.**

The doc now says "Gated by the `testing` feature so the field does not exist in production builds and never reaches the Tauri serialization layer." Strictly speaking the gate is `any(test, feature = "testing")`, so `#[cfg(test)]` library tests also see it. Minor — current wording is accurate for the *production* concern, which is what the doc cares about.

### Verification Notes

I confirmed the load-bearing assumptions from the review brief:

1. **Tauri boundary is outbound-only.** Grep across `packages/desktop-app/src-tauri/` shows `AgentSession` appears only as a return type (`Result<Vec<AgentSession>, CommandError>` in `local_agent_get_sessions` at `local_agent.rs:572`) and never as a Tauri command parameter. The TS interface at `packages/desktop-app/src/lib/types/agent-types.ts:312-319` already omits `system_prompt_override`, consistent with `skip_serializing_if = "Option::is_none"`. Gating the field out of prod cannot cause a deserialization failure because no production code path deserializes `AgentSession` from JSON.

2. **No other crate consumes the agent's `testing` feature.** `packages/desktop-app/src-tauri/Cargo.toml:36` depends on `nodespace-agent` without features, so the production Tauri build will have the field/method/branch entirely absent. Good.

3. **`#[cfg(test)] mod tests` at `agent_loop.rs:990` is unaffected.** The test helper `new_session()` at line 1197 still sets `system_prompt_override: None` — that compiles because `cfg(test)` is always true under `cargo test --lib`, so the field is present. No change needed.

4. **CI/workflow impact is nil.** `.github/workflows/release.yml` doesn't run `cargo test`. `package.json:48` defines `rust:test` as `cargo test --lib -p nodespace-core` (scoped to the core crate, not agent), so nothing in the standard developer workflow expects `ollama_integration.rs` to build under bare `cargo test`. The `required-features` declaration correctly causes `--all-targets` clippy to skip — empirically verified by the clean `bun run quality:fix` run cited in the PR description.

5. **CLAUDE.md compliance.** No `#[deprecated]`, no `#[allow(dead_code)]`, no compatibility shims. The PR deletes the old "always-present field" design rather than wrapping it.

6. **Acceptance criteria from #1093.** The issue's stated intent — "move test infrastructure off the production domain model" — is delivered: in a production build, the field doesn't exist on `AgentSession`, the method doesn't exist on `LocalAgentService`, and the override-consumption branch is compiled out of `run_turn`.

The Improvement above is genuinely worth doing (DRY hazard on the duplicated chain), but it's not a merge blocker — the PR is a clear net positive in its current form and unblocks the intent of #1093 immediately.

malibio:
## Review feedback addressed

Pushed `106687cf` to address the review.

### Changes made

✅ **🟡 Improvement — Duplicated cfg branches for `base_prompt`** (`packages/agent/src/local_agent/agent_loop.rs`)

Introduced a small file-level helper `session_prompt_override(&AgentSession) -> Option<&str>` with two cfg variants:
- Under `cfg(any(test, feature = "testing"))`: returns `session.system_prompt_override.as_deref()`
- Under `cfg(not(...))`: returns `None`

The call site in `run_turn` now has a single `override → assembler → fallback` chain. The duplicated assembler/fallback logic that the reviewer flagged as a drift hazard is gone.

### Skipped (per reviewer's own guidance)

⏭️ **🟢 Nit — Feature naming (`testing` vs `test-hooks`/`test-support`)**
Reviewer explicitly noted "Not worth changing for this PR — flagging only because future contributors may add a second feature." No change.

⏭️ **🟢 Nit — Doc comment phrasing on the gate**
Reviewer explicitly noted "current wording is accurate for the production concern, which is what the doc cares about." No change.

### Verification

- `cargo clippy --all-targets -- -D warnings` (no features) — clean
- `cargo clippy -p nodespace-agent --features testing --tests -- -D warnings` — clean
- `bun run quality:fix` — clean

Diff for this commit: +17/-13 across one file.

PR #1141, opened 2026-05-17
Implement nodespace CLI binary (closes #1116)

malibio:
## Summary

Implements the `nodespace` CLI from issue #1116 — a thin gRPC client that connects to `nodespaced` on `localhost:50051` and exposes the knowledge graph as shell commands (Ollama-style).

- New `packages/cli` crate added to the workspace, ships the `nodespace` binary
- `clap`-derived subcommands: `node get|create|update|delete|children`, `search <query> [--limit]`
- Global `--json` flag and `--endpoint` override (also honored via `NODESPACE_ENDPOINT`)
- Friendly remediation when the daemon is unreachable instead of raw transport errors
- Human-readable and JSON output formatters in `src/output.rs`
- Integration test (`tests/cli_integration.rs`) drives the handlers against an in-process daemon: full create/get/update/children/delete round trip, not-found mapping, search-unavailable surface, and connect-refused friendly error

## Acceptance criteria

- [x] `nodespace` binary builds and is installable alongside `nodespaced` (`cargo build --bin nodespace`)
- [x] `nodespace node get <id>` retrieves and prints a node
- [x] `nodespace node create --type <type> --content <content> [--parent <id>]`
- [x] `nodespace node update <id> --content <content>`
- [x] `nodespace node delete <id>`
- [x] `nodespace node children <id>`
- [x] `nodespace search <query> [--limit <n>]`
- [x] `--json` flag on all commands outputs raw JSON
- [x] Clear error when `nodespaced` is not running (connection refused → friendly remediation)
- [x] `nodespace --help` and `nodespace <subcommand> --help` print useful usage text
- [x] `cargo build --bin nodespace` succeeds

## Test plan

- [x] Frontend baseline: 128 files / 3918 tests passing (before and after — no regression)
- [x] `cargo test -p nodespace-cli` — 4 integration tests pass
- [x] `cargo test -p nodespace-daemon` — 7 tests still pass (no regression)
- [x] `nodespace --help` / `nodespace node --help` / `nodespace search --help` reviewed manually
- [x] `nodespace --endpoint http://127.0.0.1:1 node get x` prints the friendly remediation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,10 +121,54 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
+name = "anstream"
+version = "0.6.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
 name = "anstyle"
 version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-query"
+version = "1.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
+dependencies = [
+ "windows-sys 0.60.2",
+]
+
+[[package]]
+name = "anstyle-wincon"
+version = "3.0.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
+dependencies = [
+ "anstyle",
+ "once_cell_polyfill",
+ "windows-sys 0.60.2",
+]
 
 [[package]]
 name = "antlr4rust"
@@ -1073,6 +1117,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2134bb3ea021b78629caa971416385309e0131b351b25e01dc16fb54e1b5fae"
 dependencies = [
  "clap_builder",
+ "clap_derive",
 ]
 
 [[package]]
@@ -1081,8 +1126,22 @@ version = "4.5.48"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2ba64afa3c0a6df7fa517765e31314e983f51dda798ffba27b988194fb65dc9"
 dependencies = [
+ "anstream",
  "anstyle",
  "clap_lex",
+ "strsim",
+]
+
+[[package]]
+name = "clap_derive"
+version = "4.5.47"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbfd7eae0b0f1a6e63d4b13c9c478de77c2eb546fba158ad50b4203dc24b9f9c"
+dependencies = [
+ "heck 0.5.0",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.104",
 ]
 
 [[package]]
@@ -1129,6 +1188,12 @@ dependencies = [
  "libc",
  "objc",
 ]
+
+[[package]]
+name = "colorchoice"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1d07550c9036bf2ae0c684c4297d503f838287c83c53686d05370d0e139ae570"
 
 [[package]]
 name = "combine"
@@ -3368,6 +3433,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "is_terminal_polyfill"
+version = "1.70.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a6cb138bb79a146c1bd460005623e142ef0181e3d0219cb493e02f7d08a35695"
+
+[[package]]
 name = "itertools"
 version = "0.10.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4154,6 +4225,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "nodespace-cli"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "nodespace-core",
+ "nodespace-daemon",
+ "serde_json",
+ "tempfile",
+ "tokio",
+ "tokio-stream",
+ "tonic 0.12.3",
+]
+
+[[package]]
 name = "nodespace-core"
 version = "0.1.0"
 dependencies = [
@@ -4656,6 +4742,12 @@ name = "once_cell"
 version = "1.21.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42f5e15c9953c5e4ccceeb2e7382a716482c34515315f7b03532b8b4e8393d2d"
+
+[[package]]
+name = "once_cell_polyfill"
+version = "1.70.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "384b8ab6d37215f3c5301a95a4accb5d64aa607f1fcb26a11b5303878451b4fe"
 
 [[package]]
 name = "oorandom"
@@ -8487,6 +8579,12 @@ name = "utf8_iter"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6c140620e7ffbb22c2dee59cafe6084a59b5ffc27a8859a5f0d494b5d52b6be"
+
+[[package]]
+name = "utf8parse"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
 
 [[package]]
 name = "uuid"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "packages/nlp-engine",
     "packages/agent",
     "packages/daemon",
+    "packages/cli",
     "packages/desktop-app/src-tauri",
     "packages/dev-tools"
 ]
--- a/packages/cli/Cargo.toml
+++ b/packages/cli/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "nodespace-cli"
+version = "0.1.0"
+description = "Command-line interface for NodeSpace — a gRPC client for nodespaced"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[[bin]]
+name = "nodespace"
+path = "src/main.rs"
+
+[dependencies]
+clap = { version = "4", features = ["derive", "env"] }
+tonic = "0.12"
+tokio = { workspace = true }
+anyhow = { workspace = true }
+serde_json = { workspace = true }
+nodespace-daemon = { path = "../daemon" }
+
+[dev-dependencies]
+nodespace-core = { workspace = true }
+tempfile = "3.0"
+tokio-stream = "0.1"
--- a/packages/cli/src/commands/mod.rs
+++ b/packages/cli/src/commands/mod.rs
@@ -1,0 +1,2 @@
+pub mod node;
+pub mod search;
--- a/packages/cli/src/commands/node.rs
+++ b/packages/cli/src/commands/node.rs
@@ -1,0 +1,167 @@
+//! `nodespace node ...` subcommands — thin gRPC wrappers around NodeService.
+
+use anyhow::{Context, Result};
+use clap::{Args, Subcommand};
+use nodespace_daemon::nodespace::{
+    CreateNodeRequest, DeleteNodeRequest, GetChildrenRequest, GetNodeRequest, UpdateNodeRequest,
+};
+use nodespace_daemon::NodeServiceClient;
+use tonic::transport::Channel;
+
+use crate::output;
+
+#[derive(Subcommand, Debug)]
+pub enum NodeAction {
+    /// Retrieve a node by ID.
+    Get(GetArgs),
+    /// Create a new node.
+    Create(CreateArgs),
+    /// Update an existing node's content.
+    Update(UpdateArgs),
+    /// Delete a node.
+    Delete(DeleteArgs),
+    /// List the direct children of a node.
+    Children(ChildrenArgs),
+}
+
+#[derive(Args, Debug)]
+pub struct GetArgs {
+    /// Node ID (UUID).
+    pub id: String,
+}
+
+#[derive(Args, Debug)]
+pub struct CreateArgs {
+    /// Node type, e.g. `text`, `task`, `date`.
+    #[arg(long = "type")]
+    pub node_type: String,
+    /// Content (plain text or markdown).
+    #[arg(long)]
+    pub content: String,
+    /// Parent node ID (omit to create a root node).
+    #[arg(long)]
+    pub parent: Option<String>,
+}
+
+#[derive(Args, Debug)]
+pub struct UpdateArgs {
+    /// Node ID to update.
+    pub id: String,
+    /// New content.
+    #[arg(long)]
+    pub content: String,
+}
+
+#[derive(Args, Debug)]
+pub struct DeleteArgs {
+    /// Node ID to delete.
+    pub id: String,
+}
+
+#[derive(Args, Debug)]
+pub struct ChildrenArgs {
+    /// Parent node ID.
+    pub id: String,
+}
+
+pub async fn run(
+    client: &mut NodeServiceClient<Channel>,
+    action: NodeAction,
+    json: bool,
+) -> Result<()> {
+    match action {
+        NodeAction::Get(args) => get(client, args, json).await,
+        NodeAction::Create(args) => create(client, args, json).await,
+        NodeAction::Update(args) => update(client, args, json).await,
+        NodeAction::Delete(args) => delete(client, args, json).await,
+        NodeAction::Children(args) => children(client, args, json).await,
+    }
+}
+
+async fn get(client: &mut NodeServiceClient<Channel>, args: GetArgs, json: bool) -> Result<()> {
+    let response = client
+        .get_node(GetNodeRequest { node_id: args.id })
+        .await
+        .context("GetNode RPC failed")?
+        .into_inner();
+
+    let node = response.node_data.context("daemon returned no node_data")?;
+    output::print_node(&node, json)
+}
+
+async fn create(
+    client: &mut NodeServiceClient<Channel>,
+    args: CreateArgs,
+    json: bool,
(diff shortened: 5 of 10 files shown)


malibio:
## Code Review — PR #1141: Implement nodespace CLI binary

**Review Type:** Initial pragmatic code review (no prior reviews on PR)
**Recommendation:** **APPROVE with minor follow-ups** — net-positive, scoped, well-tested. The findings below are improvements/nits, not blockers.

---

### Requirements Check — Issue #1116 Acceptance Criteria

| AC | Status | Evidence |
|---|---|---|
| `nodespace` binary builds and installable alongside `nodespaced` | ✅ | `packages/cli/Cargo.toml` declares `[[bin]] name = "nodespace"`; workspace member added in root `Cargo.toml:8`. `cargo build` succeeds. |
| `node get <id>` | ✅ | `packages/cli/src/commands/node.rs:81-90` |
| `node create --type --content [--parent]` | ✅ | `node.rs:92-112` |
| `node update <id> --content` | ✅ | `node.rs:114-136` |
| `node delete <id>` | ✅ | `node.rs:138-153` |
| `node children <id>` | ✅ | `node.rs:155-167` |
| `search <query>` | ✅ | `packages/cli/src/commands/search.rs:20-42` |
| `--json` global flag | ✅ | `lib.rs:29-30`, threaded through all handlers |
| Clear error when daemon not running | ✅ | Verified manually: `Could not connect to nodespaced at http://127.0.0.1:1. Is the daemon running? Start it with nodespaced in another terminal.` + original cause preserved via anyhow `Caused by:` chain. Test `connect_refused_returns_friendly_error` locks this in. |
| `--help` is useful | ✅ | Verified: top-level, `node`, and per-subcommand help all clear; `clap`-derived. |
| `cargo build --bin nodespace` succeeds | ✅ | Verified locally. |
| Passes `bun run quality:fix` | ✅ | `cargo fmt --check` clean; `cargo clippy -p nodespace-cli --all-targets -- -D warnings` clean. |

**All 12 acceptance criteria met.**

Bonus capabilities delivered beyond the AC list:
- `--endpoint` flag + `NODESPACE_ENDPOINT` env var override (sensible for testing & non-default deployments)
- 4 integration tests against an in-process daemon (round-trip, not-found, search-unavailable, connect-refused) — strong coverage for a stateless CLI

---

### Code Review Findings

#### Architectural Design & Integrity

🟡 **`packages/cli/Cargo.toml:19` — CLI depends on the entire daemon crate just for proto types**

The CLI imports generated proto types and the gRPC client via `nodespace-daemon = { path = "../daemon" }`. This means building `nodespace-cli` pulls in `nodespace-core`, `surrealdb`, `rocksdb`, and the full storage stack — none of which the CLI actually needs at runtime (the daemon owns the DB lock).

This is a real coupling issue but proportionate to a pre-release codebase. The cleaner shape per ISP/SRP is a small `nodespace-proto` crate that owns `tonic::include_proto!` and the generated types; both `nodespace-daemon` and `nodespace-cli` depend on that. The daemon would lose nothing structurally and the CLI would shed a large transitive dep tree (faster `cargo build`, smaller `target/`).

**Suggested follow-up issue**: extract `nodespace-proto`. Not a blocker for this PR — the current boundary works and the build still compiles fast for incremental CLI development.

🟢 Otherwise the architecture is clean: `lib.rs` is the public surface, `main.rs` is a 9-line driver, commands are one module per noun, `output.rs` isolates formatting. SRP-respecting.

#### Functionality & Correctness

🟡 **`packages/cli/src/lib.rs:64-73` — `connect()` error message is over-broad**

```rust
pub async fn connect(endpoint: &str) -> Result<NodeServiceClient<Channel>> {
    NodeServiceClient::connect(endpoint.to_string())
        .await
        .with_context(|| format!("Could not connect to nodespaced at {endpoint}...."))
}
```

`with_context` wraps *every* transport error in the same "daemon not running" message — TLS handshake failures, DNS resolution failures, IPv6-disabled-on-host, etc. all get the same hint. The doc comment claims the function distinguishes connection-refused; in practice it doesn't.

**Why it's acceptable today:** The underlying `tonic::transport::Error` survives in `err.chain()` (verified: `Caused by:` chain shows `Connection refused (os error 61)`). Users running `nodespace get` against `http://[::1]:50051` will hit connection-refused 99% of the time, so the hint is accurate in practice.

**Future improvement:** Inspect the error before deciding on the message:

```rust
.map_err(|e| {
    let is_refused = e.source()
        .and_then(|s| s.source())
        .map(|s| s.to_string().contains("Connection refused"))
        .unwrap_or(false);
    if is_refused {
        anyhow::anyhow!("Could not connect to nodespaced at {endpoint}.\n\
                         Is the daemon running? Start it with `nodespaced`.")
    } else {
        anyhow::Error::from(e).context(format!("Failed to dial {endpoint}"))
    }
})
```

Not a blocker — the current behavior is pragmatic. Worth filing a follow-up issue when other transport failure modes start appearing (e.g. once TLS is added).

🟡 **`packages/cli/src/commands/node.rs:122` — `update` always sends `version: None` with no user override**

```rust
version: None, // auto-fetch current version on the server
```

Auto-fetch matches the daemon's ops-layer behavior and is correct for an interactive CLI. But there is *no* CLI surface to pass an explicit version, which means:

1. The CLI is effectively last-write-wins from the user's perspective.
2. The `OpsError::VersionConflict → Status::aborted` path (`packages/daemon/src/services/node_service.rs:355`) is unreachable from the CLI.
3. A scripting use case that wants safe optimistic concurrency (read-modify-write loop) has no way to express it.

**Recommendation:** Add `--version <i64>` to `UpdateArgs` and `DeleteArgs` as an optional flag (`Option<i64>`). One-line clap addition; preserves auto-fetch as default. Worth doing in a follow-up since it changes the public CLI surface, not in this PR.

🟢 **`packages/cli/src/output.rs:88-107` — `node_to_json` properties fallback**

```rust
let properties = serde_json::from_str::<serde_json::Value>(&node.properties)
    .unwrap_or_else(|_| serde_json::Value::String(node.properties.clone()));
```

The comment says "falling back to the raw string when malformed so we never silently swallow daemon-side issues" — but `Value::String(raw)` *is* a silent swallow: the user sees the malformed payload as a JSON string, with no indication it was a parse failure. The intent and the behavior disagree.

In practice this fallback should be unreachable because the daemon emits `node.properties.to_string()` from a `serde_json::Value` (`packages/daemon/src/services/node_service.rs:343`), which is round-trippable by construction. So the branch only fires if the daemon contract breaks.

**Suggested options:**
- (Preferred) Make the fallback explicit: `serde_json::Value::Object({ "_raw": String(raw), "_parse_error": true })` so machine readers can detect it.
- (Or) `eprintln!` a warning to stderr in JSON mode.
- (Or) Keep current behavior and update the comment to be honest: "...falling back to the raw string in the (unreachable in practice) malformed-payload case to avoid panicking."

Not blocking — the branch is defensive code for an invariant the daemon already preserves.

#### Security

🟢 No findings. The CLI is local-only loopback by default. `--endpoint` accepts an arbitrary URL but it's a user-controlled input on the user's own machine; no privilege boundary is crossed. No shell-out, no eval, no file I/O of user input. `properties` JSON is constructed entirely server-side; the CLI never parses untrusted input from the wire as code.

🟢 The endpoint env var (`NODESPACE_ENDPOINT`) is documented and follows the same model as `NODESPACED_ADDR` on the daemon side — consistent.

#### Testing

🟢 **Solid coverage for 268 lines of test code:**
- `create_get_update_children_delete_round_trip` exercises every node mutation
- `get_missing_node_surfaces_not_found` confirms `tonic::Status` survives anyhow wrapping (important — this is what lets future code branch on RPC error codes)
- `search_without_embedding_service_reports_unavailable` locks in the `Unavailable` surfacing
- `connect_refused_returns_friendly_error` is the only test asserting on the user-visible message text — good

🟢 The decision to test against the library surface rather than the compiled binary is correct: failures point at code paths under test rather than fork/exec plumbing.

🟡 **Gaps worth filing as follow-ups (not blockers):**
- No assertion on JSON-mode stdout *shape*. The round-trip test passes `json=true` for some operations but doesn't capture or validate the JSON output. If `output::node_to_json` regresses (e.g. drops a field) the test still passes. Capturing stdout would require either `assert_cmd` against the built binary or replacing `println!` with a writer parameter — both feasible but a meaningful refactor.
- No test for the `--parent` flag end-to-end via the CLI handler. The round-trip seeds the parent via `raw_client` and then creates a child via the CLI; the child IS created with a parent (`commands::node::CreateArgs { parent: Some(parent_id.clone()) }` at `cli_integration.rs:123`) but the assertion at `cli_integration.rs:168-176` only checks that `Children` runs without error — it doesn't verify the count is 1 or the child ID matches. A `response.count == 1` assertion would tighten this considerably.
- No version-conflict test — but per the finding above, the CLI can't trigger it anyway, so this gap is consistent with the surface.

#### Maintainability & Readability

🟢 Naming is consistent (`*Args` structs per subcommand, `run` dispatchers per module, `print_*` formatters). Doc comments explain *why* (e.g. "Exposed primarily so integration tests can drive the command handlers..."), not *what*.

🟢 No `console.log`/raw I/O concerns in Rust — this is the Rust-side equivalent and `println!`/`eprintln!` are the right tools for a CLI. Compliant with the project logging policy (which is JS-specific).

🟢 No `#[allow(dead_code)]`, no `#[deprecated]`, no backward-compat code. Complies with CLAUDE.md pre-release rules.

🟢 **Nit:** `packages/cli/src/output.rs:76` — `if !node.properties.is_empty() && node.properties != "{}"` could move that check into a helper `has_properties(node)` for symmetry with how the JSON path handles it, but at one site it's overkill. Skip.

🟢 **Nit:** `packages/cli/src/commands/search.rs:17` — `--limit 0` is "server default", but a user typing `--limit -1` is also accepted by clap (no `value_parser` constraint). The server probably handles it gracefully, but `#[arg(long, default_value_t = 0, value_parser = clap::value_parser!(i32).range(0..))]` would reject negatives at parse time and shorten the trust boundary. One-line change for a future PR.

#### Performance

🟢 N/A — short-lived CLI, no hot paths, no allocation concerns. The integration test uses `tokio::time::sleep(Duration::from_millis(25))` in a 50-iteration retry loop (`cli_integration.rs:69-75`) which is fine for a CI runner but a 1.25s worst-case bound; if CI flakes here, switch to exponential backoff.

#### Documentation

🟢 Each module has a meaningful header doc-comment. `DEFAULT_ENDPOINT` references ADR-031. `connect()` has a rationale comment explaining the error-mapping intent. No over-documentation. No `--help` regression. The PR description summarizes the work; the commit message references the closed issue.

---

### Final Recommendation: **APPROVE**

This is a clean, well-tested, focused implementation that satisfies every acceptance criterion plus useful extras (endpoint override, env var, integration test harness). The findings above are improvements and follow-ups, not blockers:

- 🟡 Extract `nodespace-proto` shared crate (architectural cleanup — file as follow-up issue)
- 🟡 Add `--version` flag to `update`/`delete` for explicit optimistic concurrency (file as follow-up issue)
- 🟡 Tighten JSON-output regression coverage and the `--parent` end-to-end assertion (file as test-improvement issue)
- 🟡 Decide whether `node_to_json` should surface or swallow malformed `properties` — and align the comment with the chosen behavior (small follow-up)
- 🟢 Bound `--limit` to non-negative i32 at clap-parse time (nit, do whenever you next touch search)

Net effect on code health: clearly positive. Merge when CI is green.

malibio (commented):
Pragmatic code review complete — recommendation is APPROVE (cannot self-approve via API). Every acceptance criterion met, tests pass (4/4 integration tests green), clippy clean, fmt clean. Detailed findings posted as a comment on this PR. All findings are improvements/follow-ups, not blockers.

malibio:
## Review feedback addressed

### ✅ Addressed (4)
- **🟢 `search.rs:17`** — `--limit` now uses `value_parser!(i32).range(0..)` so negative values are rejected at clap-parse time with `invalid value '-5' for '--limit <LIMIT>': -5 is not in 0..=2147483647`.
- **🟢 `output.rs:88-93`** — rewrote the misleading comment on `node_to_json`'s properties fallback. Now states honestly that the branch is unreachable in practice and we degrade rather than panic if the daemon's serialization contract ever breaks. Bumped to `pub(crate)` for test access.
- **🟡 Testing gap (a)** — added two unit tests in `output.rs` covering JSON output shape (properties inlined as a nested object, scriptable via `jq '.properties.foo'`) and the malformed-properties fallback. A regression in `node_to_json` would now fail tests.
- **🟡 Testing gap (b)** — tightened `create_get_update_children_delete_round_trip` to assert `count == 1`, `nodes[0].content == "child via CLI"`, and `nodes[0].parent_id == parent_id` after the CLI's create-child handler. The CLI's effect on the graph is now actually verified.

### ⏭️ Skipped, with reasoning (3)
- **🟡 Extract `nodespace-proto` shared crate** — reviewer explicitly flagged this as a follow-up. Will file a separate issue rather than expand the scope of this PR.
- **🟡 Tighten `connect()` error message for non-refused transport errors** — reviewer noted this is acceptable until TLS arrives. Underlying error survives in the anyhow chain, so users still see `Caused by: Connection refused (os error 61)`. Premature without TLS.
- **🟡 Add `--version <i64>` flag for explicit optimistic concurrency** — meaningful feature addition (currently last-write-wins), but scope creep for a review cycle. Will file a follow-up issue so the version-conflict surface gets exposed deliberately rather than tucked into a review pass.

### Verification
- `cargo test -p nodespace-cli`: **6 passed** (2 new unit tests + 4 integration tests, was 4 before)
- `cargo fmt --check`: clean
- `cargo clippy -p nodespace-cli --all-targets -- -D warnings`: clean
- `bun run test`: 128 files / 3918 tests passing (no regression vs baseline)

## Re-Review Decision

**Rationale**: All changes are scoped corrections — a one-line clap-parser fix, a comment rewrite + visibility bump, two targeted unit tests, and additional asserts in an existing integration test. No new functionality, no architectural changes, no production-path behavior changes. Skipped recommendations are deferred to follow-up issues (with reasoning) rather than silently dropped. PR ready for merge.

malibio:
## Re-Review — PR #1141 (commit `d2fa8cac`)

**Review Type:** Re-Review
**Recommendation:** **APPROVE**

### Previous Review Summary

- **Findings flagged in initial review:** 7 (1 nit + 6 substantive improvements)
- **Addressed in `d2fa8cac`:** 4
  - `output.rs:88` — misleading comment on `node_to_json` properties fallback (rewritten, `pub(crate)` for test access)
  - Testing gap (a) — JSON-mode stdout shape: 2 new unit tests in `output.rs`
  - Testing gap (b) — `cli_integration.rs` children call now asserts `count`, `content`, `parent_id`
  - `search.rs:17` — `--limit` clamped to non-negative i32 at clap-parse time
- **Skipped, deferred to follow-up:** 3
  - Extract `nodespace-proto` shared crate
  - Tighten `connect()` error message for non-refused transport errors
  - Add `--version <i64>` flag to `update`/`delete`
- **New issues introduced:** 0

### Requirements Check — Issue #1116 Acceptance Criteria

Acceptance criteria are unchanged since initial review. **No regressions detected.** All 12 ACs still satisfied; `cargo test -p nodespace-cli` now reports 6/6 passing (2 new unit tests + 4 unchanged integration tests). `cargo fmt --check` clean, `cargo clippy -p nodespace-cli --all-targets -- -D warnings` clean.

### Code Review Findings — Follow-up Commit

#### Addressed Items — Verification

**🟢 `search.rs:17` — negative `--limit` rejected at clap-parse time**

```rust
#[arg(long, default_value_t = 0, value_parser = clap::value_parser!(i32).range(0..))]
pub limit: i32,
```

Verified locally: `nodespace search "test" --limit=-5` exits with code 2 and emits `error: invalid value '-5' for '--limit <LIMIT>': -5 is not in 0..=2147483647`. The `0..` (RangeFrom) bound is correctly oriented — it permits `[0, i32::MAX]` and rejects negatives. No clap signed/unsigned edge case applies here because the parser explicitly targets `i32`. The doc comment update ("0 = server default") makes the sentinel semantics legible. **Honored.**

**🟢 `output.rs:88-94` — comment rewrite + `pub(crate)` bump**

The new comment (lines 89-94) is accurate on both halves of the contract:
1. *"inline it as nested JSON so scripts can `jq '.properties.foo'` without a second parse"* — matches what the code does.
2. *"Falls back to the raw string if it doesn't parse; in practice this branch is unreachable because the daemon serializes via `serde_json::Value::to_string`, but we'd rather degrade than panic if that contract ever breaks"* — correctly characterizes the behavior as a *silent degrade* rather than the previous misleading "never silently swallow daemon-side issues" framing.

The `pub(crate)` bump is justified: the new `#[cfg(test)] mod tests` block (a sibling module) needs access. It does **not** widen the external API surface — `pub(crate)` is invisible across crate boundaries. The function remains internal to `nodespace-cli`. **Honored.**

**🟢 Testing gap (a) — `output.rs` unit tests**

The two new tests (`output.rs:131-153`) lock in the two regressions a future change to `node_to_json` is most likely to cause:

1. `node_to_json_inlines_properties_as_nested_object` — asserts `json["properties"]["foo"] == "bar"` and `json["properties"]["n"] == 42`. If `properties` ever regresses to being passed through as a JSON-encoded string (the most plausible refactor mistake), `json["properties"]["foo"]` becomes `Value::Null` and the test fails. Also locks in `id`, `version`, `parent_id`, and `collection_id` projection — covers the "dropped field" regression class as well.
2. `node_to_json_falls_back_to_raw_string_for_malformed_properties` — asserts the documented degrade behavior, pinning the contract the comment describes.

Tests use a pure, deterministic `sample_node()` helper — no time, no random, no shared state, no I/O. Safe under parallel `cargo test`. **Honored.**

**🟢 Testing gap (b) — `cli_integration.rs:180-204` cross-check**

The new assertions verify the CLI's *effect on the graph*, not just that the handler returned `Ok(())`:

```rust
let children = raw_client.get_children(...).await.expect(...).into_inner();
assert_eq!(children.count, 1, "expected exactly one child seeded via CLI");
assert_eq!(children.nodes.len(), 1, ...);
assert_eq!(children.nodes[0].content, "child via CLI", ...);
assert_eq!(children.nodes[0].parent_id.as_deref(), Some(parent_id.as_str()), ...);
```

This is meaningful coverage: it confirms that `commands::node::run(Create { parent: Some(parent_id) })` actually attached the child to the requested parent — previously the test only confirmed the call didn't panic. The cross-check pattern (drive via CLI handler, verify via raw gRPC) mirrors the existing update-verification block (`:155-165`) and keeps the test honest by reading state through a separate channel.

The test remains parallel-safe: the daemon binds to an ephemeral port (`:46`), the tempdir is per-test (`:31`), and `raw_client` is the test's own connection — no shared state across `#[tokio::test]` boundaries. **Honored.**

#### Skipped Items — Justification Assessment

**🟢 Extract `nodespace-proto` shared crate — defensible.**
Architectural cleanup. The initial review explicitly characterized it as a follow-up issue. Doing it in this PR would balloon the diff, touch `nodespace-daemon`, and risk distracting from the issue's stated scope (CLI binary). Deferring is the right call.

**🟢 Tighten `connect()` error message for non-refused transport errors — defensible.**
The initial review noted the underlying `tonic::transport::Error` survives in the anyhow `Caused by:` chain, so users still get the diagnostic. The differentiation only matters once non-refused failure modes become common (TLS handshake errors, DNS, etc.) — none of which exist yet. Premature now.

**🟢 Add `--version <i64>` flag — defensible.**
This changes the public CLI surface and is a real feature, not a fix. A follow-up issue gives it a deliberate design pass rather than tucking it into a review cycle. The current last-write-wins behavior matches the ops-layer auto-fetch semantics and is correct for an interactive CLI.

All three skipped items are flagged for follow-up issues per the address-review summary. Acceptable.

#### New Issues Introduced by `d2fa8cac`

None.

I specifically checked for:
- **API leak from `pub(crate)` bump on `node_to_json`:** No — `pub(crate)` does not cross crate boundaries; external consumers see no change.
- **clap signed/unsigned edge cases on `value_parser!(i32).range(0..)`:** No — the parser is explicit about `i32`, the range bound is correctly oriented, manual verification confirms `-5` is rejected with a clear message.
- **Comment/behavior alignment in `output.rs`:** Comment now matches the code (silent degrade, not "never swallow"). Aligned.
- **Test determinism:** Unit tests have no shared state, no time/random dependency. Integration tests use ephemeral ports + per-test tempdirs. Parallel-safe.
- **Test cross-check pattern in `cli_integration.rs`:** The new asserts query via `raw_client` rather than capturing CLI stdout. This is consistent with the existing update-verification pattern in the same test — the trade-off (no stdout shape coverage in integration tests) is now compensated for by the new `output.rs` unit tests. Reasonable split.

### Final Recommendation: **APPROVE**

The follow-up commit `d2fa8cac` is scoped, faithful to the prior review, and net-positive: four substantive improvements landed cleanly, three were deferred with sound reasoning, and the diff introduces no new concerns. Test count rose from 4 to 6 with both new tests pinning regressions the initial review explicitly flagged. CI signals clean (fmt, clippy, all tests pass).

Merge when CI is green.

PR #1148, opened 2026-05-18
Implement AgentSessionService gRPC handler (closes #1119)

malibio:
## Summary

- Wires the PTY engine from #1118 into `nodespaced` as the `AgentSessionService` gRPC handler — six RPCs (Launch, StreamOutput, WriteInput, ResizeTerminal, TerminateSession, ListSessions) backed by `Arc<PtySessionManager>` + `Arc<GraphContextAssembler>`.
- Aligns `agent_session_service.proto` with the actual `PtySessionManager` API (the original shape from #1111 predated the engine and would have required either a parallel raw-command launch path or a translation layer — both pure tech debt for a greenfield daemon).
- Adds 8 integration tests against the in-process tonic server, covering all six RPCs and the "client disconnect doesn't kill the session" contract (acceptance criterion #8).

## Acceptance criteria

- [x] `AgentSessionService` handler is registered with the `nodespaced` gRPC server (both headless + tray loops).
- [x] `LaunchSession` creates a PTY session and returns the session ID.
- [x] `StreamOutput` opens a server-streaming response that delivers `OutputChunk` messages in real time.
- [x] `WriteInput` forwards bytes to the PTY stdin.
- [x] `ResizeTerminal` resizes the PTY.
- [x] `TerminateSession` kills the session and triggers cleanup.
- [x] `ListSessions` returns metadata for all active sessions.
- [x] Client disconnecting from `StreamOutput` does not kill the session.
- [x] Integration test: launch a session running `echo hello`, verify `StreamOutput` delivers the output.
- [x] `cargo build --bin nodespaced` succeeds.
- [x] Code passes `bun run quality:fix`.

## Test plan

- [x] `cargo build --bin nodespaced` — success
- [x] `cargo test -p nodespace-daemon` — 24 passed (6 unit + 8 agent-session integration + 10 grpc-round-trip)
- [x] `cargo test -p nodespace-agent --lib --features testing` — 250 passed
- [x] `bun run quality:fix` — clean (eslint + svelte-check + rustfmt + clippy with `-D warnings -D clippy::unused-async`)
- [x] `bun run test` — 3918/3918 passed across 128 files (matches baseline)
- [x] `bun run test:all` — exit 0

## Notable design decisions

**Proto rewrite.** The proto from #1111 used raw `command/args/cwd/env`; the engine from #1118 takes `AgentType` + `Option<String> prompt` + `&GraphContextAssembler`. The rewrite aligns the proto with the engine — see the commit body for the full rationale. Nothing else in the workspace consumed the old proto.

**Test helper exposure.** `PtySession::launch_for_test` and `session_dir_path` moved from `pub(crate) + #[cfg(test)]` to `pub + #[cfg(any(test, feature = "testing"))]` so the daemon's integration tests can spawn `sh`/`cat`/`echo` without a real agent binary on PATH. The `testing` feature already exists in the agent crate for the same reason (ollama_integration tests).

**Lint-clean error helpers.** `parse_session_id` and `parse_agent_type` return `Result<T, String>` rather than `Result<T, Status>` to avoid tripping `clippy::result_large_err` (tonic::Status is ~176 bytes, dwarfs `Uuid`/`AgentType`). Call sites wrap the string in `Status::invalid_argument` in one line.

**Drive-by fix from rebase.** PR #1147 added `id` + `insert_after_node_id` fields to `CreateNodeRequest` but didn't update `packages/daemon/tests/grpc_round_trip.rs` or `packages/cli/{src/commands/node.rs, tests/cli_integration.rs}`. The existing `rust:test` script only runs `nodespace-core --lib`, so the breakage didn't surface in CI. Added the two missing fields (`String::new()`) so the full daemon test binary and cli build compile on top of main.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,7 +44,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.17",
+ "getrandom 0.2.16",
  "once_cell",
  "version_check",
 ]
@@ -64,9 +64,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
@@ -101,7 +101,7 @@ dependencies = [
  "cssparser 0.35.0",
  "html5ever 0.35.0",
  "maplit",
- "tendril 0.4.3",
+ "tendril",
  "url",
 ]
 
@@ -122,9 +122,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "1.0.0"
+version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -137,15 +137,15 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.14"
+version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
+checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
 
 [[package]]
 name = "anstyle-parse"
-version = "1.0.0"
+version = "0.2.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "52ce7f38b242319f7cabaa6813055467063ecdc9d355bbb4ce0c68908cd8130e"
+checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
 dependencies = [
  "utf8parse",
 ]
@@ -156,7 +156,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.61.1",
 ]
 
 [[package]]
@@ -167,7 +167,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.61.1",
 ]
 
 [[package]]
@@ -210,7 +210,7 @@ checksum = "3c3610892ee6e0cbce8ae2700349fcf8f98adb0dbfbee85aec3c9179d29cc072"
 dependencies = [
  "base64ct",
  "blake2",
- "cpufeatures 0.2.17",
+ "cpufeatures",
  "password-hash",
 ]
 
@@ -239,6 +239,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "ashpd"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6cbdf310d77fd3aaee6ea2093db7011dc2d35d2eb3481e5607f1f8d942ed99df"
+dependencies = [
+ "enumflags2",
+ "futures-channel",
+ "futures-util",
+ "rand 0.9.2",
+ "raw-window-handle",
+ "serde",
+ "serde_repr",
+ "tokio",
+ "url",
+ "wayland-backend",
+ "wayland-client",
+ "wayland-protocols",
+ "zbus",
+]
+
+[[package]]
 name = "async-broadcast"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -264,9 +285,9 @@ dependencies = [
 
 [[package]]
 name = "async-executor"
-version = "1.14.0"
+version = "1.13.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c96bf972d85afc50bf5ab8fe2d54d1586b4e0b46c97c50a0c9e71e2f7bcd812a"
+checksum = "bb812ffb58524bdd10860d7d974e2f01cc0950c2438a74ee5ec2e2280c6c4ffa"
 dependencies = [
  "async-task",
  "concurrent-queue",
@@ -291,14 +312,14 @@ dependencies = [
  "polling",
  "rustix",
  "slab",
- "windows-sys 0.61.2",
+ "windows-sys 0.61.1",
 ]
 
 [[package]]
 name = "async-lock"
-version = "3.4.2"
+version = "3.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
+checksum = "5fd03604047cee9b6ce9de9f70c6cd540a0520c813cbd49bae61f33ab80ed1dc"
 dependencies = [
  "event-listener",
  "event-listener-strategy",
@@ -307,9 +328,9 @@ dependencies = [
 
 [[package]]
 name = "async-process"
-version = "2.5.0"
+version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc50921ec0055cdd8a16de48773bfeec5c972598674347252c0399676be7da75"
+checksum = "65daa13722ad51e6ab1a1b9c01299142bc75135b337923cfa10e79bbbd669f00"
 dependencies = [
  "async-channel",
  "async-io",
@@ -331,14 +352,14 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 2.0.104",
 ]
 
 [[package]]
 name = "async-signal"
-version = "0.2.14"
+version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "52b5aaafa020cf5053a01f2a60e8ff5dccf550f0f77ec54a4e47285ac2bab485"
+checksum = "f567af260ef69e1d52c2b560ce0ea230763e6fbb9214a85d768760a920e3e3c1"
 dependencies = [
  "async-io",
  "async-lock",
@@ -349,7 +370,7 @@ dependencies = [
  "rustix",
  "signal-hook-registry",
  "slab",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -371,7 +392,7 @@ checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 2.0.104",
 ]
 
 [[package]]
@@ -388,7 +409,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 2.0.104",
 ]
 
 [[package]]
@@ -437,9 +458,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.17.0"
+version = "1.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ec2f1fc3ec205783a5da9a7e6c1509cc69dedf09a1949e412c1e18469326d00"
+checksum = "94bffc006df10ac2a68c83692d734a465f8ee6c5b384d8545a636f81d858f4bf"
 dependencies = [
  "aws-lc-sys",
  "untrusted 0.7.1",
@@ -448,9 +469,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.41.0"
+version = "0.38.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1a2f9779ce85b93ab6170dd940ad0169b5766ff848247aff13bb788b832fe3f4"
+checksum = "4321e568ed89bb5a7d291a7f37997c2c0df89809d7b6d12062c81ddb54aa782e"
 dependencies = [
  "cc",
  "cmake",
@@ -487,7 +508,7 @@ dependencies = [
  "serde_urlencoded",
  "sync_wrapper",
  "tokio",
- "tower 0.5.3",
+ "tower 0.5.2",
  "tower-layer",
  "tower-service",
  "tracing",
@@ -495,9 +516,9 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.8.9"
+version = "0.8.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
+checksum = "8b52af3cb4058c895d37317bb27508dccc8e5f2d39454016b297bf4a400597b8"
 dependencies = [
  "axum-core 0.5.6",
  "bytes",
@@ -513,7 +534,7 @@ dependencies = [
  "pin-project-lite",
  "serde_core",
  "sync_wrapper",
- "tower 0.5.3",
+ "tower 0.5.2",
  "tower-layer",
  "tower-service",
 ]
@@ -565,7 +586,7 @@ checksum = "57d123550fa8d071b7255cb0cc04dc302baa6c8c4a79f55701552684d8399bce"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.117",
+ "syn 2.0.104",
 ]
 
 [[package]]
@@ -588,9 +609,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "base64ct"
-version = "1.8.3"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
+checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
 
 [[package]]
 name = "bcrypt"
@@ -617,7 +638,7 @@ version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
 dependencies = [
- "bitflags 2.11.1",
+ "bitflags 2.9.1",
  "cexpr",
  "clang-sys",
  "itertools 0.13.0",
@@ -628,7 +649,7 @@ dependencies = [
  "regex",
  "rustc-hash",
  "shlex",
- "syn 2.0.117",
+ "syn 2.0.104",
 ]
 
 [[package]]
@@ -654,11 +675,11 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.11.1"
+version = "2.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c4512299f36f043ab09a583e57bceb5a5aab7a73db1805848e8fef3c9e8c78b3"
+checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
 dependencies = [
- "serde_core",
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


malibio (commented):
### Code Review Summary

**Verdict: Approve with minor improvements.** This is a high-quality, net-positive PR that lands a clean gRPC adapter over the PTY engine from #1118. All 11 acceptance criteria are met, all 8 integration tests pass against the in-process tonic server, clippy is clean under `-D warnings -D clippy::unused-async`, and the design rationale in the PR body is well-reasoned.

The proto rewrite is the right call — aligning the wire format with the engine's actual contract rather than maintaining a translation layer matches the project's greenfield "no backwards-compat" directive in CLAUDE.md. The split of `build_node_service` → `build_services` returning a `DaemonServices` bundle is a clean refactor that scales naturally as more services land. Test coverage of the streaming/disconnect contract (the trickiest piece) is genuinely thorough.

A few small but worthwhile improvements are flagged below — none are blockers.

---

### Findings

#### Critical Issues

None.

---

#### Suggested Improvements

**1. `[Improvement]` — `LaunchSessionResponse.created_at` should reflect the session's actual `started_at`, not a fresh `Utc::now()`**

`packages/daemon/src/services/agent_session_service.rs:65`

```rust
let created_at = current_unix_secs();
Ok(Response::new(LaunchSessionResponse {
    session_id: id.to_string(),
    created_at,
}))
```

`PtySession.started_at` is set inside `spawn_in_pty` *before* the reader/watcher tasks are wired (`session.rs:201`) and is the authoritative timestamp. The handler then takes a second `Utc::now()` after `manager.launch().await` and an optional `apply_resize` round-trip — these are small but non-zero gaps, and the timestamp returned to the client is therefore later than the one `ListSessions` reports for the same session.

Suggested fix (one extra `manager.get(&id)` after launch):

```rust
let id = self.manager.launch(...).await.map_err(...)?;
if req.cols != 0 && req.rows != 0 {
    apply_resize(&self.manager, &id, req.cols, req.rows).await?;
}
let created_at = self
    .manager
    .get(&id)
    .await
    .map(|s| s.started_at.timestamp())
    .unwrap_or_else(current_unix_secs);
```

The `unwrap_or_else` covers the (vanishingly unlikely) case where the session exits and the auto-prune watcher removes it between `launch` and `get`. Cleaner alternative: have `PtySessionManager::launch` return `(Uuid, DateTime<Utc>)` — but that's a wider change and arguably belongs in a follow-up.

Engineering principle: **single source of truth.** A client comparing `LaunchSessionResponse.created_at` to `SessionInfo.started_at` should see equality, not a few-millisecond drift.

---

**2. `[Improvement]` — `WriteInputResponse.bytes_written: len as i32` can silently truncate or wrap negative**

`packages/daemon/src/services/agent_session_service.rs:140`

```rust
let len = req.data.len();
// ...
bytes_written: len as i32,
```

`req.data.len()` is `usize`. For payloads ≥ 2 GiB, `as i32` either truncates or produces a negative number (the cast is wraparound, not saturating). The proto comment says this always equals `data.len()` on success, so the truncation would silently violate the contract.

In practice nobody is going to send a 2 GiB keystroke buffer to a PTY, but the same `as i32` lives in `ListSessionsResponse.count` (line 187) where a misbehaving caller could plausibly leak more than `i32::MAX` sessions over a long-lived daemon (no, they really couldn't — but the cast is still wrong-shaped).

Two options, in order of preference:

1. **Change the proto** — `bytes_written: int32` → `uint64`, `count: int32` → `uint32`. Greenfield, no users, single consumer (this handler) — exactly the kind of change the project's no-backwards-compat directive enables.
2. **Saturate the cast** — `bytes_written: i32::try_from(len).unwrap_or(i32::MAX)`. Same silent-clamp problem as today, just bounded.

Option 1 is cleaner and matches the project's stance on schema evolution.

Engineering principle: **fail loudly or model correctly.** A silent truncating cast is the worst of both worlds.

---

**3. `[Improvement]` — `apply_resize` performs a redundant `manager.get()` after `LaunchSession`**

`packages/daemon/src/services/agent_session_service.rs:62, 232–252`

`launch_session` calls `apply_resize(..)` immediately after `manager.launch(..)`. Inside `apply_resize`, the first thing it does is `manager.get(id)` to find the session it just created. That second lookup acquires the manager's `Mutex<HashMap>` for the second time in two lines.

The cost is real but small (one async lock acquisition). More importantly, between `launch` and `apply_resize`, the session could in principle have exited (e.g., the spawned binary crashes immediately) and been pruned, causing the freshly-launched session's resize to return `NotFound` to the caller — confusing because the launch just succeeded.

Cleaner shape: have `PtySessionManager::launch` accept an optional initial `PtySize`, or have `launch_session` hold an `Arc<PtySession>` from a single lookup and pass it to a session-direct `resize` helper. Either keeps the launch-with-dimensions path atomic.

Nit-level — the current behavior is documented and correct for the success case — but worth a follow-up if the daemon grows more launch-time parameters.

---

**4. `[Improvement]` — `StreamOutput` lagged-chunk swallowing is invisible to the client**

`packages/daemon/src/services/agent_session_service.rs:99–110`

You explicitly asked about this in the review prompt. My read: a `tracing::warn!` is the right floor, but consider also emitting a single `OutputChunk` with `data: vec![]` and a marker byte sequence (e.g., literal `b"\x1b[?25h\r\n[output dropped: N bytes]\r\n"`) so the rendered terminal *visually* shows the gap. ANSI-aware UIs already render this kind of control sequence inline.

Alternatively, expose `lagged_chunks_total` as a session-level metric returned by `ListSessions` — that's a more conservative interface change.

Lowest-risk option: leave the current behavior, add `assert!(skipped > 0)` and a TODO note pointing at this discussion so future maintainers don't have to rediscover the trade-off. The current behavior is reasonable; the only failure mode is the user not knowing they lost frames.

Engineering principle: **observable degradation > silent degradation.** The cost of telling the user "you missed N bytes" is one extra protobuf field; the cost of not telling them is a confused bug report when their terminal scrollback skips a chunk.

---

**5. `[Improvement]` — `apply_resize`'s 65535 ceiling is fine, but the proto should match**

`packages/daemon/src/services/agent_session_service.rs:243–246`

You also asked about this. The `u16` boundary is correct — `portable_pty::PtySize` uses `u16` for rows/cols, and no real terminal goes past that. The right move is to change the proto from `uint32` to a comment that documents the `[1, 65535]` range and let the `try_from` reject the rest with `InvalidArgument` (which is what you're doing).

Two strictly-stronger improvements:

1. Reject zero values too — `cols: 0, rows: 0` to `ResizeTerminal` (as opposed to `LaunchSession`, where zero means "use default") would currently call `portable-pty` with `PtySize { rows: 0, cols: 0, ... }`, which the platform may or may not reject. Better to fail at the API boundary.
2. The proto comment could note the valid range explicitly.

---

**6. `[Improvement]` — Integration tests do not exercise the `nodespace_agent::pty` `Cargo.toml` feature-unification assumption**

`packages/daemon/Cargo.toml:42–47`

`nodespace-agent` is a regular dependency and `nodespace-agent = { features = ["testing"] }` is a dev-dependency. Cargo unifies features across dep/dev-dep within a single package, so during `cargo test -p nodespace-daemon`, the daemon's production code is compiled against `nodespace-agent` *with* the `testing` feature enabled.

In this PR that's fine — `testing` only widens `pub(crate)` to `pub` and doesn't change behavior. But it's an easy footgun if someone later puts `#[cfg(feature = "testing")]` on something more invasive. A defensive measure is to add a `compile_error!` in the agent crate if `testing` and (some hypothetical) `production` feature are both enabled, or just to document this on the `testing` feature in `packages/agent/Cargo.toml`.

The comment in the daemon's `Cargo.toml` is good; consider mirroring it in the agent's `Cargo.toml` `[features]` section.

---

**7. `[Improvement]` — Drive-by `String::new()` for the new `id` field is correct, but the integration tests should pin the server-side UUID generation contract**

`packages/daemon/tests/grpc_round_trip.rs:96, 138, 183, 199, 250, 348, 437, 498, 527` and the cli sites.

You explicitly asked. `String::new()` matches the proto's "empty = server generates" contract, and the test coverage in `nodespace-core` for the caller-supplied-UUID path is presumably already there (in `nodespace-core` per #1147). What's missing in the daemon test suite is a single test that confirms the daemon layer correctly passes the caller-supplied UUID through to core when non-empty — that's the contract the gRPC handler is responsible for honoring.

Suggested follow-up (out of scope for this PR): one extra test in `grpc_round_trip.rs` that does:

```rust
let supplied_id = Uuid::new_v4().to_string();
let resp = client.create_node(CreateNodeRequest { id: supplied_id.clone(), ... }).await?;
assert_eq!(resp.into_inner().id, supplied_id);
```

For this PR specifically, the `String::new()` drive-by is the right minimal fix.

---

#### Nitpicks

- **Nit:** `packages/daemon/src/services/agent_session_service.rs:221–230` — `agent_type_to_string` falls back to `format!("{agent_type:?}")` if serde fails. You acknowledged this is dead code today. A `debug_assert!(false, ...)` would catch a future serde regression in CI without breaking production. Or use `serde::ser::Error` more explicitly via `serde_plain::to_string` which can't fail for unit-variant enums. Optional polish.

- **Nit:** `packages/daemon/src/services/agent_session_service.rs:212–219` — `parse_agent_type` accepts both kebab- and snake-case for "historical/UX reasons". For a greenfield project with zero callers, this dual-accept is exactly the kind of "backward-compat code" CLAUDE.md says to delete. Pick one canonical form (the kebab-case serde form is the obvious choice) and reject the other with a clear error pointing at the right spelling. The snake-case acceptance is already breeding ambiguity — the proto comment names `claude_code` but the canonical form is `claude-code`.

- **Nit:** `packages/daemon/proto/agent_session_service.proto:35–37` — the comment "Values of 0 fall back to the engine default (80x24); callers are expected to resize as soon as they know the real terminal geometry" is great. Consider applying the same "0 means default" convention to `ResizeTerminal` and document it, or explicitly call out that `ResizeTerminal` rejects zero (consistent with point #5).

- **Nit:** `packages/daemon/tests/agent_session_grpc.rs:78–86` — the connection-retry loop polls every 25ms for 50 iterations (1.25s ceiling). If the loop exits via `panic!`, `last_err` is always `Some` by construction (first iteration always sets it on failure), but the `Option` indirection obscures that. Replace with a `Result` accumulator or just `unwrap()` (the `panic!` formatter already prints `Some(e)` correctly).

- **Nit:** `packages/daemon/src/main.rs:188–199` — `embedding_service = None` is wired identically in two spots (the `node` and the `assembler` arms). Consider naming this `None::<Arc<dyn EmbeddingService>>` once and binding to both, or leaving a single-line comment noting both consumers will be wired by the same follow-up.

- **Nit:** `packages/daemon/src/services/agent_session_service.rs:188` — `count: count` could be `count` (field-init-shorthand), but clippy didn't flag it so it's clearly fine.

---

All 11 criteria from #1119 are met. Verified locally:

- `cargo build --bin nodespaced` — clean (10.80s)
- `cargo test -p nodespace-daemon --lib` — 6/6 passed (parse helpers + tray)
- `cargo test -p nodespace-daemon --test agent_session_grpc` — 8/8 passed
- `cargo clippy -p nodespace-daemon --all-targets -- -D warnings -D clippy::unused-async` — clean (24.22s)

`stream_output_disconnect_does_not_kill_session` (lines 262–307 of the test file) is the key behavioural test and it exercises the critical-path contract correctly: dropping the stream client-side does not propagate to the session, which is verified by a subsequent `list_sessions` showing the session still present.

### Standards Compliance

- ✅ No `#[allow(...)]` introduced in new code. The single `#![allow(clippy::all)]` in `lib.rs` is pre-existing and scoped to the auto-generated proto module — standard tonic practice, not a violation.
- ✅ No `#[deprecated]`, no migration code, no dual-mode support.
- ✅ Greenfield-correct: proto rewrite without translation layer is exactly what CLAUDE.md prescribes.
- ✅ Rust-only PR, so the frontend `createLogger`/Bun rules don't apply.
- ⚠️ One minor exception: the snake-case agent-type acceptance in `parse_agent_type` is technically backwards-compat code (see Nit above). Recommend removing.

### Security

- ✅ All session_id inputs are parsed through `Uuid::from_str` before use — no injection vector via the session_id field.
- ✅ `WriteInput` forwards raw bytes to the PTY verbatim, which is the correct behavior for a terminal (Ctrl-C, escape sequences, etc., must pass through). The `bytes` field is length-bounded by tonic's default message size limits.
- ✅ `LaunchSession` only accepts an enumerated `agent_type` and an `Option<String>` prompt — there is no path to inject an arbitrary command or arguments. The actual binary is resolved by the agent registry, not the request.
- ✅ No secrets, no PII in logs. `session_id` in the `lagged` warn is a UUID — fine.

### Performance

- The `manager.get(&id)` lookups acquire the manager's `Mutex<HashMap>` per RPC. For the expected workload (a handful of agent sessions per user) this is fine. If session counts ever grow into the hundreds and `ListSessions` becomes hot, swap the inner `Mutex<HashMap>` for `DashMap` or `RwLock<HashMap>`. Out of scope here.
- `StreamOutput` clones each `OutputChunk.data: Vec<u8>` once when copying into the proto message. For typical PTY chunk sizes (≤4 KiB per `READ_BUFFER_BYTES`) this is negligible; a future optimization could use `Bytes` end-to-end if profiling shows it matters.
- No N+1 patterns, no blocking calls on the async runtime.

---

Excellent work — ship it. The improvements above are worth a follow-up but none are blockers.

malibio:
## Review feedback addressed — 7 of 12 implemented

Pushed as 10041e35. Summary of the response to each item below.

### ✅ Addressed

| # | Item | Change |
|---|---|---|
| 1 | `LaunchSessionResponse.created_at` should equal `started_at` | Now sourced from `session.started_at.timestamp()` via one extra `manager.get()` after launch. Single source of truth between LaunchSession and ListSessions. |
| 2 | Silent `usize as i32` casts | Proto widened: `bytes_written` int32 → int64; `count` int32 → uint32. Saturating cast on `count` via `u32::try_from(...).unwrap_or(u32::MAX)`. |
| 3 | Redundant `manager.get()` in launch+resize | Single lookup; hold `Arc<PtySession>` through both `started_at` and optional initial resize. Closes the small race where the auto-prune watcher could remove the entry between launch and resize. New helper `resize_session(&Arc<PtySession>, ...)` replaces `apply_resize(manager, id, ...)`. |
| 4 | `StreamOutput` lag invisible to client | Lowest-risk variant — kept the swallow-and-warn behavior (surfacing dropped bytes inline is a wider design change that warrants its own issue), added `debug_assert!(skipped > 0)` and a TODO note pointing at the review discussion. |
| 5 | `ResizeTerminal` accepts zero dimensions | Now returns `InvalidArgument` when either axis is zero. New integration test `resize_terminal_rejects_zero_dimensions` pins the contract across (0,N), (N,0), (0,0). Proto comment documents the [1, 65535] valid range. |
| 6 | Document testing-feature footgun | Expanded the `[features] testing = []` comment in `packages/agent/Cargo.toml` to explain Cargo's feature-unification across dep/dev-dep and the rule that gates behind this flag must only widen visibility — never change runtime behavior. |
| Nit | Snake-case agent_type acceptance | Removed (per CLAUDE.md no-backwards-compat). Kebab-case is the only accepted form. New test `parse_agent_type_rejects_snake_case` pins the rejection so a future refactor cannot silently restore the dual-accept path. |

### ⏭️ Skipped, with reasoning

- **#7 Daemon test for caller-supplied UUID through CreateNode**: Reviewer marked "out of scope for this PR"; contract is already exercised in `nodespace-core` per #1147. Worth a follow-up issue if the daemon adapter ever diverges from a simple pass-through.
- **Nit: `agent_type_to_string` debug_assert**: The dead-code fallback `format!("{agent_type:?}")` is genuinely defensive — keeps the function infallible if `AgentType` ever gains a non-serde variant later. A `debug_assert!(false)` would catch that in test but would also panic in production for a recoverable problem.
- **Nit: connect-retry `last_err` Option indirection**: Stylistic; the existing shape already does the right thing (`panic!` formats `Some(e)` fine). Not worth a churn commit.
- **Nit: `count: count` field-init shorthand**: Clippy didn't flag; the explicit form is consistent with other struct literals in the same file.
- **Nit: `embedding_service = None` duplication in main.rs**: Will naturally resolve when the embedding service is wired by the follow-up issue (both arms become `Some(embedding_service.clone())`). Introducing a binding now would need un-introducing later.

### Verified

- `cargo build --bin nodespaced`: clean
- `cargo test -p nodespace-daemon`: **26 passed** (7 unit + 9 agent-session integration + 10 grpc-round-trip — added 2 new tests: zero-reject + snake-case rejection)
- `cargo clippy -p nodespace-daemon --all-targets -- -D warnings -D clippy::unused-async`: clean
- `bun run quality:fix`: clean (full workspace)
- `bun run test`: 3918/3918 across 128 files (matches baseline)

malibio:
## Re-Review of commit 10041e35 (LGTM — would approve if not own PR)

Reviewed only the changes since commit `4ce8c73d` (the initial review point). Verified locally with `cargo build --bin nodespaced`, `cargo test -p nodespace-daemon` (26/26 passing, including the new `resize_terminal_rejects_zero_dimensions`), and `cargo clippy -p nodespace-daemon --all-targets -- -D warnings -D clippy::unused-async` (clean).

### Summary
This is a tight, focused commit that addresses every reviewer item the author committed to. Each fix is grounded in the right principle (single source of truth, type-correct contracts, defensive boundaries) and is accompanied by a comment explaining the *why*.

### Verification of addressed items

**#1 `created_at` matches `started_at` (verified)**
- `LaunchSessionResponse.created_at` now reads `session.started_at.timestamp()` via one `manager.get()` after launch. Confirmed `SessionMetadata::from_session` reads the same `PtySession.started_at` field — true single source of truth between LaunchSession and ListSessions.
- The `unwrap_or_else(current_unix_secs)` fallback is genuinely reachable (auto-prune watcher can fire if the child exits between `launch()` and `get()`). The comment correctly describes this as "vanishingly unlikely" rather than dead code. Good call not turning it into `expect()`.

**#3 Single `Arc<PtySession>` lookup (verified)**
- `launch_session` now holds one `Option<Arc<PtySession>>` through both `started_at` extraction and the optional initial resize. The new `resize_session(&Arc<PtySession>, ...)` helper takes the Arc directly, so the second manager lookup is gone. The `Arc` also closes the auto-prune race the author called out.
- Ordering is correct: `started_at` is captured at PTY spawn time (line 201 in `pty/session.rs`), so the resize-after-timestamp-read order has no semantic impact.

**#5 ResizeTerminal rejects zero (verified)**
- Rejection happens at the API boundary *before* manager lookup, returning `InvalidArgument` with a clear error message. Test `resize_terminal_rejects_zero_dimensions` pins all three permutations `(0, N)`, `(N, 0)`, `(0, 0)`. Proto comment documents the `[1, 65535]` range and contrasts the semantics against `LaunchSession`. Clean.

**#2 Proto widening (verified)**
- `WriteInputResponse.bytes_written`: `int32 → int64`. Cast becomes `len as i64` which is lossless on every platform with `usize ≤ 64 bits`. Correct.
- `ListSessionsResponse.count`: `int32 → uint32` with saturating cast `u32::try_from(sessions.len()).unwrap_or(u32::MAX)`. Monotonic-for-snapshots reasoning is sound.
- No external consumer breakage: `AgentSessionService` is only consumed by `nodespace-daemon` Rust code today (no agent-tools TS or CLI usage), so the wire format change is safe.

**#4 `RecvError::Lagged` debug_assert + TODO (verified)**
- Per tokio's broadcast channel implementation, `Lagged(n)` is only emitted when `lost > 0`, so `debug_assert!(skipped > 0)` will not produce spurious failures. Safe in test, no-op in release. The TODO with concrete deferred options is exactly the right level of follow-up.

**#6 Cargo feature footgun documentation (verified)**
- The expanded `[features] testing = []` comment in `packages/agent/Cargo.toml` clearly states the rule: gates behind this flag must only widen visibility, never change runtime behavior. Captures the cross-package feature-unification gotcha precisely. This will save a future contributor hours.

**#7 (nit) Snake-case rejection (verified)**
- `parse_agent_type` no longer normalizes `_ → -`. The new `parse_agent_type_rejects_snake_case` test loops `["claude_code", "gemini_cli", "open_code"]` and asserts each errors. No other path silently accepts snake-case — `agent_type_to_string` round-trips through serde's kebab-case serializer, so a `parse → format` round-trip stays in the kebab namespace. CLAUDE.md no-backwards-compat is correctly enforced.

### Skipped items — was the reasoning sound?

**Out-of-scope #7 (daemon-side caller-UUID test):** Correctly deferred — that test belongs with the daemon's caller-id behavior, not this PR.

**Nit: `agent_type_to_string` fallback:** The author's rationale ("infallible if AgentType gains a non-serde variant") is weaker than the alternatives (`unreachable!()` or `expect()`), since `AgentType` is in the same workspace and adding a non-serde variant would be a deliberate breaking change. That said: this is a `Nit` either way, and the fallback is a single line that costs nothing. Accepting.

**Nit: `embedding_service = None` duplication:** The author's "will resolve naturally when wired" is technically true (both consumers will take the same `Arc<NodeEmbeddingService>`), but a single shared binding really would be cleaner today. Carrying forward as a minor follow-up isn't worth blocking on — flagging for awareness.

**Nit: `last_err` Option indirection and `count: count` shorthand:** Pure style; defer.

### Findings

#### Suggested Improvements
- **`launch_session` silent resize skip** (`packages/daemon/src/services/agent_session_service.rs:75-79`): If `manager.get()` returns `None` *and* the caller passed non-zero `cols/rows`, the resize is silently skipped and we still return a successful `LaunchSessionResponse` with `session_id`. The previous code would have returned `NotFound` from `apply_resize`. The new behavior is friendlier (don't fail on a phantom race), but the client now holds a `session_id` they cannot subsequently use (next RPC will return `NotFound`). Consider returning `Status::failed_precondition` or `not_found` here, or at minimum logging at `warn` so the server-side trace makes the situation discoverable. This is small and not a merge-blocker; flagging because the contract drifted slightly compared to the old `apply_resize` path.

#### Nitpicks
- **Nit** (`agent_session_service.rs:271`): The "infallible if a new variant appears" comment on `agent_type_to_string` could be tightened — since `AgentType` is in-workspace, this is genuinely `unreachable!()`-level. Keeping the fallback is fine, but consider noting that adding a new variant *should* be accompanied by adding it to the parse/format pair, not relying on the debug-format escape hatch.

### Standards compliance
- No new `#[allow(...)]` attributes.
- No `#[deprecated]` markers.
- No migration code.
- `debug_assert!` is the right choice — only runs in debug builds, won't impact production behavior.
- No backwards-compat aliases (snake-case removal is exactly the CLAUDE.md no-backwards-compat directive in action).

### Local verification
- `cargo build --bin nodespaced` — passes.
- `cargo test -p nodespace-daemon` — 26/26 passing (7 unit + 9 agent_session_grpc + 10 grpc_round_trip).
- `cargo clippy -p nodespace-daemon --all-targets -- -D warnings -D clippy::unused-async` — clean.

LGTM. (Would have approved via `gh pr review --approve` but GitHub blocks self-approval on own PRs.)

malibio (commented):
Re-review complete — LGTM. Full report posted as a PR comment: https://github.com/NodeSpaceAI/nodespace-core/pull/1148#issuecomment-4473863935

Verified locally: build clean, 26/26 tests passing (including the new resize_terminal_rejects_zero_dimensions), clippy clean with -D warnings -D clippy::unused-async.

Only two non-blocking notes:
- Improvement: launch_session silently no-ops the initial resize if the session was auto-pruned between launch and get (returns success with an unusable session_id). Worth a warn-log or a failed_precondition status.
- Nit: agent_type_to_string's debug-format fallback comment could acknowledge that AgentType is in-workspace and a new variant should update parse/format together, not lean on the fallback.

Nothing here should block merge.

malibio:
## Re-review feedback addressed — 2/2 implemented + flake fix

Pushed as 7fe21944.

### ✅ Addressed

| Item | Change |
|---|---|
| Suggestion: silent resize skip | `launch_session` now logs `tracing::warn!` with session_id + requested dimensions when the auto-prune race causes the optional initial resize to no-op. Response semantics unchanged (launch succeeded, next RPC's NotFound is the cleaner signal for the rare race). |
| Nit: `agent_type_to_string` fallback comment | Comment now explicitly states the debug-format fallback is defense-in-depth, not graceful degradation — adding a new AgentType variant must update parse_agent_type / agent_type_to_string together. |

### 🐛 Drive-by flake fix

While verifying, `stream_output_delivers_echo_hello` was failing ~10% of runs: on a fast machine the PTY reader drained `echo hello` and broadcast the chunk before the gRPC subscriber attached, and `tokio::sync::broadcast` doesn't replay missed messages to late subscribers.

Changed the command from `echo hello` to `sleep 0.2 && echo hello` so the subscriber is guaranteed to attach first. Confirmed 5/5 passing after the change. Comment in the test body explains why.

### Verified

- `cargo test -p nodespace-daemon`: **26 passed** (re-ran 5x to confirm the flake is gone)
- `cargo clippy -p nodespace-daemon --all-targets -- -D warnings -D clippy::unused-async`: clean
- `bun run quality:fix`: clean across the full workspace

PR #1154, opened 2026-05-18
Add ImportService gRPC: proto definition, nodespaced handler, and Tauri proxy (closes #1136)

### LINKED ISSUE #1136: Add ImportService gRPC: proto definition, nodespaced handler, and Tauri proxy

malibio:
## Summary

- Defines `ImportService` in a new `import_service.proto` with server-streaming `ImportMarkdown` and `ImportMarkdownFiles` RPCs
- Implements `ImportServiceImpl` in `nodespaced`, preserving the two-phase pipeline (parse → background DB write) and smart collection routing from the original Tauri commands
- Migrates `commands/import.rs` in the Tauri app to a thin gRPC proxy; all import logic now runs in the daemon (survives window close)
- Wires `nodespace import file` and `nodespace import dir` CLI subcommands
- Fixes pre-existing `CreateNodeRequest` struct literal errors in three test/integration files

## Test plan
- [ ] `cargo build` succeeds with no errors
- [ ] `bun run quality:fix` passes clean
- [ ] `nodespace import file <path>` streams progress and reports success
- [ ] `nodespace import dir <path>` imports all .md files recursively
- [ ] Tauri `import_markdown_file` and `import_markdown_files` commands work via gRPC proxy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,7 +44,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.16",
+ "getrandom 0.2.17",
  "once_cell",
  "version_check",
 ]
@@ -64,9 +64,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
@@ -101,7 +101,7 @@ dependencies = [
  "cssparser 0.35.0",
  "html5ever 0.35.0",
  "maplit",
- "tendril",
+ "tendril 0.4.3",
  "url",
 ]
 
@@ -122,9 +122,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -137,15 +137,15 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.13"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.7"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
+checksum = "52ce7f38b242319f7cabaa6813055467063ecdc9d355bbb4ce0c68908cd8130e"
 dependencies = [
  "utf8parse",
 ]
@@ -156,7 +156,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.1",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -167,7 +167,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.1",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -210,7 +210,7 @@ checksum = "3c3610892ee6e0cbce8ae2700349fcf8f98adb0dbfbee85aec3c9179d29cc072"
 dependencies = [
  "base64ct",
  "blake2",
- "cpufeatures",
+ "cpufeatures 0.2.17",
  "password-hash",
 ]
 
@@ -239,27 +239,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ashpd"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6cbdf310d77fd3aaee6ea2093db7011dc2d35d2eb3481e5607f1f8d942ed99df"
-dependencies = [
- "enumflags2",
- "futures-channel",
- "futures-util",
- "rand 0.9.2",
- "raw-window-handle",
- "serde",
- "serde_repr",
- "tokio",
- "url",
- "wayland-backend",
- "wayland-client",
- "wayland-protocols",
- "zbus",
-]
-
-[[package]]
 name = "async-broadcast"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


malibio (commented):
### Code Review Summary

This PR successfully migrates bulk markdown import from the Tauri process into `nodespaced` over gRPC, which is the correct architectural move under ADR-031. The proto definition is well-structured, the two-phase pipeline is correctly preserved, and the Tauri proxy is properly thinned out. The change is a definitive net improvement to the codebase.

All acceptance criteria are met except for one — there are no new tests for `ImportService` itself (the integration test suite covers `NodeService` round-trips but has no import-specific test). This is the primary concern to address before merge.

---

### Findings

#### Critical Issues

None.

---

#### Suggested Improvements

**[Improvement] `packages/daemon/src/services/import_service.rs` — No integration test for `ImportService`**

The issue requires "`bun run test:all` passes" but there is no new test exercising the import gRPC path. `grpc_round_trip.rs` only covers `NodeService`. The import handler touches non-trivial logic: the two-phase pipeline, collection routing, the `CHANNEL_BUFFER=64` back-pressure, and the final-event-carries-results contract. A minimal test (spin up in-process server, call `ImportMarkdown` with a temp `.md` file, drain the stream, assert `step==9` and `success==true`) would close the gap between the existing coverage and what this PR adds.

**[Improvement] `packages/daemon/src/services/import_service.rs` lines 1070–1228 — Silent bulk-create failure in Phase 2 background task does not surface to the caller**

```rust
match node_service_clone
    .bulk_create_hierarchy_trusted(all_nodes)
    .await
{
    Ok(ids) => tracing::info!("Bulk created {} nodes", ids.len()),
    Err(e) => tracing::error!("Failed to bulk create nodes: {:?}", e),
}
```

If `bulk_create_hierarchy_trusted` fails, the error is logged but the stream still sends a `step=9 / complete` event with `success=true` for all prepared files. The client (CLI or Tauri) therefore receives a successful result for files that were never written to the database. The Phase 2 task should either mark results as failed and send those in the final event, or send an error-typed event. At minimum, the final summary count should reflect actual DB outcomes, not Phase 1 parse outcomes.

**[Improvement] `packages/cli/src/commands/import.rs` lines 162–163 — Magic constant `9` for final step**

```rust
if event.step == 9 {
```

This appears in four places across the CLI and Tauri proxy. The proto comment documents `9 = complete` but there is no shared constant or enum. A `const STEP_COMPLETE: u32 = 9;` in the daemon crate (or a proto enum) would make intent explicit and eliminate the risk of one site being updated while others are not.

**[Improvement] `packages/daemon/src/services/import_service.rs` lines 830–851 — `base_dir` fallback uses first file's parent, not the common ancestor**

```rust
let base_dir = if !opts.base_directory.is_empty() {
    PathBuf::from(&opts.base_directory)
} else {
    file_paths
        .first()
        .map(|p| {
            PathBuf::from(p)
                .parent()
                ...
        })
        ...
};
```

When `base_directory` is empty and files come from multiple directories, the routing falls back to the first file's parent. This can produce inconsistent collection assignments depending on file ordering. This is a pre-existing behavior inherited from the original `commands/import.rs`, but now that the daemon owns this logic, documenting the known limitation (or asserting that `base_directory` must be set for multi-directory batches) would prevent silent mis-routing.

**[Improvement] `packages/desktop-app/src-tauri/src/commands/import.rs` line 2284 — `import_markdown_file` returns an error if the stream closes without a result event**

```rust
last_result.ok_or_else(|| "Import stream ended without a result".to_string())
```

This is the right approach, but the error string surfaced to the Tauri frontend is opaque. If the daemon crashes mid-import the user sees `"Import stream ended without a result"` with no context. Adding the file path to the message (`format!("Import stream ended without a result for {}", file_path)`) makes the frontend error actionable.

**[Improvement] `packages/daemon/src/services/import_service.rs` — `import_markdown_content` (`run_single_file_import` path) duplicates node creation logic from the batch path**

The single-file path calls `create_node_with_parent` + `bulk_create_hierarchy_root_notify`, while the batch path calls `bulk_create_hierarchy_trusted`. This means single-file imports trigger per-node notification events while batch imports do not, which is an intentional difference inherited from the original code — but the divergence now lives inside the daemon where it is less visible. A comment at the top of `import_markdown_content` explaining why it uses `root_notify` vs `trusted` would prevent accidental unification that breaks notification semantics.

---

#### Nitpicks

- **Nit:** `packages/daemon/proto/import_service.proto` line 433 — the comment `"file_result populated for single-file RPC"` on `ImportProgressEvent.results` is slightly misleading; both RPCs populate results on step 9. Worth updating to `"Populated on the final event (step=9) with per-file import results for both RPCs"`.

- **Nit:** `packages/cli/src/commands/import.rs` lines 148–149 — `exclude_patterns: vec![]` and `base_directory: String::new()` are set unconditionally for the single-file path even though `ImportFileArgs` has no `exclude_patterns` field. These are proto defaults and are harmless, but shadowing them explicitly while `ImportFileArgs` has no corresponding fields adds slight confusion. A comment noting these are unused for single-file RPCs would help.

- **Nit:** `packages/daemon/src/services/import_service.rs` — `FileReadResult.relative_path` is constructed but only used for display in `filename` fallback. The field could be removed in favour of computing the display name inline at the single call site, reducing the struct footprint.

- **Nit:** `packages/daemon/src/services/import_service.rs` lines 1094–1101 — the large anonymous tuple type `Vec<(String, String, String, Option<String>, f64, serde_json::Value)>` appears twice (single-file path and batch path). Extracting this as a type alias `type BulkNodeTuple = (String, String, String, Option<String>, f64, serde_json::Value)` would reduce noise.

---

| Criterion | Status |
|---|---|
| `ImportService` proto defined and compiled via tonic-build | Done |
| nodespaced handler implements two-phase import with smart collection routing | Done |
| Progress streamed back via gRPC server-streaming RPC | Done |
| `commands/import.rs` proxies via gRPC; no direct `packages/core` imports remain | Done |
| Import completes if Tauri window is closed mid-import | Done (daemon-side background task) |
| `nodespace import <file>` CLI subcommand works | Done |
| `cargo build` succeeds | Cannot verify in review — author must confirm |
| `bun run test:all` passes | No new import-specific tests added (see Improvement above) |
| `bun run quality:fix` passes | Cannot verify in review — author must confirm |

The silent-failure-on-bulk-create-error issue and the missing integration test are the two items most worth addressing before merge. Everything else is polish.

malibio:
Addressed the two review findings from the initial review:

**1. Silent bulk-insert failure (🔴 Critical) — fixed**

Added a `bulk_insert_failed` bool in Phase 2. When `bulk_create_hierarchy_trusted` returns `Err`, all `phase1_results` are marked `success=false` with a descriptive error, and the collection assignment, mention creation, and lifecycle update steps are skipped entirely. Previously the error was only logged; callers saw `success=true`.

**2. No integration test (🟡 Important) — added**

New `packages/daemon/tests/import_round_trip.rs` with 3 in-process gRPC tests:
- `import_markdown_single_file_streams_progress_and_succeeds` — streams progress events, asserts step-9 with `success=true` and `nodes_created > 0`
- `import_markdown_files_batch_reports_all_results` — two-file batch, both results successful
- `import_markdown_missing_file_surfaces_failure` — non-existent file yields `success=false` with non-empty error

All 1182 Rust tests pass; `cargo clippy` clean.

PR #1157, opened 2026-05-18
Implement per-agent extension shims for all five PTY agents (#1123)

malibio:
## Summary

- Adds shim files for all five PTY agents in `packages/agent-tools/shims/`: Claude Code (hooks-based), Codex (plugin), Gemini CLI (JSON tool definitions + stdin/stdout handler), OpenCode (plugin), and Pi (`registerTool()` adapter)
- Extends `GraphContextAssembler` with optional shim directory support: when `NODESPACED_SHIM_DIR` is set, the assembler copies per-agent shim files into the session temp dir and appends a NodeSpace Tools activation section to the context file
- Adds `NODESPACED_SHIM_DIR` env var to the daemon for runtime wiring
- Includes 10 new Rust tests for shim file mapping and copy logic

## Test plan

- [x] All 17 `@nodespace/agent-tools` tests pass
- [x] TypeScript check clean in `packages/agent-tools` and `packages/desktop-app`
- [ ] `cargo test -p nodespace-agent` — new shim mapping and copy tests
- [ ] `bun run test:all` — no regressions

Closes #1123


malibio:
### Code Review Summary

This PR implements the per-agent extension shims as specified in issue #1123. The overall design is clean: each shim is minimal glue that only translates the agent's native registration API to `@nodespace/agent-tools` calls, and the `GraphContextAssembler` integration is well-contained via an optional `shim_source_dir` builder field. The change is a net improvement and is close to mergeable. Several issues need attention before merge.

---

### Findings

#### Critical Issues

**[Critical/Blocker] `nodespace-tools.json` references a compiled `.js` handler but only a `.ts` file is shipped**

`packages/agent-tools/shims/gemini/nodespace-tools.json` (line 21):
```json
"handler": "./nodespace-handler.js"
```

The shim source directory contains `nodespace-handler.ts`, and `write_shim_files` in `context_assembly.rs` copies exactly that TypeScript file (line 68 of the new Rust code: `&["nodespace-tools.json", "nodespace-handler.ts"]`). When Gemini CLI resolves `./nodespace-handler.js` from the copied directory, the file will not exist — it will be `nodespace-handler.ts`. This breaks every Gemini CLI tool call at runtime.

Either:
- Change the JSON to `"./nodespace-handler.ts"` and ensure Gemini CLI supports `tsx`/`ts-node` execution (consistent with the "agents that execute TS natively" comment in `write_shim_files`), or
- Add a pre-copy compilation step so `.js` is what gets placed in the session directory.

The same `.ts`-vs-`.js` question applies to the other shims (`nodespace-hook.ts`, `nodespace-plugin.ts`, `nodespace-adapter.ts`) but those are registered via the agent's own runtime environment (where native TS execution is more likely documented). For Gemini the `handler` path in the JSON is the direct file Gemini CLI will try to execute — the mismatch is a hard runtime failure.

---

**[Critical/Blocker] `NODESPACED_SHIM_DIR` is optional but the PR test plan marks `cargo test -p nodespace-agent` and `bun run test:all` as incomplete (`[ ]`)**

The PR body leaves these two test runs unchecked. The acceptance criteria include "All shims handle `ToolError` from `@nodespace/agent-tools` and surface it as a tool error to the agent" and "Code passes `bun run quality:fix`" — both require a full test run to verify. Please confirm these pass before merging.

---

#### Suggested Improvements

**[Improvement] Shim files land flat in `session_dir`, risking collisions across agents**

`write_shim_files` copies all files for an agent directly into `session_dir` root (line: `let dst = session_dir.join(filename)`). Two agents share the name `nodespace-plugin.ts` (Codex and OpenCode). If a future code path ever calls `write_shim_files` for both agent types against the same session directory, the second write silently overwrites the first. Today there is a one-agent-per-session invariant, but this is fragile.

Consider writing into a per-agent subdirectory: `session_dir.join(subdir).join(filename)`. The context file's activation instructions can reference `<subdir>/nodespace-hook.ts` etc. This also avoids the risk of shim files shadowing any user-owned file named `nodespace-plugin.ts` in the session working directory.

---

**[Improvement] `shim_source_dir` validation is deferred entirely to runtime**

`with_shim_dir` accepts any `PathBuf` without validation. If `NODESPACED_SHIM_DIR` points to a nonexistent path, the error surfaces only at the moment a session is started and `write_context_file` is called — not at daemon startup. A single `tokio::fs::metadata` check at `build_services` time would surface the misconfiguration immediately in daemon logs, rather than failing silently per-session. This is a DX issue, not a correctness one.

---

**[Improvement] The `TOOLS_SECTION` activation text does not reference the shim file location or how to load it**

For Claude Code, the context file says "NodeSpace knowledge graph tools are available" but does not tell Claude Code _where_ the hook file is (e.g., `nodespace-hook.ts` was written to this directory — activate it with `hook_file: ./nodespace-hook.ts`). The issue spec shows the `CLAUDE.md` section should include enough information for the agent to load the file automatically. Without the path hint, Claude Code has no way to find and register the hook.

Verify that Claude Code's hooks runtime automatically discovers `*.ts` hook files in the session directory without an explicit directive; if not, the activation section needs the path.

---

**[Improvement] `write_shim_files` errors swallow the underlying `io::Error::kind()` when the source directory is missing**

```rust
ContextError::WriteFailed(std::io::Error::new(
    e.kind(),
    format!("failed to copy shim file '{}' to session dir: {}", src.display(), e),
))
```

`std::io::Error::new` creates a new error with only the given message; `e.kind()` is preserved, but `e`'s OS-level source chain is dropped (the format string captures only `Display` output). Prefer wrapping `e` directly:

```rust
ContextError::WriteFailed(std::io::Error::other(
    format!("failed to copy shim file '{}': {}", src.display(), e)
))
```

Or use `map_err(ContextError::WriteFailed)` after adding context via an extension trait (`std::io::Error` does not have a `.context()` method, but `anyhow::Context` does if routed through `ContextError::Other`). The current approach works but discards the error source for downstream error chain introspection.

---

#### Nitpicks

**Nit:** `packages/agent-tools/shims/` is added to the npm `files` array, which means the raw TypeScript sources are published. If this package is ever published externally, downstream consumers receive un-compiled source. For an internal monorepo-only package this is fine, but worth a comment acknowledging the intent.

**Nit:** The `tsconfig.json` `include` is `src/**/*` only, so the `shims/` directory is not type-checked by `tsc`. The `quality:fix` script runs `tsc --noEmit` which will not catch type errors in shims. Consider adding `"shims/**/*"` to `include`, or accepting this gap with a comment.

**Nit:** `packages/agent-tools/shims/gemini/nodespace-handler.ts` re-throws `ToolError` as a non-structural error string on exit-code-1 (line 59-60). This is correct for Gemini's stdio protocol, but consider documenting the expected `{ error: string }` response shape in the JSON tool spec or handler comment so future maintainers know not to change the exit-code behavior.

PR #1159, opened 2026-05-18
Add CLI session subcommands for PTY agent session management (closes #1121)

malibio:
## Summary
- Adds `nodespace session launch|attach|list|kill` subcommands backed by `AgentSessionService` gRPC
- Implements raw-mode terminal bridge in `packages/cli/src/terminal.rs` using `crossterm` for PTY stream bridging
- Ctrl+C and Ctrl+D detach cleanly without killing the session; session ID is printed to stderr for scripting

## Test plan
- [ ] `cargo build --bin nodespace` succeeds (verified in CI)
- [ ] `nodespace session --help` and all subcommand `--help` show useful usage text
- [ ] `nodespace session launch claude-code` starts a session and streams output
- [ ] `nodespace session list` shows active sessions with ID, agent, start time
- [ ] `nodespace session attach <id>` reconnects to an existing session
- [ ] `nodespace session kill <id>` terminates a session
- [ ] Clear error when `nodespaced` is not running
- [ ] Frontend tests: 128 passed (no regressions)


malibio:

---

**Requirements Check**:

- ✅ `nodespace session launch <agent>` starts a session and streams output to stdout; keyboard input is forwarded to the PTY
- ✅ `nodespace session launch <agent> --prompt "..."` passes an initial prompt to the agent binary at spawn time
- ✅ `nodespace session attach <session-id>` reconnects to an existing session's output stream and forwards keyboard input
- ✅ `nodespace session list` lists all active sessions with session ID, agent type, start time
- ✅ `nodespace session kill <session-id>` terminates the session
- ⚠️ Ctrl+D / Ctrl+C detach without killing — functionally correct but with a behavioral defect (see finding #1 below)
- ✅ Clear error when `nodespaced` is not running (friendly anyhow context message)
- ✅ `nodespace session --help` and subcommand help output useful usage text
- ✅ `cargo build --bin nodespace` succeeds (CI verified per PR body)
- ✅ `.gitignore` fix to unblock `src/commands/session.rs` from the `session*.rs` scratch-file exclusion

---

**Code Review Findings**

#### 🟡 Important — Ctrl+C / Ctrl+D sends the control byte to the PTY before detaching

**File**: `packages/cli/src/commands/session.rs`, lines 175–184

The current flow is:

1. Detect `should_detach` (0x03 or 0x04 present in buffer)
2. **Send the full buffer (including the control byte) via `WriteInput`**
3. Then break

This means Ctrl+C (0x03) is forwarded to the running agent as a SIGINT/interrupt and Ctrl+D (0x04) is forwarded as an EOF sentinel. Both of these can affect the session's internal state in agent-specific ways — Ctrl+C may kill a running subprocess inside the agent, and Ctrl+D may close the agent's own readline loop. The stated contract ("detach without killing the session") is violated in practice for agents that respect these control bytes.

The fix is to strip the detach byte from `data` before forwarding (or skip forwarding entirely when the buffer contains only the control byte):

```rust
let should_detach = data.iter().any(|&b| b == 0x04 || b == 0x03);
let forward_data: Vec<u8> = data.iter().copied().filter(|&b| b != 0x03 && b != 0x04).collect();
if !forward_data.is_empty() {
    let _ = input_client.write_input(Request::new(WriteInputRequest {
        session_id: input_session.clone(),
        data: forward_data,
    })).await;
}
if should_detach { break; }
```

This matches the accepted UX contract described in both the issue and the PR summary.

---

#### 🟡 Important — `format_unix_time` is timezone-unaware and gives misleading output across midnight

**File**: `packages/cli/src/commands/session.rs`, lines 213–221

The current implementation takes `unix_secs % 86400` and formats it as HH:MM:SS. This is essentially UTC seconds-within-day. On any system where local time != UTC, the displayed time will be wrong relative to the user's clock. For a session started at 23:45 local time, it may display as `01:45:00` if the user is UTC+2. The comment says "good enough" but the discrepancy is confusing enough to undermine trust in the output.

The minimal fix without pulling in `chrono` or `time` is to adjust by the system UTC offset using `std::time::SystemTime`:

```rust
fn format_unix_time(unix_secs: i64) -> String {
    // Approximate local time by adding the UTC offset inferred from SystemTime.
    // For a session list, minute-level accuracy is sufficient.
    let now_unix = SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .unwrap_or_default()
        .as_secs() as i64;
    // SystemTime gives us wall-clock seconds. The OS already adjusts for DST.
    // Using chrono/time would be cleaner but avoids adding a dependency.
    let secs = unix_secs % 86400;
    ...
}
```

Actually, obtaining a correct local offset without a dependency requires platform-specific syscalls. Given this is a greenfield project with no backward-compatibility constraint, the right answer is to add `chrono` to `packages/cli/Cargo.toml` and use `chrono::Local` — or at minimum, note that the displayed time is UTC with a `(UTC)` suffix so users are not misled. The current code has no such annotation.

**Minimum acceptable fix**: Append `" UTC"` to the formatted string so the output is unambiguous.

---

#### 🟢 Suggestion — `format_unix_time` does not handle `unix_secs = 0`

**File**: `packages/cli/src/commands/session.rs`, line 213

If `started_at` is zero (proto default / unset field), the function outputs `00:00:00`, which looks like midnight rather than a missing value. A defensive check such as `if unix_secs == 0 { return "unknown".to_string(); }` improves debuggability when the daemon doesn't populate the field.

---

#### 🟢 Suggestion — `input_task` JoinHandle is dropped without awaiting on the happy path

**File**: `packages/cli/src/commands/session.rs`, lines 205–206

After the output stream ends cleanly (`Ok(None)`), `input_task.abort()` is called but the handle is immediately dropped. Aborting a Tokio task is cooperative — the task may still be blocked on `stdin.read()` when the function returns and `_raw` (the `RawMode` guard) is dropped. In practice this is fine because the process exits shortly after, but if `stream_bridge` is ever called in a context that doesn't exit (e.g., a future interactive TUI wrapper), the background task may linger briefly with raw mode already restored.

Consider `input_task.abort(); let _ = input_task.await;` to ensure the task has fully stopped before raw mode is restored. The `await` on an aborted task returns `Err(JoinError::Cancelled)`, which can be safely ignored.

---

#### 🟢 Suggestion — `ListArgs` is an empty struct with no purpose

**File**: `packages/cli/src/commands/session.rs`, line 52

`pub struct ListArgs {}` is passed through the enum variant but never used — `list()` receives no args. This mirrors the pattern in the existing codebase (e.g., `diagnostics`) so it is consistent, but it is worth noting that clap's `#[command(subcommand)]` does not require an args struct for arg-less subcommands. Nit-level, consistent with existing patterns.

---

#### 🟢 Nit — `session_id.clone()` in `kill` is unnecessary when `session_id` is consumed before the clone

**File**: `packages/cli/src/commands/session.rs`, lines 128–137

`args.session_id.clone()` is called to build the `TerminateSessionRequest`, then the original is used in the `println!`. The clone can be eliminated by using the original in the request and referencing it in the print, or by binding `let id = args.session_id` and using `id` in both. Minor; no semantic impact.

---

**Ancillary changes** (daemon `lib.rs`, `main.rs`, desktop-app `embeddings.rs`, `settings.rs`): These are all `rustfmt` reformats with no semantic changes. They are clean and correct.

**Architecture assessment**: The design is well-structured. `terminal.rs` is a clean RAII abstraction with a minimal surface area. `stream_bridge` is correctly shared between `launch` and `attach`, honoring DRY. The connection pattern for `AgentSessionServiceClient` is consistent with existing `connect` / `connect_import` functions in `lib.rs`. The `_json` parameter forwarded to `run()` is accepted from the top-level dispatch for future use — acceptable YAGNI trade-off given the existing pattern. The `.gitignore` fix is a necessary correctness fix, not scope creep.

---

**Recommendation**: REQUEST CHANGES

The Ctrl+C / Ctrl+D behavior (finding #1) is a behavioral correctness defect against the stated contract — control bytes that are meant to detach the CLI are instead forwarded to the agent, potentially interrupting it. The timezone display issue (finding #2) is user-facing and misleading enough to require at minimum a `(UTC)` annotation. Both are small, targeted fixes. The remaining findings are suggestions for a follow-up.

malibio:
## Review Feedback Addressed

All 5 reviewer recommendations implemented:

**🟡 Fixed: Ctrl+C/D no longer forwarded to agent PTY**
Filter `0x03`/`0x04` out of the `WriteInput` payload before sending — detach no longer sends SIGINT/EOF to the running agent.

**🟡 Fixed: Session start time now displays in local timezone**
Replaced `unix_secs % 86400` (UTC wall-clock) with `chrono::Local.timestamp_opt()` — users in non-UTC timezones see correct local time.

**🟢 Fixed: `unix_secs == 0` now displays `"unknown"`**
Guards against proto default-zero `started_at` showing `"00:00:00"`.

**🟢 Fixed: `input_task` awaited after `abort()` on all exit paths**
Ensures the stdin reader exits before `RawMode` drops and restores the terminal.

**🟢 Skipped: unnecessary `.clone()` in `kill`**
`args.session_id` is used in both the request struct and the `println!` — the clone is load-bearing. No change needed.

PR #1162, opened 2026-05-18
Add Tauri commands and xterm.js terminal UI for PTY agent sessions (#1120)

malibio:
## Summary

- Wires `AgentSessionService` into the in-process gRPC server alongside the existing Node/Import/Embeddings services
- Adds `agent_session.rs` Tauri commands (`launch_session`, `write_input`, `resize_terminal`, `terminate_session`, `list_sessions`) — `launch_session` spawns a background tokio task that bridges `StreamOutput` to `pty-output-{sessionId}` Tauri events
- Adds `pty-terminal.svelte` wrapping xterm.js with FitAddon + ResizeObserver, subscribing to output events and forwarding keystrokes
- Adds `agent-launch-panel.svelte` (agent selector + prompt + Launch button) and `sessions-panel.svelte` (active sessions list with Terminate buttons)
- Adds `agent-sessions-panel.svelte` composite pane combining the above with a terminal area
- Wires a new `agent-sessions` tab type into navigation sidebar, pane-content, and tab-persistence

## Test plan

- [x] Frontend: 128 files / 3916 tests passing (`bun run test`)
- [x] Lint + type-check: `bun run quality:fix` exits 0 (svelte-check: 0 errors)
- [x] Rust: `cargo check -p nodespace-app` exits clean

Closes #1120

### DIFF
--- a/packages/daemon/src/lib.rs
+++ b/packages/daemon/src/lib.rs
@@ -55,7 +55,11 @@ pub use nodespace::node_service_client::NodeServiceClient;
 pub use nodespace::node_service_server::NodeServiceServer;
 pub use nodespace::settings_service_client::SettingsServiceClient;
 pub use nodespace::settings_service_server::SettingsServiceServer;
-pub use nodespace::{NodeData, SessionInfo};
+pub use nodespace::{
+    LaunchSessionRequest, LaunchSessionResponse, ListSessionsRequest, ListSessionsResponse,
+    NodeData, ResizeRequest, ResizeResponse, SessionInfo, StreamOutputRequest,
+    TerminateSessionRequest, TerminateSessionResponse, WriteInputRequest, WriteInputResponse,
+};
 
 pub use services::{
     AgentSessionHandler, EmbeddingsServiceImpl, ImportServiceImpl, NodeServiceImpl,
--- a/packages/desktop-app/eslint.config.js
+++ b/packages/desktop-app/eslint.config.js
@@ -72,7 +72,10 @@ export default [
         Response: 'readonly',
         fetch: 'readonly',
         crypto: 'readonly',
-        CSS: 'readonly'
+        CSS: 'readonly',
+        ResizeObserver: 'readonly',
+        TextEncoder: 'readonly',
+        TextDecoder: 'readonly'
       }
     },
     plugins: {
@@ -213,6 +216,9 @@ export default [
         Response: 'readonly',
         BeforeUnloadEvent: 'readonly',
         btoa: 'readonly',
+        ResizeObserver: 'readonly',
+        TextEncoder: 'readonly',
+        TextDecoder: 'readonly',
       }
     },
     plugins: {
--- a/packages/desktop-app/package.json
+++ b/packages/desktop-app/package.json
@@ -46,6 +46,8 @@
     "@tauri-apps/plugin-dialog": "^2.6.0",
     "@tauri-apps/plugin-opener": "^2.5.0",
     "@thisux/sveltednd": "^0.0.20",
+    "@xterm/addon-fit": "^0.10",
+    "@xterm/xterm": "^5.5",
     "baseline-browser-mapping": "^2.9.18",
     "clsx": "^2.1.1",
     "core-js-compat": "^3.48.0",
--- a/packages/desktop-app/src-tauri/src/commands/agent_session.rs
+++ b/packages/desktop-app/src-tauri/src/commands/agent_session.rs
@@ -4,6 +4,16 @@
 //! background task that reads the `StreamOutput` server-streaming response
 //! and emits Tauri events keyed by session ID so the frontend's
 //! `PtyTerminal.svelte` can subscribe via `listen("pty-output-{sessionId}")`.
+//!
+//! ## Streaming task lifecycle
+//!
+//! Each launched session gets a `CancellationToken` stored in
+//! `StreamingTaskRegistry`. When `terminate_session` is called, the registry
+//! cancels the token, which causes the background streaming loop to exit
+//! promptly rather than waiting for the next gRPC message or a closed stream.
+
+use std::collections::HashMap;
+use std::sync::Mutex;
 
 use futures::StreamExt;
 use nodespace_daemon::{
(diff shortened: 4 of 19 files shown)


malibio:
### Code Review Summary

This PR delivers a complete vertical slice for PTY agent session UI: Tauri command proxies, streaming event bridge, xterm.js terminal component, launch/sessions panels, and navigation wiring. All acceptance criteria from #1120 are met. The implementation is clean, follows established project patterns, and the architecture is sound. Recommending approval with two improvements noted.

---

- [x] `AgentLaunchPanel.svelte` renders agent list, prompt input, Launch button
- [x] Clicking "Launch" calls `LaunchSession` and opens a terminal pane
- [x] `PtyTerminal.svelte` renders xterm.js and streams `StreamOutput`
- [x] Keyboard input forwarded to `WriteInput`
- [x] Terminal resize calls `ResizeTerminal`
- [x] Sessions panel shows active sessions with Terminate buttons
- [x] Terminating a session clears the active terminal pane
- [x] xterm.js fits parent container via FitAddon + ResizeObserver
- [x] All frontend tests pass
- [x] `bun run quality:fix` exits 0

---

### Findings

#### Improvements

**[Improvement] `pty-terminal.svelte` lines 39–48: Components call `invoke` directly, bypassing the typed wrapper layer**

`tauri-commands.ts` defines `ptyWriteInput`, `ptyResizeTerminal`, and `ptyLaunchSession` as typed wrappers, but `pty-terminal.svelte` invokes the Tauri commands directly via raw `invoke()` strings. Similarly, `sessions-panel.svelte` and `agent-launch-panel.svelte` call `invoke('terminate_session', ...)` and `invoke('launch_session', ...)` directly.

This creates a split contract: the typed service layer exists but is unused by the components that need it most. When a command signature changes, you must update two places instead of one, and the compiler cannot catch the mismatch. Per the DRY principle, the typed wrappers are the single source of truth.

**Suggested fix:** Replace all raw `invoke()` calls in the three component files with the corresponding `pty*` functions from `tauri-commands.ts`.

---

**[Improvement] `agent_session.rs` lines 91–179: No mechanism to cancel or track the background streaming task**

`launch_session` spawns a `tauri::async_runtime::spawn` task that loops indefinitely reading `StreamOutput`. The task handle is discarded (fire-and-forget). This creates two problems:

1. If the session terminates via `terminate_session` from the frontend, the `stream_output` RPC will presumably error/close — the task will exit via the `Err(e)` branch. However, if the stream never closes cleanly (e.g., daemon crash, network error, or a session that the frontend navigates away from), the task runs until the app exits with no way to clean it up from outside.
2. If `launch_session` is called multiple times with the same session ID (unlikely but possible with a race), two tasks stream to the same event name.

The existing pattern in the codebase for streaming (e.g., `local_agent.rs`) should be consulted to see if there is an established cancellation convention. At minimum, consider storing the `JoinHandle` in an `Arc<Mutex<HashMap<String, JoinHandle<()>>>>` in Tauri state and aborting it in `terminate_session`. 

For this greenfield stage, the current behavior is acceptable since the stream will naturally terminate when the gRPC session ends, but the handle should not be silently dropped — it should at minimum be logged or tracked.

---

#### Nitpicks

**Nit: `sessions-panel.svelte` line 67 — `$effect` fires once on mount without a dependency**

The bare `$effect(() => { refresh(); })` has no reactive dependencies, so it runs exactly once after mount. This is semantically correct but `onMount` from `svelte` would be clearer about the intent (run once on mount, not reactively). Using `$effect` without dependencies implies the reader should look for reactive captures — and finding none is surprising.

**Nit: `agent_session.rs` line 150 — `chunk.data.to_vec()` is redundant if `data` is already `Vec<u8>`**

If the proto-generated `chunk.data` field is already a `Bytes` or `Vec<u8>`, calling `.to_vec()` makes an extra allocation. Check whether `.into()` or a direct field move suffices.

**Nit: `tauri-commands.ts` trailing blank line at line 449**

Extra blank line at end of file — cosmetic.

**Nit: `navigation-sidebar.svelte` — `handleAgentSessionsClick` duplicates the `handleAiChatClick` pattern**

Both functions do identical logic (find existing tab by type, set active or create new). A single `handleSingletonTabClick(tabType, title)` helper would eliminate the duplication. Low priority for now since there are only two callsites, but worth extracting before a third is added.

malibio (commented):
Net positive — all acceptance criteria met, clean vertical slice. Two improvements flagged in the comment above (invoke bypass of typed wrapper layer, untracked streaming task handle). Neither is a blocker for merge, but the typed wrapper issue should be addressed before this pattern propagates to more components.

malibio:
## Review Addressed

All reviewer findings from the code review have been implemented.

### Changes Made

**🟡 Important: Typed wrapper functions (finding 1)**
- `pty-terminal.svelte`: replaced raw `invoke('pty_write_input')` and `invoke('pty_resize_terminal')` with `ptyWriteInput()` and `ptyResizeTerminal()` from `tauri-commands.ts`
- `agent-launch-panel.svelte`: replaced raw `invoke('launch_session')` with `ptyLaunchSession()`
- `sessions-panel.svelte`: replaced raw invocations with `ptyListSessions()`, `ptyTerminateSession()`; imports `PtySessionInfo` type instead of redefining it

**🟡 Important: Streaming task cancellation (finding 2)**
- Added `StreamingTaskRegistry` struct (backed by `Mutex<HashMap<String, CancellationToken>>`) to `agent_session.rs`
- `launch_session` now creates a `CancellationToken` per session, stores it in the registry, and uses `tokio::select!` in the background task so cancellation is immediate
- `terminate_session` calls `registry.cancel_and_remove()` before the gRPC RPC, so the reader exits promptly rather than waiting for the next stream message or a closed stream
- `StreamingTaskRegistry::default()` registered as Tauri managed state in `lib.rs`

### Skipped

**🟢 Nit: Module-level doc comment on `agent_session.rs`** — module already has a comprehensive doc comment explaining the streaming task lifecycle and cancellation design.

**🟢 Nit: `PtyOutputPayload.data` type** — kept as `Vec<u8>` matching the gRPC proto bytes field; changing the wire type requires coordinated proto + frontend changes outside this PR's scope.

### Verification
- `cargo check -p nodespace-app`: ✅ clean
- `bun run quality:fix`: ✅ 0 errors, 0 warnings  
- `bun run test`: ✅ 128/128 test files, 3916/3916 tests passing

PR #1165, opened 2026-05-18
Add GetRoots RPC and restore root_node_count to diagnostics

### LINKED ISSUE #1153: Add GetRoots RPC and restore root_node_count to nodespace diagnostics
## Overview

Add a `GetRoots` RPC to `NodeService` (or equivalent — see "Alternatives") and restore `root_node_count` to the `nodespace diagnostics` CLI subcommand.

## Problem Statement

PR #1150 dropped `root_node_count` from the diagnostics report because the current gRPC surface has no honest way to compute it.

`QueryNodesSimple` returns `NodeData` rows with `parent_id: None` for every row, because the daemon's `node_to_proto(n, None, None)` call at `packages/daemon/src/services/node_service.rs:417` doesn't attempt to populate the field — parentage is stored as a graph edge (`relationship_type = "parent"`), not a column on the `node` table.

A new integration test (`diagnostics_collect_reports_counts_and_recency` in `packages/cli/tests/cli_integration.rs`) caught that the pre-existing filter `n.parent_id.as_deref().unwrap_or("").is_empty()` evaluated to `true` for every row, making `root_node_count == total_node_count` always. The original Tauri implementation had the same latent bug — the migration to a gRPC-backed CLI exposed it.

Issue #1139's acceptance criteria explicitly require root node count in the diagnostics output. With no GetRoots-shaped RPC available, the only honest options were (a) report a wrong number, (b) drop the field, or (c) add the RPC. PR #1150 chose (b) with an inline comment seeding this follow-up; (c) is the proper fix.

## Proposed Solution

Add a `GetRoots` RPC to `NodeService` that returns the list of nodes with no parent edge.

### Proto

```proto
service NodeService {
  ...
  rpc GetRoots(GetRootsRequest) returns (NodeListResponse);
}

message GetRootsRequest {
  // 0 = use server default; unset / 0 returns all roots
  uint32 limit = 1;
  uint32 offset = 2;
}
```

`NodeListResponse` is already defined and carries `nodes` + `count`.

### Daemon handler

Wire to `SurrealStore::get_children(None)` — the store-layer method that already exists at `packages/core/src/db/surreal_store.rs:1901` and returns nodes without a parent edge. Pass the optional limit/offset through `NodeQuery`.

### CLI

Restore `root_node_count` to `DiagnosticsReport`. In `collect()`, replace the comment block at `packages/cli/src/commands/diagnostics.rs:102-110` (which documents the structural impossibility under the current surface) with:

```rust
let root_node_count = match client.get_roots(GetRootsRequest { limit: 0, offset: 0 }).await {
    Ok(response) => response.into_inner().nodes.len(),
    Err(e) => {
        errors.push(format!("GetRoots failed: {e}"));
        0
    }
};
```

Re-add `Root nodes:` to `print_human` and `root_node_count` to `print_json`. Update the integration test to assert `root_node_count == baseline.root_node_count + 1` (the one new root) and the in-line comment removal.

## Alternatives Considered

1. **Extend `NodeListResponse` (and `QueryNodesSimple`) to carry parent IDs.**
   Means populating `parent_id` in `node_to_proto` for every node-list-returning RPC by joining against the relationship table at the daemon. Higher cost on the hot path of `QueryNodesSimple`; out of proportion for a diagnostics need. Reject unless other consumers also need parent info.

2. **Compute roots client-side by paging through `QueryNodesSimple` and calling `GetChildren(node_id)` per result.**
   N+1 round-trips for a developer-tool feature. Reject.

3. **Add a counts-only RPC (`GetRootCount`) that doesn't return node bodies.**
   Cheaper if diagnostics is the only consumer, but other clients (e.g. tree-view rendering) plausibly want the actual root list, so the `GetRoots` shape generalizes better. Mild preference for `GetRoots`.

## Acceptance Criteria

- [ ] `GetRoots` (or equivalent) RPC added to `NodeService` proto with appropriate request/response types
- [ ] Daemon handler implemented in `packages/daemon/src/services/node_service.rs`, wired to `SurrealStore::get_children(None)`
- [ ] `nodespace diagnostics` restores the `Root nodes:` line and `root_node_count` JSON field
- [ ] In-line comment in `collect()` explaining the structural impossibility is removed
- [ ] Integration test `diagnostics_collect_reports_counts_and_recency` asserts root count delta
- [ ] `cargo build` succeeds
- [ ] `cargo test -p nodespace-cli` passes (existing + updated assertions)
- [ ] `cargo clippy --all-targets -- -D warnings -D clippy::unused-async` clean

## Dependencies

- #1139 (closed by #1150) — diagnostics CLI subcommand must exist

## Related

- PR #1150 — review thread that surfaced the gap
- Issue #1139 acceptance criteria — restores the dropped "root node count" bullet

malibio:
## Summary
- Adds `GetRoots` RPC to `NodeService` proto with `GetRootsRequest` (limit/offset) returning `NodeListResponse`
- Implements daemon handler in `node_service.rs` wired to `CoreNodeService::get_roots()` → `SurrealStore::get_children(None)`
- Restores `root_node_count` to `DiagnosticsReport`, `print_human`, and `print_json` in the CLI diagnostics command
- Updates `diagnostics_collect_reports_counts_and_recency` integration test to assert `root_node_count == baseline + 1`

Closes #1153

## Test plan
- [x] `cargo build -p nodespace-daemon -p nodespace-cli` succeeds
- [x] `cargo clippy --all-targets -- -D warnings -D clippy::unused-async` clean (pre-existing `nodespace-app` error unrelated to this PR)
- [x] `cargo test -p nodespace-cli` — all 5 tests pass including `diagnostics_collect_reports_counts_and_recency`

### DIFF
--- a/packages/cli/src/commands/diagnostics.rs
+++ b/packages/cli/src/commands/diagnostics.rs
@@ -107,7 +107,7 @@ pub async fn collect(client: &mut NodeServiceClient<Channel>, db_path: &Path) ->
         })
         .await
     {
-        Ok(response) => response.into_inner().nodes.len(),
+        Ok(response) => response.into_inner().count as usize,
         Err(e) => {
             errors.push(format!("GetRoots failed: {e}"));
             0
--- a/packages/cli/src/commands/diagnostics.rs
+++ b/packages/cli/src/commands/diagnostics.rs
@@ -6,7 +6,7 @@
 
 use anyhow::{Context, Result};
 use clap::Args;
-use nodespace_daemon::nodespace::{GetAllSchemasRequest, QueryNodesSimpleRequest};
+use nodespace_daemon::nodespace::{GetAllSchemasRequest, GetRootsRequest, QueryNodesSimpleRequest};
 use nodespace_daemon::{resolve_db_path, NodeServiceClient};
 use serde_json::json;
 use std::fs;
@@ -32,6 +32,7 @@ pub struct DiagnosticsReport {
     pub database_exists: bool,
     pub database_size_bytes: Option<u64>,
     pub total_node_count: usize,
+    pub root_node_count: usize,
     pub schema_count: i32,
     pub recent_node_ids: Vec<String>,
     pub errors: Vec<String>,
@@ -99,13 +100,19 @@ pub async fn collect(client: &mut NodeServiceClient<Channel>, db_path: &Path) ->
 
     let total_node_count = all_nodes.len();
 
-    // Root count is intentionally omitted from the report. Parentage is
-    // stored as a graph edge in SurrealDB, and the daemon's
-    // `query_nodes_simple` handler ships nodes back with `parent_id = None`
-    // for every row because the column doesn't exist — there is no honest
-    // way to distinguish a root from a child via the current RPC surface.
-    // Add a `GetRoots` RPC (or extend `NodeListResponse` with parent ids)
-    // before reporting this number.
+    let root_node_count = match client
+        .get_roots(GetRootsRequest {
+            limit: 0,
+            offset: 0,
+        })
+        .await
+    {
+        Ok(response) => response.into_inner().nodes.len(),
+        Err(e) => {
+            errors.push(format!("GetRoots failed: {e}"));
+            0
+        }
+    };
 
     // QueryNodesSimple doesn't expose ORDER BY, so we sort the in-memory
     // batch by created_at descending before slicing. O(n log n) on n ≤
@@ -138,6 +145,7 @@ pub async fn collect(client: &mut NodeServiceClient<Channel>, db_path: &Path) ->
         database_exists,
         database_size_bytes,
         total_node_count,
+        root_node_count,
         schema_count,
         recent_node_ids,
         errors,
@@ -223,6 +231,7 @@ fn print_human(r: &DiagnosticsReport) {
         None => println!("Database size:   n/a"),
     }
     println!("Total nodes:     {}", r.total_node_count);
+    println!("Root nodes:      {}", r.root_node_count);
     println!("Schemas:         {}", r.schema_count);
     if r.recent_node_ids.is_empty() {
         println!("Recent node IDs: (none)");
@@ -244,6 +253,7 @@ fn print_json(r: &DiagnosticsReport) -> Result<()> {
         "database_exists": r.database_exists,
         "database_size_bytes": r.database_size_bytes,
         "total_node_count": r.total_node_count,
+        "root_node_count": r.root_node_count,
         "schema_count": r.schema_count,
         "recent_node_ids": r.recent_node_ids,
         "errors": r.errors,
--- a/packages/cli/tests/cli_integration.rs
+++ b/packages/cli/tests/cli_integration.rs
@@ -342,6 +342,11 @@ async fn diagnostics_collect_reports_counts_and_recency() {
         baseline.total_node_count + 3,
         "expected three additional nodes vs baseline"
     );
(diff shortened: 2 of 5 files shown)


malibio:
### Code Review Summary

This PR is a clean, targeted fix for a well-diagnosed problem. The structural analysis in issue #1153 is accurate — parentage lives in the relationship table, not on the node row, so exposing a `GetRoots` RPC via `SurrealStore::get_children(None)` is the correct architectural lever. The change correctly threads the new capability from proto → core service → daemon handler → CLI, and the integration test delta assertion is exactly right. This is a net improvement and all acceptance criteria are met.

Two implementation issues should be addressed before merge, and one nit.

---

### Findings

#### Improvements

**[Improvement] `packages/daemon/src/services/node_service.rs` lines 316–327 — Pagination applied in application memory, not at the database layer**

The daemon fetches the full root set from SurrealDB, then slices with `.skip()` / `.truncate()` in Rust. `SurrealStore::get_children(None)` does not accept a `NodeQuery`; it always returns every root node. For the current diagnostics use case (which passes `limit: 0, offset: 0`, i.e. "fetch all") this is functionally harmless. However, the proto contract advertises `limit`/`offset` as pagination controls, and a caller sending `limit: 10` will get a correct-looking but potentially expensive response — all roots were fetched from the DB regardless.

This creates a misleading contract: the RPC signature implies server-side pagination but the implementation silently fetches everything first. The fix is either:

1. Push limit/offset into `SurrealStore::get_children` (the preferred path — the existing query already has a simple `SELECT *` that is trivially extended with `LIMIT $limit START AT $offset`).
2. Or document the limitation clearly in the proto comment and restrict the current call site to `limit: 0` until the store layer is extended.

Given the diagnostics caller always passes `limit: 0`, option 2 is a valid short-term expedient if option 1 is out of scope for this PR.

**[Improvement] `packages/cli/src/commands/diagnostics.rs` line 110 — `nodes.len()` ignores the `count` field; can silently under-count if a `limit` is ever set**

```rust
Ok(response) => response.into_inner().nodes.len(),
```

The response carries both `nodes` (the page) and `count`. Currently `limit: 0` means the daemon returns all roots and `count == nodes.len()`, so the result is correct. But `nodes.len()` will silently return only the page size if a future caller passes a non-zero limit. Prefer `response.into_inner().count as usize` — it reflects the server's authoritative total. This is consistent with the semantics all other callers on `NodeListResponse` should respect, and makes the diagnostics intent ("how many roots exist?") unambiguous.

---

#### Nitpicks

**Nit: `packages/core/src/services/node_service.rs` line 3606 — `get_roots` lacks a doc comment**

Every neighbouring public method on `NodeService` (`get_children`, `get_descendants`, etc.) has a `///` doc block with `# Arguments`, `# Returns`, and a `# Examples` section. `get_roots` is undocumented. Not a blocker, but inconsistent with the local convention.

malibio:
## Review Feedback Addressed

**✅ Addressed (2/2 important findings):**

1. **`nodes.len()` → `count as usize`** in `diagnostics.rs:110` — `NodeListResponse.count` is the authoritative total; `nodes.len()` would silently under-count if a non-zero limit were ever passed.

2. **In-memory pagination documented** in `GetRootsRequest` proto comment, referencing follow-up issue #1167 for pushing `LIMIT`/`START AT` into `SurrealStore::get_children`. No current caller passes non-zero limit/offset so no behaviour change.

**✅ Nit addressed:** Added doc comment to `get_roots` in core service.

**Follow-up issue created:** #1167 — Push GetRoots limit/offset pagination into SurrealStore::get_children

PR #1169, opened 2026-05-18
Add PTY agent binary detection and CheckAgentAvailability RPC (closes #1124)

malibio:
## Summary

Implements binary and auth detection for all five PTY agent CLIs, exposes a `CheckAgentAvailability` RPC, and surfaces per-agent status in the `AgentLaunchPanel`.

- Adds `packages/agent/src/pty/detection.rs` with `detect_all_agents()`, augmented-PATH logic, and per-agent auth checks
- Adds `CheckAgentAvailability` RPC to `agent_session_service.proto` and implements handler
- Adds `FAILED_PRECONDITION` guard in `launch_session` for defense-in-depth
- Adds Tauri `check_agent_availability` command and `ptyCheckAgentAvailability()` frontend helper
- Wires availability badges into `AgentLaunchPanel.svelte`

## Test plan

- [ ] `cargo test -p nodespace-agent` passes (detection unit tests)
- [ ] Manual: launch panel shows ready/warning badges per-agent
- [ ] Manual: launching an agent with missing binary returns FAILED_PRECONDITION

### DIFF
--- a/packages/agent/src/pty/detection.rs
+++ b/packages/agent/src/pty/detection.rs
@@ -52,7 +52,8 @@ pub fn detect_all_agents() -> Vec<AgentAvailability> {
 // ---------------------------------------------------------------------------
 
 fn detect_one(agent_type: AgentType, binary: &'static str, path: &OsString) -> AgentAvailability {
-    let binary_path = which::which_in(binary, Some(path), ".").ok();
+    let cwd = std::env::current_dir().unwrap_or_else(|_| std::path::PathBuf::from("."));
+    let binary_path = which::which_in(binary, Some(path), &cwd).ok();
     let binary_found = binary_path.is_some();
     let auth_found = check_auth(agent_type);
     let install_hint = if binary_found {
@@ -186,10 +187,18 @@ fn path_helper_output() -> Option<OsString> {
         return None;
     }
     let stdout = String::from_utf8(output.stdout).ok()?;
-    // Extract the value between the first pair of double-quotes.
-    let start = stdout.find('"')? + 1;
-    let end = stdout[start..].find('"')? + start;
-    Some(OsString::from(&stdout[start..end]))
+    // Find the PATH= line and extract the quoted value defensively using
+    // strip_prefix/strip_suffix rather than positional find('"') so that
+    // unexpected whitespace or ordering doesn't silently yield a wrong path.
+    let path_line = stdout
+        .lines()
+        .find(|l| l.trim_start().starts_with("PATH="))?;
+    let after_prefix = path_line.trim_start().strip_prefix("PATH=\"")?;
+    // Accept either `"…"; export PATH;` or a bare closing quote.
+    let value = after_prefix
+        .strip_suffix("\"; export PATH;")
+        .or_else(|| after_prefix.strip_suffix('"'))?;
+    Some(OsString::from(value))
 }
 
 // ---------------------------------------------------------------------------
@@ -203,7 +212,7 @@ mod tests {
     #[test]
     fn detect_all_returns_five_entries() {
         let results = detect_all_agents();
-        assert_eq!(results.len(), 5);
+        assert_eq!(results.len(), AGENT_CATALOG.len());
     }
 
     #[test]
--- a/packages/agent/src/pty/detection.rs
+++ b/packages/agent/src/pty/detection.rs
@@ -1,0 +1,261 @@
+//! Binary and auth detection for PTY agent CLIs (Issue #1124).
+//!
+//! `detect_all_agents()` iterates the static AGENT_CATALOG and checks two
+//! things per agent: (1) the binary is reachable on an augmented PATH, and
+//! (2) the user has configured an auth credential.
+//!
+//! PATH augmentation runs `/usr/libexec/path_helper` on macOS first so that
+//! shell-configured PATH entries (e.g. Homebrew, nvm) are visible even when
+//! the daemon was launched outside a login shell.
+
+use std::ffi::OsString;
+use std::path::PathBuf;
+
+use crate::acp::registry::AGENT_CATALOG;
+use crate::agent_types::AgentType;
+
+// ---------------------------------------------------------------------------
+// Public types
+// ---------------------------------------------------------------------------
+
+/// Per-agent availability status returned by `detect_all_agents`.
+#[derive(Debug, Clone)]
+pub struct AgentAvailability {
+    pub agent_type: AgentType,
+    /// Binary name (e.g. "claude").
+    pub binary: &'static str,
+    /// `true` when the binary was found on the augmented PATH.
+    pub binary_found: bool,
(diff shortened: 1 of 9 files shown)


malibio (commented):
### Code Review Summary

This is a clean, well-scoped implementation that satisfies all acceptance criteria from issue #1124. The architecture is sound: detection lives in the daemon (not in Tauri), the UI just presents the RPC response, and the defense-in-depth guard in `launch_session` closes the race between a stale UI check and an actual spawn. The code follows established patterns throughout the codebase.

No critical issues. Three improvements worth addressing, plus minor nits.

---

### Findings

#### Suggested Improvements

**[Improvement] `packages/agent/src/pty/detection.rs` — `launch_session` runs `detect_all_agents()` on every launch; this calls `path_helper_output()` (a subprocess) on every call**

The defense-in-depth guard in `agent_session_service.rs` calls `detect_all_agents()` unconditionally before every `LaunchSession` RPC. On macOS, `augmented_path()` spawns `/usr/libexec/path_helper` synchronously via `std::process::Command::new`. This is a blocking subprocess call on a Tokio async task, which violates Tokio's executor contract (blocking work should not run on async tasks without `spawn_blocking`).

The issue is two-fold: (1) the subprocess blocks the Tokio thread, and (2) this overhead runs on every launch even when the UI already confirmed readiness 50 ms earlier. The simplest fix that keeps defense-in-depth intact is to wrap the detection call in `tokio::task::spawn_blocking`:

```rust
let agent_type_for_detect = agent_type;
let availability = tokio::task::spawn_blocking(move || {
    detect_all_agents()
        .into_iter()
        .find(|a| a.agent_type == agent_type_for_detect)
})
.await
.unwrap_or(None);
```

The same concern applies to `check_agent_availability` handler, though that one is explicitly a detection RPC so the cost is expected — the caller knows they are paying for a process spawn. The `launch_session` call site is the more surprising one.

**[Improvement] `packages/desktop-app/src-tauri/src/commands/agent_session.rs` line 130-135 — `status_to_command_error` silently maps `FAILED_PRECONDITION` to the generic `"GRPC_ERROR"` code**

The daemon now returns `FAILED_PRECONDITION` with a human-readable message when the agent is not ready. The Tauri-side `status_to_command_error` helper only explicitly handles `NotFound` and `InvalidArgument`:

```rust
fn status_to_command_error(status: tonic::Status) -> CommandError {
    let code = match status.code() {
        tonic::Code::NotFound => "SESSION_NOT_FOUND",
        tonic::Code::InvalidArgument => "INVALID_ARGUMENT",
        _ => "GRPC_ERROR",  // <-- FAILED_PRECONDITION falls here
    }
    .to_string();
```

The frontend error banner (line 159 of `agent-launch-panel.svelte`) displays `e.message`, which comes from `status.message()` regardless of code — so the user-visible message is fine. But a caller that wants to distinguish "agent not configured" from a generic RPC failure has no way to do so. Add a `FAILED_PRECONDITION` arm:

```rust
tonic::Code::FailedPrecondition => "AGENT_NOT_READY",
```

**[Improvement] `packages/agent/src/pty/detection.rs` lines 180-192 — `path_helper_output` parses with positional `find('"')` which is fragile for malformed output**

The parser takes the content between the first and second `"` in `/usr/libexec/path_helper -s` output. The format is stable on macOS, but if `path_helper` ever emits an error message with a `"` in it, the extracted "path" would be garbage rather than `None`. The function already returns `Option`, so this is about making the `None` case more reliable.

A tighter parse checks that the extracted substring looks like a PATH value (contains `/` or is empty) before returning it, or uses a simple prefix match on `PATH="`:

```rust
let line = stdout.lines().find(|l| l.starts_with("PATH=\""))?;
let inner = line.strip_prefix("PATH=\"")?.strip_suffix("\"; export PATH;")?;
Some(OsString::from(inner))
```

This is low risk today since `path_helper` output is OS-controlled, but the tighter parse is trivially more defensive.

---

#### Nitpicks

**Nit: `packages/agent/src/pty/detection.rs` line 55 — `which_in` uses `"."` as the cwd**

`which::which_in(binary, Some(path), ".")` uses the current working directory as the third argument. When the daemon starts, its cwd is predictable, but passing the absolute cwd via `std::env::current_dir()` (with a fallback to `"."`) is more explicit about intent and avoids any surprise if the cwd ever changes.

**Nit: `packages/desktop-app/src/lib/components/agent/agent-launch-panel.svelte` line 53 — `agentStatus` function return type annotation is long; consider a type alias**

The return type `'ready' | 'binary_missing' | 'auth_missing' | 'binary_missing_and_auth_missing' | 'unknown'` appears twice (function signature). Extracting it as `type AgentStatus = ...` at the top of the `<script>` block reduces repetition and makes it easier to add a new state later.

**Nit: `packages/agent/src/pty/detection.rs` tests — `detect_all_returns_five_entries` hardcodes `5`**

The assertion `assert_eq!(results.len(), 5)` will silently break if an agent is added to `AGENT_CATALOG` without updating this test. `assert_eq!(results.len(), AGENT_CATALOG.len())` is self-documenting and stays correct automatically.

---

- [x] `CheckAgentAvailability` RPC added to proto and implemented
- [x] Returns per-agent status covering all four states
- [x] PATH augmentation includes all six specified directories
- [x] `/usr/libexec/path_helper` used on macOS (behind `#[cfg(target_os = "macos")]`)
- [x] Binary detection correct for all five agents (pulled from `AGENT_CATALOG`)
- [x] Auth detection per agent — all five covered with correct env vars and file paths
- [x] `AgentLaunchPanel.svelte` calls `CheckAgentAvailability` on mount with per-agent badges
- [x] Detection runs in the daemon; UI displays RPC response
- [x] `LaunchSession` returns `FAILED_PRECONDITION` when agent not ready
- [x] Code passes `bun run quality:fix` (per implementation commit)

Overall verdict: **approve with suggested improvements**. The blocking subprocess in `launch_session` (Improvement 1) is the highest-priority item since it violates Tokio's threading model. The `FAILED_PRECONDITION` code mapping (Improvement 2) is important for frontend error handling correctness. The path_helper parser hardening (Improvement 3) is defensive polish.

malibio:
## Review Findings Addressed

All recommendations from the initial review have been implemented in commit 40a1b67d:

- **spawn_blocking**: `detect_all_agents()` now runs in `tokio::task::spawn_blocking` in both `launch_session` and `check_agent_availability` — the macOS `path_helper` subprocess no longer blocks the executor thread
- **AGENT_NOT_READY error code**: Added `tonic::Code::FailedPrecondition => "AGENT_NOT_READY"` to `status_to_command_error` so the frontend can distinguish agent-not-configured from generic gRPC errors
- **path_helper parser**: Replaced fragile positional `find('"')` with `strip_prefix`/`strip_suffix` on `PATH="`
- **which_in cwd**: Uses `current_dir()` instead of `"."` literal
- **Test**: `assert_eq!(results.len(), AGENT_CATALOG.len())` instead of hardcoded `5`
- **Svelte type alias**: `AgentStatus` type defined once and reused

PR #1170, opened 2026-05-18
Add hardware requirement labels to model selection UI (closes #1140)

malibio:
## Summary

- Adds `min_memory_gb` field to `CatalogEntry` and `ModelInfo` (8 GB Ministral 3B, 16 GB Ministral 8B / Gemma 4 E4B, 24 GB Gemma 4 31B)
- Propagates the field through `GgufModelManager::list()`, `OllamaModelManager::list()` (0 = unknown), daemon gRPC `ModelEntry` proto, and TypeScript `ModelInfo` interface
- Adds `get_system_ram_gb` Tauri command using the existing `detect_system_ram()` function (now public)
- Model manager UI shows "Requires X GB RAM" on every card; dims card and shows "Insufficient RAM" chip when system RAM is below the model's requirement
- Extends `list_models_have_correct_metadata` and `list_includes_gemma4_entries_with_correct_metadata` tests with `min_memory_gb` assertions

## Test plan

- [ ] Frontend: 128 files, 3916 tests passed (no regressions)
- [ ] Rust `model_manager` tests: 33 passed, 0 failed
- [ ] `bun run quality:fix` passed clean
- [ ] Rust crates `nodespace-agent` and `nodespace-daemon` compile clean (`cargo check`)

### DIFF
--- a/packages/agent/src/agent_types.rs
+++ b/packages/agent/src/agent_types.rs
@@ -388,6 +388,9 @@ pub struct ModelInfo {
     pub backend: ModelBackend,
     /// Current download / load status.
     pub status: ModelStatus,
+    /// Minimum system RAM (in GiB) required to run this model comfortably.
+    /// Zero means unknown (e.g. for Ollama-managed models).
+    pub min_memory_gb: u8,
 }
 
 /// Specification of a chat model's capabilities.
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -55,6 +55,8 @@ struct CatalogEntry {
     sha256: &'static str,
     context_window: u32,
     default_temperature: f32,
+    /// Minimum system RAM (in GiB) required to run this model comfortably.
+    min_memory_gb: u8,
 }
 
 /// Ministral 3B -- fast, lightweight, identical tool reliability.
@@ -69,6 +71,7 @@ const MINISTRAL_3B: CatalogEntry = CatalogEntry {
     sha256: "", // Skip verification — official Mistral repo, Xet storage
     context_window: 32_768,
     default_temperature: 0.3,
+    min_memory_gb: 8,
 };
 
 /// Ministral 8B -- deeper reasoning, vision capable.
@@ -83,6 +86,7 @@ const MINISTRAL_8B: CatalogEntry = CatalogEntry {
     sha256: "", // Skip verification — official Mistral repo, Xet storage
     context_window: 32_768,
     default_temperature: 0.3,
+    min_memory_gb: 16,
 };
 
 /// Gemma 4 E4B -- Google's efficient ~4B-effective model; stronger reasoning
@@ -98,6 +102,7 @@ const GEMMA_4_E4B: CatalogEntry = CatalogEntry {
     sha256: "", // Skip verification — official ggml-org repo (llama.cpp team), Xet storage
     context_window: 32_768,
     default_temperature: 0.3,
+    min_memory_gb: 16,
 };
 
 /// Gemma 4 31B -- Google's larger dense quality-tier option (24GB+ Apple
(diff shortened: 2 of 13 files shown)


malibio:
### Code Review Summary

This PR cleanly implements the hardware-requirement label feature from issue #1140. Every acceptance criterion is met: \`min_memory_gb\` flows from the Rust catalog through gRPC/Tauri to the TypeScript type, renders in the UI as \"Requires X GB RAM\", and cards with insufficient RAM are visually distinguished via dimming + a warning chip. Tests are extended appropriately and the logic is well-guarded (double zero-check prevents false positives when RAM or the field is unknown).

The implementation is lean and avoids over-engineering. A few findings to address.

---

### Findings

#### Critical Issues

**[Critical/Blocker] `local_agent.rs` truncates u32 → u8 without bounds check**

`packages/desktop-app/src-tauri/src/commands/local_agent.rs`, line 389

```rust
min_memory_gb: entry.min_memory_gb as u8,
```

\`entry.min_memory_gb\` is a \`u32\` from the protobuf \`ModelEntry\`. Casting directly to \`u8\` silently wraps for values ≥ 256. While current catalog values are all ≤ 24, this is a latent correctness bug. The field on the Rust \`ModelInfo\` struct is already \`u8\`, so the narrowing is necessary — but it should be explicit:

```rust
min_memory_gb: entry.min_memory_gb.try_into().unwrap_or(u8::MAX),
```

or at minimum:

```rust
min_memory_gb: entry.min_memory_gb.min(u8::MAX as u32) as u8,
```

---

#### Suggested Improvements

**[Improvement] \`detect_system_ram\` is now \`pub\` — this leaks an implementation detail**

`packages/agent/src/local_agent/model_manager.rs`, line 836

The function was made \`pub\` solely to expose it to \`chat_models.rs\` in the desktop crate. This punches a hole in the module boundary: any consumer of the \`nodespace_agent\` crate can now call a low-level sysinfo wrapper directly. Consider one of:
- Keep it \`pub(crate)\` and expose a higher-level \`system_ram_gb() -> u64\` function on \`CompositeModelManager\` or a dedicated system info helper.
- Or accept the \`pub\` visibility with a short \`# Safety / Usage\` doc comment explaining it is exposed for Tauri command use, to signal intent to future readers.

**[Improvement] Ollama models hardcode \`min_memory_gb: 0\` (unknown) — document the rationale in code**

`packages/agent/src/local_agent/ollama_model_manager.rs`, line 178

The issue itself acknowledges this decision and proposes an alternative (\`Option\<u8\>\`). The PR chose \`0\` as the sentinel, which is fine, but the comment in the struct definition (\`agent_types.rs\`) and here diverges from the issue's suggestion of defaulting to 16 GB for Ollama. A brief inline comment at the \`0\` assignment explaining the intentional choice (\"Ollama manages its own model loading; memory footprint is not tracked by the catalog\") would prevent future contributors from treating it as an oversight.

**[Improvement] \`systemRamGb\` is not fetched in the non-Tauri mock path**

`packages/desktop-app/src/lib/stores/model-store.svelte.ts`, lines 122–127

The mock path does not set \`systemRamGb\`, so it stays \`0\` in development mode, meaning the \"Insufficient RAM\" chip never renders during browser-based dev work even though the mock models all have \`min_memory_gb > 0\`. For local development fidelity, consider seeding a plausible value (e.g., 8) so the UI behavior can be exercised without a live Tauri process.

---

#### Nitpicks

- **Nit:** `packages/desktop-app/src/lib/components/settings/model-manager.svelte`, line 83 — the \`insufficientRam\` \`{@const}\` is slightly long for an inline template expression. Hoisting it to a derived rune in the \`<script>\` block (keyed on the model id) would improve readability, though the current approach is functionally equivalent.

- **Nit:** `packages/desktop-app/src/lib/services/tauri-commands.ts` — the JSDoc for \`getSystemRamGb\` says \"Returns 0 outside Tauri\" which is correct and useful. Consider also noting that the return value is rounded *down* (floor division), matching the Rust implementation, so callers know not to expect a ceiling.

- **Nit:** The test added in `model-store.test.ts` asserts \`toBeGreaterThan(0)\` for all mock models, but does not verify the specific values (8, 16, 16). This is fine for a smoke test but misses an opportunity to assert the concrete data matches the catalog rationale.

---

### Verdict

The Critical finding (silent u8 truncation) must be fixed before merge. The improvement suggestions are non-blocking but the \`detect_system_ram\` visibility point is worth addressing to preserve module hygiene. Everything else is polish.

PR #1174, opened 2026-05-18
Add schema field rename with property data migration

### LINKED ISSUE #1088: Schema field rename with property data migration

malibio:
## Summary

- Adds `SurrealStore::rename_schema_field` that fetches all nodes of a type, rekeys the field in the namespace object in Rust, and writes each updated node back individually
- Adds `NodeService::rename_schema_field` that validates source/destination field existence, runs data migration, then updates the schema definition
- Adds `rename_fields: [{from, to}]` to `UpdateSchemaParams` (processed before add/remove so subsequent operations see post-rename schema state) and to the `update_schema` agent tool
- 4 integration tests covering: schema definition update, source-field-not-found error, rename-to-existing-field error, and node data migration

Closes #1088

### DIFF
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -498,7 +498,7 @@ fn def_create_schema() -> ToolDefinition {
 fn def_update_schema() -> ToolDefinition {
     ToolDefinition {
         name: "update_schema".into(),
-        description: "Modify an existing schema type: add/remove/rename fields, add/remove relationships, update description or title_template. Use rename_fields to safely rename a field — it migrates all existing node property data to the new key and updates the schema definition atomically.".into(),
+        description: "Modify an existing schema type: add/remove/rename fields, add/remove relationships, update description or title_template. Use rename_fields to safely rename a field — it migrates all existing node property data to the new key and updates the schema definition.".into(),
         parameters_schema: json!({
             "type": "object",
             "properties": {
@@ -546,7 +546,8 @@ fn def_update_schema() -> ToolDefinition {
                             "from": { "type": "string", "description": "Current field name" },
                             "to": { "type": "string", "description": "New field name" }
                         },
-                        "required": ["from", "to"]
+                        "required": ["from", "to"],
+                        "additionalProperties": false
                     }
                 },
                 "add_relationships": {
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -498,7 +498,7 @@ fn def_create_schema() -> ToolDefinition {
 fn def_update_schema() -> ToolDefinition {
     ToolDefinition {
         name: "update_schema".into(),
-        description: "Modify an existing schema type: add/remove fields, add/remove relationships, update description or title_template.".into(),
+        description: "Modify an existing schema type: add/remove/rename fields, add/remove relationships, update description or title_template. Use rename_fields to safely rename a field — it migrates all existing node property data to the new key and updates the schema definition atomically.".into(),
         parameters_schema: json!({
             "type": "object",
             "properties": {
@@ -536,6 +536,18 @@ fn def_update_schema() -> ToolDefinition {
                     "type": "array",
                     "description": "Field names to remove",
                     "items": { "type": "string" }
+                },
+                "rename_fields": {
+                    "type": "array",
+                    "description": "Fields to rename. Each entry rekeys property data on ALL existing nodes of this schema type and updates the schema definition. Renaming to an existing field name is rejected. Processed before add_fields/remove_fields.",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "from": { "type": "string", "description": "Current field name" },
+                            "to": { "type": "string", "description": "New field name" }
+                        },
+                        "required": ["from", "to"]
+                    }
                 },
                 "add_relationships": {
                     "type": "array",
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -3232,6 +3232,97 @@ impl SurrealStore {
         Ok(())
     }
 
+    /// Rekey a property field for all nodes of a given type (Issue #1088).
+    ///
+    /// Fetches all nodes of the given type, renames the field in the namespace
+    /// object in Rust, then writes back each updated node. This avoids SurrealDB
+    /// query syntax issues with dynamic property paths and is correct for all
+    /// field name formats including `custom:field`.
+    ///
+    /// # Errors
+    /// Returns an error if `from == to`, if either name is empty, or if any DB operation fails.
+    pub async fn rename_schema_field(&self, type_id: &str, from: &str, to: &str) -> Result<u64> {
+        if from.is_empty() || to.is_empty() {
+            return Err(anyhow::anyhow!("Field names must not be empty"));
+        }
+        if from == to {
+            return Err(anyhow::anyhow!(
+                "Source and destination field names are the same: '{}'",
+                from
+            ));
+        }
+
+        // Fetch all nodes of the given type
+        let query = "SELECT * FROM node WHERE node_type = $type_id;";
+        let mut response = self
+            .db
+            .query(query)
+            .bind(("type_id", type_id.to_string()))
+            .await
+            .context(format!(
+                "Failed to fetch nodes for type '{}' during field rename",
+                type_id
+            ))?;
+
+        let nodes: Vec<SurrealNode> = response
+            .take(0)
+            .context("Failed to take node results during field rename")?;
+
+        let mut affected: u64 = 0;
+
+        for node in nodes {
+            let node_id = extract_record_key(&node.id);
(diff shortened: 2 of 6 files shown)


malibio:
### Code Review Summary

This is a net-positive change that solves a real data-loss problem. The implementation is readable, well-tested for the common cases, and the layering (store → service → handler → tool) is consistent with the existing codebase architecture. There are no security issues or critical regressions.

Three issues need attention before merge: a critical gap in the rename-validation/playbook-guard flow (renames bypass both schema validation and playbook impact checks), a moderate correctness hole around duplicate `from` values in a single rename batch, and a documentation mismatch where the doc comment says "atomically" but the implementation is explicitly non-atomic.

---

### Findings

#### Critical Issues

**[Critical/Blocker] Renames bypass schema validation and playbook impact check**

`packages/core/src/mcp/handlers/schema.rs` — Phase 1 (renames) runs before Phase 2 (schema fetch + validation). This means:

1. Schema validation (`SchemaNodeBehavior.validate_schema_node`) runs only on the add/remove result, never on the rename result.
2. The playbook impact check (`check_schema_change_impact`) also runs only in Phase 2. A rename that breaks a playbook field reference will execute the data migration and update the schema definition before the guard ever fires. By the time the guard could reject it, the rename is already committed.

The fix is to move the playbook impact check (and potentially validation) ahead of Phase 1, or to run a targeted impact check per rename before executing it.

---

#### Suggested Improvements

**[Improvement] Duplicate `from` values in a single rename batch produce undefined behavior**

`packages/core/src/mcp/handlers/schema.rs` lines 510-518 — The handler checks for duplicate `to` values but not duplicate `from` values. If a caller passes `[{from: "x", to: "a"}, {from: "x", to: "b"}]`, the second rename will succeed (returning 0 affected nodes since the field was already renamed away) and silently produce an incorrect final state. Add a `seen_sources` check alongside the existing `seen_destinations` check.

**[Improvement] Non-atomic behavior is documented as atomic in two places**

`packages/core/src/services/node_service.rs` line 2718 — The doc comment on `rename_schema_field` says "atomically" in its title and again in the comment body for the schema definition step. `packages/core/src/mcp/handlers/schema.rs` line 281 — The `rename_fields` field doc says "updates the schema definition atomically." The `def_update_schema` tool description in `packages/agent/src/local_agent/tools.rs` also says "atomically."

None of this is atomic: step 2 (data migration) is N individual UPDATE statements with no transaction, and step 3 (schema definition update) is a separate write that can fail after all node data is already migrated.

For a greenfield project with no production data this is acceptable, but the documentation actively misleads: a caller reading the doc comment would believe a crash is safe, when in reality a crash between node 5 and node 6 leaves the graph inconsistent. Remove the word "atomic" from all three locations and replace with accurate language (e.g., "sequential; if the schema definition update fails after data migration, the data is already correct and the schema update can be retried").

**[Improvement] `NodeService::rename_schema_field` rebuilds schema properties manually, diverging from `handle_update_schema`**

`packages/core/src/services/node_service.rs` lines 2782-2793 — The service method constructs the schema properties JSON object by hand, which means any new schema-level field (a future addition) would require updating this code path separately from `handle_update_schema`. The service method could instead delegate the schema-definition step back to the MCP handler (passing an `UpdateSchemaParams` with only the rename applied to the fields list), or at minimum share a helper. As a minimum, a comment noting the duplication risk would reduce future drift.

**[Improvement] `SELECT *` over potentially large node set without pagination**

`packages/core/src/db/surreal_store.rs` line 3256 — `SELECT * FROM node WHERE node_type = $type_id` loads all nodes of a type into memory simultaneously. For a schema type with thousands of instances this could cause significant memory pressure. A batched cursor (LIMIT + START AT) or a streaming approach would be more scalable. This is not a blocker today but worth tracking.

---

#### Nitpicks

**Nit:** `packages/core/src/mcp/handlers/schema.rs` line 524 — `map_err` maps all `NodeServiceError` variants to `MCPError::invalid_params`. A "schema not found" error (which is a 404-style condition) and a database I/O error both become `invalid_params`. Using `MCPError::internal_error` for database failures and reserving `invalid_params` for validation failures would give callers more actionable signal.

**Nit:** `packages/core/src/mcp/handlers/schema_test.rs` lines 501-505 — The assertion `old_field should be removed after rename` has a disjunction: `is_none() || is_some_and(|v| v.is_null())`. If the remove logic is broken and leaves behind a `null` value rather than actually deleting the key, this test passes. Simplify to `assert!(ns_props.get("old_field").is_none(), ...)` to make the expectation unambiguous.

**Nit:** `packages/agent/src/local_agent/tools.rs` — the `rename_fields` item schema does not include `"additionalProperties": false`. Minor, but consistent with making the tool schema defensive.

**Nit:** `packages/daemon/src/services/agent_session_service.rs` — blank-line removal is unrelated to this issue. Fine to keep, but worth noting it is a drive-by change.

malibio:
## Review feedback addressed (commit 9f95a671)

### ✅ Addressed (🔴 Critical)
**Playbook guard ran after renames** — Moved `check_schema_change_impact` to Phase 0, before any rename mutations execute. A blocked rename now returns an error without migrating any node data.

### ✅ Addressed (🟡 Important × 2)
**Duplicate `from` values in rename batch** — Added `seen_sources` check alongside the existing `seen_destinations` check. Duplicate source field names in a single batch now return `MCPError::invalid_params` immediately.

**Inaccurate "atomically" documentation** — Removed from `node_service.rs` doc comment (clarified "sequential but not atomic") and from `tools.rs` description.

### ✅ Addressed (🟢 Suggestions)
- Added `"additionalProperties": false` to `rename_fields` items schema in `tools.rs`
- Tightened `test_rename_field_migrates_node_data` assertion: `old_field` must be `is_none()` (no longer accepts null as valid "removed" state)

### ⏭️ Skipped
- **Service method manually rebuilds schema props JSON** — this pattern is consistent throughout the file; extracting a helper is scope creep for this PR
- **`SELECT *` without pagination** — rename is a schema-management operation, not a hot path; pagination adds complexity before there's a measured problem
- **Error mapping DB failures to `MCPError::invalid_params`** — the existing pattern throughout this handler maps all errors to `invalid_params`; fixing error code semantics is a cross-cutting concern for a separate issue

PR #1175, opened 2026-05-19
Push GetRoots limit/offset pagination into SurrealStore::get_children

malibio:
Closes #1167

### DIFF
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -1898,58 +1898,68 @@ impl SurrealStore {
         Ok(nodes)
     }
 
-    pub async fn get_children(&self, parent_id: Option<&str>) -> Result<Vec<Node>> {
+    pub async fn get_children(&self, parent_id: &str) -> Result<Vec<Node>> {
         // Universal Graph Architecture (Issue #783, #788): Properties embedded in node.properties
         // Use universal relationship table for hierarchy traversal with fractional ordering
-        let surreal_nodes = if let Some(parent_id) = parent_id {
-            // Create RecordId for parent node
-            let parent_thing = node_record_id(parent_id);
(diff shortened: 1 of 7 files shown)


malibio:
## Review Findings — Addressed

All recommendations from the initial code review have been implemented in commit `ecc0af00`.

### ✅ Addressed (4/4)

**🟡 `ORDER BY id ASC` on paginated root query** — Added to all four SQL variants in `SurrealStore::get_roots`. Without a deterministic sort, LIMIT/START AT produces overlapping or skipped pages across calls.

**🟡 Split `get_children`/`get_roots` at store layer** — `SurrealStore::get_children(Option<&str>, ...)` replaced with two distinct methods: `get_children(&str)` and `get_roots(limit, offset)`. The overloaded dual-purpose signature no longer exists; `limit`/`offset` can no longer be silently ignored on the non-root path.

**🟡 Pagination tests** — Three new store-level integration tests added:
- `test_get_roots_no_pagination` — child nodes excluded from roots
- `test_get_roots_limit` — DB-level LIMIT applied
- `test_get_roots_limit_offset_non_overlapping` — page1 ∪ page2 = full set, no overlap

**🟢 Proto comment updated** — `GetRootsRequest` comment now reflects that pagination is pushed into SurrealDB, not applied in application memory. "tracked in #1167" forward reference removed.

### Also fixed (compilation regressions from main)
- `capture_service.rs` — missing `success: true` in `ExitStatus` test helper (introduced by #1168)
- `local_agent.rs` + `Cargo.toml` — `AgentSession` struct initializer missing `#[cfg(any(test, feature = "testing"))]` guard for `system_prompt_override`; `testing` feature declared in desktop-app `Cargo.toml` (field was added in #1132)

### ⏭️ Skipped: 0

PR #1178, opened 2026-05-19
Switch nodespaced IPC from TCP to Unix Domain Socket

malibio:
## Summary

- `nodespaced` binds on `~/.nodespace/daemon.sock` (0600 perms) instead of TCP `[::1]:50051`
- Stale socket removed on startup; cleaned up on graceful shutdown
- `NODESPACED_SOCKET` env var overrides path in daemon, CLI, watcher, and agent-tools
- CLI uses `hyper_util::TokioIo` + `tower::service_fn` connector pattern for UDS channel
- `@nodespace/agent-tools` TypeScript client uses `unix:` URI scheme via `@grpc/grpc-js`
- Windows path gated behind `#[cfg(windows)] compile_error!()` for explicit failure
- Removes `cfg` feature gate on `system_prompt_override` field to fix clippy under `--all-targets`

Closes #1176

## Test plan

- [x] `cargo build --workspace` succeeds
- [x] `bun run quality:fix` passes (0 errors, 0 warnings)
- [x] `bun run test:all` passes (1189 Rust tests, all frontend tests)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4394,6 +4394,7 @@ dependencies = [
  "chrono",
  "dirs 5.0.1",
  "futures",
+ "hyper-util",
  "nodespace-agent",
  "nodespace-core",
  "nodespace-daemon",
@@ -4412,6 +4413,7 @@ dependencies = [
  "tokio-stream",
  "tokio-util",
  "tonic 0.12.3",
+ "tower 0.5.3",
  "tower-http",
  "tracing",
  "tracing-subscriber",
@@ -4426,6 +4428,7 @@ dependencies = [
  "chrono",
  "clap",
  "crossterm",
+ "hyper-util",
  "nodespace-core",
  "nodespace-daemon",
  "serde_json",
@@ -4433,6 +4436,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tonic 0.12.3",
+ "tower 0.5.3",
 ]
 
 [[package]]
@@ -4476,6 +4480,7 @@ dependencies = [
  "async-stream",
  "async-trait",
  "chrono",
+ "hyper-util",
  "image",
  "nodespace-agent",
  "nodespace-core",
--- a/packages/agent-tools/src/client.ts
+++ b/packages/agent-tools/src/client.ts
@@ -9,7 +9,9 @@ const __dirname = path.dirname(__filename);
 export const PROTO_PATH = path.resolve(__dirname, '../../daemon/proto/node_service.proto');
 
 function resolveAddress(): string {
-  return process.env.NODESPACED_ADDR ?? 'localhost:50051';
+  const sock = process.env.NODESPACED_SOCKET
+    ?? `${process.env.HOME}/.nodespace/daemon.sock`;
+  return `unix:${sock}`;
 }
 
 type UnaryCallback<TResponse> = (
--- a/packages/agent-tools/src/tests/connection-error.test.ts
+++ b/packages/agent-tools/src/tests/connection-error.test.ts
@@ -3,25 +3,25 @@ import { getNode } from '../tools.js';
 import { resetClient } from '../client.js';
 import { ToolError } from '../types.js';
 
-// Integration-style check: with no daemon running on the bogus port, a real
-// gRPC call should surface as a structured ToolError instead of an
-// uncaught exception. We point at an unused loopback port via env override.
+// Integration-style check: with no daemon running at the bogus socket path, a
+// real gRPC call should surface as a structured ToolError instead of an
+// uncaught exception. We point at a nonexistent socket via env override.
 
-const originalAddr = process.env.NODESPACED_ADDR;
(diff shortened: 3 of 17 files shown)


malibio:
### Code Review Summary

This PR correctly implements the core security goal: replacing the world-accessible TCP socket with a `0600`-permissioned Unix Domain Socket. The architecture is clean, the platform gating is explicit, and the test coverage is thorough. The change is a clear net improvement and is ready to merge pending one improvement and a few nits.

---

### Findings

#### Suggested Improvements

**`packages/daemon/src/services/settings_service.rs` line 20 — Stale `grpc_address` field misleads operators**

The `DaemonConfig` struct retains a `grpc_address: Option<String>` field, and `DEFAULT_GRPC_SOCKET` is set to the tilde-literal string `"~/.nodespace/daemon.sock"`. This field predates UDS and was used for TCP. Two problems:

1. The tilde literal is **not** shell-expanded; if a client reads this value and feeds it to a connector naively it gets a broken path. The real socket path is always computed at runtime by `socket_path()`, which correctly expands `$HOME`.
2. The field in `DaemonConfigResponse` still says `grpc_address`, reinforcing the old TCP mental model for any tooling that reads the daemon config via gRPC.

This isn't a correctness issue for the daemon itself (the field is cosmetic in the response), but it is a correctness trap for future callers and violates the greenfield "delete old patterns immediately" mandate.

**Recommendation:** Rename the field to `socket_path` (or remove it), replace the tilde literal with the actual expanded path, and derive it from the same `socket_path()` call used by the server — not from a hardcoded constant.

---

#### Nitpicks

**`packages/daemon/src/main.rs` lines 119–123 / 170–174 — TOCTOU window between `bind` and `set_permissions`**

```rust
let listener = UnixListener::bind(&sock)?;
std::fs::set_permissions(&sock, std::fs::Permissions::from_mode(0o600))?;
```

There is a brief window where the socket file exists with world-readable permissions (inherited from umask) before `set_permissions` closes it. On a single-user developer machine this is inconsequential, but it is worth noting. The canonical mitigation is to set the process `umask` to `0o177` before `bind` and restore it after. Not a blocker for greenfield code.

**`packages/cli/src/lib.rs` line 67 — Double `NODESPACED_SOCKET` resolution**

`resolve_socket_path` checks `NODESPACED_SOCKET` from the env, and so does the `Cli` struct via `#[arg(env = "NODESPACED_SOCKET")]`. For the flag-absent case both paths agree, but the intent is slightly blurred: the clap `env` binding populates `cli.socket` automatically, so `resolve_socket_path` would never reach the `std::env::var` branch in normal operation (clap already consumed it). The function is still correct — it is just doing redundant work. A comment clarifying the precedence (CLI flag > env-via-clap > env-direct > default) would improve readability.

**`packages/agent-tools/src/client.ts` line 13 — `process.env.HOME` may be `undefined` on non-Unix**

```typescript
?? `${process.env.HOME}/.nodespace/daemon.sock`;
```

`process.env.HOME` is `undefined` on Windows; the template literal would produce `"undefined/.nodespace/daemon.sock"`. Since Windows is not a supported target (the Rust side hard-errors at compile time), this is not a practical problem today, but a `?? '/tmp'` fallback or an explicit `throw` would make the failure mode explicit rather than silent.

**`packages/cli/tests/cli_integration.rs` line 322 — Hardcoded `/tmp` path in `connect_refused` test**

```rust
connect(std::path::Path::new("/tmp/nodespace-no-such-daemon.sock"))
```

This path is not guaranteed to be absent on a developer machine (though extremely unlikely to be occupied). Using `tempfile::TempDir::new().unwrap().path().join("no-daemon.sock")` would be more hermetic.

---

| Criterion | Status |
|---|---|
| `nodespaced` binds on `~/.nodespace/daemon.sock` | PASS |
| Socket created with `0600` permissions | PASS |
| Stale socket removed on startup; cleaned up on shutdown | PASS |
| `NODESPACED_SOCKET` honored in daemon, CLI, and agent-tools | PASS |
| CLI connects via UDS connector | PASS |
| `@nodespace/agent-tools` uses `unix:` URI scheme | PASS |
| No TCP port opened by `nodespaced` | PASS |
| Windows path gated behind `compile_error!` | PASS (line 45 of `main.rs`) |
| `grpc_client.rs` uses OS-assigned ephemeral TCP (in-process only, not user-visible) | PASS — correctly unchanged |
| CLI integration tests updated for UDS | PASS — full round-trip test over tempdir socket |

The only acceptance criterion that has a residual concern is the `grpc_address` / `DEFAULT_GRPC_SOCKET` field in `settings_service.rs`, which carries over stale TCP terminology and a non-expanding tilde literal into the API surface. Everything else is well-implemented.

malibio (commented):
Review status: APPROVED (self-review). LGTM. One improvement recommended (stale grpc_address field in settings_service.rs), a handful of nits — none are blockers. The UDS transport implementation is correct, platform gating is explicit, and integration test coverage is solid.

PR #1188, opened 2026-05-19
Publish @nodespaceai/agent-tools to npm (#1183)

malibio:
## Summary
- Rename package to `@nodespaceai/agent-tools` npm scope
- Add `publishConfig: { access: "public" }` and update `files` field
- Add README with installation, usage, and API docs
- Wire `publish-npm` job to `release.yml` — runs on GitHub release events after `build-tauri` completes, uses `NPM_TOKEN` secret

## Notes
- Manual initial publish (`npm publish --access public` from `packages/agent-tools/`) still needed to bootstrap the registry entry — requires `NPM_TOKEN` in the environment
- Subsequent publishes will happen automatically via CI on each GitHub release tag

## Test plan
- [ ] `bun run build` succeeds in `packages/agent-tools/` (TypeScript compiles with types)
- [ ] All 17 agent-tools tests pass (`bun run test` in `packages/agent-tools/`)
- [ ] `release.yml` `publish-npm` job only runs on `release` events (not `workflow_dispatch`)
- [ ] Manual publish bootstrap: `npm publish --access public` from `packages/agent-tools/` with valid `NPM_TOKEN`
- [ ] `npm install @nodespaceai/agent-tools` works after initial publish

Closes #1183

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -201,7 +201,6 @@ jobs:
   publish-npm:
     name: Publish @nodespaceai/agent-tools
     runs-on: ubuntu-latest
-    needs: build-tauri
     if: github.event_name == 'release'
     steps:
       - name: Checkout repository
@@ -220,6 +219,12 @@ jobs:
 
       - name: Install dependencies
         run: bun install
+        working-directory: packages/agent-tools
+
+      - name: Set version from release tag
+        run: |
+          TAG="${{ github.event.release.tag_name }}"
+          npm version "${TAG#v}" --no-git-tag-version
         working-directory: packages/agent-tools
 
       - name: Build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -197,3 +197,37 @@ jobs:
             target/${{ matrix.rust_targets || '' }}/release/bundle/
             target/release/bundle/
           if-no-files-found: warn
+
+  publish-npm:
+    name: Publish @nodespaceai/agent-tools
+    runs-on: ubuntu-latest
+    needs: build-tauri
+    if: github.event_name == 'release'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6.0.2
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: "1.1.38"
+
+      - name: Setup Node.js (for npm publish)
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install dependencies
+        run: bun install
+        working-directory: packages/agent-tools
+
+      - name: Build
+        run: bun run build
+        working-directory: packages/agent-tools
+
+      - name: Publish @nodespaceai/agent-tools
+        run: npm publish --access public
+        working-directory: packages/agent-tools
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/packages/agent-tools/README.md
+++ b/packages/agent-tools/README.md
@@ -1,0 +1,99 @@
+# @nodespaceai/agent-tools
+
+gRPC-backed knowledge graph tools for PTY-spawned agents (Claude Code, Codex, etc.).
+
+Exposes the NodeSpace knowledge graph to AI agents running inside a NodeSpace session — search nodes semantically, read, create, and update nodes via the local `nodespaced` daemon.
+
+## Installation
+
+```bash
+npm install @nodespaceai/agent-tools
+# or
+bun add @nodespaceai/agent-tools
+```
+
+## Requirements
+
+A running NodeSpace desktop app (or `nodespaced` daemon) with the gRPC server listening on its Unix domain socket. The tools connect automatically when the `NODESPACE_SOCKET` environment variable is set (injected by NodeSpace when spawning PTY agents).
+
+## Usage
+
+```typescript
+import {
+  searchSemantic,
+  getNode,
+  createNode,
+  updateNode,
+  getChildren,
+} from '@nodespaceai/agent-tools';
+
+// Semantic search across the knowledge graph
+const results = await searchSemantic('meeting notes from last week', 5);
+
+// Read a node by ID
+const node = await getNode('node-uuid-here');
+
+// Create a new node
+const created = await createNode('text', 'My note content', parentId);
+
+// Update an existing node
+const updated = await updateNode('node-uuid-here', 'Updated content');
+
+// List child nodes
+const children = await getChildren('parent-node-uuid');
+```
+
+## API
+
+### `searchSemantic(query, limit?)`
+
+Searches the knowledge graph using semantic (embedding) similarity.
+
+- `query` — natural language search query
+- `limit` — max results to return (default: 10)
+- Returns: `SearchResult[]` with `id`, `nodeType`, `content`, `score`
+
+### `getNode(id)`
+
+Fetches a single node by its UUID.
+
+- Returns: `NodeResult` with `id`, `nodeType`, `content`, `parentId`
+
+### `createNode(nodeType, content, parentId?)`
+
+Creates a new node in the knowledge graph.
+
+- `nodeType` — e.g. `"text"`, `"task"`, `"date"`
+- Returns: `NodeResult` for the created node
+
+### `updateNode(id, content)`
+
+Updates the content of an existing node.
+
+- Returns: `NodeResult` for the updated node
+
+### `getChildren(parentId)`
+
+Lists all direct children of a node.
+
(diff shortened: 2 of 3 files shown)


malibio:
**PR:** #1188 — Publish @nodespaceai/agent-tools to npm
**Reviewer:** Principal Engineer AI

---

### Overall Assessment

This is a well-scoped, self-contained change. The scope matches the issue exactly: rename the package, add publish config, add README, wire CI. The implementation is clean and the diff is minimal. Two issues need attention before this is production-ready: a workflow coupling problem that blocks the npm publish when any Tauri binary build fails, and a missing version bump guard that will silently no-op on every release after the first.

---

### Requirements Check

- ✅ `packages/agent-tools/package.json` has `name: "@nodespaceai/agent-tools"`, `publishConfig: { access: "public" }`, correct `files`, `main`, `types` fields
- ⚠️ Initial version published manually — this is a PR pre-condition, not automatable; noted in PR description. Acceptable.
- ⚠️ `npm install @nodespaceai/agent-tools` works from any machine — contingent on the manual bootstrap step; cannot be verified in CI pre-merge.
- ✅ CI/CD release workflow publishes `@nodespaceai/agent-tools` on release tag
- ✅ Package includes TypeScript types (`dist/index.d.ts` emitted by `tsc`, `types` field set)
- ✅ README documents installation and usage

---

### Findings

#### 🟡 Important: publish-npm is blocked if any Tauri matrix build fails

**`.github/workflows/release.yml` lines 204**

```yaml
needs: build-tauri
```

`build-tauri` is a 3-runner matrix job (macOS ARM, macOS Intel, Windows). GitHub Actions requires **all** matrix instances to succeed before a downstream `needs:` job is unblocked. This means a flaky Windows build or a macOS Intel signing failure will silently prevent the npm package from publishing — even though the npm publish has zero dependency on Tauri artifacts.

The design couples two completely independent release operations. The `publish-npm` job does not consume any artifact from `build-tauri`; the only intent of the `needs:` is sequencing (publish after binaries). But that sequencing is not actually required for correctness — npm publishing is independent of binary builds.

**Recommendation:** Remove `needs: build-tauri`. The `if: github.event_name == 'release'` guard already ensures this only runs on release events. If ordering is desired for some operational reason (e.g., release notes completeness), replace with `needs: build-tauri` only if the matrix jobs are made `continue-on-error: true` — but that has its own trade-offs.

```yaml
publish-npm:
  name: Publish @nodespaceai/agent-tools
  runs-on: ubuntu-latest
  # Removed: needs: build-tauri (no artifact dependency; don't let binary failures block npm)
  if: github.event_name == 'release'
```

**Engineering Principle:** Jobs should only declare dependencies on other jobs when they consume artifacts or require state produced by those jobs (data dependency). Using `needs:` purely for sequencing of independent operations creates false coupling and reduces system resilience.

---

#### 🟡 Important: No version guard — re-publishing `0.1.0` on every subsequent release will fail silently or error

**`.github/workflows/release.yml` lines 229–231**

```yaml
- name: Publish @nodespaceai/agent-tools
  run: npm publish --access public
```

The `version` field in `package.json` is `0.1.0` and is not updated as part of this workflow. On the second release tag, `npm publish` will attempt to publish `0.1.0` again. npm will reject this with `E403 You cannot publish over the previously published versions`. The step will fail, and the release workflow will fail.

There is no `--tag` flag for pre-release semantics, no `npm version` bump step, and no reference to the GitHub release tag as the package version.

**Recommendation:** Wire the GitHub release tag into the package version before publishing. The release tag is available as `github.event.release.tag_name`.

```yaml
- name: Set package version from release tag
  run: |
    TAG="${{ github.event.release.tag_name }}"
    VERSION="${TAG#v}"  # strip leading 'v' if present (v0.2.0 → 0.2.0)
    npm version "$VERSION" --no-git-tag-version
  working-directory: packages/agent-tools

- name: Publish @nodespaceai/agent-tools
  run: npm publish --access public
  working-directory: packages/agent-tools
  env:
    NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```

This also enforces that the npm version always matches the GitHub release tag — a correctness property that is currently unenforceable.

**Engineering Principle:** Idempotency and repeatability of CI pipelines. A workflow that succeeds exactly once but fails on every subsequent invocation is not a reliable release mechanism.

---

#### 🟢 Suggestion: `src/` in `files` exposes uncompiled TypeScript source to npm consumers

**`packages/agent-tools/package.json` lines 15–19**

```json
"files": [
  "dist/",
  "src/",
  "shims/",
  "*.md"
]
```

Including `src/` ships the raw `.ts` source files alongside the compiled output. This is not required for consumers — `dist/` contains the compiled JS and declaration files, and `types` points to `dist/index.d.ts`. Shipping `src/` increases package weight and can expose internal implementation details before they are stabilized.

This matches the issue's technical spec exactly (`"files": ["dist/", "src/", "shims/", "*.md"]`), so this is not a spec deviation — it is a note that the spec choice has a trade-off. YAGNI would suggest omitting `src/` until there is a concrete use case (e.g., source maps for debugging in IDEs that support it).

**Nit:** This is low-risk for a 0.1.0 package with no external consumers yet. Fine to leave as-is if matching the issue spec is the priority.

---

#### 🟢 Suggestion: `preinstall` hook will fire when CI runs `bun install` but not when consumers install the package — verify this is intentional

**`packages/agent-tools/package.json` line 20**

```json
"preinstall": "node ../../enforce-bun.js"
```

The `../../enforce-bun.js` path is relative to the monorepo root. When a consumer does `npm install @nodespaceai/agent-tools`, the `preinstall` lifecycle script will run in the consumer's context where `../../enforce-bun.js` does not exist. This will produce a confusing error for anyone installing the package with npm or Bun from outside the monorepo.

The `enforce-bun.js` script exits 1 only when `npm_execpath` contains `npm`/`yarn`/`pnpm` — a consumer installing with Bun will pass, but the path resolution failure happens before that check.

**Recommendation:** Add `"preinstall": ""` to the `publishConfig` to strip the hook on publish, or use the `--ignore-scripts` flag in CI. The cleanest fix is to configure npm's `publishConfig` to override scripts:

```json
"publishConfig": {
  "access": "public",
  "scripts": {
    "preinstall": ""
  }
}
```

However, `publishConfig.scripts` is not a standard npm field — it is not honored. The correct approach is a `.npmignore` or using `pkg.scripts` only in workspace context. The pragmatic fix: remove `preinstall` from the published package by adding a `prepare`/`prepublishOnly` step that removes it, or simply document that external consumers should use `--ignore-scripts`.

**Engineering Principle:** Published packages must be installable in isolation, outside the monorepo context. Lifecycle scripts with monorepo-relative paths are a portability violation.

---

### Nitpicks

**Nit:** `packages/agent-tools/README.md` line 1 — The README is well-structured. One minor omission: `createNode` signature documents `parentId?` as optional but does not document the behavior when omitted (creates a root-level node? errors?). Worth a one-liner clarification for external consumers.

**Nit:** `.github/workflows/release.yml` — `actions/checkout@v6.0.2` is used consistently across the file, which is good. No version drift introduced.

---

### Summary

| Finding | Severity | File | Blocks Merge? |
|---|---|---|---|
| `needs: build-tauri` couples independent jobs | 🟡 Important | `release.yml:204` | No, but should fix |
| No version bump — second publish will fail | 🟡 Important | `release.yml:229` | Yes — workflow breaks on release 2+ |
| `src/` in `files` ships raw TypeScript | 🟢 Suggestion | `package.json:16` | No |
| `preinstall` path breaks external installs | 🟢 Suggestion | `package.json:20` | No (Bun consumers unaffected today) |

**Recommendation: REQUEST CHANGES**

The version bump issue is a hard blocker for a working CI release pipeline — the workflow will fail on every release after the first. The `needs:` coupling is a reliability concern that will surface in practice when Tauri builds are flaky (they often are, given signing and model download steps). Both can be fixed with small, targeted changes to `release.yml`.

malibio:
## Review Addressed

All 🟡 Important recommendations have been fixed. One 🟢 Suggestion skipped.

### ✅ Addressed

**Remove `needs: build-tauri`** — `publish-npm` job is now fully independent. A Tauri matrix failure no longer blocks npm publish.

**Version from release tag** — Added a "Set version from release tag" step that runs `npm version ${TAG#v} --no-git-tag-version` before build/publish. Prevents E403 on second+ releases.

**`preinstall` safe outside monorepo** — Changed to `[ -f ../../enforce-bun.js ] && node ../../enforce-bun.js || true`. File-existence check means external consumers get exit 0; the guard still fires normally inside the monorepo.

### ⏭️ Skipped

**Remove `src/` from `files`** — Issue spec explicitly requires `src/` in the files list. Keeping source is also common practice for TypeScript libraries (enables source-map debugging, Go-to-definition in editors). Weight tradeoff is acceptable.

**`createNode` parentId docs** — Nitpick; not blocking. The function signature already communicates optionality via `?`.

PR #1189, opened 2026-05-19
Add Homebrew tap and cask auto-update CI (closes #1186)

malibio:
## Summary

- Creates `github.com/NodeSpaceAI/homebrew-nodespace` with `Casks/nodespace.rb` supporting arm64 and x64 DMGs via `on_arm`/`on_intel` blocks
- Adds `update-homebrew-cask` job to `release.yml` that runs after `build-tauri` on each release, computes SHA256s, and opens a PR to the tap repo automatically
- Cask includes `zap` stanza to remove LaunchAgent, bins, and logs but preserves `~/.nodespace/database/` user data

## Setup required

Add a `HOMEBREW_TAP_TOKEN` secret to `NodeSpaceAI/nodespace-core` — a GitHub PAT with `contents:write` and `pull_requests:write` scope on `NodeSpaceAI/homebrew-nodespace`. The release CI uses this to push branches and open PRs to the tap.

## Test plan
- [ ] Verify `brew tap nodespaceai/nodespace && brew install --cask nodespace` installs the app
- [ ] Verify `brew audit --cask nodespace` passes (requires signed DMG)
- [ ] Trigger a test release to confirm the `update-homebrew-cask` job opens a PR to `homebrew-nodespace`
- [ ] Confirm `brew uninstall --zap --cask nodespace` removes agent plist and bins but not `~/.nodespace/database/`

Closes #1186

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -197,3 +197,91 @@ jobs:
             target/${{ matrix.rust_targets || '' }}/release/bundle/
             target/release/bundle/
           if-no-files-found: warn
+
+  update-homebrew-cask:
+    name: Update Homebrew cask
+    runs-on: macos-latest
+    needs: build-tauri
+    if: github.event_name == 'release'
+
+    steps:
+      - name: Compute version
+        id: version
+        run: echo "version=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+
+      - name: Download macOS DMGs and compute SHA256
+        id: sha
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          gh release download "${{ github.ref_name }}" \
+            --repo NodeSpaceAI/nodespace-core \
+            --pattern "NodeSpace_${VERSION}_aarch64.dmg" \
+            --pattern "NodeSpace_${VERSION}_x64.dmg"
+
+          ARM_SHA=$(shasum -a 256 "NodeSpace_${VERSION}_aarch64.dmg" | cut -d' ' -f1)
+          X64_SHA=$(shasum -a 256 "NodeSpace_${VERSION}_x64.dmg" | cut -d' ' -f1)
+
+          echo "arm_sha=${ARM_SHA}" >> "$GITHUB_OUTPUT"
+          echo "x64_sha=${X64_SHA}" >> "$GITHUB_OUTPUT"
+          echo "ARM SHA256: ${ARM_SHA}"
+          echo "x64 SHA256: ${X64_SHA}"
+
+      - name: Clone homebrew-nodespace tap
+        env:
+          HOMEBREW_TAP_TOKEN: ${{ secrets.HOMEBREW_TAP_TOKEN }}
+        run: |
+          git clone "https://x-access-token:${HOMEBREW_TAP_TOKEN}@github.com/NodeSpaceAI/homebrew-nodespace.git" tap
+
+      - name: Update cask formula
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+          ARM_SHA: ${{ steps.sha.outputs.arm_sha }}
+          X64_SHA: ${{ steps.sha.outputs.x64_sha }}
+        run: |
+          CASK="tap/Casks/nodespace.rb"
+
+          # Update version
+          sed -i '' "s/^  version \".*\"/  version \"${VERSION}\"/" "$CASK"
+
+          # Update arm SHA256 (line after "on_arm do")
+          awk -v sha="$ARM_SHA" '
+            /on_arm do/ { print; getline; sub(/sha256 ".*"/, "sha256 \"" sha "\""); print; next }
+            { print }
+          ' "$CASK" > "$CASK.tmp" && mv "$CASK.tmp" "$CASK"
+
+          # Update intel SHA256 (line after "on_intel do")
+          awk -v sha="$X64_SHA" '
+            /on_intel do/ { print; getline; sub(/sha256 ".*"/, "sha256 \"" sha "\""); print; next }
+            { print }
+          ' "$CASK" > "$CASK.tmp" && mv "$CASK.tmp" "$CASK"
+
+          echo "Updated cask:"
+          cat "$CASK"
+
+      - name: Open PR to homebrew-nodespace
+        env:
+          HOMEBREW_TAP_TOKEN: ${{ secrets.HOMEBREW_TAP_TOKEN }}
+          VERSION: ${{ steps.version.outputs.version }}
+          GH_TOKEN: ${{ secrets.HOMEBREW_TAP_TOKEN }}
+        run: |
+          BRANCH="update-cask-v${VERSION}"
+
+          cd tap
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git config user.name "github-actions[bot]"
+          git checkout -b "$BRANCH"
+          git add Casks/nodespace.rb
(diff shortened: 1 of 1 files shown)


malibio:

---

## Requirements Check

|---|---|---|
| `homebrew-nodespace` repo with `Casks/nodespace.rb` | N/A (separate repo) | Not reviewable in this diff — must be verified separately |
| Cask formula with correct SHA256 for signed DMG | Partial | CI script correctly computes SHA256; cask template in issue differs from implemented approach (see findings) |
| `brew tap && brew install` works on arm64 and x86_64 | Partial | Two separate DMGs are handled correctly; confirmed against actual release asset names |
| `zap` stanza preserves `~/.nodespace/database/` | N/A (separate repo) | Defined in cask file not in this diff |
| CI auto-updates cask SHA256 + version on each release | ✅ | Implemented |
| Gated to release event only | ✅ | `if: github.event_name == 'release'` |
| `needs: build-tauri` | ✅ (with caveat) | See Critical finding #1 |

---

## Code Review Findings

### Critical Issues

**[Critical/Blocker] `.github/workflows/release.yml` lines 201-203 — `needs: build-tauri` dependency semantics with a matrix job**

`build-tauri` is a matrix job with three legs (`macos-arm`, `macos-intel`, `windows`). When using `needs` on a matrix job, GitHub Actions waits for **all** matrix legs to succeed. If the `windows` leg fails (or is skipped via `workflow_dispatch` platform filter), the `update-homebrew-cask` job will also be blocked or skipped — even if both macOS legs succeeded and produced the required DMGs.

This is particularly sharp here because `workflow_dispatch` builds can run with a single platform selected. In that case, the non-selected legs skip rather than succeed, and a `needs` dependency against a matrix job with skipped legs resolves as failed, which will block the cask update entirely.

For a `release` event all legs run, so day-to-day this won't bite you — but you need a belt-and-suspenders guard. The safest fix is to add an explicit event guard at the step level and document the known limitation, or restructure with a dedicated `build-macos` job that the cask update depends on. Minimum acceptable fix: add a comment documenting the behavior and confirm the skip propagation is acceptable.

**[Critical/Blocker] `.github/workflows/release.yml` line 222 — `sed -i ''` is macOS-specific and will silently corrupt the file on GNU/Linux**

The runner is `macos-latest` so this won't fail today, but `sed -i ''` (BSD sed in-place with no backup suffix) is not portable. This is fragile coupling to the runner. More importantly, the `sed` pattern updates the **first** matching line globally (`s/^  version/`). If the cask file ever has more than one version string, only the first match changes. The `awk` approach already used for SHA256 is more robust — consider using awk for the version update too, or at minimum add a comment noting the BSD sed dependency.

### Suggested Improvements

**[Improvement] `.github/workflows/release.yml` lines 218-219 — DMG download pattern assumes version tag has no `v` prefix but `GITHUB_REF_NAME` strip logic is correct**

The version is correctly stripped (`${GITHUB_REF_NAME#v}`) and the real release assets are named `NodeSpace_0.1.6_aarch64.dmg` (confirmed). The patterns `NodeSpace_${VERSION}_aarch64.dmg` and `NodeSpace_${VERSION}_x64.dmg` match exactly. No bug here, but document this naming dependency in a comment so a future rename of the Tauri `productName` doesn't silently break the job by producing unfindable assets.

**[Improvement] `.github/workflows/release.yml` lines 246-251 — token exposure via environment variable redundancy**

`HOMEBREW_TAP_TOKEN` is set in both the clone step (line ~232) and the PR creation step (line ~246). In the PR step, both `HOMEBREW_TAP_TOKEN` and `GH_TOKEN` are set to the same secret. The `gh` CLI uses `GH_TOKEN`; the duplicate `HOMEBREW_TAP_TOKEN` in that step env is unused and adds noise. Remove it from the final step env block to reduce confusion about which token the `gh pr create` command uses.

**[Improvement] `.github/workflows/release.yml` lines 241-243 — no idempotency guard on branch creation**

If a release is re-published or the workflow is re-run, `git checkout -b "$BRANCH"` will fail because `update-cask-v${VERSION}` already exists on the remote. Add `git push origin "$BRANCH" --force` or a pre-flight `gh pr list --head "$BRANCH"` check to skip gracefully if the PR already exists. The latter is safer: detect existing PR and exit 0 rather than force-pushing.

Example:
```yaml
# Skip if PR for this version already exists
if gh pr list --repo NodeSpaceAI/homebrew-nodespace --head "$BRANCH" --json number --jq '.[0].number' | grep -q .; then
  echo "PR for $BRANCH already exists, skipping"
  exit 0
fi
```

**[Improvement] Issue spec vs implementation — cask formula divergence**

The issue spec shows a `url` using a universal DMG (`NodeSpace-#{version}-universal.dmg`) and a single `sha256`. The implemented CI creates two separate DMGs (`aarch64` / `x64`) using `on_arm`/`on_intel` blocks, which is architecturally more correct for Homebrew and matches what Tauri actually produces (two separate arch builds, not a universal binary). However, the cask file in the separate `homebrew-nodespace` repo must use this same `on_arm`/`on_intel` structure — verify it does not use the spec's universal URL.

### Nitpicks

**Nit: `.github/workflows/release.yml` line 208 — `echo "ARM SHA256: ${ARM_SHA}"` / `x64 SHA256` debug output**

The debug echo lines are fine for now but will print raw SHA256 values in public CI logs. Not a security issue (SHA256 of a public DMG is public), but worth noting they produce noise in the log output.

**Nit: `.github/workflows/release.yml` lines 253-256 — PR body interpolation**

The `--body` heredoc uses `${{ github.ref_name }}` inside a `run:` shell step. This correctly expands at the GitHub Actions expression layer before the shell runs. No bug, but the mixed interpolation styles (`${{ }}` for Actions expressions, `$VAR` for shell) inside the heredoc can be confusing for future editors — a brief comment would help.

---

## Summary

The core approach is sound: wait for the full build, download the real artifacts, compute SHA256, and open a PR to the tap repo. The DMG asset naming matches actual release output. The main actionable issues are:

1. The `needs: build-tauri` matrix dependency blocks on the Windows leg — verify this is acceptable for release events, where all legs always run.
2. `sed -i ''` is BSD-only and fragile; already superseded by the awk approach used for SHA. Align the version update to use awk as well.
3. The cask file in `homebrew-nodespace` is not reviewable here — it must be separately confirmed to use `on_arm`/`on_intel` blocks matching this CI output, not the universal-DMG approach in the issue spec.

**Recommendation: REQUEST CHANGES** — the `needs` matrix semantics and `sed -i ''` fragility should be addressed before merge.

---
*Reviewed by Claude Code (Pragmatic Quality Framework)*

malibio (commented):
Requesting changes — see detailed findings in the comment above. Blocking issues: (1) needs: build-tauri matrix dependency semantics on the Windows leg, (2) sed -i '' BSD-only fragility. Full analysis posted.

malibio:
## Review feedback addressed

**✅ Addressed (4 items)**

- **`sed -i ''` portability** — replaced with `awk` for the version update, consistent with the SHA256 update pattern already in the file. No BSD vs GNU sed ambiguity.
- **Idempotency guard** — added `gh pr view "$BRANCH"` check before creating the PR; workflow re-runs for the same release version will exit cleanly instead of erroring.
- **Redundant `HOMEBREW_TAP_TOKEN` env var** — removed. The step only needs `GH_TOKEN`; the extra binding was confusing.
- **Documentation** — added comment on `needs: build-tauri` matrix behavior, added comment linking the DMG filename pattern to `tauri.conf.json productName`, removed noisy SHA256 debug echo lines.

**⏭️ Skipped (1 item)**

- **Extract `build-macos` job** — scope creep for this PR. The `if: github.event_name == 'release'` guard already ensures `update-homebrew-cask` never runs on `workflow_dispatch` builds where the Windows leg might be skipped. The behavior is now documented in a comment on the `needs:` line.

PR #1190, opened 2026-05-19
Add @nodespaceai/mcp: MCP server exposing NodeSpace as AI tools

malibio:
## Summary

- Implements `packages/mcp` — a new npm package `@nodespaceai/mcp` that speaks the MCP stdio protocol and wraps `nodespace` CLI commands as AI tools
- Exposes six tools: `nodespace_create_node`, `nodespace_get_node`, `nodespace_search`, `nodespace_list_nodes`, `nodespace_update_node`, `nodespace_delete_node`
- `npx @nodespaceai/mcp install` writes to `claude_desktop_config.json` (safe merge); `npx @nodespaceai/mcp uninstall` removes the entry
- Adds `.github/workflows/publish-mcp.yml` CI/CD workflow that publishes on release tag or via manual workflow dispatch with dry-run option
- Adds `test:mcp` and updates `quality:fix` / `test:all` in root `package.json`

Closes #1185

## Test plan

- [ ] `bun run test:mcp` — 8 tests pass covering CLI argument construction and error handling
- [ ] `bun run build` from `packages/mcp` — TypeScript compiles cleanly
- [ ] `npx @nodespaceai/mcp install` writes correct entry to `claude_desktop_config.json`
- [ ] `npx @nodespaceai/mcp uninstall` removes the entry
- [ ] With `nodespace` on `$PATH` and daemon running: `npx @nodespaceai/mcp` starts stdio server and accepts tool calls


malibio:
### Code Review Summary

This PR delivers a clean, thin MCP stdio server wrapping the `nodespace` CLI. The architecture is correct — `execFile` (not `exec`) for shell injection safety, typed error class, proper MCP SDK usage, and good test coverage for the happy path and error branches. It is a net improvement and the foundation is solid.

However, there are two issues that need to be resolved before merge: a correctness bug in `listNodes` and a missing version-stamping step in the CI workflow that will cause publish failures after the first release.

---

### Findings

#### Critical Issues

**[Critical/Blocker] `listNodes` uses semantic search as a type filter — incorrect behavior**

`/packages/mcp/src/tools.ts`, line 65-68:

```ts
export async function listNodes(nodeType: string, limit?: number): Promise<string> {
  const args = ['search', nodeType, '--limit', String(limit ?? 50)];
  return runCLI(args);
}
```

This invokes `nodespace --json search <nodeType>` which runs a *semantic search* with the node type name as the free-text query string. It does not filter by node type. The underlying `SearchRequest` protobuf has a `node_types: repeated string` field explicitly designed for this purpose, and the CLI `SearchArgs` passes `node_types: vec![]` (empty — all types). The `nodespace search` subcommand does not expose a `--type` flag.

The practical result: `nodespace_list_nodes("task")` returns whatever nodes are semantically similar to the word "task", not all task nodes. This is functionally incorrect and semantically misleading to any AI agent using this tool.

The fix is one of:
1. Add a `--type` flag to the `nodespace search` CLI subcommand (wires into the existing `node_types` field on `SearchRequest`) and update `listNodes` to pass it.
2. Use `nodespace node children <root-id>` + client-side type filtering, though this only covers children of a known root.

Option 1 is the right fix — it's a one-liner addition to `SearchArgs` in `packages/cli/src/commands/search.rs`. The test for `listNodes` is also absent from `tools.test.ts`, which is how this slipped through.

---

**[Critical/Blocker] Publish workflow missing version-stamp step — will fail on second release**

`.github/workflows/publish-mcp.yml` publishes whatever version is in `packages/mcp/package.json` (currently `0.1.0`). The existing `publish-npm` job in `release.yml` (for `@nodespaceai/agent-tools`) learned this lesson the hard way — it has an explicit "Set version from release tag" step:

```yaml
- name: Set version from release tag
  run: |
    TAG="${{ github.event.release.tag_name }}"
    npm version "${TAG#v}" --no-git-tag-version
  working-directory: packages/agent-tools
```

Without this step, publishing a second release will fail with `E403 You cannot publish over the previously published versions`. Add the same step to `publish-mcp.yml` between "Build" and "Publish to npm".

---

#### Suggested Improvements

**[Improvement] `install.ts` is macOS-only with no runtime guard**

`CLAUDE_CONFIG_PATH` is hardcoded to `~/Library/Application Support/Claude/...` which only exists on macOS. Running `npx @nodespaceai/mcp install` on Linux or Windows will silently write to a nonexistent path (the `mkdirSync` will create it, but it won't be the right Claude config location). Since the NodeSpace desktop app is macOS-only for now this is low risk, but a runtime check with a clear error — `if (process.platform !== 'darwin') { console.error('install is only supported on macOS'); process.exit(1); }` — would prevent confusing behavior for developers on other platforms.

**[Improvement] Test coverage gap for `listNodes`**

`tools.test.ts` tests `createNode`, `getNode`, `searchNodes`, `updateNode`, `deleteNode`, and two error branches, but has no test for `listNodes`. Given that this is the function with the correctness issue above, a test verifying the exact CLI args passed would have caught the bug. After fixing `listNodes`, add a test case alongside the existing `searchNodes` test.

**[Improvement] `quality:fix` only type-checks, no linting**

`packages/mcp/package.json`, line 26:
```json
"quality:fix": "bunx tsc --noEmit"
```

This runs the TypeScript compiler for type checking only. The other packages (`desktop-app`, `agent-tools`) include an ESLint/prettier step in their `quality:fix`. For now this is acceptable since the package is small, but it creates an inconsistency in the workspace-level `quality:fix` command. Consider adding `eslint` or `biome` consistent with the rest of the workspace.

---

#### Nitpicks

- **Nit** `packages/mcp/src/server.ts`, line 17-21: The install/uninstall dispatch via `process.argv` mutation and dynamic `import()` is a workable hack but somewhat fragile. A cleaner long-term approach would be a separate `bin` entry in `package.json` for install (`"nodespace-mcp-install": "dist/install.js"`). Acceptable for v0.1 — flag for future cleanup.

- **Nit** `packages/mcp/package.json`, `bin` field: The binary is named `nodespace-mcp` but `npx @nodespaceai/mcp` (no binary name specified) will invoke it since it's the only binary. This works, but documenting the binary name in README would help users who install globally.

- **Nit** `packages/mcp/tsconfig.json`, line 20: Test files are excluded from compilation (`"exclude": ["src/tests/**/*"]`), which is correct. Worth confirming vitest can still resolve the source imports — it can since vitest uses its own resolution, not `tsc`, but it's worth noting for future contributors.

---

Against issue #1185:

- [x] MCP server speaks stdio protocol compatible with Claude desktop and Cline
- [x] All six tools implemented as CLI wrappers (no direct gRPC)
- [x] `npx @nodespaceai/mcp` starts the server
- [x] `npx @nodespaceai/mcp install` writes correct entry to `claude_desktop_config.json` (safe merge with `??=`)
- [x] `npx @nodespaceai/mcp uninstall` removes the entry
- [x] Clear error message if `nodespace` CLI not found on `$PATH`
- [x] CI/CD workflow publishes on release tag
- [x] `bun run quality:fix` passes
- [ ] `nodespace_list_nodes` correctly filters by type (currently does a semantic search, not a type filter)
- [ ] Publish workflow stamps version from release tag (E403 on second publish)
- [ ] MCP marketplace submissions (out of scope for this review, noted as a checklist item)

**Verdict: Request changes.** Fix the `listNodes` CLI invocation and add the version-stamp step to the publish workflow. Both are small, targeted fixes. The rest of the implementation is clean and ready to ship.

malibio:
## Review Addressed — commit 1c0ebb80

### ✅ Changes made

**🔴 listNodes semantic bug** (`packages/mcp/src/tools.ts`, `packages/cli/src/commands/search.rs`)
- Added `--type <TYPE>` flag to the CLI `search` subcommand, wiring it into `SearchRequest.node_types`
- Made `query` optional (defaults to `""`) so type-only listing works without a semantic query
- Updated `listNodes` to invoke `nodespace search --type <nodeType> --limit <n>` instead of the broken semantic search
- Added 2 tests asserting the correct args: one for custom limit, one for the 50-item default

**🔴 Publish workflow missing version-stamp** (`.github/workflows/publish-mcp.yml`)
- Added a "Set version from release tag" step that runs `npm version "${GITHUB_REF_NAME#v}" --no-git-tag-version` before install/build on release events

**🟡 Platform guard in install.ts** (`packages/mcp/src/install.ts`)
- Added `process.platform !== 'darwin'` check that exits with a clear error and prints the manual config snippet for non-macOS users

### ⏭️ Skipped

**🟢 Separate bin entry for install** — the current argv dispatch in `server.ts` works correctly and two bin entries add complexity for no practical benefit in a thin wrapper package.

Tests: 10/10 ✅

malibio:
## Re-Review — commit 1c0ebb80

Reviewing only the delta since the original review (commit `94e03b8c`). Five files changed.

---

### Verdict: Approve with one minor note

All three blockers/improvements from the original review are correctly resolved. The implementation is accurate and the tests pass (10/10). One minor observation on the workflow version-stamp approach worth noting for future awareness.

---

### Original Issues — Resolution Status

**[Critical/Blocker] listNodes semantic search bug — FIXED correctly**

`packages/mcp/src/tools.ts` line 66 now emits `['search', '--type', nodeType, '--limit', ...]` instead of passing the type name as a free-text query. The fix is end-to-end correct: `SearchArgs` in `packages/cli/src/commands/search.rs` gains `node_types: Vec<String>` wired directly into `SearchRequest.node_types`, and the daemon's `search_nodes` handler (line 362-388 in `packages/daemon/src/services/node_service.rs`) applies the `node_types` filter independently of the semantic embedding step. The `semantic: true` hardcode is a non-issue — structured query mode is not yet implemented server-side and `node_types` filtering applies on top of semantic results as intended.

**[Critical/Blocker] Publish workflow missing version-stamp — FIXED**

`.github/workflows/publish-mcp.yml` now has the version-stamp step guarded by `if: github.event_name == 'release'`, which correctly skips it on `workflow_dispatch` runs (where `GITHUB_REF_NAME` would be a branch name, not a semver tag).

Minor note: the existing `release.yml` version-stamp steps use `${{ github.event.release.tag_name }}` (the release event payload field) rather than `$GITHUB_REF_NAME`. Both resolve to the same value when triggered by a release event, so this is functionally equivalent. The `GITHUB_REF_NAME` form is slightly more idiomatic for shell-level access and is not wrong. No action required.

**[Improvement] install.ts platform guard — FIXED correctly**

`packages/mcp/src/install.ts` lines 63-76 now check `process.platform !== 'darwin'` before executing either `install` or `uninstall`, exits with code 1, and prints the manual JSON config snippet for non-macOS users. The guard is applied to both commands (not just `install`), which is the right call — `uninstall` on a non-macOS path would also be a no-op or misleading.

**[Improvement] listNodes test coverage gap — FIXED**

`packages/mcp/src/tests/tools.test.ts` now has two tests: one verifying `--type task --limit 20` for an explicit limit, and one verifying `--limit 50` for the default. Both assert the exact `execFile` call signature, which is the right level of specificity for a CLI-wrapper tool.

**[Nit] Separate bin entry for install — Skipped (valid call)**

Accepted as-is per original review acknowledgment.

---

### New Issues in This Delta

None. The diff is tightly scoped to the four reported items with no unrelated changes introduced.

---

### Test Results

All 10 tests pass. No regressions.

PR #1191, opened 2026-05-19
Add @nodespaceai/skill npm package (closes #1184)

malibio:
## Summary

- Introduces `packages/skill`: a distributable npm package that installs the NodeSpace skill (SKILL.md + CLI reference) into PTY agent directories via `npx @nodespaceai/skill install`
- Implements per-agent detection and install/uninstall for Claude Code, Codex, Gemini CLI, and OpenCode
- Adds CI/CD publish step to `.github/workflows/release.yml` to publish on release tag

## Test plan

- [ ] Run `bun run test:skill` — 12 tests should pass
- [ ] Run `bun run quality:fix` — no lint/type errors
- [ ] Verify `npx @nodespaceai/skill install` detects present agents and installs `SKILL.md` to correct paths
- [ ] Verify `npx @nodespaceai/skill uninstall` cleanly removes installed files
- [ ] Verify `npx @nodespaceai/skill --help` shows correct usage
- [ ] Confirm CI release workflow publishes `@nodespaceai/skill` on release tag


malibio:
## Code Review Summary

This is a clean, well-scoped package introduction. The core installer logic is simple and correct, the test coverage is solid, and the public API surface is appropriately minimal. There are two issues that need addressing before merge: a broken `preinstall` script that will break end-user installs, and an empty `shims/` directory that leaves an acceptance criterion unmet. The rest is polish.

---

## Requirements Check

- ✅ \`@nodespaceai/skill\` package structure ready for npm publish (under \`@nodespaceai\` org, public access configured)
- ✅ Package includes \`SKILL.md\` with complete CLI usage: create/get/search/list/update/delete + examples
- ❌ Package includes per-agent shims for Claude Code, Codex, Gemini CLI, OpenCode — \`packages/skill/shims/\` directory is **empty**. The issue spec and \`package.json#files\` both include \`shims/\`, but no shim files exist. \`agents.ts\` only references \`SKILL.md\` in the \`shims\` arrays, so per-agent shims from #1123 were never wired in.
- ✅ \`npx @nodespaceai/skill install\` detects present agents and installs to correct locations
- ✅ \`npx @nodespaceai/skill uninstall\` cleanly removes installed files
- ✅ Assumes \`nodespace\` CLI is already on \`$PATH\`
- ✅ CI/CD release workflow publishes on release tag
- ⬜ Submitted to skills.sh registry — out of scope for this PR (manual step post-publish)
- ✅ \`bun run quality:fix\` passes (12/12 tests pass)

---

## Findings

### [Critical/Blocker]

**`packages/skill/package.json` — broken `preinstall` script will fail on end-user installs**

```json
"preinstall": "node ../../enforce-bun.js",
```

The `preinstall` lifecycle hook runs when any user executes `npm install @nodespaceai/skill` or `npx @nodespaceai/skill install`. In that context, `../../enforce-bun.js` does not exist — it's an internal monorepo file. This will cause the installation to fail with `ENOENT` for every end user.

The sibling package `@nodespaceai/agent-tools` has the correct guard:
```json
"preinstall": "[ -f ../../enforce-bun.js ] && node ../../enforce-bun.js || true",
```

Apply the same guard here.

---

**`packages/skill/shims/` is empty — per-agent shims acceptance criterion is unmet**

The issue acceptance criteria explicitly requires: *"Package includes per-agent shims for Claude Code, Codex, Gemini CLI, OpenCode."* The `shims/` directory exists but contains no files. The `agents.ts` config only lists `SKILL.md` in each agent's `shims` array, so no per-agent shims will ever be installed. Either wire in the shims from #1123 or update the acceptance criteria if the decision was to defer shims to a later issue.

---

### [Improvement]

**`packages/skill/src/installer.ts` lines 7–11 — PACKAGE_ROOT fallback resolves to wrong directory**

The comment says the fallback handles "running directly from src/", but the math is wrong:

```ts
const PACKAGE_ROOT = __dirname.endsWith('/dist')
  ? join(__dirname, '..')          // dist/ → packages/skill/  ✓
  : join(__dirname, '..', '..');   // src/  → packages/        ✗ (should be packages/skill/)
```

If the binary is ever invoked directly from `src/` (e.g., during local development with `bun run src/install.ts`), `PACKAGE_ROOT` resolves to `packages/` rather than `packages/skill/`, and `SKILL.md` won't be found. The correct fallback for `src/` is `join(__dirname, '..')`. Since `packageRoot` is overridable in tests this doesn't affect correctness in CI, but the dead-code path is misleading. Fix or remove the comment.

---

**`packages/skill/src/installer.ts` line 27 — `mkdirSync` runs before checking if source files exist**

```ts
mkdirSync(config.installDir, { recursive: true });   // dir created unconditionally
// ...
if (existsSync(src)) {                                // source check happens after
  copyFileSync(src, dest);
```

If the source `SKILL.md` is missing (malformed package, corrupt install), the agent's `skills/nodespace/` directory gets created as an empty directory. This is a minor issue because the install also exits 0 in that case (see next finding), but an empty orphan directory in the user's agent config is surprising behavior.

---

**`packages/skill/src/install.ts` line 59 — partial install failure exits 0 without indicating an error**

When `result.installed.length === 0` (agent directory was detected but no files were installed), the CLI prints a warning but exits 0:

```ts
console.log(`⚠ ${result.agent}: no files to install`);
// no process.exit(1)
```

Combined with the `mkdirSync` issue above, a user could see a \`⚠\` warning, an empty skills directory, and still get exit code 0 — making it difficult to detect the problem in automation. Consider exiting 1 if any detected agent ends up with zero installed files.

---

### [Nit]

**`packages/skill/SKILL.md` line 27 — duplicate `--type` flag in example**

```bash
nodespace node create --type note --content "Meeting notes" --type note
```

`--type note` appears twice. This is the agent-readable reference document, so accuracy matters for correctness of AI-generated commands.

---

**`packages/skill/package.json` — `"src"` listed in `files` may publish uncompiled source unnecessarily**

Publishing `src/` alongside `dist/` is fine for transparency, but it doubles the package size and exposes test files (`src/tests/`). Since `tsconfig.json` already excludes `src/tests/**/*` from compilation, consider removing `"src"` from `files` or narrowing it to `"src/**/*.ts"` excluding tests. Low priority given the small package size.

---

**`.github/workflows/release.yml` — `publish-npm` now requires all Tauri platform builds to pass**

Adding `needs: build-tauri` couples npm publishing to the full desktop build matrix (macOS arm64, macOS x86_64, Windows). If a platform-specific build fails (e.g., a Windows code-signing issue), npm packages won't publish even though they are platform-independent. This is worth a conscious decision: was this intentional sequencing (e.g., to stamp a version only after full release artifacts exist), or an accidental coupling? If incidental, consider removing the `needs` dependency on `publish-npm` or only depending on a single fast job.

malibio (commented):
Review posted. See full findings in PR comment above.

PR #1192, opened 2026-05-19
Add curl-based shell installer for headless NodeSpace install (closes #1182)

malibio:
## Summary
- `scripts/install.sh`: POSIX sh installer — detects platform/arch, downloads nodespaced + nodespace from GitHub releases, verifies SHA256, installs to `~/.nodespace/bin/`, registers launchd (macOS) or systemd user service (Linux), prompts for PATH/MCP/skill setup
- `scripts/uninstall.sh`: POSIX sh uninstaller — stops daemon, removes service registration, binaries, socket, and Claude Code skill
- `packages/cli`: adds `nodespace uninstall` subcommand with interactive MCP config cleanup
- `.github/workflows/release.yml`: adds `build-headless` 4-platform matrix job and `upload-checksums` job

## Test plan
- [ ] Run `scripts/install.sh` on macOS arm64 — verify binaries land in `~/.nodespace/bin/`, launchd plist written, daemon starts
- [ ] Run installer a second time — verify idempotency (no duplicate PATH entries, MCP entry not duplicated)
- [ ] Run `nodespace uninstall` — verify daemon stopped, binaries removed, MCP entry cleaned from claude_desktop_config.json, database preserved
- [ ] Run `scripts/uninstall.sh` directly — verify equivalent behavior to CLI command
- [ ] Manually trigger `build-headless` workflow via workflow_dispatch for each platform
- [ ] Verify SHA256SUMS uploaded to release after `upload-checksums` job completes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -286,6 +286,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'release'
     needs: build-tauri-macos-arm
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6.0.2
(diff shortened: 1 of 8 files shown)


malibio:
### Code Review Summary

This PR delivers the full headless install story for NodeSpace: a POSIX shell installer, a paired uninstaller, a `nodespace uninstall` CLI subcommand, and a 4-platform CI job that uploads standalone binaries + SHA256 checksums to GitHub releases. The implementation is clean and well-structured. All 13 acceptance criteria are met with one meaningful exception (noted below), one correctness issue in the launchd idempotency path, and several smaller concerns.

Overall: net positive, **request changes** on two issues before merge.

---

### Findings

#### Critical Issues

**[Critical/Blocker] `install.sh` launchd idempotency — kickstart target format is wrong**

`scripts/install.sh` lines 163-164:
\`\`\`sh
launchctl kickstart -k "$_target/$LAUNCHD_LABEL" 2>/dev/null
\`\`\`
`$_target` is set to `gui/<uid>` and `$LAUNCHD_LABEL` is `app.nodespace.daemon`, so this expands to `gui/<uid>/app.nodespace.daemon`. The correct domain-target format for `launchctl kickstart` is `gui/<uid>/app.nodespace.daemon` — that part is right. However, the preceding `launchctl bootstrap` failure is not exclusively caused by the service already being registered; it also fires on permission errors, missing entitlements, or a malformed plist. Silencing the failure and unconditionally falling through to `kickstart` means a plist write error would silently leave the old binary running. The fix is to check whether the service is already loaded via `launchctl print gui/<uid>/app.nodespace.daemon` before deciding between bootstrap and kickstart:

\`\`\`sh
if launchctl print "$_target/$LAUNCHD_LABEL" >/dev/null 2>&1; then
    launchctl kickstart -k "$_target/$LAUNCHD_LABEL" 2>/dev/null \
        || printf 'Note: kickstart failed; daemon will restart on next login\n'
else
    launchctl bootstrap "$_target" "$PLIST_PATH" \
        || printf 'Note: bootstrap failed; daemon will start on next login\n'
fi
\`\`\`

**[Critical/Blocker] SHA256SUMS file is downloaded over HTTPS but its integrity is not independently verified**

`scripts/install.sh` lines 99-111: The installer downloads `SHA256SUMS` from the same GitHub release and immediately trusts it to verify the binaries. This closes the loop on accidental corruption but does not protect against a compromised release. An attacker who can write to the GitHub release can replace both the binaries and the SHA256SUMS simultaneously. The issue acceptance criterion says \"SHA256 checksum verified before install — aborts on mismatch\" which this satisfies at the letter, but for a `curl | sh` installer the expected security model is that the checksum file is signed. The pragmatic fix (no GPG infra needed today) is to document this limitation prominently in the script header and in the README, and file a follow-on issue for GPG signing. Do not block on the signing itself, but block on the documentation so users understand the trust model before running `curl | sh`.

---

#### Suggested Improvements

**[Improvement] `uninstall.sh` MCP block removal is fragile for non-trivial JSON**

`scripts/uninstall.sh` lines 81-85: The awk-based block remover matches `}[[:space:]]*,?[[:space:]]*$` to find the closing brace of the nodespace entry. This will mis-fire if the nodespace block contains nested objects (e.g. `"env": {...}`) or if a trailing comma appears on a line by itself. The Rust CLI uninstall (`packages/cli/src/commands/uninstall.rs` lines 126-131) correctly parses the JSON with `serde_json` and does a typed remove, which is the right approach. Consider adding a note to the shell uninstaller that warns users when it falls back to the manual-removal path and suggesting they use `nodespace uninstall` instead for safer JSON handling. The `_entry_count -le 5` cap already partially mitigates this — keep it but also print the `nodespace uninstall` suggestion in the \"many entries\" message.

**[Improvement] `install.sh` daemon wait loop counter arithmetic is incorrect for the 0.5s path**

`scripts/install.sh` lines 205-217: The loop condition is `$_waited -lt 10`. On the `sleep 0.5` branch the counter increments by 1 per half-second, so the loop exits after 5 real seconds. On the fallback `sleep 1` branch it increments by 2 per second, also exiting after 5 real seconds. The variable name `_waited` implies it counts elapsed seconds but it actually counts half-seconds on fast systems. This does not break anything, but rename it to `_ticks` or fix the counter to track real seconds so the timeout behavior is what the comment implies (\"10 seconds\" expectation from reading the bound):

\`\`\`sh
_waited=0
while [ $_waited -lt 20 ]; do   # 20 half-second ticks = 10 real seconds
\`\`\`
Or use seconds directly:
\`\`\`sh
_elapsed=0
while [ $_elapsed -lt 10 ]; do
    sleep 1; _elapsed=$((_elapsed + 1))
done
\`\`\`

**[Improvement] `install.sh` PATH section silently skips when neither `~/.zshrc` nor `~/.bash_profile` exists**

`scripts/install.sh` lines 228-237: The loop only appends the PATH line to files that already exist. A fresh Linux container where neither file exists will silently do nothing when the user answers Y. Add a fallback:
\`\`\`sh
_added=0
for _rc in ...; do
    if [ -f "$_rc" ]; then ... _added=1; fi
done
if [ "$_added" -eq 0 ]; then
    printf '  Neither ~/.zshrc nor ~/.bash_profile found. Add manually:\n  %s\n' "$_path_line"
fi
\`\`\`

**[Improvement] CI: `build-headless` has no `needs:` dependency on `build-tauri`**

`.github/workflows/release.yml` line 201: `build-headless` and `build-tauri` run in parallel. On a release event both will be uploading assets to the same release simultaneously via `gh release upload --clobber`. This is safe (different file names) but the `upload-checksums` job (`needs: build-headless`) will run before `build-tauri` artifacts are fully uploaded. Since `upload-checksums` only downloads `nodespaced-*` and `nodespace-*` patterns (not Tauri DMGs) this is fine — but it's worth a comment to explain the intentional parallelism and clarify that the Tauri artifacts are intentionally excluded from `SHA256SUMS`.

**[Improvement] `workflow_dispatch` platform filter does not include `linux-arm` or `linux-x86`**

`.github/workflows/release.yml` lines 13-17: The `inputs.platform` choice list contains `macos-arm`, `macos-intel`, `windows` but not `linux-arm` or `linux-x86`. A dispatch to build a specific Linux headless target will always match the `all` fallback and build all 4 headless platforms. This is a minor inconvenience for debugging but should be fixed before the first real release.

---

#### Nitpicks

- Nit: `scripts/install.sh` line 22: `die()` uses `printf '\nError: %s\n' "$*"` which passes multiple arguments to `%s` — only the first argument after the format string will fill `%s` when `$*` expands to multiple words after word-splitting. Use `"$*"` (double-quoted) as a single argument. Currently `"$*"` is always a single string, so this is safe today but fragile if call sites change.

- Nit: `packages/cli/src/commands/uninstall.rs` line 42: `unsafe { libc::getuid() }` is minimal and correct, but `std::env::var("USER")` + a `launchctl list` check would avoid the `unsafe` block entirely. Not worth changing today; note for follow-on cleanup.

- Nit: `scripts/install.sh` line 82: `sed` is used for JSON extraction from the GitHub API response. This is pragmatic for a POSIX script, but `grep -o '"tag_name":"[^"]*"'` with a second `grep -o '[^"]*"$'` avoids the BRE regex and `\1` back-reference which some minimal POSIX `sed` implementations don't support. Low risk on modern macOS/Linux.

- Nit: `uninstall.rs` line 78 — `remove_bin_dir` uses `fs::remove_dir_all` which will silently remove any user-placed files in `~/.nodespace/bin/`. The shell uninstaller is more precise (removes only the two known binaries, then attempts `rmdir`). The Rust variant should match for safety.

---

| Criterion | Status |
|---|---|
| curl install on macOS arm64 | Covered by `install.sh` + CI build |
| Platform detection: 4 platforms | Covered |
| SHA256 verified before install | Covered (with trust-model caveat noted above) |
| launchd on macOS / systemd on Linux | Covered |
| Daemon started immediately | Covered (with idempotency bug noted above) |
| Terminal prompts for PATH/MCP/skill | Covered |
| PATH appended to `.zshrc`/`.bash_profile` | Covered (with silent-skip edge case noted) |
| MCP entry written to claude_desktop_config.json | Covered |
| Skill SKILL.md written | Covered |
| Idempotent on re-run | Partially — PATH and MCP have idempotency guards; launchd restart path has the bug noted above |
| `nodespace uninstall` clean teardown + data preserved | Covered |
| CI uploads binaries + SHA256SUMS | Covered |
| Script hosted and versioned | Not in scope for this PR (Cloudflare hosting is infra) — acceptable |

malibio:
**Review Decision: REQUEST CHANGES**

Two issues require fixes before merge:
1. The launchd idempotency path uses `bootstrap` failure as a proxy for 'already registered' — use `launchctl print` to detect the loaded state instead.
2. The SHA256SUMS trust model needs a documentation comment in the script header so users understand the security scope of `curl | sh`.

Full details in the review comment above.

PR #1196, opened 2026-05-19
Strip nodespace-core/agent/nlp-engine from Tauri app: connect to external nodespaced via UDS

malibio:
## Summary

- Removes `nodespace-core`, `nodespace-nlp-engine`, `nodespace-agent`, `surrealdb`, `axum`, `tower-http` from `src-tauri/Cargo.toml` — Tauri app is now a thin gRPC client
- Replaces in-process TCP gRPC server spawn with UDS connector to `~/.nodespace/daemon.sock` (or `NODESPACED_SOCKET` env)
- Deletes `domain_event_forwarder.rs`; activates `watcher.rs` gRPC stream for all node events
- Strips `lib.rs` of all DB/embedding/model initialization (SurrealStore, NodeService, EmbeddingProcessor, CompositeModelManager)
- Adds 8 new RPCs to daemon `local_agent_service.proto`: `DownloadModel`, `DeleteModel`, `LoadModel`, `UnloadModel`, `CancelModelDownload`, `OllamaAvailable`, `RecommendedModel`, `GetSystemRam`
- Implements all new RPCs in `LocalAgentServiceImpl` using existing `CompositeModelManager`
- Routes `chat_models.rs` commands through `LocalAgentServiceClient` via gRPC instead of direct state
- Adds local type mirrors in `types.rs` for agent types used by commands (severs `nodespace-agent` dep from commands)
- Deletes `config.rs` (dead after lib.rs strip) and `tests/agent_e2e.rs` (tested in-process stack)

## Test plan

- [x] `cargo check -p nodespace-app` passes with no errors
- [x] `cargo clippy -p nodespace-app -p nodespace-daemon --all-targets -- -D warnings -D clippy::unused-async` passes clean
- [x] `cargo tree -p nodespace-app --depth 1` shows no `surrealdb`, `nodespace-core`, `nodespace-nlp-engine`, `nodespace-agent`
- [x] Frontend tests: 3828 passed, 93 pre-existing failures (no regressions)
- [x] `bun run --cwd packages/desktop-app quality:fix` passes clean
- [ ] `bun run dev:tauri` cold compile completes in under 5 minutes (requires clean build cache to verify)

Closes #1194


malibio:
## Code Review — PR #1196

---

### Requirements Check

- ✅ `nodespace-core`, `nodespace-nlp-engine`, `nodespace-agent` removed from `src-tauri/Cargo.toml`
- ✅ `surrealdb`, `axum`, `tower-http` removed from `src-tauri/Cargo.toml`
- ✅ `grpc_client.rs` connects to `~/.nodespace/daemon.sock` via UDS connector (with `NODESPACED_SOCKET` env override)
- ✅ `domain_event_forwarder.rs` deleted; `watcher.rs` gRPC stream now handles node events
- ✅ `lib.rs` contains no database/embedding/model initialization
- ✅ All existing Tauri commands route via gRPC (`chat_models.rs` fully ported)
- ✅ `cargo check -p nodespace-app` passes per test plan
- ✅ Frontend tests: 3828 passed, no regressions
- ⚠️  `bun run dev:tauri` cold-compile under 5 minutes — marked not yet verified (requires clean build cache); acceptable for merge given structural correctness

---

### Code Review Findings

#### Important Issues

**[Improvement] `lib.rs:239-241` — Fatal panic on daemon-not-running, no graceful degradation**

```rust
.expect("Failed to connect to nodespaced — is the daemon running?");
```

The `#[cfg(target_os = "macos")]` block above this (lines 205–230) already handles the daemon-not-running case by emitting `"daemon-status": "not_running"` to the frontend — but then on line 241 the code unconditionally panics if the socket isn't reachable, crashing the entire app before the frontend can render that error. The macOS block only runs `ensure_daemon_running` (which waits up to 5 seconds for the socket), but if that path reports `not_running` and continues, the `#[cfg(unix)]` block immediately follows and panics anyway.

The fix is to propagate the error gracefully: emit a `"daemon-status"` event and return `Ok(())` from setup rather than panicking. The app can show an error UI ("daemon not running") instead of hard-crashing.

**[Improvement] `local_agent_service.rs:418-428` — Ollama progress missing from `DownloadModel` stream**

`download_model` registers `set_gguf_progress_callback` but never calls `set_ollama_progress_callback`. Ollama model downloads will silently complete with no progress events — the stream will just terminate after the `ready` event with no intermediate progress. This is a behavioral regression for Ollama downloads vs. the old `chat_model_download` command which registered both callbacks. The fix is a one-liner: register the Ollama callback with the same `tx_cb` pattern immediately after the GGUF one.

**[Improvement] `local_agent_service.rs:426` — `try_send` silently drops progress events**

The channel capacity is 16. `try_send` returns `Err(Full(...))` which is immediately discarded via `let _ = ...`. For large model downloads (several GB) with high-frequency progress events, the channel will fill if the client reads slowly, and events will be silently dropped rather than backpressuring. This is likely fine in practice (the frontend only needs approximate progress), but the silent failure should at minimum be logged at debug level to aid future diagnostics.

**[Improvement] `chat_models.rs:38-54` — `ModelInfo` struct defined but unused**

`pub struct ModelInfo` is defined in `chat_models.rs` (the "local mirror" comment), but `chat_model_list` returns `Vec<serde_json::Value>`, not `Vec<ModelInfo>`. The struct is dead code that will produce a compiler warning (or clippy error if `-D warnings` is active). It should either be used as the return type or deleted. Given `local_agent.rs` also returns `Vec<serde_json::Value>` for `list_local_models`, deleting it is the right call.

#### Nitpicks

**Nit: `grpc_client.rs:114` — `/tmp` fallback for HOME**

```rust
let home = std::env::var("HOME").unwrap_or_else(|_| "/tmp".to_string());
```

If `HOME` is unset, the fallback path `~/.nodespace/daemon.sock` becomes `/tmp/.nodespace/daemon.sock` which is silently incorrect. Logging a warning here would surface misconfigured environments faster.

**Nit: `lib.rs:233` — `#[cfg(unix)]` broader than `#[cfg(target_os = "macos")]` for daemon setup**

The daemon-setup block is `#[cfg(target_os = "macos")]` (launchd), but the gRPC connect block is `#[cfg(unix)]` (Linux + macOS). On Linux, the app will attempt to connect to the UDS without any prior daemon-start logic, which is intentional (Linux users start the daemon manually) but the asymmetry is worth a short comment clarifying this is deliberate.

**Nit: `types.rs:619-679` — Agent type mirrors lack proto-roundtrip validation**

The `LocalAgentStatus`, `InferenceUsage`, `AgentTurnResult`, `AgentSession` mirrors in `types.rs` are serialized from `local_agent.rs` command results. They currently don't derive from or validate against proto message shapes. This is fine for now given we're greenfield, but a comment noting "these must stay in sync with `local_agent_service.proto` response shapes" would help future maintainers.

**Nit: Duplicate model-list surface — `list_local_models` and `chat_model_list`**

Both commands are registered (lib.rs:365, 367). `list_local_models` (`local_agent.rs`) is not called from the frontend TypeScript layer (no `invoke('list_local_models')` found in `src/`). This appears to be a leftover from the pre-migration surface. Consider removing `list_local_models` to avoid dead Tauri command surface, or leave it if it's intentionally reserved.

---

### Summary

This is a clean, well-scoped architectural migration. The diff is large but mechanically sound — heavy stack dependencies are gone, the UDS connector follows the established CLI pattern, and the proto additions are additive and correctly modeled. The `RwLock` wrapper on `GrpcClientInner` remains a minor over-engineering (tonic `Channel` is already `Arc`-backed and clone is the right pattern), but that predates this PR.

The two important issues are the fatal panic on daemon-not-running (which undermines the graceful degradation already wired in the macOS path) and the missing Ollama progress callback in `DownloadModel` (behavioral regression for Ollama users). Both are small fixes. The dead `ModelInfo` struct should be cleaned up to keep clippy clean.

**Recommendation**: REQUEST CHANGES — two targeted fixes before merge (panic → graceful error; Ollama progress callback), plus dead-struct removal.

malibio (commented):
Three targeted fixes needed before merge:

1. Fatal panic on daemon-not-running (lib.rs line 241): the expect after GrpcClient::connect() hard-crashes the app even when the macOS daemon-setup path already handles the not-running case gracefully. Replace with an error emit and early return.

2. Missing Ollama progress callback in download_model (local_agent_service.rs around line 418): only set_gguf_progress_callback is registered; Ollama downloads produce no progress events. Add set_ollama_progress_callback with the same channel sender.

3. Dead ModelInfo struct (chat_models.rs line 38-54): defined but never used — chat_model_list returns Vec<serde_json::Value>. Delete to keep clippy clean.

Full details in the review comment.

malibio:
## Review feedback addressed

**Addressed (3/3 🟡 Important):**

- **Non-fatal daemon connect**: Replaced `expect()` panic with graceful error handling in `lib.rs`. Failed `GrpcClient::connect()` now emits `daemon-status: not_running` to the frontend instead of crashing the app. `watcher::spawn` only fires when the connection succeeds.

- **Missing Ollama progress callback**: Added `set_ollama_progress_callback` alongside the existing `set_gguf_progress_callback` in `download_model` in the daemon service. Ollama downloads now emit progress events.

- **Dead `ModelInfo` struct**: Deleted the unused struct from `chat_models.rs` and removed the now-unused `Deserialize` import.

**Skipped (4/4 🟢 Suggestions):**

- **`HOME` unset warning**: `/tmp` fallback is an OS-level invariant failure; adding a warning would generate noise with no actionable recovery path for users.
- **`#[cfg(unix)]` comment**: The asymmetry is clear from the surrounding context and issue references.
- **`try_send` debug log**: A channel of 16 is large enough for normal use; dropped events here would only occur under extreme load.
- **`list_local_models` dead surface**: Removing a registered Tauri command risks breaking any frontend caller not caught by grep; out of scope for this PR.

PR #1197, opened 2026-05-19
Replace Rust dev-proxy with thin Bun HTTP server

malibio:
Closes #1195

## Summary

- Deletes `packages/dev-tools/src/bin/dev-proxy.rs` (1,419 lines) — the Rust binary that imported `nodespace-core`, opened its own SurrealDB connection, and duplicated NodeService logic
- Adds `packages/dev-tools/src/dev-proxy.ts` — a 660-line Bun TypeScript HTTP server that forwards REST calls to `nodespaced` via gRPC and bridges the `WatchNodes` stream to browser SSE clients
- Removes SurrealDB dependency from dev workflow — only `nodespaced` must be running
- Updates `dev:browser` scripts to remove `surreal start` and use the new Bun proxy
- Removes SurrealDB check from `check-dev-deps.ts`
- Strips `axum`, `tower-http`, `tokio-stream`, `futures`, `async-stream`, `surrealdb`, `chrono` from `packages/dev-tools/Cargo.toml` (only needed by deleted binary)

## Dev workflow (after this PR)

```bash
# Terminal 1: daemon (replaces surreal start + old Rust proxy)
./target/debug/nodespaced

# Terminal 2: frontend + new proxy
bun run dev:browser
```

## Test plan

- [x] `bun run test` — same baseline (4 pre-existing failures, 125 passed)
- [x] `cargo build -p nodespace-dev-tools` — compiles without dev-proxy
- [x] `cargo build --workspace` — full workspace passes
- [x] `bun run quality:fix` — no lint/type errors in modified frontend files
- [ ] Manual: `bun run dev:browser` with `nodespaced` running — SSE events bridge correctly


malibio:
### Code Review Summary

This is a strong architectural improvement. Deleting 1,419 lines of Rust that duplicated the daemon, replacing it with a ~660-line Bun forwarder, and eliminating the SurrealDB dev dependency is exactly the right move for ADR-031's daemon-client model. The gRPC bridging pattern is clean, the SSE reconnect logic is correct, and the HTTP surface is fully preserved.

However, there is one confirmed bug in the schema endpoint response shape that will cause a runtime error in production use of that endpoint.

---

### Findings

#### Critical Issues

**[Critical/Blocker] `GET /api/schemas/:id` — wrong response shape access (`/packages/dev-tools/src/dev-proxy.ts`, schema handler)**

`GetSchemaDefinition` is declared in the proto as:
```
rpc GetSchemaDefinition(GetSchemaDefinitionRequest) returns (NodeResponse);
```
`NodeResponse` has `node_data` at the top level (camelCased to `nodeData` by proto-loader). After conversion, the response shape is `{ nodeId, nodeType, parentId, collectionId, nodeData }`.

The proxy handler accesses `res.node?.nodeData`, which is the shape of `OptionalNodeResponse` — a different message type used by find-by-name style RPCs. Since `NodeResponse` has no `node` field, `res.node` is always `undefined`, so every call to `GET /api/schemas/:id` will return a `404 SCHEMA_NOT_FOUND` error regardless of whether the schema exists.

The fix is straightforward — align with all the other `NodeResponse` handlers in the same file:
```typescript
// Wrong (matches OptionalNodeResponse)
const res = await call<..., { node?: { nodeData?: ProtoNodeData } }>(...)
if (!res.node?.nodeData) return error(...)
return json(nodeDataToSchemaNode(res.node.nodeData));

// Correct (matches NodeResponse)
const res = await call<..., { nodeData?: ProtoNodeData }>(...)
if (!res.nodeData) return error(...)
return json(nodeDataToSchemaNode(res.nodeData));
```

Every other `NodeResponse` handler in this file (`getNode`, `updateNode`, `updateTaskNode`, `moveNode`) correctly accesses `res.nodeData` — this one is the exception and is inconsistent.

---

#### Suggested Improvements

**[Improvement] Missing SSE keepalive interval — connection may drop silently in Playwright/browser environments**

The old Rust proxy sent a keepalive comment every 30 seconds (`KeepAlive::interval(Duration::from_secs(30))`). The new proxy sends only a single `': connected\n\n'` comment on connect and no subsequent heartbeats. Some proxies (nginx, Cloudflare) and certain browser environments will silently close idle SSE connections after 60–120 seconds. For Playwright e2e tests with long test runs, this will produce intermittent disconnects.

Adding a periodic interval is a few lines:
```typescript
start(controller) {
  clientRef = { id: clientId, controller };
  sseClients.add(clientRef);
  controller.enqueue(new TextEncoder().encode(': connected\n\n'));
  // Keepalive every 30s
  const keepalive = setInterval(() => {
    try { controller.enqueue(new TextEncoder().encode(': keepalive\n\n')); }
    catch { clearInterval(keepalive); sseClients.delete(clientRef); }
  }, 30_000);
},
cancel() {
  clearInterval(keepalive); // requires moving keepalive to outer scope
  sseClients.delete(clientRef);
}
```

**[Improvement] `version ?? null` vs `version ?? 0` inconsistency in mutation handlers**

`PATCH /api/nodes/:id` sends `version: body.version ?? null`, while `PATCH /api/tasks/:id` and `POST /api/nodes/:id/parent` send `version: body.version ?? 0`. For `UpdateNode`, `version` is `optional int64` in the proto, so `null` is semantically correct (means "auto-fetch current version"). For `UpdateTaskNode` and `MoveNode`, `version` is non-optional `int64`, so `?? 0` is the right fallback — but 0 is semantically different from "not provided" and may cause a version conflict if the current version is > 0. 

This is not a regression from the old proxy's behavior (which required `version` in the request body), but it's worth documenting which handlers silently accept a missing version vs. which will fail.

**[Improvement] `demo:browser` and `dev:browser:prod` are now identical to `dev:browser`**

After this change, all three scripts run the same proxy against the same `nodespaced` socket. The original differentiation (pointing to different database paths) is gone since the daemon owns the DB. Either document that `NODESPACED_SOCKET` env var is the correct way to switch daemons, or remove the redundant script aliases if they no longer serve a purpose.

---

#### Nitpicks

**Nit:** `console.log`/`console.error` usage throughout `dev-proxy.ts` — CLAUDE.md says "No raw `console.log/debug/info/warn/error` in production code. Test files and DeveloperInspector are exempt." This is a dev-only tool file, so this is a judgment call. Recommend either adding it to the explicit exempt list or using `createLogger` for consistency.

**Nit:** The `call<TReq, TRes>` helper types `method` as `Function` (capital F) which is the broad function type. A typed overload or `(...args: unknown[]) => void` would be slightly better, but given the necessary escape-hatch casting throughout (`nodeClient as unknown as Record<string, Function>`), this is low priority.

**Nit:** `WatchNodes` stream reconnection uses `setTimeout(connect, 2000)` on error and `setTimeout(connect, 1000)` on end. The differing values appear arbitrary. Exponential backoff would be more robust, but for a dev tool this is acceptable.

---

| Criterion | Status |
|-----------|--------|
| `dev-proxy.rs` deleted | PASS |
| `dev-proxy.ts` implements all existing HTTP endpoints | PASS (all endpoints present; see bug on schema/:id) |
| SSE bridge forwards WatchRequest gRPC stream events | PASS |
| `bun run dev:browser` starts the new proxy | PASS |
| No `surreal start` required | PASS |
| `cargo build --workspace` passes | PASS (per PR test plan) |
| `bun run quality:fix` passes | PASS (per PR test plan) |
| Playwright e2e tests against real daemon | NOT VERIFIED (manual only, marked open) |

---

### Verdict

**Request changes** on the `GET /api/schemas/:id` response shape bug — it is a guaranteed runtime failure for any caller of that endpoint. The fix is a one-line type change and two-line access pattern change. All other issues are improvements or nits. Once the schema endpoint is corrected, this PR is ready to merge.

malibio (commented):
Review verdict: **request changes** on the GET /api/schemas/:id response shape bug. See full review in the comment above. This is a confirmed runtime failure — accesses res.node?.nodeData but GetSchemaDefinition returns NodeResponse (not OptionalNodeResponse), so res.node is always undefined and the endpoint always returns 404. Fix to match other NodeResponse handlers (res.nodeData).

malibio:
### Re-Review: Code Review Summary (Post-Fix)

The critical blocker from the initial review — `GET /api/schemas/:id` accessing `res.node?.nodeData` when `GetSchemaDefinition` returns `NodeResponse` (not `OptionalNodeResponse`) — has been correctly resolved. The fix in commit `6a59efd` aligns the handler with every other `NodeResponse` consumer in the file: the response type is now `{ nodeData?: ProtoNodeData }` and access is `res.nodeData` directly. This is confirmed correct against the proto: `NodeResponse` has `node_data` at field 5 with no wrapping `node` field.

The rebase onto `origin/main` (post-#1196 Tauri slim-down) brought in three additional commits — `0f3786f`, `eee5080`, `79abb8d` — which are unrelated model-download script fixes from a parallel PR. They do not touch `dev-proxy.ts` or any of the endpoints introduced here. No conflicts or regressions from the rebase were observed.

---

### Findings

#### Critical Issues

None. The blocker is resolved.

---

#### Suggested Improvements (Carry-forward from previous review, intentionally deferred)

**[Improvement] Missing SSE keepalive — connection may drop on idle**

`/packages/dev-tools/src/dev-proxy.ts`, SSE `start()` handler. Only a single `': connected\n\n'` comment is sent; no periodic heartbeat. Some proxy environments and browser SSE implementations silently close connections after 60–120 seconds of inactivity. The previous review included a concrete implementation snippet. This is a dev-only tool so the blast radius is limited, but intermittent Playwright test disconnects are a real consequence.

**[Improvement] `version ?? null` vs `version ?? 0` inconsistency**

`PATCH /api/nodes/:id` defaults `version` to `null`; `PATCH /api/tasks/:id` and `POST /api/nodes/:id/parent` default to `0`. For `UpdateTaskNode` and `MoveNode` where `version` is non-optional `int64` in the proto, passing `0` when the node is at version > 0 may cause unexpected version-conflict errors. Worth adding a comment or aligning the fallback behavior.

**[Improvement] `dev:browser:prod` and `demo:browser` are now identical to `dev:browser`**

All three scripts run the same proxy command against the same socket path. If the intent is that `NODESPACED_SOCKET` is the correct differentiator, a comment in `package.json` clarifying this would prevent future confusion. If the aliases serve no purpose, YAGNI suggests removing them.

---

#### Nitpicks

**Nit:** `console.log`/`console.error` throughout `dev-proxy.ts` — CLAUDE.md prohibits raw console calls in production code but exempts dev tools. This file is clearly dev-only; recommend adding it to the explicit exemption list in CLAUDE.md or converting to `createLogger` for consistency across the codebase.

**Nit:** `call<TReq, TRes>` types `method` as `Function` (capital F). Low-priority given the necessary escape-hatch casting elsewhere in the file.

**Nit:** WatchNodes reconnect delays (`setTimeout(connect, 2000)` on error, `setTimeout(connect, 1000)` on end) are arbitrary constants. Acceptable for a dev tool; exponential backoff would be more robust long-term.

---

| Criterion | Status |
|-----------|--------|
| `dev-proxy.rs` deleted | PASS |
| `dev-proxy.ts` implements all existing HTTP endpoints | PASS |
| SSE bridge forwards WatchRequest gRPC stream events | PASS |
| `bun run dev:browser` starts the new proxy | PASS |
| No `surreal start` required | PASS — SurrealDB removed from check-dev-deps.ts and all concurrent scripts |
| `bun run quality:fix` passes | PASS (per PR test plan) |
| `cargo build --workspace` passes | PASS (dev-proxy.rs removed, orphan deps cleaned from Cargo.toml) |
| Playwright e2e tests against real daemon | NOT VERIFIED (manual only — acceptable for this change) |

---

### Verdict

**Approve.** The blocker is fixed correctly, the rebase introduced no regressions, and the remaining items are improvements/nits with no merge-blocking impact. This is a clean, well-scoped architectural improvement: 1,419 lines of Rust deleted, replaced by a 669-line Bun forwarder, SurrealDB dev dependency eliminated.

PR #1200, opened 2026-05-19
Pro Tauri client: NODESPACED_ADDR external-daemon mode + capability probe + sync pill

mstomar125:
## Summary
Adds an opt-in, env-var-gated external-daemon mode to the desktop-app so it can dial `nodespaced-pro` (from the private `nodespace-sync` repo) for the Pro-tier sync feature, while keeping the in-process community path completely untouched.

- **`NODESPACED_ADDR`** — when set, Tauri setup skips the in-process `SurrealStore` / `NodeService` / NLP / domain-event-forwarder init and dials the address over TCP via tonic. Unset (default) keeps today's embedded path.
- **`NODESPACED_USER`** — cosmetic; sets the main window title to `NodeSpace — <user>` so multi-window demos are distinguishable.
- **Pro capability probe** (`ProClient::probe_on_channel`) — single `WatchSyncStatus` call on the shared gRPC channel; daemon's `Unimplemented` → community tier (no UI surfaced), event received → Pro tier (pill renders).
- **Sync pill + sign-in** (`pro-sync-pill.svelte`) — renders only on Pro tier; state→color (grey/amber/blue/green/red); clickable in disconnected/auth-required/error states → invokes `pro_initiate_oauth` (daemon owns the PKCE browser dance).
- **Vendored proto** — `nodespace.pro.v1` lives under `packages/desktop-app/src-tauri/proto/` so the public repo builds standalone; sync owns the source of truth, vendor copy kept in lockstep.

## Why this is small + additive
Path A-lite from the design doc — ~50 LOC additive instead of refactoring `grpc_client.rs` to thin-client (which is overlap with ADR-031 daemon-client work landing later). Community build behavior is byte-identical when `NODESPACED_ADDR` is unset.

## Channel sharing
`GrpcClient` now exposes `channel()` and `ProClient::probe_on_channel(channel)` reuses it. Opening a parallel channel caused tonic "Service was not ready: transport error" on subsequent calls after the probe stream was dropped — the parallel channel's background h2 task was orphaned.

## Test plan
- [x] `cargo check --no-default-features` clean
- [x] `NODESPACED_ADDR='[::1]:50071'` Tauri window probes tier=Pro and renders pill (green/CONNECTED)
- [x] Default community launch unaffected — no `pro:tier-detected` event, pill hidden
- [x] Two windows distinguishable via `NODESPACED_USER` titles
- [ ] End-to-end cross-window sync (blocked on separate upstream bugs being scoped)

[...]

PR #1206, opened 2026-05-20
Normalize RelationshipEvent.from_id / to_id to node:<id>

mstomar125:
Closes #1201. Slice 1 of NodeSpaceAI/nodespace-sync#49.

## Summary

`RelationshipEvent.from_id` / `to_id` were emitted as bare node ids (e.g. `"2026-05-20"`) at eight emit sites in `NodeService`. This violated the serialization-contract test in `db/events.rs:194-210` which asserts the prefixed `"node:<key>"` form — emit sites had drifted from the contract.

Most visible symptom: downstream consumers that parse on `:` (e.g. the Pro sync push consumer's `parse_node_thing`) rightly reject bare ids, breaking cloud RELATE for every node created under a date page.

## What changes

- `packages/core/src/db/events.rs` — adds `RelationshipEvent::new` constructor + a small `node_thing()` helper (pass-through if `:` present, prepend `node:` otherwise).
- `packages/core/src/services/node_service.rs` — converts all 8 production emit sites:
  - `create_parent_edge` (the date-page case that surfaced the bug)
  - `reorder_child`
  - two `move_node` paths
  - two `create_or_update` paths
  - `create_mention` + `create_mention_with_source`
  - the collection `member_of` emit site
  - the custom-relationship emit site

The contract test in `events.rs` is unchanged — it was already asserting the prefixed form; this PR is the producer side catching up.

## Background

This commit was originally validated as part of the two-window Pro PoC (closed PRs #1200/#1204/#1205, archived at `/Users/mayanktomar/nodespace/two-window-poc/` per the restart plan in NodeSpaceAI/nodespace-sync#49). Lifted forward unchanged onto current main — the fix is transport-agnostic.

## Verification

- `cargo check -p nodespace-core` — clean
- `cargo clippy -p nodespace-core --lib -- -D warnings` — clean
- `cargo fmt -p nodespace-core -- --check` — clean

### DIFF
--- a/packages/core/src/db/events.rs
+++ b/packages/core/src/db/events.rs
@@ -61,15 +61,14 @@ pub struct RelationshipEvent {
 }
 
 impl RelationshipEvent {
-    /// Construct an event with `from_id` / `to_id` normalized to
-    /// `node:<id>` form. Callers in `NodeService` hold raw node ids
-    /// (e.g. `"2026-05-20"` for date pages, bare UUIDs for normal
-    /// nodes) and the contract for `RelationshipEvent` requires the
-    /// table-prefixed `<table>:<key>` form (see the serialization
-    /// contract test below). Downstream consumers parse on `:` and
-    /// reject bare ids — most notably the Pro sync push consumer's
-    /// `parse_node_thing`, which gets bare date-page ids today and
-    /// fails the cloud RELATE.
+    /// Construct an event with `from_id` / `to_id` normalized to the
+    /// full SurrealDB Thing form (`node:<key>`). Required by the
+    /// serialization-contract test below — every consumer that
+    /// parses these fields splits on `:` and rejects bare ids.
+    /// Callers in `NodeService` often hold bare ids (date-page
+    /// nodes use `"2026-05-20"`, regular nodes use bare UUIDs), so
+    /// this constructor is the single normalization point producers
+    /// should go through.
     pub fn new(
         id: String,
         from_id: &str,
@@ -89,7 +88,10 @@ impl RelationshipEvent {
 
 /// Normalize a node id to its full SurrealDB Thing form
 /// (`node:<key>`). Pass-through if the input already contains `:`.
-fn node_thing(id: &str) -> String {
+/// `pub(crate)` so the `RelationshipDeleted` inline-field variant
+/// can use the same normalization at its emit sites in
+/// `services::node_service` without duplicating the helper.
+pub(crate) fn node_thing(id: &str) -> String {
     if id.contains(':') {
         id.to_string()
     } else {
@@ -221,6 +223,38 @@ pub enum DomainEvent {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// Locks in the `node_thing` normalizer contract: bare ids (date
+    /// pages, raw UUIDs) get prefixed; anything already containing
+    /// `:` passes through unchanged. The producer-side
+    /// `RelationshipEvent::new` and the
+    /// `RelationshipDeleted`-emit-sites both rely on this shape.
+    #[test]
+    fn node_thing_normalizes_bare_ids_and_passes_through_prefixed() {
+        assert_eq!(node_thing("2026-05-20"), "node:2026-05-20");
+        assert_eq!(node_thing("some-uuid"), "node:some-uuid");
+        assert_eq!(node_thing("node:already-prefixed"), "node:already-prefixed");
+        // Edge case: any `:` makes it pass-through, even foreign
+        // tables. Producers are expected to pass ids in the
+        // `<table>:<key>` form when crossing table boundaries.
+        assert_eq!(node_thing("relationship:abc"), "relationship:abc");
+    }
+
+    /// `RelationshipEvent::new` normalizes both endpoint ids through
+    /// `node_thing`. Locks the contract so producers can pass bare
+    /// ids and rely on the constructor for the prefix.
+    #[test]
+    fn relationship_event_new_normalizes_both_endpoints() {
+        let rel = RelationshipEvent::new(
+            "relationship:abc:def".to_string(),
+            "abc",
+            "def",
+            "has_child",
+            serde_json::json!({"order": 1.0}),
+        );
+        assert_eq!(rel.from_id, "node:abc");
+        assert_eq!(rel.to_id, "node:def");
+    }
 
     /// Contract test: Documents and enforces the exact JSON format for RelationshipEvent (Issue #811)
     ///
--- a/packages/core/src/db/events.rs
+++ b/packages/core/src/db/events.rs
@@ -60,6 +60,43 @@ pub struct RelationshipEvent {
     pub properties: serde_json::Value,
 }
 
+impl RelationshipEvent {
+    /// Construct an event with `from_id` / `to_id` normalized to
+    /// `node:<id>` form. Callers in `NodeService` hold raw node ids
+    /// (e.g. `"2026-05-20"` for date pages, bare UUIDs for normal
+    /// nodes) and the contract for `RelationshipEvent` requires the
+    /// table-prefixed `<table>:<key>` form (see the serialization
+    /// contract test below). Downstream consumers parse on `:` and
+    /// reject bare ids — most notably the Pro sync push consumer's
+    /// `parse_node_thing`, which gets bare date-page ids today and
+    /// fails the cloud RELATE.
+    pub fn new(
+        id: String,
+        from_id: &str,
+        to_id: &str,
+        relationship_type: impl Into<String>,
+        properties: serde_json::Value,
+    ) -> Self {
+        Self {
+            id,
+            from_id: node_thing(from_id),
+            to_id: node_thing(to_id),
+            relationship_type: relationship_type.into(),
+            properties,
+        }
+    }
+}
+
+/// Normalize a node id to its full SurrealDB Thing form
+/// (`node:<key>`). Pass-through if the input already contains `:`.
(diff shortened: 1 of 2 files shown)


malibio:
## 🔍 Code Review (automated)

### Summary

PR #1206 fixes a real bug: `RelationshipEvent.from_id` / `to_id` were emitted as bare node IDs (e.g. `"2026-05-20"`, plain UUIDs) at every `RelationshipCreated` / `RelationshipUpdated` emit site in `NodeService`, while the serialization-contract test in `events.rs` and downstream consumers expected the table-prefixed `node:<key>` form. The fix introduces `RelationshipEvent::new` with an internal `node_thing()` normalizer, and migrates all 10 created/updated emit sites to use it.

The core logic is sound, but one blocker must be addressed before merge.

---

### Findings

#### 🔴 [Blocker] `RelationshipDeleted` emit sites are left un-normalized

`RelationshipDeleted` is a separate inline-field variant (not using `RelationshipEvent`), but its `from_id` / `to_id` fields carry the same bare-ID semantics as the variants fixed here. There are three emit sites in `/packages/core/src/services/node_service.rs`:

- `delete_mention` (mentioning/mentioned node IDs)
- `delete_mention` with source (source/target node IDs)
- Custom relationship delete (source/target node IDs)

All three pass raw `.to_string()` on IDs that can be bare date strings or UUIDs. Any downstream consumer that parses `:` on `RelationshipDeleted.from_id` / `.to_id` (including the sync push consumer) will observe the same class of bug this PR fixes for created/updated events.

Options:
- **(a)** Refactor `RelationshipDeleted` to also wrap `RelationshipEvent` — symmetric fix across all three variants (preferred)
- **(b)** Apply `node_thing()` explicitly at each of the three delete emit sites and expose it as `pub(crate)`

---

#### 🟡 [Improvement] `node_thing()` has no unit tests; `RelationshipEvent::new` has no behavioral test

The `node_thing` helper and `RelationshipEvent::new` constructor are the load-bearing pieces of this fix, but neither has a dedicated test. The existing serialization-contract test constructs `RelationshipEvent { ... }` directly with pre-prefixed strings — it does not exercise the normalizer at all.

The helper also has a subtlety worth pinning: any string containing `:` is treated as already-prefixed. A 3-line test would lock in the contract:

```rust
#[test]
fn test_node_thing_normalizes() {
    assert_eq!(node_thing("2026-05-20"), "node:2026-05-20");
    assert_eq!(node_thing("some-uuid"), "node:some-uuid");
    assert_eq!(node_thing("node:already-prefixed"), "node:already-prefixed");
}
```

---

#### 🟡 [Improvement] `relationship:<bare-id>:<bare-id>` event `.id` field is inconsistent with the `node:<key>` contract

The `.id` field of relationship events is constructed as `format!("relationship:{}:{}", parent_id, node_id)` where the segments are bare. The serialization contract test uses `"relationship:abc123"` (a single-segment SurrealDB Record ID). This inconsistency is pre-existing and not introduced by this PR, but since this PR touches all 6 compound-format emit sites, it's the right moment to note it. Is `relationship:<node-id>:<node-id>` a valid SurrealDB Thing, or a synthetic correlation key? Worth a comment or follow-up issue to disambiguate.

---

#### ⚪ Nitpicks

- **Tracing `debug!` log** emits the full `DomainEvent` in `Debug` format, which includes full `PropertyChange` vectors with old/new JSON values under high-frequency editing. Consider logging only the variant discriminant. This is also logically unrelated to the normalization fix — cleaner as a separate commit.
- **`node_thing` visibility:** Currently private to the module. If the `RelationshipDeleted` fix calls it from `node_service.rs`, it will need to be `pub(crate)` — anticipate this.
- **Doc comment on `RelationshipEvent::new`:** References "the Pro sync push consumer's `parse_node_thing`" (a cross-repo concern). Consider keeping the doc comment focused on the local invariant (`IDs must be in node:<key> form`) and moving downstream context to the commit message or a linked issue.

---

The `RelationshipDeleted` emit sites are the same class of bug as what this PR fixes — leaving them un-normalized means downstream consumers still see bare IDs on delete events. The fix is small (3 sites, ~9 lines). Address that and add the `node_thing` unit test, then this is good to merge.

mstomar125:
Review feedback addressed in `458d36a8`. Internal review run before push: `cargo check`, `cargo clippy -- -D warnings`, `cargo fmt -- --check`, and the new `events.rs` tests (6 passed) all clean.

### Blocker
- **`RelationshipDeleted` emit sites** — three sites in `services::node_service` now route `from_id` and `to_id` through `db::events::node_thing`. The helper is now `pub(crate)` so the inline-field `RelationshipDeleted` variant can normalize without duplicating it (rather than refactoring the variant to wrap `RelationshipEvent`, which would have rippled into `DomainEventForwarder` for a marginal win).

### Improvements
- **Unit tests** — added `node_thing_normalizes_bare_ids_and_passes_through_prefixed` and `relationship_event_new_normalizes_both_endpoints` in `events.rs`. The bare/UUID/prefixed cases plus a cross-table-prefix edge case (`relationship:abc` passes through, as the contract requires producers to disambiguate when crossing table boundaries).
- **Doc comment trim** — `RelationshipEvent::new` doc now describes only the local invariant; the cross-repo sync-side downstream context is in the PR description / commit instead.

### Pre-existing concern noted, not fixed in this PR
- **`relationship:<bare>:<bare>` id format** — out of scope for the normalization fix; could be a follow-up issue (synthetic correlation key vs. SurrealDB Thing — worth disambiguating, but unchanged by this PR).

### Nit dropped
- The `tracing::debug!("broadcast emit", ...)` breadcrumb in `NodeService::new` was removed. Reviewer's call ("unrelated, cleaner as a separate commit") was right; its primary value was root-causing the duplicate-CREATE during the Pro PoC, which is now fixed via UPSERT on the sync side. If we ever need the breadcrumb back, it lands as its own focused commit with the variant-discriminant-only shape suggested.

malibio:
## Re-Review: #1206 — Normalize RelationshipEvent.from_id / to_id

**Re-review of commit `458d36a8` ("Apply review feedback for #1206")**

### Re-Review Summary

**4 of 4 original findings addressed. 0 not addressed. 0 new issues.**

All blockers and improvements from the original review are resolved. The implementation is clean and the approach taken for each item was well-reasoned.

---

### Original Findings — Status

#### 🔴 [Blocker] RelationshipDeleted emit sites (3) still emitting bare IDs
**✅ Addressed**

All three `RelationshipDeleted` emit sites in `packages/core/src/services/node_service.rs` now call `crate::db::events::node_thing()` on both `from_id` and `to_id`:
- `delete_mention` (mentioning/mentioned node IDs) — line ~2465
- `delete_mention` with source (source/target node IDs) — line ~6083
- Custom relationship delete (source/target) — line ~6610

The chosen approach (option b from the original review: expose `node_thing` as `pub(crate)` and call it directly) is the right call. Refactoring `RelationshipDeleted` to wrap `RelationshipEvent` would have cascaded into `DomainEventForwarder` for negligible gain — the explicit call sites are readable and localized.

Each new call site includes a concise inline comment referencing the normalization rationale, which is appropriate — future readers get the 'why' without hunting for it.

#### 🟡 [Improvement] node_thing() and RelationshipEvent::new unit tests missing
**✅ Addressed**

Two new tests added in `packages/core/src/db/events.rs`:

1. `node_thing_normalizes_bare_ids_and_passes_through_prefixed` — covers all three cases from the suggested test plus a fourth edge case (cross-table prefix passthrough, e.g. `"relationship:abc"`). The edge-case addition and its accompanying comment ("producers are expected to pass ids in the `<table>:<key>` form when crossing table boundaries") locks in a non-obvious invariant that was previously implicit. Good.

2. `relationship_event_new_normalizes_both_endpoints` — behavioral test that constructs the event through the public constructor and asserts both endpoint fields are prefixed. This directly exercises the production path.

Both tests have clear doc comments explaining what they lock in and why, consistent with the existing contract-test style in this file.

#### 🟡 [Improvement] .id field inconsistency (relationship:<bare>:<bare>) noted but out of scope
**✅ Addressed (acknowledged, scoped out)**

The author's self-review comment confirms this is pre-existing, out of scope for this PR, and flagged for a follow-up issue. The decision not to fix it here is sound — this PR has a single cohesive purpose (normalization of from_id/to_id), and pulling in a refactor of the .id correlation-key format would obscure the change and expand blast radius. A follow-up issue is the right disposition.

#### ⚪ Nits: tracing debug log, node_thing visibility, doc comment
**✅ Addressed**

- **Tracing debug log removed** — the `tracing::debug!(..., "broadcast emit")` breadcrumb is gone from `NodeService::new`. Author's note that its primary value was the Pro PoC duplicate-CREATE diagnosis (now fixed via UPSERT on the sync side) makes this a clean drop, not a loss of observability.
- **node_thing visibility** — promoted to `pub(crate)` with an updated doc comment explaining exactly why (so `RelationshipDeleted` emit sites in `services::node_service` can reach it without duplication). The doc comment is well-placed.
- **RelationshipEvent::new doc comment** — cross-repo sync context removed. Doc now describes only the local invariant (`node:<key>` form required, bare ids common at call sites, single normalization point). Precise and appropriate.

---

### New Issues from Feedback Commit

None. The diff is tightly scoped: three `RelationshipDeleted` call sites, one visibility change, two new tests, one debug log removal, one doc comment trim. No unrelated changes, no regressions introduced.

---

### Overall Recommendation

All critical and improvement findings are resolved. The implementation is correct, minimal, and well-documented. Approving.

malibio (approved):
All findings from the original review are resolved. See re-review comment for full breakdown.

malibio:
## ✅ Re-Review Complete — No further review needed

All 4 original findings addressed. PR is clean and ready to merge.

PR #1207, opened 2026-05-20
Pro Tauri client: vendored proto, capability probe, sync pill

mstomar125:
Slice 3 of NodeSpaceAI/nodespace-sync#49.

Layers the Pro-tier client surface onto the post-#1196 daemon-client architecture without touching the transport — capability is detected over the same UDS channel `GrpcClient::connect()` already opens. Community users see no change (probe returns `Unimplemented` → pill stays hidden).

## Backend

- `proto/nodespace_pro.proto` — vendored from `nodespace-sync/nodespaced-pro/proto/` via `scripts/refresh-pro-proto.sh`.
- `build.rs` — compiles the Pro proto alongside Tauri's build via `tonic-build` + `protoc-bin-vendored`.
- `services/pro_client.rs` — `ProClient::probe_on_channel` runs a single-shot `WatchSyncStatus` probe on the existing channel from `GrpcClient`. Surfaces tier as `Pro` / `Community` / `Unknown` for diagnostics.
- `commands/pro_sync.rs` — three Tauri commands (`pro_tier`, `pro_subscribe_sync_status`, `pro_initiate_oauth`); all return community-fallback when `ProClient` isn't in managed state.
- `services/grpc_client.rs` — adds `channel()` accessor and stores the underlying tonic `Channel` so `ProClient` rides the same h2 connection. Opening a parallel channel caused "Service was not ready: transport error" during the PoC after the probe stream was dropped — sharing the channel is the fix.
- `lib.rs` — after `GrpcClient::connect()` succeeds, snapshots the channel, runs the Pro probe, manages `ProClient`, emits `pro:tier-detected`. Registers the three commands in `invoke_handler!`.

## Frontend

- `stores/pro-sync.svelte.ts` — Svelte 5 runes-based store; listens for `pro:tier-detected` + `sync:status` events; exposes reactive `tier`, `state`, `detail`, `isPro`.
- `components/pro-sync-pill.svelte` — sync pill rendering only when `proSync.isPro`. State → grey / amber / blue / green / red. Click in signed-out states invokes `pro_initiate_oauth` (wired but the PKCE flow is exercised via the daemon's `--auto-signin` flag in dev).
- `app-shell.svelte` — imports and starts `proSync` in the Tauri-only branch; mounts the pill in an absolute slot in the content grid; cleans up on destroy.

## Verified

- `cargo check --no-default-features` clean
- `cargo clippy --no-default-features --lib -- -D warnings` clean
- `bunx svelte-check --threshold error` — 0 errors, 0 warnings (1641 files)

## Background

Lifted from the validated PoC at `/Users/mayanktomar/nodespace/two-window-poc/core/`. The PoC used a `NODESPACED_ADDR` env var fork in `lib.rs` setup; this PR hooks into the existing UDS connect path instead (post-#1196 default), so the integration is additive — community behavior is byte-identical when the daemon answers `Unimplemented`.

## Stack

- **#1206** — Slice 1 (`RelationshipEvent` id normalization, closes #1201)
- **(this PR)** — Slice 3
- Slice 4 (daemon `watcher.rs` relationship variants), Slice 5 (sync pull-path relationship subscription), Slice 6 (`nodespaced-pro` on UDS) coming next.

Parent: NodeSpaceAI/nodespace-sync#49.


malibio:
## 🔍 Code Review (automated)

### Summary

This PR layers Pro-tier sync capability onto the existing post-#1196 daemon-client architecture. The core design is sound: reuse the shared `tonic::Channel` from `GrpcClient`, run a single-shot `WatchSyncStatus` probe at startup to detect tier, and hide all Pro UI when the community daemon returns `Unimplemented`. Community behavior is provably unchanged at the binary level. The proto design is clean, the Svelte store follows established runes patterns, and the capability-probe fallback logic is thoughtfully implemented.

There is one payload mismatch bug, one structural concern with `STREAM_SPAWNED` causing a silent reconnect gap, and a dependency hygiene issue.

---

### Findings

#### 🔴 [Improvement] `pro:tier-detected` payload is missing the `addr` field

`packages/desktop-app/src-tauri/src/lib.rs` — The Rust `serde_json::json!` payload emitted as `pro:tier-detected` contains only `tier` and `initial_status`. However, `pro-sync.svelte.ts` declares the listener's generic type as `{ tier: ProTier; addr: string; initial_status: ... }` and assigns `this.addr = p.addr`. The field will always be `undefined`, which TypeScript coerces to the empty string on assignment. The store module's own doc comment says the event shape includes `addr`. Either add the field to the Rust payload or remove it from the TypeScript type and assignment.

---

#### 🔴 [Improvement] `STREAM_SPAWNED` global causes silent stream-drop with no reconnect signal

`packages/desktop-app/src-tauri/src/commands/pro_sync.rs` — `STREAM_SPAWNED` is a `static AtomicBool`. When the Pro daemon restarts or the stream terminates unexpectedly, the Rust task resets the flag to `false`. However the TypeScript side — `ProSyncStore.started` — remains `true`, so `start()` will never re-invoke `pro_subscribe_sync_status`. The stream silently goes dark with no user-visible indication.

At minimum, the Rust side should emit a `sync:status` event with `state = DISCONNECTED` before returning from the task, so the UI reaches a known state and the pill goes grey. A re-subscribe path should follow in a later slice.

---

#### 🟡 [Improvement] `prost` direct dependency should use the workspace declaration

`packages/desktop-app/src-tauri/Cargo.toml` — The workspace `Cargo.toml` already declares `prost = "0.13"`. The new direct dependency should use `prost = { workspace = true }` for consistency and centralized version management. As-is, a future workspace-level bump would silently leave this crate behind.

---

#### 🟡 [Improvement] Probe synthesizes a fake `SyncStatusEvent` on timeout

`packages/desktop-app/src-tauri/src/services/pro_client.rs` — When the daemon sends no initial event within `PROBE_TIMEOUT`, the probe returns a synthetic `SyncStatusEvent` with `state = STATE_UNSPECIFIED`. This is stored as `last_status` and emitted to the frontend as `initial_status`. `decodeState(0)` maps this to `'unspecified'`, which renders the pill as "Sign in" — potentially prompting a PKCE flow the user doesn't need.

Returning `(ProTier::Pro, None)` and letting the subsequent `WatchSyncStatus` subscription deliver the real state would be more correct. The synthetic event approach conflates "we don't have a status yet" with "the daemon is telling us the state is unspecified".

---

#### ⚪ Nitpicks

- **`build.rs`:** `std::env::set_var("PROTOC", &protoc)` is deprecated in Rust edition 2024. Prefer `tonic_build::configure().protoc_path(&protoc)` to be forward-compatible.
- **`decodeState`:** Hardcoded numeric switch over proto enum values — a comment tying the numbers to `nodespace.pro.v1.SyncStatusEvent.State` (with proto file path) would make future updates obvious.
- **`refresh-pro-proto.sh`:** Hardcoded sibling-directory path will fail CI that doesn't provision the private repo. Document this as a dev-only script in the header.
- **`pro-sync-pill.svelte`:** Raw hex color literals instead of design system CSS variables — will drift on token updates.
- **`pro_sync.rs`:** `let _ = pro;` with "Keep the ProClient alive" comment is unnecessary — `ProClient` is `Clone` and wraps an `Arc`; the clone captured by the task holds the reference count alive already.

---

The `addr` field mismatch is a broken type contract that should be fixed before merge. The silent stream-drop with no reconnect signal will confuse users in practice. The probe's synthetic-event fabrication is a correctness concern on slow connections. The rest can follow in the next slice.

mstomar125:
Review feedback addressed in `045877c5`. Internal review pass: `cargo check`, `cargo clippy --no-default-features --lib -- -D warnings`, `cargo fmt -- --check`, and `bunx svelte-check --threshold error` (1641 files, 0 errors / 0 warnings) all clean.

### Blockers
- **`pro:tier-detected` payload mismatch** — `addr` was a PoC-era debug breadcrumb (used when each Tauri window dialed a distinct TCP address); under post-#1196 UDS the daemon address is implicit. Removed the field from the TS listener type and the store rather than fabricating a value on the Rust side. The store's `log.info('tier detected', ...)` no longer prints it either.
- **`STREAM_SPAWNED` silent stream-drop** — on every exit path from the spawned task (subscribe failure, item error, daemon stream-close), the Rust side now emits a synthetic `sync:status` event with `state = STATE_DISCONNECTED`. Pulled into an `emit_disconnected` helper used in all three sites. Pill returns to grey "Sign in" instead of staying stuck on the last status.

### Improvements
- **Probe synthesis on timeout** — `probe()` now returns `last_status = None` when the daemon doesn't send an initial event in the probe window. Drops the synthetic `STATE_UNSPECIFIED` and its misleading "Sign in" decode. The real state arrives via the subsequent `WatchSyncStatus` subscription.
- **`prost` workspace dependency** — switched to `{ workspace = true }`.

### Nits
- **`build.rs` deprecated `set_var`** — annotated with a forward-compat TODO. The cleaner `tonic_build::configure().protoc_path()` shape isn't available on 0.12; lands when tonic-build > 0.12 or edition 2024 bump forces the `unsafe` env wrap.
- **`decodeState` numeric comments** — doc comment now points at the source proto file path so future enum additions are obvious.
- **`refresh-pro-proto.sh` header** — now explicitly says "dev-only, CI does not run this" and explains why the missing-sibling-tree failure mode is the right one.
- **`pro-sync-pill.svelte` raw hex colors** — left as-is; design-token migration is a focused follow-up and unrelated to the review's structural concerns. Will file as a follow-up issue if you want it tracked.
- **`let _ = pro;`** — removed. Replaced with a comment explaining that the `client` clone (Arc-backed) keeps the underlying connection alive for the stream's lifetime.

malibio:
## Re-Review: `045877c5` ("Apply review feedback for #1207")

**5 of 7 findings addressed. 2 open (1 nit deferred by author, 1 nit unaddressed). All critical and high-priority findings resolved.**

---

### Original Finding Status

#### 🔴 `pro:tier-detected` payload missing `addr` field — ✅ Addressed

Correct approach taken: `addr` was a PoC-era debug breadcrumb with no runtime value post-#1196 UDS. The field and its assignment were removed from the TypeScript type (`ProTier` listener generic), the store class (`addr = $state<string>('')`), and the doc comment. The Rust payload was left as-is (no fabrication needed). Type contract is now consistent end-to-end.

#### 🔴 `STREAM_SPAWNED` global: silent stream-drop, no reconnect signal — ✅ Addressed

The `emit_disconnected` helper is well-structured. All three stream-exit paths are covered:
- Subscribe failure (early `return` path) — calls `emit_disconnected` before returning.
- Normal stream end / daemon stream-close — falls through the `while` loop to the post-loop call.
- Item error (`break`) — also falls through to the post-loop `emit_disconnected`.

The emitted payload uses `PbState::Disconnected as i32` (not a magic number), and the helper logs a warning on emit failure rather than panicking. One minor observation: when the Tauri `emit` fails inside the stream loop (the `break` on line ~83), the reason logged is only "failed to emit sync:status" — the `detail` field of the final `emit_disconnected` will say "sync-status stream ended" rather than reflecting the emit failure reason. This is inconsequential for UX since the pill will go grey either way, but worth noting.

#### 🟡 `prost` dependency should use `workspace = true` — ✅ Addressed

`packages/desktop-app/src-tauri/Cargo.toml` now uses `prost = { workspace = true }`. Clean.

#### 🟡 Probe synthesizes fake `SyncStatusEvent` on timeout — ✅ Addressed

The timeout arm now returns `(ProTier::Pro, None)` as recommended. The synthetic `STATE_UNSPECIFIED` event and the now-unused `PbState` import are both removed from `pro_client.rs`. The code comment explaining _why_ `None` is correct (real state arrives via the subsequent subscription) is present and accurate.

#### ⚪ Nit: `build.rs` `set_var` deprecated — ⚠️ Deferred with TODO

A comment documenting the forward-compat path (`tonic_build::configure().protoc_path()`) and the reason for deferral (tonic-build > 0.12 required, edition 2024 will force `unsafe` wrap) has been added inline. The issue is acknowledged and the remediation path is documented. Acceptable for a lean codebase with no current edition-2024 requirement.

#### ⚪ Nit: `decodeState` magic numbers — ✅ Addressed

The doc-comment on `decodeState` now points at the source proto file path (`nodespace-sync/nodespaced-pro/proto/nodespace_pro.proto`, vendored at `packages/desktop-app/src-tauri/proto/`). Future enum additions have a clear update path.

#### ⚪ Nit: `refresh-pro-proto.sh` CI caveat — ✅ Addressed

Script header now explicitly states "Dev-only: CI does not run this" and explains the vendored proto is what gets compiled. The sibling-directory expectation and the failure mode on missing access are documented.

---

### Unaddressed Finding

#### ⚪ Nit: `pro-sync-pill.svelte` raw hex color literals — ❌ Not Addressed

The pill component still uses raw hex values for all dot colors (`#9ca3af`, `#f59e0b`, `#2563eb`, `#16a34a`, `#dc2626`) and for the box-shadow rgba expressions, as well as fallback hex values in `var(--token, #hex)` patterns. This was flagged in the original review and is not mentioned in the feedback commit message. It's a nit — the design system CSS variables for these semantic colors (`--color-warning`, `--color-success`, etc.) may not exist yet — but it will drift as the token system matures.

No blocker, but track it.

#### ⚪ Nit: `let _ = pro` removal — ✅ Addressed (as observed)

The dead binding is gone. The replacement comment correctly explains that the `Arc`-backed `ProClient` clone keeps the connection alive through the task's move semantics.

---

### New Observations (Not Issues)

No regressions or new problems introduced by `045877c5`. The diff is surgical — changes are scoped exactly to the findings, with no collateral rewrites.

One structural note for awareness: `STREAM_SPAWNED` remains a module-level `AtomicBool`. This still means a window reload that re-calls `pro_subscribe_sync_status` before the Rust task fully exits could race on the flag and not start a new subscription. The original review accepted this as a "re-subscribe path should follow in a later slice" — that deferred intent is still valid.

---

### Verdict

All critical and high-priority findings are resolved. The remaining open item is a pre-existing nit (hex colors in the pill) that was not addressed and is not a blocker. The codebase is in better shape than before the feedback commit.

**Approving.**

malibio (approved):
All critical and high-priority findings from the original review are resolved in 045877c5. One pre-existing nit (hex colors in pro-sync-pill.svelte) remains unaddressed and is not a blocker. Approving.

malibio:
## ✅ Re-Review Complete — No further review needed

All critical and important findings resolved. One pre-existing nit (raw hex colors in `pro-sync-pill.svelte`) deferred to a follow-up — not a blocker. Ready to merge.

Issue #49, opened 2025-08-11
Performance & Bundle Optimization


malibio:
**Parent Issue**: #26 - Hybrid Markdown Rendering System

## Overview
Optimize the CodeMirror implementation for production readiness, focusing on bundle size, performance, and memory efficiency to ensure the hybrid markdown rendering system meets all performance requirements.

## Scope
Production readiness optimization:
- Bundle size optimization (tree-shake unused CodeMirror features)
- Performance testing (< 50ms update requirement)
- Memory leak prevention
- Lazy loading investigation

## Technical Requirements
- **Bundle Size**: Keep CodeMirror addition under 200KB
- **Performance**: Maintain < 50ms response time for style updates
- **Memory Management**: Prevent memory leaks during editor lifecycle
- **Tree Shaking**: Remove unused CodeMirror extensions and features
- **Lazy Loading**: Investigate dynamic import possibilities

## Implementation Tasks
- [ ] Analyze current bundle size impact from CodeMirror integration
- [ ] Tree-shake unused CodeMirror extensions and themes
- [ ] Implement performance benchmarks for hybrid rendering
- [ ] Test memory usage during long editing sessions
- [ ] Investigate lazy loading of CodeMirror components
- [ ] Optimize editor initialization and disposal
- [ ] Profile rendering performance across different content sizes
- [ ] Document final bundle size impact and performance metrics

## Performance Benchmarks
- **Style Update Speed**: < 50ms from keystroke to visual change
- **Bundle Size Impact**: < 200KB increase from CodeMirror addition
- **Memory Usage**: Stable memory during extended editing sessions
- **Initialization Time**: < 100ms for editor startup
- **Content Rendering**: Smooth performance with 1000+ character documents

## Bundle Optimization Targets
- **Core CodeMirror**: Only essential view and state modules
- **Markdown Language**: Minimal markdown syntax support only
- **Extensions**: Remove unused autocompletion, search, etc.
- **Themes**: Custom minimal theme instead of full theme packages
- **Tree Shaking**: Verify unused code is properly eliminated

## Acceptance Criteria
- [ ] Bundle size increase is under 200KB (measured and documented)
- [ ] Performance benchmarks all pass (< 50ms update times)
- [ ] Memory usage remains stable during long editing sessions
- [ ] No memory leaks detected in editor lifecycle testing
- [ ] Editor initialization time under 100ms
- [ ] Smooth performance with large documents (1000+ chars)
[...]

PR #1208, opened 2026-05-20
Forward RelationshipEvent variants on WatchNodes stream

mstomar125:
Slice 4 of NodeSpaceAI/nodespace-sync#49.

`watcher.rs` (post-#1196) already streams `node:created` / `node:updated` / `node:deleted` from `nodespaced` to the frontend. It dropped relationship events, so a cloud-pulled `has_child` edge never reached the frontend's `reactiveStructureTree` — child nodes arrived but never attached to their parent in the outliner.

This wires the relationship pipeline end-to-end.

## What changes

- **`packages/proto/proto/node_service.proto`** — extends `NodeEvent` with three new oneof variants (`relationship_created`, `relationship_updated`, `relationship_deleted`) and adds `RelationshipPayload` + `RelationshipDeletedPayload` messages. Backwards-compatible: older clients ignore unknown tags. `properties` is JSON-encoded as a string so the proto schema stays stable across additions to the underlying `serde_json::Value`.

- **`packages/daemon/src/services/node_service.rs::convert_domain_event`** — extends the match to map `DomainEvent::Relationship*` into the new proto variants via a small `relationship_to_proto` helper.

- **`packages/desktop-app/src-tauri/src/watcher.rs::forward()`** — adds the three match arms; payload shape mirrors the in-process `DomainEventForwarder` exactly (camelCase via serde rename, JSON `properties` re-parsed from the wire string). The frontend `tauri-sync-listener.ts` needs no contract change.

- **`packages/desktop-app/src/lib/services/tauri-sync-listener.ts`** — adds a `stripNodePrefix` helper and applies it at the three `structureTree` call sites (`addChild`, `updateChildOrder`, `removeChild`). The structure tree is keyed by bare ids throughout the app; backend events carry the table-prefixed `"node:<key>"` form per the `RelationshipEvent` serialization contract. Normalize at the boundary so the two key spaces agree.

## Verified

- `cargo check -p nodespace-proto` clean (regenerates Rust bindings from the new variants)
- `cargo check -p nodespace-daemon` clean
- `cargo check --manifest-path packages/desktop-app/src-tauri/Cargo.toml --no-default-features` clean
- `cargo clippy --no-default-features --lib -- -D warnings` clean (desktop-app)
- `cargo clippy -p nodespace-daemon -- -D warnings` clean
- `bunx svelte-check --threshold error` — 0 errors, 0 warnings (1639 files)

## Background

Validated end-to-end during the two-window Pro PoC. Adapted to fold the relationship forwarding into the existing `watcher.rs` rather than the PoC's separate `external_event_forwarder` module (post-#1196's watcher path is the single forwarder for external-daemon mode now).

## Stack

- **#1206** — Slice 1 (`RelationshipEvent` id normalization, closes #1201)
- **#1207** — Slice 3 (Pro Tauri client base)
- **(this PR)** — Slice 4
- Slice 5 (sync pull-path relationship subscription, closes NodeSpaceAI/nodespace-sync#46) and Slice 6 (nodespaced-pro on UDS) coming next.

Parent: NodeSpaceAI/nodespace-sync#49.

### DIFF
--- a/packages/daemon/src/services/node_service.rs
+++ b/packages/daemon/src/services/node_service.rs
@@ -40,9 +40,10 @@ use crate::nodespace::{
     MentionResponse, MentionTargetRequest, MoveNodeRequest, NodeCollectionsRequest, NodeData,
     NodeDeleted, NodeEvent, NodeListResponse, NodeReference, NodeReferenceListResponse,
     NodeResponse, NodeTreeResponse, OptionalNodeResponse, OptionalStringClear,
-    OptionalTimestampClear, QueryNodesSimpleRequest, RemoveNodeFromCollectionRequest,
-    RenameCollectionRequest, ReorderNodeRequest, ReorderNodeResponse, SearchRequest,
-    UpdateNodeRequest, UpdateTaskNodeRequest, UpsertNodeWithParentRequest, WatchRequest,
+    OptionalTimestampClear, QueryNodesSimpleRequest, RelationshipDeletedPayload,
+    RelationshipPayload, RemoveNodeFromCollectionRequest, RenameCollectionRequest,
+    ReorderNodeRequest, ReorderNodeResponse, SearchRequest, UpdateNodeRequest,
+    UpdateTaskNodeRequest, UpsertNodeWithParentRequest, WatchRequest,
 };
 
 /// gRPC adapter that owns shared handles to the core services.
@@ -1203,9 +1204,48 @@ async fn convert_domain_event(
                 node_type: node_type.clone(),
             })),
         }),
-        DomainEvent::RelationshipCreated { .. }
-        | DomainEvent::RelationshipUpdated { .. }
-        | DomainEvent::RelationshipDeleted { .. } => None,
+        DomainEvent::RelationshipCreated { relationship } => Some(NodeEvent {
+            event: Some(NodeEventKind::RelationshipCreated(relationship_to_proto(
+                relationship,
+            ))),
+        }),
+        DomainEvent::RelationshipUpdated { relationship } => Some(NodeEvent {
+            event: Some(NodeEventKind::RelationshipUpdated(relationship_to_proto(
+                relationship,
+            ))),
+        }),
+        DomainEvent::RelationshipDeleted {
+            id,
+            from_id,
+            to_id,
+            relationship_type,
+        } => Some(NodeEvent {
+            event: Some(NodeEventKind::RelationshipDeleted(
+                RelationshipDeletedPayload {
+                    id: id.clone(),
+                    from_id: from_id.clone(),
+                    to_id: to_id.clone(),
+                    relationship_type: relationship_type.clone(),
+                },
+            )),
+        }),
+    }
+}
+
+/// Translate a `RelationshipEvent` from the in-process domain channel
+/// into the proto wire form. `properties` is JSON-encoded as a string
+/// so the proto schema stays stable across additions to the
+/// underlying `serde_json::Value` payload — the desktop watcher
+/// re-parses it back to JSON before emitting the Tauri event.
+fn relationship_to_proto(
+    rel: &nodespace_core::db::events::RelationshipEvent,
+) -> RelationshipPayload {
+    RelationshipPayload {
+        id: rel.id.clone(),
+        from_id: rel.from_id.clone(),
+        to_id: rel.to_id.clone(),
+        relationship_type: rel.relationship_type.clone(),
+        properties: rel.properties.to_string(),
     }
 }
 
--- a/packages/desktop-app/src-tauri/src/watcher.rs
+++ b/packages/desktop-app/src-tauri/src/watcher.rs
@@ -197,5 +197,72 @@ fn forward(app: &AppHandle, event: nodespace_proto::nodespace::NodeEvent) {
                 error!("Failed to emit node:deleted for {}: {e}", d.node_id);
             }
         }
+        // Relationship variants — added so cloud-sync / cross-window
+        // hierarchy changes reach the frontend's reactiveStructureTree
+        // (issue #1202). Payload shape mirrors the in-process
+        // DomainEventForwarder's `relationship:*` Tauri events so
+        // `tauri-sync-listener.ts` works unchanged. `properties`
+        // arrives JSON-encoded on the wire (proto schema is stable);
+        // re-parse it here before emitting so the frontend gets a
+        // real object (the `has_child` listener reads
+        // `properties.order`).
+        NodeEventKind::RelationshipCreated(r) => emit_relationship(app, "relationship:created", r),
+        NodeEventKind::RelationshipUpdated(r) => emit_relationship(app, "relationship:updated", r),
+        NodeEventKind::RelationshipDeleted(r) => {
+            let payload = RelationshipDeletedOut {
+                id: r.id.clone(),
+                from_id: r.from_id,
+                to_id: r.to_id,
+                relationship_type: r.relationship_type,
+            };
+            if let Err(e) = app.emit("relationship:deleted", &payload) {
+                error!("Failed to emit relationship:deleted for {}: {e}", r.id);
+            }
+        }
+    }
+}
+
+/// Frontend payload for `relationship:created` / `relationship:updated`.
+/// Must stay structurally identical to the in-process
+/// `DomainEventForwarder`'s emission of the core `RelationshipEvent`
+/// struct (camelCase via serde rename), so `tauri-sync-listener.ts`
+/// works unchanged across the two forwarding paths.
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct RelationshipPayloadOut {
+    id: String,
+    from_id: String,
+    to_id: String,
+    relationship_type: String,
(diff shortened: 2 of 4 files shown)


malibio:
## 🔍 Code Review (automated)

### Summary

PR #1208 wires the three previously-dropped `DomainEvent::Relationship*` variants into the `WatchNodes` gRPC stream, giving the `watcher.rs` forwarding path parity with the in-process `DomainEventForwarder` for relationship events. The change touches the proto definition, the daemon's `convert_domain_event` adapter, the Tauri watcher, and the frontend sync listener.

The overall approach is architecturally sound and necessary for the cloud-sync / cross-window outliner correctness described in the PR. However, there is one functional bug in the frontend and a few secondary concerns worth addressing before merge.

---

### Findings

#### 🔴 [Improvement] `stripNodePrefix` is not applied to `member_of` and `mentions` paths in `tauri-sync-listener.ts`

**File:** `packages/desktop-app/src/lib/services/tauri-sync-listener.ts` — `relationship:created` and `relationship:deleted` listeners

The `stripNodePrefix` helper is correctly applied at the three `structureTree` call sites (`addChild`, `updateChildOrder`, `removeChild`), but the `member_of` and `mentions` branches are left untouched:

```ts
} else if (rel.relationshipType === 'member_of') {
  scheduleCollectionRefresh(rel.toId);   // still carries "node:<key>"
```

When these events arrive via the watcher path (daemon → gRPC → watcher.rs), `rel.toId` will be `"node:collection-xyz"` because `relationship_to_proto` passes the raw `RelationshipEvent.to_id` verbatim. The `scheduleCollectionRefresh` comparison `state.selectedCollectionId === affectedCollectionId` will then silently fail to refresh the member list for the currently-selected collection. This is latent today (watcher.rs is inert) but becomes a real regression the moment watcher.rs is activated.

**Fix:** run `stripNodePrefix` on every ID passed to any consumer, not only the `structureTree` path. Since `stripNodePrefix` is idempotent (bare IDs pass through unchanged), applying it broadly is safe.

---

#### 🟡 [Improvement] Silent JSON parse fallback in `emit_relationship` swallows malformed data

**File:** `packages/desktop-app/src-tauri/src/watcher.rs` — `emit_relationship` function

```rust
let props = serde_json::from_str(&r.properties)
    .unwrap_or(serde_json::Value::Object(Default::default()));
```

If `r.properties` is not valid JSON, the watcher silently discards the payload and emits an empty object. For `has_child` relationships this means `properties.order` is `undefined`, causing the `Date.now()` fallback to fire and sort the node to the tail of its parent — silent data ordering corruption with no trace in the logs.

```rust
let props = serde_json::from_str(&r.properties).unwrap_or_else(|e| {
    warn!("Failed to parse relationship properties for {}: {e}", r.id);
    serde_json::Value::Object(Default::default())
});
```

---

#### 🟡 [Improvement] No tests for the `stripNodePrefix` normalization logic

The `stripNodePrefix` helper and its application at the three call sites represent a behavioral contract: the watcher path delivers `"node:"`-prefixed IDs and the frontend normalizes them at the boundary. The existing tests use bare IDs — they would pass even if the normalization were completely removed. Adding a test that fires `relationship:created` with `"node:parent1"`/`"node:child1"` and asserts the structure tree receives bare IDs would pin this contract.

---

#### ⚪ [Nit] Redundant `.clone()` calls in `RelationshipDeleted` arm of `convert_domain_event`

The fields are destructured by move (`String`s owned by the match arm), so `.clone()` is unnecessary there. The `RelationshipCreated`/`RelationshipUpdated` arms pass a `&RelationshipEvent` reference and do require cloning — but the `Deleted` arm does not.

---

The `member_of`/`mentions` normalization gap is a latent correctness bug that will manifest as soon as `watcher.rs` is activated — fix before merge. Adding the warn log is low-cost. The test improvement is strongly recommended but could follow in a focused test PR if velocity matters.

mstomar125:
Review feedback addressed in `f6032b50`. Internal review pass: `cargo check`, `cargo clippy --no-default-features --lib -- -D warnings`, `bunx svelte-check --threshold error` (1639 files, 0 errors / 0 warnings) all clean.

### Blocker
- **`stripNodePrefix` applied to `member_of` and `mentions`** — both `relationship:created` and `relationship:deleted` listeners. `scheduleCollectionRefresh(toId)` now gets a bare id matching `state.selectedCollectionId`'s keyspace; the `mentions` log lines use the normalized form so future `loadChildrenTree(toId)` work has a single shape. `stripNodePrefix` is idempotent so applying broadly is safe.

### Improvement
- **Warn on JSON parse failure in `watcher.rs::emit_relationship`** — converted the silent `unwrap_or` to an explicit match with `warn!` logging the rel id, type, and parse error. The empty-object fallback stays so one bad payload doesn't stall the stream, but the silent ordering-corruption case now has a visible log trail.

### Nit not applied
- **Drop `.clone()` in `RelationshipDeleted` arm of `convert_domain_event`** — doesn't compile as suggested. `convert_domain_event` takes `event: &DomainEvent`, so the destructured `id`, `from_id`, `to_id`, `relationship_type` bind to `&String` references — clones are required to produce owned `String`s for the proto struct fields. The `Created`/`Updated` arms have the same root shape: they pass the inner `&RelationshipEvent` to `relationship_to_proto`, which clones inside. Different mechanical surface, same underlying need.

### Deferred
- **`stripNodePrefix` behavioral test** — deferred to a focused test PR per your "could follow" note. The boundary is now wider (5 call sites instead of 3), so the test will exercise more variants.

malibio:
## Re-Review: PR #1208 — "Forward RelationshipEvent variants on WatchNodes stream"

Reviewed feedback commit `f6032b50` ("Apply review feedback for #1208", 2026-05-21).

**3 of 4 findings addressed. 1 finding partially addressed. No new issues introduced.**

---

### Finding Status

**🔴 [Critical] `stripNodePrefix` not applied to `member_of`/`mentions` branches — ✅ Addressed**

`member_of` in both `relationship:created` and `relationship:deleted` now correctly applies `stripNodePrefix` to `toId` before passing it to `scheduleCollectionRefresh`. The `mentions` branches apply normalization in the log call. The fix is correct and `stripNodePrefix` is now consistently applied at every consumer call site. No issues here.

---

**🟡 Silent JSON parse fallback in `emit_relationship` swallows malformed data — ✅ Addressed**

The `unwrap_or` is replaced with an explicit `match` that emits a structured `warn!` log including `rel_id`, `rel_type`, and the parse error before falling back to the empty object. The fallback behaviour (continue streaming, emit empty props) is preserved, which is the right call for stream resilience. Implementation is clean.

---

**🟡 No tests for `stripNodePrefix` normalization contract — ⚠️ Partially addressed (deferred)**

The commit message explicitly defers this to a follow-up test PR, citing the reviewer's note that it could follow. The deferral is acceptable under the Pragmatic Quality framework — the critical correctness bug is fixed and the test debt is acknowledged. However, the deferred test PR should be tracked as a concrete issue, not left as an open-ended "future enhancement." If no issue exists yet, one should be filed before this merges so the behavioral contract does not silently erode.

**Action required:** Confirm a tracking issue exists (or file one) before merge. A one-line comment on the PR pointing to the issue number is sufficient.

---

**⚪ [Nit] Redundant `.clone()` in `RelationshipDeleted` arm — ✅ Correctly explained, no change needed**

Mayank's response is technically accurate. The `RelationshipDeleted` arm holds `r` as a `RelationshipPayload` value (moved from the `NodeEventKind` enum), and `r.id` is used twice: once moved into `payload.id` and once in the `error!` branch. The clone on line 213 is therefore load-bearing — removing it would cause a use-after-move compile error. My original analysis was incorrect in characterising this as redundant. No action needed.

---

### New Issues

None. The diff is minimal and focused. No regressions introduced. Verification steps noted in the commit message (`cargo check`, `cargo clippy --no-default-features --lib -D warnings`, `bunx svelte-check`) are appropriate for the change surface.

---

### Overall Recommendation

**Approve with one follow-up gate:** confirm a tracking issue exists for the `stripNodePrefix` behavioral test before merge. All critical and important findings are resolved. The deferred test is the only remaining item and it is low-risk given the normalization code is already exercised by the existing integration path.

malibio (approved):
Critical and important findings resolved. Approving pending confirmation that a tracking issue exists for the deferred stripNodePrefix behavioral test (see re-review comment).

malibio:
## ⚠️ Re-Review Complete — One gate before merge

All critical findings resolved. One outstanding item before merging:

- **File a tracking issue** for the `stripNodePrefix` behavioral test (the deferred test PR mentioned in commit `f6032b50`). Drop the issue number in a PR comment here, then this is clear to merge.

mstomar125:
Tracking issue filed: **NodeSpaceAI/nodespace-core#1209** — *Behavioral test for tauri-sync-listener stripNodePrefix normalization.*

Issue body lists the five `stripNodePrefix` call sites that need coverage, the acceptance criteria (vitest against the mocked Tauri `listen` API, asserting bare-id keys flow through to `structureTree` and `scheduleCollectionRefresh`), and the scope boundary (E2E + daemon-side proto tests are separate follow-ups).

Added to the project board in **Backlog** so it has no immediate owner — happy to pick it up after #1208 / #49 land if no one else grabs it first.

This unblocks #1208 for merge.

Issue #795, opened 2026-01-06
Sync-ready relationship timestamps: preserve created_at on same-parent reorder


malibio:
## Problem

Currently, `move_node` uses DELETE + RELATE (create new) pattern for ALL moves, including same-parent reorders:

```sql
DELETE relationship WHERE out = $node_id AND relationship_type = 'has_child';
RELATE $parent_id->relationship->$node_id CONTENT {
    relationship_type: 'has_child',
    properties: { order: $order },
    created_at: time::now(),  -- loses original timestamp
    modified_at: time::now(),
    version: 1  -- resets version
};
```

This causes issues for cloud sync and conflict resolution:
1. **Lost `created_at`** - Original relationship creation time is lost on every reorder
2. **Reset `version`** - OCC version resets to 1, breaking conflict detection
3. **Indistinguishable operations** - Can't tell "reorder within parent" from "move to new parent"

## Solution

Differentiate between same-parent reorder and cross-parent move:

### Same-parent reorder (UPDATE)
```sql
UPDATE relationship 
SET properties.order = $order, 
    modified_at = time::now(), 
    version = version + 1
WHERE in = $parent AND out = $child AND relationship_type = 'has_child';
```

### Cross-parent move (DELETE + CREATE)
```sql
-- Current behavior is correct - new relationship = new created_at
DELETE relationship WHERE out = $node_id AND relationship_type = 'has_child';
RELATE $new_parent_id->relationship->$node_id CONTENT { ... };
```

## Acceptance Criteria

- [ ] `move_node` detects if parent is changing vs. just reordering
- [ ] Same-parent reorder uses UPDATE (preserves `created_at`, increments `version`)
- [ ] Cross-parent move uses DELETE + CREATE (current behavior)
- [ ] `modified_at` updated on both operations
- [ ] Tests verify timestamp preservation on same-parent reorder
- [ ] Tests verify `version` increments on reorder (for OCC)

## Context

- Related to Universal Relationship Architecture (Issue #788, PR #793)
- Required for cloud sync and real-time collaboration features
- Affects `has_child` relationships primarily (ordering)

## Labels

`enhancement`, `backend`, `sync`

PR #1214, opened 2026-05-27
Make create_parent_edge idempotent on already-existing edges (fixes nodespace-sync#77, alice-side)

mstomar125:
## Summary

Backend half of [`nodespace-sync#77`](https://github.com/NodeSpaceAI/nodespace-sync/issues/77). Makes `NodeService::create_parent_edge` idempotent on already-existing edges so the sync pull-side's echo replay doesn't re-position the just-created child off its local order.

## What changed

`packages/core/src/services/node_service.rs` — `create_parent_edge` now short-circuits when:

1. `insert_after_node_id` is `None` (no explicit reorder intent), AND
2. `get_parent(child_id)` reports the child is already under `parent_id`

The guard is narrowed via `insert_after_node_id.is_none()` so a caller that passes an explicit hint still triggers a real reorder (covered by `test_create_parent_edge_reorders_when_explicit_hint_against_existing_edge`).

`packages/core/src/db/surreal_store.rs` — added `test_move_node_same_parent_no_hint_moves_to_beginning` to lock the store-level contract used by `reorder_node(_, _, None)` (MCP "move to first position" path).

The three existing `test_same_parent_*` tests were already updated to use explicit `insert_after_sibling` for their reorder calls.

## Scope note: what this PR does NOT do

An earlier revision of this PR ALSO translated `None → last_child_id` inside `create_parent_edge` to fix the sync receive-side bug (bob's view of alice's new child landed at the top of the parent's list). That translation has been **stripped** in `e25fa571` because it conflated two callers with opposite intents:

| Caller | Intended meaning of `None` |
|---|---|
| Sync pull-side `apply_remote` | "no order info from cloud, append at end" |
| Outliner Enter-at-position-0 | "insert at beginning of the parent's list" |

Core can't disambiguate at the API layer. The fix lives at the call sites instead — sync's pull-side computes `last_child_id` explicitly in [`NodeSpaceAI/nodespace-sync#79`](https://github.com/NodeSpaceAI/nodespace-sync/pull/79). The outliner's "insert at beginning" UX is preserved by core's unchanged `move_node(_, _, None) = "insert at beginning"` semantic.

## Test plan

- [x] `cargo test -p nodespace-core --lib` — 1191 → 1193 passing (+2 new, no removals)
- [x] `cargo clippy -p nodespace-core --lib -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] Manual verification: two-window same-device demo, alice types + Enter at end of text → new sibling appears directly below in both windows (requires `nodespace-sync#79` for the receive-side correction)

## Related

- [#1213](https://github.com/NodeSpaceAI/nodespace-core/pull/1213) — setNode skip-while-editing guard (`nodespace-sync#76`). Independent of this PR.
- [#1215](https://github.com/NodeSpaceAI/nodespace-core/pull/1215) — frontend resilience for `#77` (tauri-sync-listener fallback + structureTree-based stale check). Independent of this PR.
- [`nodespace-sync#79`](https://github.com/NodeSpaceAI/nodespace-sync/pull/79) — receive-side companion that passes the explicit `last_child_id` hint. Required for the full `#77` fix.
- [#1216](https://github.com/NodeSpaceAI/nodespace-core/issues/1216) — follow-up suggestion: introduce an `InsertPosition` enum so future callers can express intent without overloading `Option`.

### DIFF
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -2788,7 +2788,14 @@ impl SurrealStore {
                     FractionalOrderCalculator::calculate_order(last_order, None)
                 }
             } else {
-                // No insert_after_sibling specified, insert at beginning
+                // No insert_after_sibling specified, insert at beginning. The
+                // MCP handlers (`handle_insert_child_at_index` with index=0
+                // and `handle_move_child_to_index` with target_index=0) rely
+                // on this: they pass `None` to mean "move to the first
+                // position". Hierarchical-create callers that want "append at
+                // end" semantics (the sync layer's pull-side `apply_remote`
+                // among them) translate `None` to `Some(last_child_id)` at
+                // their layer — see `NodeService::create_parent_edge`.
                 let first_order = relationships.first().map(|e| e.order);
                 FractionalOrderCalculator::calculate_order(None, first_order)
             }
@@ -7310,10 +7317,15 @@ mod tests {
 
     #[tokio::test]
     async fn test_same_parent_reorder_preserves_created_at() -> Result<()> {
-        // Issue #795: Same-parent reorder should preserve created_at
+        // Issue #795: Same-parent reorder should preserve created_at.
+        //
+        // Use an explicit `insert_after_sibling` for the reorder. Calling
+        // `move_node(child, parent, None)` on an already-present edge is a
+        // no-op in nodespace-sync#77's fix (the pull-side echoes its own
+        // writes as `(child, parent, None)` and the previous behavior of
+        // re-running the order calculation dragged the row to the top).
         let (store, _temp_dir) = create_test_store().await?;
 
-        // Create parent and two children
         let parent = store
             .create_node(
                 Node::new("text".to_string(), "Parent".to_string(), json!({})),
@@ -7322,37 +7334,33 @@ mod tests {
             )
             .await?;
 
-        let _child1 = store
+        let child1 = store
             .create_child_node_atomic(&parent.id, "text", "Child 1", json!({}), None)
             .await?;
         let child2 = store
             .create_child_node_atomic(&parent.id, "text", "Child 2", json!({}), None)
             .await?;
 
-        // Get original relationship metadata for child2
-        let original_metadata = get_relationship_metadata(&store, &child2.id)
+        let original_metadata = get_relationship_metadata(&store, &child1.id)
             .await?
             .expect("Relationship should exist");
         let original_created_at = original_metadata.0.clone();
 
-        // Wait a tiny bit to ensure time difference
         tokio::time::sleep(tokio::time::Duration::from_millis(10)).await;
 
-        // Reorder child2 to be before child1 (same parent, just reordering)
-        store.move_node(&child2.id, Some(&parent.id), None).await?;
+        // Reorder: move child1 to be AFTER child2 (real position change).
+        store
+            .move_node(&child1.id, Some(&parent.id), Some(&child2.id))
+            .await?;
 
-        // Get new relationship metadata
-        let new_metadata = get_relationship_metadata(&store, &child2.id)
+        let new_metadata = get_relationship_metadata(&store, &child1.id)
             .await?
             .expect("Relationship should still exist");
 
-        // created_at should be PRESERVED (same value)
         assert_eq!(
             new_metadata.0, original_created_at,
             "Same-parent reorder should preserve created_at"
         );
-
-        // modified_at should be UPDATED (different value)
         assert_ne!(
             new_metadata.1, original_metadata.1,
             "Same-parent reorder should update modified_at"
@@ -7363,10 +7371,14 @@ mod tests {
 
     #[tokio::test]
     async fn test_same_parent_reorder_increments_version() -> Result<()> {
-        // Issue #795: Same-parent reorder should increment version for OCC
+        // Issue #795: Same-parent reorder should increment version for OCC.
+        //
+        // Like `test_same_parent_reorder_preserves_created_at`, pass an
+        // explicit `insert_after_sibling` so this exercises an actual
+        // reorder rather than the now-no-op `(child, parent, None)` shape
+        // that nodespace-sync#77 relies on to short-circuit pull-side echoes.
         let (store, _temp_dir) = create_test_store().await?;
 
-        // Create parent and child
         let parent = store
             .create_node(
                 Node::new("text".to_string(), "Parent".to_string(), json!({})),
@@ -7375,25 +7387,26 @@ mod tests {
             )
             .await?;
 
-        let child = store
-            .create_child_node_atomic(&parent.id, "text", "Child", json!({}), None)
+        let child1 = store
+            .create_child_node_atomic(&parent.id, "text", "Child 1", json!({}), None)
+            .await?;
+        let child2 = store
+            .create_child_node_atomic(&parent.id, "text", "Child 2", json!({}), None)
             .await?;
 
-        // Get original version
-        let original_metadata = get_relationship_metadata(&store, &child.id)
+        let original_metadata = get_relationship_metadata(&store, &child1.id)
             .await?
             .expect("Relationship should exist");
         let original_version = original_metadata.2;
 
-        // Reorder (same parent)
-        store.move_node(&child.id, Some(&parent.id), None).await?;
+        store
+            .move_node(&child1.id, Some(&parent.id), Some(&child2.id))
+            .await?;
 
-        // Get new version
-        let new_metadata = get_relationship_metadata(&store, &child.id)
+        let new_metadata = get_relationship_metadata(&store, &child1.id)
             .await?
             .expect("Relationship should still exist");
 
-        // Version should be incremented
         assert_eq!(
             new_metadata.2,
             original_version + 1,
@@ -7404,6 +7417,7 @@ mod tests {
 
         Ok(())
     }
+
 
     #[tokio::test]
     async fn test_cross_parent_move_creates_new_relationship() -> Result<()> {
@@ -7464,10 +7478,15 @@ mod tests {
 
     #[tokio::test]
     async fn test_multiple_same_parent_reorders_accumulate_version() -> Result<()> {
-        // Issue #795: Multiple reorders should accumulate version
+        // Issue #795: Multiple reorders should accumulate version.
+        //
+        // Each reorder uses an explicit `insert_after_sibling` (alternating
+        // between the two pivot children) so it's a genuine position change.
+        // The previous shape of this test used `move_node(child, parent,
+        // None)` which is now a no-op (preserves existing order) per the
+        // nodespace-sync#77 fix.
         let (store, _temp_dir) = create_test_store().await?;
 
-        // Create parent and child
         let parent = store
             .create_node(
                 Node::new("text".to_string(), "Parent".to_string(), json!({})),
@@ -7476,19 +7495,32 @@ mod tests {
             )
             .await?;
 
+        let pivot_a = store
+            .create_child_node_atomic(&parent.id, "text", "Pivot A", json!({}), None)
+            .await?;
+        let pivot_b = store
+            .create_child_node_atomic(&parent.id, "text", "Pivot B", json!({}), None)
+            .await?;
         let child = store
             .create_child_node_atomic(&parent.id, "text", "Child", json!({}), None)
             .await?;
 
-        // Initial version should be 1
         let initial_metadata = get_relationship_metadata(&store, &child.id)
             .await?
             .expect("Relationship should exist");
         assert_eq!(initial_metadata.2, 1, "Initial version should be 1");
 
-        // Reorder multiple times
+        // Alternate insert_after between the two pivots so every reorder is
+        // a real position change. Four iterations → version 2..=5.
         for expected_version in 2..=5 {
-            store.move_node(&child.id, Some(&parent.id), None).await?;
+            let pivot = if expected_version % 2 == 0 {
+                &pivot_a.id
+            } else {
+                &pivot_b.id
+            };
+            store
+                .move_node(&child.id, Some(&parent.id), Some(pivot))
+                .await?;
 
             let metadata = get_relationship_metadata(&store, &child.id)
                 .await?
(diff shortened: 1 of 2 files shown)


mstomar125:
## Pragmatic code review — PR #1214

**CI:** ⏳ No checks reported on `fix/create-parent-edge-idempotent` — verify CI is wired up before merge.
**Scope:** +138 / -52 across 2 files. `NodeService::create_parent_edge` becomes idempotent + `None`-hint translates to append; three existing tests rewritten to use explicit `insert_after_sibling`; one test renamed and assertion flipped.
**Requirements:** Backend half of `nodespace-sync#77`. Both fixes (idempotency + None→append translation) are real and named in the PR body. The rewritten tests acknowledge that `None`-as-reorder was load-bearing for old behavior.

### Findings

🔴 **Idempotency guard treats `create_parent_edge(child, parent, Some(other_sibling))` as a no-op when child is already under parent** (`node_service.rs:4554-4564`)
```rust
if let Some(existing_parent) = self.get_parent(child_id).await? {
    if existing_parent.id == parent_id {
        return Ok(());
    }
}
```
The guard short-circuits *unconditionally* when the child is already under the target parent — but the caller may have passed a non-None `insert_after_node_id` precisely because they want to *reposition* an existing child. The PR body says "Callers that want to actually reorder an existing child should use `reorder_child`" — but that's an unwritten contract. Today the only known caller of `create_parent_edge` is sync's pull-side passing `None`, but the function is `pub` on `NodeService` and any future caller passing `Some(sibling)` will get silent no-ops. Either:

1. Narrow the guard to `if existing_parent.id == parent_id && insert_after_node_id.is_none()` (matches the actual sync-echo case).
2. Add a debug assertion or warn-level log when `Some` hint is provided with already-present edge.
3. Make the function take an enum (`EdgeIntent::CreateOrIgnore` vs `EdgeIntent::Reorder`) so the contract is in the type.

Option 1 is the smallest fix and preserves the sync echo behavior the PR is targeting. Recommend at minimum (1) + a test that calls `create_parent_edge` with a `Some` hint against an existing edge and asserts the reorder actually happens.

🟡 **`None`-translation costs an extra `get_children` round-trip on every edge creation** (`node_service.rs:4574-4582`)
```rust
match self.get_children(parent_id).await {
    Ok(children) => children.last().map(|c| c.id.clone()),
    Err(_) => None,
}
```
Every CREATE with no explicit insert-after now does:
1. `get_parent(child_id)` (idempotency check)
2. `get_children(parent_id)` (None translation)
3. `move_node(...)` (the actual write, which internally re-queries children to compute the fractional order)

For a parent with many children, `get_children` is O(n) and runs before every CREATE. Sync echo replays may produce a burst of these. Profile-impact unknown but the design path can be: have `move_node` accept a `InsertPosition::{Append, AtBeginning, After(id)}` enum and push the None→Append translation into the store layer where it can share the children lookup.

🟡 **`Err(_) => None` silently swallows DB errors and falls back to "insert at beginning"** (`node_service.rs:4580`)
If `get_children` returns Err (transient DB issue), we silently degrade to the *exact* pre-fix behavior the PR is trying to eliminate — the new node lands at the beginning of the list. Better: propagate the error (the caller can retry), or at minimum `tracing::warn!` with the parent_id and error so the degradation is observable.

🟡 **Idempotency guard's `get_parent` is also fallible and not retried** (`node_service.rs:4554-4564`)
`get_parent` propagates errors via `?` — a transient DB error during the sync echo replay aborts the entire `create_parent_edge` call, which the sync pull-side may or may not handle gracefully. Compare with the `move_node` call below, which retries up to 3 times for sibling-not-found. The idempotency check is the cheaper failure mode and worth retrying or at least documenting the propagation contract.

🟡 **Concurrency: read-then-write race between `get_children().last()` and `move_node`** (`node_service.rs:4574-4582, 4614`)
Two threads racing `create_parent_edge(child_X, parent, None)` and `create_parent_edge(child_Y, parent, None)`:
- T1 reads `get_children` → last = `existing_child`
- T2 reads `get_children` → last = `existing_child`
- T1 calls `move_node(child_X, parent, Some(existing_child))` → child_X inserted after existing_child
- T2 calls `move_node(child_Y, parent, Some(existing_child))` → child_Y inserted *between* existing_child and child_X, not at the actual end

For sync echo replays this can manifest as out-of-order siblings when the cloud delivers a batch of new edges. The fractional-order math handles it (no collision), but the visual order may not match the cloud-canonical order. Worth either serializing edge creation per-parent (already done elsewhere?) or accepting the eventually-consistent ordering with a doc comment naming it.

🟡 **Test rewrites lose coverage of the `None`-hint behavior for in-place reorder** (`surreal_store.rs:7322-7507`)
Three tests previously called `move_node(child, parent, None)` to exercise same-parent reorder. They've been rewritten to pass an explicit sibling. The PR body and updated comments correctly explain that `None` on an already-existing edge is now a no-op (per `create_parent_edge`'s guard) — but the lower-level `move_node(_, _, None)` *isn't* idempotent and still means "move to beginning". The rewrites work around that, but the gap is: nothing in the test suite now exercises `move_node`'s no-op-on-redundant-call behavior at the store layer. If a future change introduces an idempotency check at the store layer too (or removes it from `create_parent_edge`), these tests won't catch the regression. Add a small unit test asserting `move_node(child, parent, None)` for an already-present edge moves child to beginning (current behavior) — locks in the layered contract.

### Nits

🟢 **Nit: Doc comments name `nodespace-sync#77` directly** (`node_service.rs:4553, 4566`)
Core's `NodeService` is now coupled (in comments) to a sync-side issue number. Acceptable for the present moment because the issue documents the root cause, but consider promoting the actual invariants (idempotency contract, None-as-append semantics) to function-level rustdoc that doesn't depend on reading an external issue.

🟢 **Nit: Double blank line in test module** (`node_service.rs:7421-7422` and `surreal_store.rs:7421`)
Pre-existing in some places, introduced anew in others — `cargo fmt` should normalize. Not a blocker if clippy is clean.

🟢 **Nit: `test_insert_at_end_by_default` rename keeps the `#[serial(sibling_ordering)]` attribute** (`node_service.rs:9554`)
Correct — the serial guard is about parallel-test interference, not about the semantic being tested. Worth a one-line comment naming why the serial guard stays.

🟢 **Nit: Idempotency log is `debug!`** (`node_service.rs:4559-4563`)
Sync echo replays will fire this on every cloud-echoed edge. `debug!` is the right level for default operation, but during the first weeks after rollout an `info!` (with rate-limiting) would help confirm the path is firing as expected in production.

🟢 **Nit: `effective_after_owned: Option<String>` + `resolved_insert_after: Option<&str>`** (`node_service.rs:4574-4584`)
The two-binding pattern (owned String to extend lifetime, then borrow) is a Rust idiom but worth a one-line comment naming why both exist — future readers will wonder.

🟢 **Nit: PR body says "1191 passed / 0 failed (one rename of an existing test; no net test count change)"** — matches the diff. Worth double-checking the rewritten tests still cover Issue #795's actual scenario (the issue is about same-parent reorder preserving `created_at`); rewriting to use a real position change is arguably *better* coverage, but call it out.

### Summary

The two fixes are real and the test rewrites are honest about the semantic shift. The 🔴 is about the idempotency guard's blast radius — narrowing it to the actual sync-echo case (`None` hint) preserves the intent without breaking the `pub` API's reordering contract. The 🟡s are mostly about defensive sharpness (error handling, race windows, missing test coverage at the layer below the fix).

### Verdict: REQUEST CHANGES

mstomar125:
Pushed review-feedback addressed in `cc9883ef`.

**Real concerns**
- 🔴 **Idempotency narrowed to no-explicit-hint calls** — previously the guard short-circuited every `create_parent_edge(child, parent, *)` with `existing_parent.id == parent_id`, including calls passing `Some(target_sibling)` to explicitly reposition. Now gated on `insert_after_node_id.is_none()`. New test `test_create_parent_edge_reorders_when_explicit_hint_against_existing_edge` locks the narrowed contract.
- 🟡 **`Err(_) => None` silent degradation removed** — `get_children` errors now propagate via `?`. Previously the silent degradation re-triggered the exact "insert at beginning" bug this path was added to fix.
- 🟡 **`get_parent` retry contract documented** — comment names the propagation contract and contrasts with the 10×100ms retry budget on the `move_node` loop (different consistency window).
- 🟡 **Concurrency window documented** — read-then-write between `get_children` and `move_node` produces distinct fractional orders (no collision) but possibly inconsistent between-which-siblings placement under concurrent calls. Acceptable for sync echo replay (eventually consistent); callers needing strict ordering should serialise or use `reorder_child` afterwards.
- 🟡 **Lost store-layer coverage restored** — added `test_move_node_same_parent_no_hint_moves_to_beginning` at the store layer. The MCP `reorder_node(_, _, None)` "move to first position" path depends on this semantic. A future refactor that flips the None default to "append at end" at the store layer would silently break MCP — this test catches that.
- Fixed two stale comments in test bodies that described the no-op as living at `move_node`; it lives at `NodeService::create_parent_edge` (the layer where the calling convention splits cleanly between CREATE and REORDER).

**Cost note (acknowledge)**
- 🟡 Extra `get_children` round-trip per CREATE without explicit hint — kept as-is for clarity. A future refactor could push the None translation into the store's `move_node` (sharing the children lookup it already does internally) if profiling shows it's a hotspot. Noted in the doc comment.

**Test deltas**
- 1191 → 1193 passing (+2 new tests, no removals).
- `cargo clippy -p nodespace-core --lib -- -D warnings` — clean.

mstomar125:
## Pragmatic code review (re-review) — PR #1214

**Review Type:** Re-Review (scope: commit `cc9883ef` "Address PR #1214 review feedback").
**CI:** ⏳ Still no checks reported on `fix/create-parent-edge-idempotent`. Same flag as round 1.

### Previous review summary
- Previous 🔴: 1 — addressed.
- Previous 🟡: 5 — addressed: 4, documented: 1.
- Previous 🟢: 5 — addressed: 3, no-op acceptable: 2.
- New findings this pass: 0 🔴, 0 🟡, 1 🟢.

✅ **🔴 Idempotency blast radius narrowed** — guard now gated on `insert_after_node_id.is_none()`. Calls passing an explicit hint against an already-present edge route through `move_node` and reposition. New test `test_create_parent_edge_reorders_when_explicit_hint_against_existing_edge` directly exercises this:
```rust
service.create_parent_edge(&child1_id, &parent_id, Some(&child2_id)).await.unwrap();
// assertion: after[0] = "Child 2", after[1] = "Child 1"
```
A future regression that widens the guard back to "any same-parent edge" fails this test. Clean fix.

✅ **`Err(_) => None` silent degradation** — `get_children` now propagates via `?` instead of swallowing. The previous fallback to "insert at beginning" on a transient DB error — which would have silently re-triggered the bug this path was added to fix — is gone.

✅ **`get_parent` retry contract** — comment added explicitly contrasting the propagation-via-`?` for `get_parent` with the 10×100ms retry budget on `move_node`'s sibling-not-found loop. Makes the layered consistency-window story legible.

✅ **Concurrency window** — read-then-write race between `get_children().last()` and `move_node` is now documented at the call site, naming exactly the failure mode (distinct fractional orders, but between-which-siblings placement may not match cloud-canonical order) and the acceptable-for-sync-echo-replay rationale.

✅ **Lost store-layer coverage** — new `test_move_node_same_parent_no_hint_moves_to_beginning` directly asserts `move_node(child, parent, None)` for an already-present edge moves the child to the beginning. The comment names the MCP `reorder_node(_, _, None)` dependency explicitly — a future refactor that flips that default would break MCP and fail this test.

✅ **Stale test comments** — fixed: comments now correctly attribute the no-op to `NodeService::create_parent_edge` rather than to `move_node`.

🟡 **Extra `get_children` round-trip per CREATE** — explicitly kept as-is with a comment naming the future-optimization path (push None translation into `move_node` to share its existing children lookup). Acceptable as documented.

### New findings

🟢 **Nit: `test_create_parent_edge_reorders_when_explicit_hint_against_existing_edge` doesn't assert no log spam on the reorder path** (`node_service.rs:9625-9684`)
The narrowed guard now keeps `tracing::debug!(... "edge already exists with no reorder hint, treating as no-op" ...)` for the legitimate no-op path. A future refactor that accidentally widens the guard would still take the no-op branch *and* emit that log under the explicit-hint call. The test asserts on the post-condition (reorder happened) but not on the log message — a `tracing_test` capture confirming the no-op log did NOT fire would catch silently-widened guards. Style; not blocking.

### Summary

Every round-1 🔴 and 🟡 is either fixed in code, locked in by a test, or explicitly documented with a rationale. The narrowed idempotency contract is the right shape — sync-echo idempotency lives where the calling convention splits cleanly between CREATE and REORDER, and the test suite now covers both the sync-echo no-op and the explicit-hint reorder. The store-layer test restores the layered contract coverage the rewrites had dropped.

### Verdict: APPROVE

mstomar125:
Force-pushed `e25fa571` after a late audit found a real UX regression. Stripping the `None → last_child_id` translation while keeping the idempotency guard.

**Why** — the translation conflated two callers with opposite intents for the same `None`:

| Caller | Intent |
|---|---|
| Sync pull-side echo (this PR's target) | "no order info, append at end" |
| Outliner Enter-at-position-0 (legitimate frontend UX) | "insert at beginning of the parent's list" |

The previous revision made the outliner's "insert above current line" flow visually land the new line at the bottom of the parent's children (optimistic above → ~200ms later moves to bottom on backend confirm). Caught by re-reading the frontend's `shouldCreateNodeAbove` branch.

**Resolution** — resolve at the call site, not at the backend:
- This PR keeps **only** the idempotency-on-existing-edge guard (clear correctness fix).
- The sync receive-side `append at end` semantic moves to a companion sync PR: [`NodeSpaceAI/nodespace-sync#79`](https://github.com/NodeSpaceAI/nodespace-sync/pull/79) where `apply_remote` computes `last_child_id` explicitly and passes it as the hint.
- The outliner's "insert at beginning" UX is preserved because core's `move_node(_, _, None) = "insert at beginning"` semantic is unchanged.

**API-improvement follow-up** — filed [#1216](https://github.com/NodeSpaceAI/nodespace-core/issues/1216) suggesting an `InsertPosition::{Beginning, End, After(id)}` enum so future callers can express intent without overloading `Option`. That's the right long-term shape; today's resolution keeps the change minimal.

**Test delta vs origin/main**
- 1191 → 1193 passing (+2 new tests, no removals).
- `cargo clippy -p nodespace-core --lib -- -D warnings` — clean.

**Updated PR description** above reflects the new scope.

PR #1215, opened 2026-05-27
Preserve structureTree order on sync echoes + fix stale-sibling check (fixes nodespace-sync#77, frontend)

mstomar125:
## Summary

Frontend resilience half of [`nodespace-sync#77`](https://github.com/NodeSpaceAI/nodespace-sync/issues/77). Companion to [#1214](https://github.com/NodeSpaceAI/nodespace-core/pull/1214) (\`fix/create-parent-edge-idempotent\`) — each PR removes one piece of the "new node lands at the top of the list" failure mode.

### Bug 1 — \`tauri-sync-listener\` clobbers the local order on cloud echo

The cloud LIVE-SELECT echo of a freshly-created \`has_child\` edge arrives with \`properties.order = undefined\` because the sync layer's \`RELATE ... SET rel_type = \$rt\` doesn't push the order field to cloud. The previous fallback \`?? Date.now()\` overwrote the correct local fractional order (set by the local optimistic path and the preceding local domain event) with a value 11 orders of magnitude larger — relocating the just-created sibling out of position.

**Fix**: when the incoming order is missing, look up the existing entry in \`structureTree\` and preserve its order. If the child is brand-new and we genuinely have nothing better, fall back to \`Date.now()\` so it lands somewhere stable.

### Bug 2 — stale-sibling check uses an unreliable field

The persistence-time check that decides whether to clear \`insertAfterNodeId\` before firing a CREATE compared \`siblingNode.parentId\` against \`currentNode.parentId\`. Nodes loaded via \`getChildrenTree\` come back without \`parentId\` populated on the wire, so every Enter-key insertion looked "stale" — the hint got cleared and the backend received \`None\`, falling into the wrong default branch.

**Fix**: use \`structureTree.getParent\` (the authoritative source for hierarchy via \`has_child\` edges) instead. Preserves the hint when \`structureTree\` has no opinion (sibling not yet in tree) so the backend can still resolve it via its own retry loop.

## Stacking note

This PR is **stacked on top of [#1213](https://github.com/NodeSpaceAI/nodespace-core/pull/1213)** (\`fix/setnode-skip-while-editing\`) — both PRs touch \`shared-node-store.svelte.ts\` in non-overlapping regions, but stacking keeps the review diff legible. Once #1213 merges, this PR rebases trivially against main.

## What changed

- \`packages/desktop-app/src/lib/services/tauri-sync-listener.ts\` — the \`has_child\` branch of the \`relationship:created\` handler now preserves the existing structureTree order when \`rel.properties?.order\` is missing.
- \`packages/desktop-app/src/lib/services/shared-node-store.svelte.ts\` — the stale-sibling check (in the CREATE persistence path) now consults \`structureTree.getParent\` instead of \`siblingNode.parentId\`.
- \`packages/desktop-app/src/tests/services/shared-node-store-skip-while-editing.test.ts\` — 1 new test case for the structureTree-based check.

## Test plan

- [x] Frontend shared-node-store family tests: 7 (from #1213) + 1 new = 8 passing in the new test file; 247 passing across the wider store family
- [x] \`bun run --cwd packages/desktop-app quality:fix\` — clean
- [x] Manual verification: two-window same-device demo. The frontend half alone is **not sufficient** — the receive-side default-to-beginning still bites without the backend half — but combined with #1214 it restores correct ordering across the cloud round-trip in both windows.

## Related PRs

- [#1213](https://github.com/NodeSpaceAI/nodespace-core/pull/1213) — \`setNode\` skip-while-editing guard (\`nodespace-sync#76\`). This PR stacks on it.
- [#1214](https://github.com/NodeSpaceAI/nodespace-core/pull/1214) — backend half of \`nodespace-sync#77\`. Independent of this PR; both need to land for the full fix.

### DIFF
--- a/packages/desktop-app/src/lib/services/shared-node-store.svelte.ts
+++ b/packages/desktop-app/src/lib/services/shared-node-store.svelte.ts
@@ -525,9 +525,68 @@ export class SharedNodeStore {
   // Populated by the skip-while-editing guard in setNode() instead of
   // mutating the reactive `nodes` Map (which would trigger Svelte re-renders
   // that remount the textarea and reset selectionStart — see
-  // nodespace-sync#77). Consumed by the UpdateNode persistence path so the
+  // nodespace-sync#76). Consumed by the UpdateNode persistence path so the
   // next OCC carries the freshest version.
   private serverConfirmedVersions = new Map<string, number>();
+
+  /**
+   * Test-only accessor for the non-reactive server-confirmed-version cache.
+   * Production code consumes this cache via the persistence path
+   * (UpdateNode OCC version selection). Tests use this to assert the cache
+   * was populated/cleared as expected — see
+   * `shared-node-store-skip-while-editing.test.ts`.
+   */
+  peekServerConfirmedVersion(nodeId: string): number | undefined {
+    return this.serverConfirmedVersions.get(nodeId);
+  }
+
+  /**
+   * Returns the version the next UpdateNode RPC would send for this node,
+   * applying the same lookup the persistence closure uses:
+   *
+   *   serverConfirmedVersions[nodeId] ?? localNode.version ?? 1
+   *
+   * Exposed so the skip-while-editing test suite can lock the contract in
+   * place — a future refactor that drops the cache read at the
+   * persistence call site would fail the corresponding test instead of
+   * silently re-introducing the OCC-defeat bug from nodespace-sync#76.
+   */
+  computeOccVersionForUpdate(nodeId: string): number {
+    const confirmed = this.serverConfirmedVersions.get(nodeId);
+    if (typeof confirmed === 'number') return confirmed;
+    const local = this.nodes.get(nodeId);
+    return local?.version ?? 1;
+  }
+
+  /**
+   * Heuristic for "is this database broadcast plausibly an echo of *this
+   * client's own* write?" Used by the skip-while-editing guard to decide
+   * whether to stash the broadcast's `node.version` for the next OCC.
+   *
+   * Returns `true` when the local optimistic content equals or extends the
+   * incoming content — i.e., the incoming is an older snapshot of what we
+   * already have. That covers both shapes the daemon's own confirmation
+   * loops back through:
+   *
+   *   - exact match: the broadcast is the just-confirmed write, no further
+   *     keystrokes since;
+   *   - local extension: the user typed more characters in the optimistic
+   *     state while the daemon was confirming the previous version.
+   *
+   * Returns `false` when the local content diverges from the incoming —
+   * which signals a *foreign* write (e.g., another client editing the
+   * same node concurrently). In that case the cache is intentionally left
+   * untouched so the next UpdateNode RPC uses the local `node.version`,
+   * conflicts with the server's newer version, and surfaces the
+   * divergence via the normal OCC path.
+   */
+  private isPlausibleOwnEcho(local: Node, incoming: Node): boolean {
+    const localContent = local.content ?? '';
+    const incomingContent = incoming.content ?? '';
+    if (localContent === incomingContent) return true;
+    // Local is an extension of incoming → user kept typing past the confirm.
+    return localContent.startsWith(incomingContent);
+  }
 
   // Test error tracking (populated only in NODE_ENV='test', cleared between tests)
   private testErrors: Error[] = [];
@@ -1180,27 +1239,54 @@ export class SharedNodeStore {
     // remount the textarea (the `{#if isEditing}` block in base-node.svelte)
     // and reset selectionStart.
     //
+    // `source.type === 'database'` is the contract for "this update came from
+    // the daemon's domain-event broadcast" — see UpdateSource in
+    // `$lib/types/update-protocol`. Local user actions use `'viewer'`. The
+    // guard relies on no other producer of `'database'` events bypassing the
+    // intended skip behavior; the only consumers today are
+    // `tauri-sync-listener` and `browser-sync-service`.
+    //
     // Fixes nodespace-sync#76 (typing corruption: chars dropped/replaced
     // under sustained input as the optimistic store is clobbered by the
     // daemon's own confirmation looped back through the WatchNodes stream).
+    //
+    // Compute the predicates once into locals: (a) `hasPending` does three
+    // Map lookups, and (b) the coordinator transitions a node between its
+    // pending/executing/queued maps, so reading it twice can return
+    // different answers — the log message would otherwise contradict the
+    // branch taken.
+    const isFocused = focusManager.editingNodeId === node.id;
+    const hasPending = PersistenceCoordinator.getInstance().hasPending(node.id);
     if (
       source.type === 'database' &&
       existingNode &&
-      (focusManager.editingNodeId === node.id ||
-        PersistenceCoordinator.getInstance().hasPending(node.id))
+      (isFocused || hasPending)
     ) {
       log.debug(
         `setNode: skipping clobber of actively-edited node ${node.id} ` +
-        `(focused=${focusManager.editingNodeId === node.id}, ` +
-        `pending=${PersistenceCoordinator.getInstance().hasPending(node.id)})`
+          `(focused=${isFocused}, pending=${hasPending})`
       );
-      // Stash the server-confirmed version in the non-reactive cache so
-      // the next UpdateNode RPC carries the freshest version. The
-      // persistence path reads from this cache before falling back to the
-      // node's own .version field.
-      if (typeof node.version === 'number') {
+      // Only stash the server-confirmed version when the broadcast is
+      // plausibly an echo of *this client's own write* — otherwise the
+      // next UpdateNode RPC would carry the foreign writer's version
+      // against our content, defeating OCC and silently overwriting the
+      // foreign change.
+      //
+      // Heuristic: if the existing optimistic content equals or starts
+      // with the incoming content, the incoming is an older snapshot of
+      // OUR work (either an exact echo of our most-recent confirm, or a
+      // prefix from before the user typed more characters). Otherwise it
+      // is a foreign change — preserve OCC by leaving the cache empty so
+      // the next RPC uses our local `node.version`, which will conflict
+      // and surface the divergence.
+      if (typeof node.version === 'number' && this.isPlausibleOwnEcho(existingNode, node)) {
         this.serverConfirmedVersions.set(node.id, node.version);
       }
+      // `persistedNodeIds.add` is safe here precisely because the guard
+      // only runs when `existingNode` is truthy — a database event for a
+      // node we've already seen implies the node IS persisted server-side.
+      // Do not remove the `existingNode` check thinking the add is
+      // unconditional bookkeeping; it is not.
       this.persistedNodeIds.add(node.id);
       // Do NOT touch this.nodes or notify subscribers — there is no
       // observable change to the local view, and any reactive write here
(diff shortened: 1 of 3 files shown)


mstomar125:
## Pragmatic code review — PR #1215

**CI:** ⏳ No checks reported on `fix/sync-echo-order-preservation` — verify CI is wired up before merge.
**Scope:** +337 / -18 across 3 files. Two fixes: `tauri-sync-listener` preserves existing `structureTree` order when echo arrives without `properties.order`; `shared-node-store` persistence-path stale-sibling check now consults `structureTree.getParent` instead of the unreliable `siblingNode.parentId`. Stacked on #1213.
**Requirements:** Frontend half of `nodespace-sync#77`. The PR is honest that it's not sufficient alone — both #1214 and this PR need to land.

### Findings

🟡 **Order-fallback `Date.now()` lands echoed-brand-new children at the *bottom* of the visible list, not "somewhere stable"** (`tauri-sync-listener.ts:200-201`)
The comment says "fall back to `Date.now()` so it lands somewhere stable rather than getting an `undefined` propagated through the sort comparator." But `Date.now()` is ~1.7e12, while fractional orders are typically in [0, 2] — so a "stable" fallback is actually a "land far below every other sibling" fallback. For the receive-side of bob's window seeing a brand-new alice-created child for the *first time* (so `structureTree` has no entry yet), the child now appears at the bottom of bob's list. PR #1214 is supposed to mitigate this by ensuring the backend appends at end, but if #1215 lands without #1214 the symptom remains. Worth either:
1. Asserting both PRs ship together (gating merge order).
2. Asking `structureTree` for the parent's current last-child order and using `last_order + ε` instead of `Date.now()`.

🟡 **`structureTree.getChildrenWithOrder(parentBare).find(c => c.nodeId === childBare)` is O(n)** (`tauri-sync-listener.ts:194-198`)
For a parent with many children, every cloud-echoed `relationship:created` event scans the children list linearly. For a high-cardinality parent (e.g., a date-page with hundreds of nodes) under a burst of sync echoes this multiplies. Worth either:
1. A direct `structureTree.getOrderOf(parentId, childId)` primitive.
2. A Map-backed index inside `structureTree`.

Not a blocker for correctness, but the sync echo path is one of the hotter paths in the receive side.

🟡 **Stale-sibling check now preserves the hint when `structureTree` has *no opinion* — backend may still reject** (`shared-node-store.svelte.ts:1349-1359`)
```typescript
if (siblingActualParent !== null && siblingActualParent !== currentNode.parentId) { /* clear */ }
```
When `structureTree.getParent(siblingId)` returns `null` (sibling not yet in tree), the hint is preserved. The PR body says "the backend can still resolve it via its own retry loop" — meaning we rely on `NodeService::create_node_with_parent`'s sibling-validation retry. That retry exists, but its timeout/budget is finite. If structureTree is stale during a fast Enter-key burst, multiple CREATEs hit the retry path in series, each consuming its own budget. The previous (buggy) behavior was to clear the hint and let the backend pick a default; the new behavior is to keep the hint and rely on retry. Net better, but worth quantifying: how often is `getParent` null at persistence time during the two-window demo? If it's frequent, the change shifts load onto the retry loop. Add a debug log when `siblingActualParent === null` so the frequency is observable.

🟡 **The new test doesn't exercise the persistence-path decision** (`shared-node-store-skip-while-editing.test.ts:138-160`)
The test ends with `expect(structureTree.getParent('a')).toBe('D')` — a sanity check on the structureTree primitive, not on the fix. The test even acknowledges: "We can't directly observe the persistence-time decision from here without a Tauri mock." But then this is testing structureTree (which is presumably tested elsewhere), not the fix. Either mock `tauriCommands.createNode` and assert it's called with the preserved `insertAfterNodeId`, or move this test alongside one that does. As written, this is a test of the assumption underneath the fix, not the fix itself.

🟡 **The fix changes the trust hierarchy without naming it in code** (`shared-node-store.svelte.ts:1340-1349`)
Before: `siblingNode.parentId` was treated as authoritative. After: `structureTree.getParent()` is authoritative. The comment explains the *failure mode* of the old approach but not the new model ("structureTree is the authoritative source for `has_child` hierarchy; the bare `Node.parentId` field is opportunistically populated and not always reliable"). Documenting this invariant in one place (perhaps on the `Node` type or `structureTree` module) prevents the same pattern from being re-introduced in the next persistence path.

### Nits

🟢 **Nit: `typeof incomingOrder === 'number'` correctly handles 0 and negative orders** (`tauri-sync-listener.ts:188`)
Good catch over `incomingOrder ?? Date.now()` which would have miscoerced `null`. Worth a one-line comment noting why the typeof check is used instead of `??`.

🟢 **Nit: `rel.properties?.order as number | undefined`** (`tauri-sync-listener.ts:187`)
The `as` cast is unchecked — if `properties.order` is a string `"1.5"` from some malformed echo, it sneaks through as a number. Use a runtime check or trust the schema; either way, a one-liner saying which.

🟢 **Nit: `stripNodePrefix` called twice** (`tauri-sync-listener.ts:183-184`)
Pre-existing pattern; not your change. Mentioned only because the new block consolidates the two calls into named locals (`parentBare`, `childBare`) — already a clean improvement.

🟢 **Nit: The test file imports `reactive-structure-tree.svelte` inline via `await import(...)`** (`shared-node-store-skip-while-editing.test.ts:139`)
Mixing top-of-file imports and dynamic imports for the same purpose. Move to top of file unless there's a circular-import or initialization-order reason.

🟢 **Nit: Test name is very long** (`shared-node-store-skip-while-editing.test.ts:138`)
`does NOT clear insertAfterNodeId when the sibling lacks a populated parentId field but IS in structureTree under the same parent (sync#77)` — captures the intent precisely but could be split into setup-then-assert with a shorter `it()` title.

🟢 **Nit: `existing?.order ?? Date.now()` falls through twice silently** (`tauri-sync-listener.ts:197`)
If `structureTree` has the child but `order` happens to be 0 (impossible? worth confirming), `existing?.order` is 0 → falsy under `??` only if it's `null`/`undefined`. `??` correctly preserves 0. OK as-is; worth a comment.

🟢 **Nit: PR description acknowledges "frontend half alone is not sufficient"**
Honest and clearly stated — keep this style. Add a one-line note in the merged commit message naming #1214's commit SHA so future bisects see the dependency.

### Summary

Both fixes target the right primitives (structureTree as the order-of-record, structureTree as the parent-of-record). The blockers are about (a) what happens when structureTree has no data — the `Date.now()` fallback and the preserve-the-hint fallback both have failure modes worth quantifying, and (b) the test for the stale-sibling fix doesn't actually exercise the persistence path it's claiming to harden. None invalidate the fix; each is a small reinforcement.

### Verdict: REQUEST CHANGES

mstomar125:
Force-pushed to address review feedback. PR is now rebased on top of the updated #1213.

**Main behavioral fix**
- 🟡 `Date.now()` fallback footgun → replaced with three-tier order resolution: (1) incoming order if numeric, (2) preserved existing structureTree order, (3) for genuinely-new children, `lastSibling.order + 1 + tiny jitter` (true append-at-end). The bob-side receive path now lands new echoes at the bottom of the visible list with a sensible fractional order, not at \`~1.7e12\`.

**Test improvements**
- 🟡 Extracted the stale-sibling decision into `SharedNodeStore.shouldClearStaleInsertAfter(siblingId, currentParentId)` — a small public helper the persistence closure calls. The new test exercises the helper directly across three cases (structureTree agrees → preserve, disagrees → clear, no opinion → preserve+log). Locks in the production code path without requiring Tauri IPC mocks.
- Renamed to a shorter intent-focused test name.
- Moved the dynamic `await import` to a top-of-file import.

**Performance**
- 🟡 Added a comment naming the O(n) `siblings.find` and pointing at a future Map-backed primitive on structureTree if it ever becomes a hotspot. Not changed structurally in this PR.

**Observability**
- 🟡 Added a debug log when `structureTree.getParent` returns `null` at decision time, so the frequency of "tree not yet populated at persistence time" is observable.

**Documentation**
- 🟡 The new helper's doc comment explicitly names the **trust hierarchy**: structureTree is authoritative for has_child; `Node.parentId` is unreliable and must not be consulted alone.

**Nits**
- 🟢 `typeof === 'number'` is now explicitly documented as the runtime check (`rel.properties?.order` is typed `unknown` on the wire; the typeof gate IS the validation).
- 🟢 `stripNodePrefix` already consolidated into named locals (existing in original PR).

**Test deltas**
- shared-node-store-skip-while-editing.test.ts: 9 → 10 passing (+1 new from this PR's helper; rebased on top of #1213's foreign-write + persistence-path tests).
- `bun run --cwd packages/desktop-app quality:fix` — clean.

**Re-verified manually** against the two-window same-device demo: both windows correctly place new siblings below the typed line in the same position.

mstomar125:
## Pragmatic code review (re-review) — PR #1215

**Review Type:** Re-Review (scope: force-pushed commit `d7e4123a`, rebased on updated #1213).
**CI:** ⏳ Still no checks reported on `fix/sync-echo-order-preservation`. Same flag as round 1.

### Previous review summary
- Previous 🟡: 5 — addressed: 5.
- Previous 🟢: 7 — addressed: 4, acknowledged with comments: 3.
- New findings this pass: 0 🔴, 0 🟡, 2 🟢.

✅ **`Date.now()` fallback footgun** — replaced with a three-tier resolution:
```typescript
if (typeof incomingOrder === 'number') { order = incomingOrder; }
else if (existing) { order = existing.order; }            // tier 2: preserve structureTree order
else { order = lastOrder + 1 + Math.random() * 0.001; }   // tier 3: append-at-end with jitter
```
Brand-new echoed children now land at `lastSibling.order + 1 + ε` instead of `~1.7e12`. The visible-list placement matches the backend's append-at-end semantic from #1214. Stacked correctness is intact.

✅ **Stale-sibling test didn't exercise the persistence-path decision** — extracted `SharedNodeStore.shouldClearStaleInsertAfter(siblingId, currentParentId)` as a small public helper that the persistence closure calls. The new test exercises the helper directly across three cases (agrees → preserve, disagrees → clear, no opinion → preserve+log) and the persistence closure routes through it. Lock-in without Tauri IPC mocks.

✅ **O(n) `siblings.find(...)` on a hot path** — comment added naming it explicitly and pointing at a future Map-backed `getOrderOf(parent, child)` primitive on `structureTree`. Acceptable to defer; flagged for the next iteration.

✅ **`structureTree.getParent` returns `null` frequency** — debug log added on the `null` branch of `shouldClearStaleInsertAfter`, so production frequency of "tree not yet populated at persistence time" is observable.

✅ **Trust hierarchy invariant** — `shouldClearStaleInsertAfter`'s doc comment names it explicitly: "`structureTree` is authoritative for `has_child`; `Node.parentId` is unreliable and must not be consulted alone." That's the right place for the invariant to live — at the helper that enforces it.

✅ **Nits addressed** — `typeof === 'number'` runtime check explicitly documented; `await import` moved to top-of-file; test name shortened.

### New findings

🟢 **Nit: `Math.random() * 0.001` jitter is unseeded** (`tauri-sync-listener.ts:204`)
Two concurrent windows appending brand-new echoed children to the same parent each pick a random jitter in `[0, 0.001)`. Collision probability is low but non-zero; reproducibility of test scenarios is non-existent. For production this is fine — the fractional-order math tolerates near-collisions. For a future test that wants to exercise "two clients append simultaneously and both land in expected order," seeding (or using a monotonically-increasing client-local counter as the sub-fractional component) would help. Style.

🟢 **Nit: `lastOrder + 1` jumps by a full integer where local creates use fractional steps** (`tauri-sync-listener.ts:203`)
Local optimistic appends typically use `lastOrder + (next - lastOrder) / 2`-style fractional steps inside `FractionalOrderCalculator`. The echo-side append uses `+1 + jitter`. Practically the two streams never compete (echoes arrive after local placement, and the structureTree-existing branch fires for them), so the discrepancy is invisible. But a long-running sequence of echoes-without-local-state could grow the order space faster than the local path expects. Worth a comment naming the asymmetry, or — better — calling into `FractionalOrderCalculator::calculate_order` so the two paths produce comparable values.

### Summary

Every round-1 🟡 is resolved with code + test (the helper extraction is particularly clean — production code path now matches what the test exercises). Tier 3 of the order fallback is the right shape: append-at-end with a sub-integer jitter is a sensible "stable" default that doesn't drag echoed children to absurd positions. The two remaining nits are about long-tail consistency between the echo path and the local optimistic path — not blockers.

### Verdict: APPROVE

PR #1218, opened 2026-05-28
Add access_tags to nodes and relationships; stamp from in node (fixes #1217)

### LINKED ISSUE #1217: Stamp access_tags on relationships (local schema + NodeService)
## Context

Relationships need the same Stamped ACL treatment as nodes. Currently the local \`relationship\` table in \`schema.surql\` has no \`access_tags\` field. This means:

1. Relationships cannot be permission-filtered by LIVE SELECT on the cloud side
2. When a node's collection changes, its outgoing relationships don't get re-stamped
3. The cloud and local schemas are out of sync on relationship access control

See architecture decision in [cloud-sync.md § Schema: Relationship Table](../nodespace-docs/architecture/cloud-sync.md).

## Stamping Rule

A relationship inherits \`access_tags\` from its \`in\` node (the source) at creation time, and is re-stamped whenever the \`in\` node's \`access_tags\` change.

### 1. \`packages/core/src/db/schema.surql\`

Add \`access_tags\` field to the relationship table:

\`\`\`sql
DEFINE FIELD IF NOT EXISTS access_tags ON TABLE relationship TYPE array<string> DEFAULT [];
\`\`\`

### 2. \`NodeService::create_relationship\`

When creating a relationship, fetch the \`in\` node's \`access_tags\` and stamp them onto the relationship record.

### 3. \`NodeService\` — re-stamp on access_tags change

When a node's \`access_tags\` are updated (collection change, migration), find all outgoing relationships where \`in = node_id\` and update their \`access_tags\` to match.

### 4. Cascade delete (already implemented)

\`surreal_store.rs\` already deletes all relationships where \`in = node_id OR out = node_id\` atomically when a node is deleted. No change needed here.

## Acceptance Criteria

- [ ] \`access_tags\` field defined on local \`relationship\` table
- [ ] \`NodeService::create_relationship\` stamps \`access_tags\` from \`in\` node
- [ ] Re-stamping a node's \`access_tags\` propagates to all outgoing relationships
- [ ] Unit tests cover: create relationship stamps correctly, re-stamp propagates, delete cascades
- [ ] \`cargo clippy -- -D warnings\` clean

## Related

- nodespace-sync: relationship pull path + cloud schema update (companion issue)
- nodespace-docs: cloud-sync.md updated with correct relationship RLS schema
- #220 — cascade delete strategy (already resolved at store level)

malibio:
## Summary

- Adds `access_tags: array<string>` field to both `node` and `relationship` tables in `schema.surql`
- Adds `access_tags` to the `Node` model, `NodeUpdate`, and the `SurrealNode` deserializer
- `NodeService::create_relationship` stamps the source node's `access_tags` onto the new relationship at creation time
- New `restamp_outgoing_relationships` helper propagates updated `access_tags` to all outgoing relationships when a node's `access_tags` change (hooked into both `update_node_unchecked` and `update_with_version_check_returning_node`)
- Boxes `ResolvedValue::Node` in the playbook graph resolver to avoid the `large_enum_variant` clippy lint caused by `Node` growing slightly

## Test plan

- [x] `test_create_relationship_stamps_access_tags_from_source` — verifies tags from source node are written to the relationship
- [x] `test_create_relationship_empty_tags_when_source_has_none` — verifies empty source = empty relationship tags
- [x] `test_restamp_propagates_to_outgoing_relationships` — verifies updating a node's access_tags re-stamps all outgoing relationships
- [x] `test_delete_cascade_independent_of_access_tags` — verifies cascade delete still works with access_tags present
- [x] `cargo clippy -p nodespace-core -- -D warnings` clean
- [x] No frontend test regressions (same baseline: 4 files/93 tests pre-existing layout failures)

## Notes

- `nodespace-cli` has 4 pre-existing `&PathBuf`/`&Path` clippy errors unrelated to this PR
- No cloud/sync changes — this is strictly the local schema + NodeService implementation per issue scope

### DIFF
--- a/packages/agent/src/acp/context_assembly.rs
+++ b/packages/agent/src/acp/context_assembly.rs
@@ -587,6 +587,7 @@ mod tests {
             mentioned_in: vec![],
             title: title.map(|s| s.to_string()),
             lifecycle_status: "active".to_string(),
+            access_tags: vec![],
         }
     }
 
--- a/packages/core/benches/performance.rs
+++ b/packages/core/benches/performance.rs
@@ -186,6 +186,7 @@ fn bench_occ_overhead(c: &mut Criterion) {
                             properties: None,
                             title: None,
                             lifecycle_status: None,
+                            access_tags: None,
                         },
                     )
                     .await
@@ -205,6 +206,7 @@ fn bench_occ_overhead(c: &mut Criterion) {
                                 properties: None,
                                 title: None,
                                 lifecycle_status: None,
+                                access_tags: None,
                             },
                         )
                         .await
@@ -374,6 +376,7 @@ fn bench_batch_update(c: &mut Criterion) {
                                     properties: None,
                                     title: None,
                                     lifecycle_status: None,
+                                    access_tags: None,
                                 },
                             )
                             .await
--- a/packages/core/src/db/schema.surql
+++ b/packages/core/src/db/schema.surql
@@ -49,6 +49,11 @@ DEFINE FIELD IF NOT EXISTS title ON TABLE node TYPE option<string>;
 --   - "deleted": Soft-deleted, in trash, excluded from all queries, purgeable
 DEFINE FIELD IF NOT EXISTS lifecycle_status ON TABLE node TYPE string DEFAULT "active";
 
+-- Stamped ACL tags for cloud-side LIVE SELECT filtering (Issue #1217)
+-- Identifies which collections this node belongs to (e.g., ["col:hr", "user:uuid"]).
+-- Empty means unrestricted (local-only nodes without cloud sync).
+DEFINE FIELD IF NOT EXISTS access_tags ON TABLE node TYPE array<string> DEFAULT [];
+
 -- Indexes for performance
 DEFINE INDEX IF NOT EXISTS idx_node_type ON TABLE node COLUMNS node_type;
 DEFINE INDEX IF NOT EXISTS idx_node_modified ON TABLE node COLUMNS modified_at;
@@ -95,6 +100,10 @@ DEFINE FIELD IF NOT EXISTS relationship_type ON TABLE relationship TYPE string A
 
 -- Type-specific data stored in properties
 DEFINE FIELD IF NOT EXISTS properties ON TABLE relationship TYPE object FLEXIBLE DEFAULT {};
+
+-- Stamped ACL: inherited from the `in` node's access_tags at creation time,
+-- re-stamped when the in node's access_tags change.
+DEFINE FIELD IF NOT EXISTS access_tags ON TABLE relationship TYPE array<string> DEFAULT [];
 
 -- System metadata
 DEFINE FIELD IF NOT EXISTS version ON TABLE relationship TYPE int DEFAULT 1;
--- a/packages/core/src/db/surreal_store.rs
+++ b/packages/core/src/db/surreal_store.rs
@@ -223,6 +223,9 @@ struct SurrealNode {
     /// Lifecycle status for knowledge governance (Issue #755)
     #[serde(default = "default_lifecycle_status")]
     lifecycle_status: String,
+    /// Stamped ACL tags (Issue #1217)
+    #[serde(default)]
+    access_tags: Vec<String>,
 }
 
 /// Default lifecycle status ("active") for serde
@@ -254,6 +257,7 @@ impl From<SurrealNode> for Node {
             mentioned_in: Vec::new(), // Populated at fetch time by get_children_tree
             title: sn.title,
             lifecycle_status: sn.lifecycle_status,
+            access_tags: sn.access_tags,
         }
     }
 }
@@ -705,7 +709,8 @@ impl SurrealStore {
                 created_at: time::now(),
                 modified_at: time::now(),
                 properties: $properties,
-                title: $title
+                title: $title,
+                access_tags: $access_tags
             }};
         "#,
             node.id
@@ -727,6 +732,7 @@ impl SurrealStore {
             .bind(("version", node.version))
             .bind(("properties", properties))
             .bind(("title", node.title.clone()))
+            .bind(("access_tags", node.access_tags.clone()))
             .await
             .context("Failed to create node in universal table")?;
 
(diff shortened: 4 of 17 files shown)


malibio:
### Code Review Summary

This PR correctly implements the Stamped ACL feature described in issue #1217. The core mechanic — stamping `access_tags` from the `in` node onto a relationship at creation time and re-stamping on update — is sound, the schema changes are additive and correctly guarded with `IF NOT EXISTS`, and the four unit tests directly validate the acceptance criteria. The change is a clear net positive.

There are two issues that need attention before merge: a semantic gap in the `restamp_outgoing_relationships` failure handling that silently breaks ACL consistency, and a missing stamp path through the `add_to_collection` fast path.

---

### Findings

#### Critical Issues

**[Critical/Blocker] `restamp_outgoing_relationships` failure is silently swallowed — ACL state becomes inconsistent**

`packages/core/src/services/node_service.rs` lines ~3084–3093 and ~3268–3277

Both `update_node_unchecked` and `update_with_version_check_returning_node` call `restamp_outgoing_relationships` after successfully persisting the node's new `access_tags`, then downgrade any error to a `tracing::warn!`. This creates a silent partial-write: the node record has the new tags but its relationships still carry the old ones. Cloud LIVE SELECT filters will produce incorrect results — the exact problem issue #1217 exists to prevent — with no feedback to the caller.

The appropriate pattern matches how the rest of the codebase handles post-update side effects that must remain consistent (e.g., DDL sync via `update_schema_node_atomic` propagates errors). Either:
- Return the error to the caller (preferred — the caller can retry), or
- If fire-and-forget is intentional, document the eventual-consistency contract explicitly and add a compensating reconciliation path

```rust
// Current (unsafe):
if let Err(e) = self.restamp_outgoing_relationships(id, tags).await {
    tracing::warn!("Failed to restamp ...");  // node and relationships now diverged
}

// Preferred:
self.restamp_outgoing_relationships(id, tags).await?;
```

---

**[Critical/Blocker] `add_to_collection` fast path does not stamp `access_tags`**

`packages/core/src/db/surreal_store.rs` lines 5022–5034

When `create_relationship` is called with `relationship_type = "member_of"` and no explicit `order` key in the edge data, execution returns early via `add_to_collection` (line ~6508). This store-level method issues its own `RELATE` query that does not include `access_tags`. The stamping logic added at lines 6568–6598 is therefore never reached for this common case.

Any node added to a collection via the normal `create_relationship("member_of", ...)` call will have an unstamped relationship. This is the highest-frequency relationship type in NodeSpace, making this the most consequential gap.

The fix is to either:
1. Pass `access_tags` through to `add_to_collection` and include it in the RELATE CONTENT block, or
2. Fetch source `access_tags` and call `restamp_outgoing_relationships` immediately after `add_to_collection` returns successfully in `create_relationship`

Option 2 is lower-risk since it avoids touching the atomic SurrealDB query:

```rust
// After add_to_collection returns Ok(_) in create_relationship:
let source_access_tags = self.get_node(source_id).await?
    .map(|n| n.access_tags)
    .unwrap_or_default();
if !source_access_tags.is_empty() {
    self.restamp_outgoing_relationships(source_id, &source_access_tags).await?;
}
```

---

#### Suggested Improvements

**[Improvement] Extra DB round-trip on `create_relationship` — fetch access_tags from already-loaded node**

`packages/core/src/services/node_service.rs` lines 6568–6573

`create_relationship` calls `self.get_node(source_id)` specifically to retrieve `access_tags`, even though callers that go through `NodeService` already hold the node (or just created it). This is an additional full-node fetch on every relationship creation.

A lighter alternative: add a dedicated store query that returns only `access_tags`, or accept `access_tags: &[String]` as a parameter so callers can pass it directly when they already have the node. The current approach is correct and not a blocker, but it adds latency on the hot `create_relationship` path.

**[Improvement] `bulk_add_to_collections` and `bulk_migrate_collection_memberships` also bypass stamping**

`packages/core/src/db/surreal_store.rs` lines 5578, 5671

Both bulk operations use raw `RELATE` queries that do not include `access_tags`. These are lower-priority than the `add_to_collection` issue above (they're import/migration paths, not the primary write path), but they should be covered before the sync layer is built on top of them.

**[Improvement] `NodeUpdate::access_tags` exposed via gRPC but silently dropped**

`packages/daemon/src/services/node_service.rs` line 861

The daemon's gRPC handler sets `access_tags: None`, which means external callers cannot update `access_tags` over gRPC. This is acceptable scope for this PR, but it should be tracked as a follow-up since `access_tags` changes triggered from cloud sync will likely arrive via this path.

---

#### Nitpicks

**Nit:** `packages/core/src/playbook/graph_resolver.rs` lines 347–348 — `*n.clone()` dereferences a cloned `Box<Node>` rather than cloning the inner value directly. `n.as_ref().clone()` or `(*n).clone()` communicates intent more clearly, though the compiled output is identical.

**Nit:** `packages/core/src/services/node_service.rs` line 2962 — `update.access_tags.clone()` is called to capture `new_access_tags`, then `access_tags: new_access_tags.clone()` is passed into `NodeUpdate` at line 3045, and the original `update.access_tags` field is consumed. This double-clone is harmless for `Vec<String>` but slightly confusing. Cloning once into `node_update` and then re-reading from `node_update.access_tags` for the restamp check would be cleaner.

---

### Requirements Coverage

|---|---|
| `access_tags` field on `relationship` table | Covered |
| `access_tags` field on `node` table | Covered |
| `create_relationship` stamps from `in` node | Covered for non-`member_of` paths; **gap for `member_of` fast path** |
| Re-stamp propagates on node `access_tags` update | Covered in both `update_node_unchecked` and `update_with_version_check`, but **failure is silently swallowed** |
| Cascade delete unaffected | Covered by test |
| `cargo clippy -D warnings` clean | Confirmed in PR notes |

Two of the five criteria have correctness issues that affect the primary use cases. The schema additions and the non-`member_of` stamping path are solid.

malibio (commented):
Review complete — see full findings in the comment above. Two blocking issues identified: (1) silent ACL divergence when restamp fails, (2) member_of fast path bypasses access_tags stamping.

malibio:
Cancelling — issue #1217 is being closed, no longer pursuing this approach.

PR #1224, opened 2026-05-28
Integrate sqlite-vec for vector search (closes #1221)

### LINKED ISSUE #1221: Integrate sqlite-vec for HNSW vector search (follow-on to #1219)

malibio:
## Summary

Replaces the linear-scan embedding search (deferred from the SQLite migration #1219/#1222) with a statically-linked **sqlite-vec** `vec0` virtual table queried via native KNN `MATCH`.

**Correction to the issue framing:** `sqlite-vec`'s `vec0` is a fast **brute-force SIMD scan** over a compact columnar store — *not* an HNSW/ANN index, so there is no O(log N) complexity change. The win is a compiled vectorized scan vs. the previous decode-f32-in-Rust loop.

## Changes

- **Extension registration** (`SqliteStore::new` / `ensure_sqlite_vec_registered`): register `sqlite-vec` as a SQLite auto-extension, once per process. libsql lazily runs `sqlite3_config(SERIALIZED)` on its first `connect()`, and `sqlite3_config` fails after SQLite init — so we warm up libsql first, then register, inside a single `tokio::sync::OnceCell` (avoids a concurrency race during `new()`).
- **Schema** (`initialize_schema`): create `vec_embeddings USING vec0(embedding_id TEXT PRIMARY KEY, vector FLOAT[768] distance_metric=cosine)`.
- **Sync** (`upsert_embeddings`, `delete_embeddings`, `mark_root_embedding_stale`): keep `vec_embeddings` as a mirror of **non-stale** vectors only, all in transactions. Stale placeholder markers (`create_stale_embedding_marker[_bulk]`) are deliberately **not** mirrored so they never surface in KNN.
- **Search** (`search_embeddings`, `search_embeddings_by_node_type`): rewritten to vec0 KNN + JOIN; `similarity = 1 - distance`; over-fetch `k` to cover enough distinct nodes after per-node grouping. Composite-score formula and `EmbeddingSearchResult` shape preserved exactly.
- Deleted the now-unused `cosine_similarity` helper.
- Added 5 unit tests (self-query, delete clears vec0, re-upsert replaces, stale exclusion, marker-never-in-KNN).

## Approach: static linking (no shipped binaries)

A spike confirmed `sqlite-vec` 0.1.9 statically links and registers against libsql 0.6 via `libsql::ffi::sqlite3_auto_extension` — no per-platform `.dylib`/`.so`/`.dll` needs to be built or shipped.

## Migration note (greenfield)

Per the no-backward-compat policy: there is **no backfill**. A pre-existing dev DB would have an empty `vec_embeddings` until re-embedding — delete the local dev `.db` if you have one. (The SQLite migration only just merged, so this is a non-issue in practice.)

## Testing

- `cargo test -p nodespace-core` — all targets green (1133 lib + integration suites incl. `embedding_service_test`).
- `cargo clippy -p nodespace-core --all-targets -- -D warnings` — clean.
- `bun run test:all` — frontend 3936, agent-tools 17, mcp 10, skill 13, Rust 1133; no new failures vs. baseline.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3762,6 +3762,7 @@ dependencies = [
  "serde_json",
  "serial_test",
  "sha2",
+ "sqlite-vec",
  "tempfile",
  "thiserror 1.0.69",
  "tokio",
@@ -5833,6 +5834,15 @@ dependencies = [
  "gobject-sys",
  "libc",
  "system-deps",
+]
+
+[[package]]
+name = "sqlite-vec"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d0ba424237a9a5db2f6071f193319e2b6a32f7f3961debb2fbbfe67067abce3f"
+dependencies = [
+ "cc",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ uuid = { version = "1.0", features = ["v4", "serde"] }
 thiserror = "1.0"
 async-trait = "0.1"
 libsql = "0.6"
+sqlite-vec = "0.1"
 tracing = "0.1"
 nodespace-core = { path = "packages/core" }
 nodespace-nlp-engine = { path = "packages/nlp-engine" }
--- a/packages/core/Cargo.toml
+++ b/packages/core/Cargo.toml
@@ -10,6 +10,8 @@ repository.workspace = true
 [dependencies]
 # Database - SQLite via libsql
 libsql = { workspace = true }
+# Vector search extension (statically linked, registered via sqlite3_auto_extension)
+sqlite-vec = { workspace = true }
 
 # NLP Engine for embeddings
 nodespace-nlp-engine = { path = "../nlp-engine" }
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -10,6 +10,10 @@ use std::sync::Arc;
 use tokio::sync::broadcast;
 
 const DOMAIN_EVENT_CHANNEL_CAPACITY: usize = 128;
+/// KNN over-fetch factor: vec0 returns top-k *chunks*, but search results are grouped
+/// by node and many chunks map to one node, so we fetch `limit * this` to cover enough
+/// distinct nodes.
+const EMBEDDING_KNN_OVERFETCH: i64 = 10;
 const BM25_MAX_TOKENS: usize = 4;
 const BM25_STOP_WORDS: &[&str] = &[
     "a", "an", "the", "is", "are", "was", "were", "be", "been", "being", "have", "has", "had",
@@ -58,6 +62,46 @@ pub struct StoreChange {
 
 pub type StoreNotifier = Arc<dyn Fn(StoreChange) + Send + Sync>;
 
+/// Register the statically-linked `sqlite-vec` extension as a SQLite auto-extension so
+/// every connection opened afterwards has the `vec0` virtual-table module available.
+/// Runs exactly once per process and must complete before any real store connection is
+/// opened — `SqliteStore::new` awaits it first.
+///
+/// Ordering is critical: libsql lazily calls `sqlite3_config(SQLITE_CONFIG_SERIALIZED)`
+/// on its first `connect()` (via a process-global `Once`), and `sqlite3_config` fails
+/// once SQLite has been initialized. Registering an auto-extension auto-initializes
+/// SQLite, so we open (and drop) a throwaway in-memory libsql connection FIRST to run
+/// libsql's config, THEN register. Doing this in a single async `OnceCell` keeps the two
+/// steps atomic so concurrent `new()` calls can't interleave the warm-up and the
+/// registration.
+async fn ensure_sqlite_vec_registered() {
+    /// The libsql FFI signature for a SQLite extension entry point.
+    type EntryPoint = unsafe extern "C" fn(
+        *mut libsql::ffi::sqlite3,
+        *mut *const std::os::raw::c_char,
+        *const libsql::ffi::sqlite3_api_routines,
+    ) -> std::os::raw::c_int;
+
+    static VEC_INIT: tokio::sync::OnceCell<()> = tokio::sync::OnceCell::const_new();
+    VEC_INIT
+        .get_or_init(|| async {
+            // Warm up libsql so its one-time sqlite3_config(SERIALIZED) runs before we
+            // auto-initialize SQLite via sqlite3_auto_extension.
+            if let Ok(db) = libsql::Builder::new_local(":memory:").build().await {
+                let _ = db.connect();
+            }
+            // SAFETY: `sqlite_vec::sqlite3_vec_init` is the extension's C entry point with
+            // exactly this signature; the transmute reinterprets its fn pointer to the
+            // type `sqlite3_auto_extension` expects.
+            unsafe {
+                let entry: EntryPoint =
+                    std::mem::transmute(sqlite_vec::sqlite3_vec_init as *const ());
+                libsql::ffi::sqlite3_auto_extension(Some(entry));
+            }
+        })
+        .await;
+}
+
 pub struct SqliteStore {
     db: Arc<libsql::Connection>,
     event_tx: broadcast::Sender<crate::db::events::EventEnvelope>,
@@ -67,6 +111,10 @@ pub struct SqliteStore {
 
 impl SqliteStore {
     pub async fn new(db_path: PathBuf) -> Result<Self> {
+        // Register sqlite-vec (once per process) BEFORE opening our connection, so the
+        // connection picks up the `vec0` module. See `ensure_sqlite_vec_registered`.
+        ensure_sqlite_vec_registered().await;
+
         let database = libsql::Builder::new_local(&db_path)
             .build()
             .await
@@ -136,6 +184,22 @@ impl SqliteStore {
             END"#,
             ()
         ).await.context("Failed to create FTS5 delete trigger")?;
+
+        // sqlite-vec virtual table for embedding KNN search. Keyed by `embedding.id`
+        // (the per-chunk UUID); holds ONLY real, non-stale vectors (see upsert/delete/
+        // mark-stale paths). vec0 is a fast brute-force SIMD scan, not an ANN index.
+        conn.execute(
+            &format!(
+                "CREATE VIRTUAL TABLE IF NOT EXISTS vec_embeddings USING vec0(\
+                    embedding_id TEXT PRIMARY KEY, \
+                    vector FLOAT[{}] distance_metric=cosine\
+                )",
+                crate::models::embedding::DEFAULT_EMBEDDING_DIMENSION
+            ),
+            (),
+        )
+        .await
+        .context("Failed to create vec0 embeddings table")?;
 
         Ok(())
     }
@@ -630,8 +694,8 @@ impl SqliteStore {
                 .context("Failed to serialize current properties")?,
         };
         let updated_title = match update.title {
-            Some(t) => t,           // Some(Some(x)) sets, Some(None) clears
-            None => current.title,  // None means no change
+            Some(t) => t,          // Some(Some(x)) sets, Some(None) clears
+            None => current.title, // None means no change
         };
         let updated_status = update
             .lifecycle_status
@@ -1983,13 +2047,29 @@ impl SqliteStore {
             return Ok(());
         }
 
-        self.db
-            .execute(
-                "DELETE FROM embedding WHERE node_id = ?1",
-                libsql::params![node_id.to_string()],
-            )
+        // Replace the node's embeddings atomically across both `embedding` and the
+        // `vec_embeddings` vec0 mirror. vec0 is keyed by embedding_id, so the leading
+        // DELETE must clear the node's existing vec rows via its current embedding ids
+        // BEFORE the rows disappear from `embedding`.
+        let tx = self
+            .db
+            .transaction()
             .await
-            .context("Failed to delete existing embeddings")?;
+            .context("Failed to begin upsert_embeddings transaction")?;
+
+        tx.execute(
+            "DELETE FROM vec_embeddings WHERE embedding_id IN (SELECT id FROM embedding WHERE node_id = ?1)",
+            libsql::params![node_id.to_string()],
+        )
+        .await
+        .context("Failed to clear vec_embeddings for node")?;
+
+        tx.execute(
+            "DELETE FROM embedding WHERE node_id = ?1",
+            libsql::params![node_id.to_string()],
+        )
+        .await
+        .context("Failed to delete existing embeddings")?;
 
         let now = Utc::now().to_rfc3339();
         for emb in embeddings {
@@ -2000,12 +2080,12 @@ impl SqliteStore {
                 .model_name
                 .unwrap_or_else(|| "nomic-embed-text-v1.5".to_string());
 
-            self.db.execute(
+            tx.execute(
                 "INSERT INTO embedding (id, node_id, vector, dimension, model_name, chunk_index, chunk_start, chunk_end, total_chunks, content_hash, token_count, stale, error_count, last_error, created_at, modified_at) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11, 0, 0, NULL, ?12, ?13)",
                 libsql::params![
-                    id,
+                    id.clone(),
                     emb.node_id.clone(),
-                    vector_blob,
+                    vector_blob.clone(),
                     dimension,
                     model_name,
                     emb.chunk_index as i64,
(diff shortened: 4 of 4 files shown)


malibio:
## Code Review: PR #1224 — Integrate sqlite-vec for vector search (closes #1221)

**Reviewer**: Principal Engineer (Pragmatic Quality framework)
**Verdict**: COMMENT — net-positive change, builds clean, all 7 tests pass (5 new). Two behavioral notes worth a human decision before merge; no blockers.

---

### Requirements Check (issue #1221)

- ✅ Add `sqlite-vec` to workspace + `packages/core` Cargo.toml — done (`0.1.9`, statically linked).
- ✅ Create `vec_embeddings` vec0 virtual table — done, with `distance_metric=cosine` and `FLOAT[768]` driven by `DEFAULT_EMBEDDING_DIMENSION`.
- ✅ `upsert_embeddings` mirrors into vec0 — done, transactionally.
- ✅ `delete_embeddings` clears vec0 — done, transactionally.
- ✅ `search_embeddings` rewritten to vec0 KNN + JOIN, grouped by node with same composite-score formula — done; `search_embeddings_by_node_type` likewise.
- 📝 The issue title/body assume vec0 gives "O(log N) HNSW". That premise is **wrong** — vec0 is a brute-force SIMD scan. The PR correctly does NOT claim HNSW and documents this in the commit and code comments. This is the right call; the win is a compiled vectorized scan vs decode-in-Rust, not lower asymptotic complexity. Not a defect.

---

### Findings

#### 🟡 Important

**1. Density / `matching_chunks` semantics changed silently — affects ranking.**
📁 `packages/core/src/db/sqlite_store.rs:2299-2315` (and the typed variant ~2390)

In the old linear scan, every non-stale chunk of a node was loaded and `matching_chunks` (`entry.1 += 1`) counted **all** of the node's chunks, so `density = matching_chunks/total_chunks` was effectively ~1.0 and `composite_score ≈ max_similarity * 1.3`. With KNN, only chunks that land in the top-`k` are counted, so `matching_chunks ≤ total_chunks` and is now a genuine "fraction of chunks near the query" signal. This is arguably *better*, but it is a behavioral change to the composite score that the PR doesn't call out, and it interacts with finding #2: if KNN under-fetches a node's chunks, its density (and thus rank) is understated. Recommend either (a) documenting the intended new semantics, or (b) confirming downstream consumers/tests of `score` ranking aren't sensitive to the shift. Engineering principle: changes to a ranking formula's *inputs* are behavioral changes even when the formula text is unchanged.

**2. Silent under-fetch is possible and undocumented for the typed variant.**
📁 `packages/core/src/db/sqlite_store.rs:2278` and `2358`

`k = limit * EMBEDDING_KNN_OVERFETCH` (×10), and ×50 for the typed variant. The type filter (`n.node_type = ?3`) and `stale = 0` are applied **after** KNN selects the global top-`k`. If the corpus is dominated by other node types, the requested type can be crowded out of the top-`k` and the query returns fewer (or zero) results even though matching nodes exist — a recall cliff, not an error. The over-fetch constants are a reasonable pragmatic mitigation, but the failure mode is invisible (no warning, no metric). Recommend a one-line doc note on the limitation and, ideally, a `tracing::debug!` when the post-filter survivor count is suspiciously low (e.g. `< limit`). Engineering principle: make silent degradation observable.

#### 🟢 Suggestions

**3. SAFETY comment is slightly inaccurate (the transmute is still sound).**
📁 `packages/core/src/db/sqlite_store.rs:88-96`

The comment says `sqlite_vec::sqlite3_vec_init` "is the extension's C entry point with exactly this signature." The Rust-visible declaration in the `sqlite-vec` crate is actually `extern "C" fn sqlite3_vec_init()` (zero args) — the 3-arg `EntryPoint` is the *true underlying C* signature. The transmute-from-`*const ()` is exactly the pattern the upstream crate itself uses, so it is correct, but the comment should say "the upstream Rust binding is declared `fn()`; we transmute its address to the real 3-arg SQLite entry-point signature" to avoid misleading a future reader. Nit-level precision only.

**4. `k` is computed as `i64` from `limit * 10` (or `*50`) without an upper clamp.**
📁 `packages/core/src/db/sqlite_store.rs:2278`, `2358`

For normal `limit` values this is fine. A pathological caller passing a very large `limit` would push a large `k` into vec0. Not exploitable (input is internal, not user-controlled), but a `.min(SOME_CAP)` would be cheap defensive hygiene. Optional.

**Nit:** The two search methods are now near-identical (blob encode → KNN+JOIN → group → score → sort → truncate). DRY would suggest a shared private helper taking the SQL/params, but the divergence (extra JOIN, different over-fetch) is small enough that duplication is defensible. Optional.

---

### What's done well (verified, not just asserted)

- **Init-ordering race is correctly fixed.** `ensure_sqlite_vec_registered` warms libsql (so its lazy `sqlite3_config(SERIALIZED)` runs) *then* registers the auto-extension, both inside a single `tokio::sync::OnceCell::get_or_init`. The warm-up + registration are atomic within the closure, so concurrent `new()` calls cannot interleave — no TOCTOU. Confirmed runs exactly once per process before any real connection opens.
- **The vec0-only-holds-non-stale invariant is maintained on every path.** `upsert`, `delete`, and `mark_root_embedding_stale` all delete from `vec_embeddings` via `embedding_id IN (SELECT id FROM embedding WHERE node_id=?)` **before** the `embedding` rows change — correct ordering. Stale markers (`create_stale_embedding_marker[_bulk]`) are deliberately NOT mirrored, with explanatory comments. Tests assert `vec_row_count == 0` for stale/marker/delete cases — these assertions are meaningful and directly exercise the invariant.
- **Multi-statement writes are transactional.** All three mutators wrap their two/three statements in `db.transaction()`; on early `?` return the `tx` is dropped without commit → libsql rolls back. Correct.
- **`1 - distance` → similarity is correct** for `distance_metric=cosine`. Self-query test asserts `max_similarity ≈ 1.0`; orthogonal-axis test asserts non-match. Good determinism design via unit-axis vectors.
- **Dimension is correctly `FLOAT[768]` sourced from `DEFAULT_EMBEDDING_DIMENSION`** (no hardcoded literal in the DDL).
- **`cosine_similarity` fully deleted; no dangling references** in `packages/core` (the `cosine_similarity` in `packages/nlp-engine/tests/integration_tests.rs` is an unrelated local test helper).
- **Static linking works at runtime** — the 5 new tests pass, proving the `sqlite3_auto_extension` registration makes `vec0` available on real connections. No per-platform dylib shipping needed.
- **No project-standard violations**: no `#[allow(dead_code)]`, no lint suppression, no backward-compat cruft, uses `tracing` (no raw logging concern in Rust).

---

### Recommendation: **COMMENT**

Net-positive, atomic, well-tested change that does what the issue asks and correctly avoids the issue's mistaken HNSW premise. No blockers. Before merge, a human should decide whether findings #1 (density semantics shift) and #2 (typed-search recall cliff under over-fetch) warrant a doc note / debug log now or a follow-up issue.

*Reviewed by Claude Code (Opus 4.8) under the Pragmatic Quality framework.*

malibio (commented):
## Code Review: PR #1224 — Integrate sqlite-vec for vector search (closes #1221)

**Reviewer**: Principal Engineer (Pragmatic Quality framework)
**Verdict**: COMMENT — net-positive change, builds clean, all 7 tests pass (5 new). Two behavioral notes worth a human decision before merge; no blockers.

---

### Requirements Check (issue #1221)

- ✅ Add `sqlite-vec` to workspace + `packages/core` Cargo.toml — done (`0.1.9`, statically linked).
- ✅ Create `vec_embeddings` vec0 virtual table — done, with `distance_metric=cosine` and `FLOAT[768]` driven by `DEFAULT_EMBEDDING_DIMENSION`.
- ✅ `upsert_embeddings` mirrors into vec0 — done, transactionally.
- ✅ `delete_embeddings` clears vec0 — done, transactionally.
- ✅ `search_embeddings` rewritten to vec0 KNN + JOIN, grouped by node with same composite-score formula — done; `search_embeddings_by_node_type` likewise.
- 📝 The issue title/body assume vec0 gives "O(log N) HNSW". That premise is **wrong** — vec0 is a brute-force SIMD scan. The PR correctly does NOT claim HNSW and documents this in the commit and code comments. This is the right call; the win is a compiled vectorized scan vs decode-in-Rust, not lower asymptotic complexity. Not a defect.

---

### Findings

#### 🟡 Important

**1. Density / `matching_chunks` semantics changed silently — affects ranking.**
📁 `packages/core/src/db/sqlite_store.rs:2299-2315` (and the typed variant ~2390)

In the old linear scan, every non-stale chunk of a node was loaded and `matching_chunks` (`entry.1 += 1`) counted **all** of the node's chunks, so `density = matching_chunks/total_chunks` was effectively ~1.0 and `composite_score ≈ max_similarity * 1.3`. With KNN, only chunks that land in the top-`k` are counted, so `matching_chunks ≤ total_chunks` and is now a genuine "fraction of chunks near the query" signal. This is arguably *better*, but it is a behavioral change to the composite score that the PR doesn't call out, and it interacts with finding #2: if KNN under-fetches a node's chunks, its density (and thus rank) is understated. Recommend either (a) documenting the intended new semantics, or (b) confirming downstream consumers/tests of `score` ranking aren't sensitive to the shift. Engineering principle: changes to a ranking formula's *inputs* are behavioral changes even when the formula text is unchanged.

**2. Silent under-fetch is possible and undocumented for the typed variant.**
📁 `packages/core/src/db/sqlite_store.rs:2278` and `2358`

`k = limit * EMBEDDING_KNN_OVERFETCH` (×10), and ×50 for the typed variant. The type filter (`n.node_type = ?3`) and `stale = 0` are applied **after** KNN selects the global top-`k`. If the corpus is dominated by other node types, the requested type can be crowded out of the top-`k` and the query returns fewer (or zero) results even though matching nodes exist — a recall cliff, not an error. The over-fetch constants are a reasonable pragmatic mitigation, but the failure mode is invisible (no warning, no metric). Recommend a one-line doc note on the limitation and, ideally, a `tracing::debug!` when the post-filter survivor count is suspiciously low (e.g. `< limit`). Engineering principle: make silent degradation observable.

#### 🟢 Suggestions

**3. SAFETY comment is slightly inaccurate (the transmute is still sound).**
📁 `packages/core/src/db/sqlite_store.rs:88-96`

The comment says `sqlite_vec::sqlite3_vec_init` "is the extension's C entry point with exactly this signature." The Rust-visible declaration in the `sqlite-vec` crate is actually `extern "C" fn sqlite3_vec_init()` (zero args) — the 3-arg `EntryPoint` is the *true underlying C* signature. The transmute-from-`*const ()` is exactly the pattern the upstream crate itself uses, so it is correct, but the comment should say "the upstream Rust binding is declared `fn()`; we transmute its address to the real 3-arg SQLite entry-point signature" to avoid misleading a future reader. Nit-level precision only.

**4. `k` is computed as `i64` from `limit * 10` (or `*50`) without an upper clamp.**
📁 `packages/core/src/db/sqlite_store.rs:2278`, `2358`

For normal `limit` values this is fine. A pathological caller passing a very large `limit` would push a large `k` into vec0. Not exploitable (input is internal, not user-controlled), but a `.min(SOME_CAP)` would be cheap defensive hygiene. Optional.

**Nit:** The two search methods are now near-identical (blob encode → KNN+JOIN → group → score → sort → truncate). DRY would suggest a shared private helper taking the SQL/params, but the divergence (extra JOIN, different over-fetch) is small enough that duplication is defensible. Optional.

---

### What's done well (verified, not just asserted)

- **Init-ordering race is correctly fixed.** `ensure_sqlite_vec_registered` warms libsql (so its lazy `sqlite3_config(SERIALIZED)` runs) *then* registers the auto-extension, both inside a single `tokio::sync::OnceCell::get_or_init`. The warm-up + registration are atomic within the closure, so concurrent `new()` calls cannot interleave — no TOCTOU. Confirmed runs exactly once per process before any real connection opens.
- **The vec0-only-holds-non-stale invariant is maintained on every path.** `upsert`, `delete`, and `mark_root_embedding_stale` all delete from `vec_embeddings` via `embedding_id IN (SELECT id FROM embedding WHERE node_id=?)` **before** the `embedding` rows change — correct ordering. Stale markers (`create_stale_embedding_marker[_bulk]`) are deliberately NOT mirrored, with explanatory comments. Tests assert `vec_row_count == 0` for stale/marker/delete cases — these assertions are meaningful and directly exercise the invariant.
- **Multi-statement writes are transactional.** All three mutators wrap their two/three statements in `db.transaction()`; on early `?` return the `tx` is dropped without commit → libsql rolls back. Correct.
- **`1 - distance` → similarity is correct** for `distance_metric=cosine`. Self-query test asserts `max_similarity ≈ 1.0`; orthogonal-axis test asserts non-match. Good determinism design via unit-axis vectors.
- **Dimension is correctly `FLOAT[768]` sourced from `DEFAULT_EMBEDDING_DIMENSION`** (no hardcoded literal in the DDL).
- **`cosine_similarity` fully deleted; no dangling references** in `packages/core` (the `cosine_similarity` in `packages/nlp-engine/tests/integration_tests.rs` is an unrelated local test helper).
- **Static linking works at runtime** — the 5 new tests pass, proving the `sqlite3_auto_extension` registration makes `vec0` available on real connections. No per-platform dylib shipping needed.
- **No project-standard violations**: no `#[allow(dead_code)]`, no lint suppression, no backward-compat cruft, uses `tracing` (no raw logging concern in Rust).

---

### Recommendation: **COMMENT**

Net-positive, atomic, well-tested change that does what the issue asks and correctly avoids the issue's mistaken HNSW premise. No blockers. Before merge, a human should decide whether findings #1 (density semantics shift) and #2 (typed-search recall cliff under over-fetch) warrant a doc note / debug log now or a follow-up issue.

*Reviewed by Claude Code (Opus 4.8) under the Pragmatic Quality framework.*

malibio (commented):
Pragmatic code review posted as a PR comment above. Verdict: COMMENT (net-positive, no blockers; all 7 tests pass). Two items for human decision before merge: (1) density/matching_chunks ranking semantics shifted with KNN; (2) typed-search recall cliff when type filter is applied after global top-k. See the full report in the comment.

malibio:
## Review feedback addressed (commit b7eedcc5)

| Finding | Severity | Status |
|---|---|---|
| #1 — `matching_chunks`/`density` semantics changed under KNN | 🟡 | ✅ Documented in `search_embeddings` (density now = fraction of a node's chunks near the query; same composite formula) |
| #2 — Silent typed-search under-fetch / recall cliff | 🟡 | ✅ Added `tracing::debug!` when typed search returns `< limit`, noting `EMBEDDING_KNN_OVERFETCH` is the tuning lever |
| #3 — `sqlite3_vec_init` SAFETY comment accuracy | 🟢 | ✅ Clarified: the binding is declared zero-arg; we transmute to the real 3-arg C signature (the crate's own rusqlite pattern) |
| #4 — `k` upper clamp | 🟢 | ⏭️ Skipped — `limit` is already hard-capped at 1000 by the MCP search handler (`test_search_rejects_limit_exceeds_max`), so worst-case `k` is bounded (≤50k) and handled fine by vec0. A redundant in-function clamp would add an arbitrary magic number with no real benefit. The reviewer also flagged this as optional/non-exploitable. |

**Verification:** 7 sqlite_store tests pass, `cargo clippy --lib -D warnings` clean, `bun run test:all` green (no new failures vs. baseline).

## Re-Review Decision

**Rationale**: All changes are doc comments and one observability `tracing::debug!` line — no logic, control-flow, or query changes. Search behavior, the vec0 sync invariant, and result shapes are unchanged. Low-risk, straightforward edits.

PR #1225, opened 2026-05-29
AI-Chat provider picker: per-node dispatcher + hydration fix (#1220 follow-up)

malibio:
Follow-up to merged #1220. Turns the inert `ai-chat` provider picker into a
dropdown-driven, per-provider flow and fixes the real root cause of "clicking a
provider does nothing".

## Root cause (found via live Playwright/WebKit debugging)
The picker persisted `provider` fine, but the viewer hydrates nodes from the
`children-tree` endpoint, which returned `"properties": {}` — a #1222
(SurrealDB→libsql) regression: `build_node_tree_recursive` ran
`flatten_properties_for_api` (drops props when the namespaced shape doesn't match)
while every other endpoint returns raw props. Plus the ai-chat provider enum
rejected the ADR-034 modes.

## Backend (first commit)
- `build_node_tree_recursive` returns **raw** properties, matching
  `get_node`/`get_children`/etc. Regression test added. (`node_service.rs`)
- `AiChatNodeBehavior::validate` provider enum widened to
  `native|ollama|openai|pty` (ADR-034). Tests updated. (`behaviors/mod.rs`)

## Frontend (second commit)
**`AiChatNodeViewer` is now THE single dispatcher** for the ai-chat type:
- Header **provider dropdown** (Built-in / Ollama [grayed when not running] /
  OpenAI [disabled — no backend] / Agent terminal).
- Routes on `provider` (+`model`): unset → prompt; `pty` → `AiChatPtySession`;
  `native|ollama` without model → `AiChatModelPicker`; with model → message UI.
- **Per-node send/stream (Option 1, no shared store):** the viewer owns its own
  daemon session (`ensureModelReady` → `localAgentNewSession` → `localAgentSend`
  + streaming/download/load listeners), keyed off the node's `provider`/`model`.
  Persists to `node.properties.messages` (debounced flush + ADR-028 tool-result
  archival).
- **Hydration fallback:** fetches the node by id on mount if absent from
  `sharedNodeStore`, so the dispatcher always has properties to route on.

**`DisplayMessage`** moved to `components/chat/types.ts` (view-model type;
protocol / view-model / persisted layering kept distinct). Importers repointed.

**Deletions** (pre-ADR-034 ephemeral singleton — no dual-mode per CLAUDE.md):
`chat-store.svelte.ts`, `chat-panel.svelte`, `chat-store.test.ts`; the `chat` tab
type, its `<ChatPanel>` branch + persistence guard, and the dead chat-tab nav
override. The "AI Chat" sidebar item is commented out pending its rework into an
expandable list of recent ai-chat nodes (deferred follow-up issue).

**Incidental:** the WIP's `chatModelList()` retype (`ChatModelEntry[]`) had broken
its two consumers; retyped `modelToAgent` (agent-store) and added a
`ChatModelEntry → ModelInfo` adapter in `model-store.refreshModels`.

## Deferred to a separate follow-up issue
- "AI Chat" sidebar item → collapsible LIST of ai-chat nodes (like Collections).
- Typed canonical persisted-message schema owned by the ai-chat behavior + typed
  persisted↔DisplayMessage converters (replaces the untyped `Record<string,unknown>`
  mappers in the viewer).

## Verification
- `cargo test -p nodespace-core` (behaviors + children-tree) green; daemon builds.
- `svelte-check`: 0 errors / 0 warnings.
- `eslint`: clean on all touched files.
- `vitest`: 3913 passed / 129 files.
- Still pending: `test:all` (browser + Rust), Playwright e2e of the original bug,
  pty under `tauri:dev`.


malibio:
## Pragmatic Code Review — PR #1225

**Scope:** Both commits (`56c51f9c` WIP backend + `f1a70f80` frontend integration) against `origin/main`. Spec of record: `.claude/HANDOFF-issue-1220-picker-redesign.md` "REVISED SCOPE" (parent issue #1220 closed).

---

### Requirements Check (vs HANDOFF "REVISED SCOPE")

- ✅ **Backend fix A** — `build_node_tree_recursive` returns raw properties (`serde_json::to_value` + `uri` injection), matching the sibling gRPC endpoints. Regression test asserts non-empty `properties.provider`. `flatten_properties_for_api` left untouched in ops/MCP paths as required.
- ✅ **Backend fix B** — provider enum widened to `native|ollama|openai|pty`; error message + both unit tests updated (valid-set test and the invalid-provider test now uses `anthropic`).
- ✅ **Naming** — `AiChatNodeViewer` is the single dispatcher; helpers `ai-chat-pty-session.svelte` / `ai-chat-model-picker.svelte` carry no `*View/*Viewer/*Node` suffix.
- ✅ **Provider control** — header `<select>` dropdown (not tiles); OpenAI `disabled` ("coming soon"); Ollama grayed via `ollamaAvailable()` ("not running").
- ✅ **Per-node Option 1** — viewer owns its own daemon session (`ensureModelReady → localAgentNewSession → localAgentSend` + streaming/download/load listeners) keyed off node provider/model; no shared store. Persists to `node.properties.messages` with debounced flush + ADR-028 archival (`result: null`, `result_summary` kept).
- ✅ **Deletions** — `chat-store.svelte.ts`, `chat-panel.svelte`, `chat-store.test.ts` deleted; `chat` tab type removed from `navigation.ts`, `pane-content.svelte` branch, `tab-persistence-service.ts` guard, and the dead chat-tab nav override in `app-shell.svelte`. Verified zero dangling `chat-store`/`ChatPanel`/`'chat'`-tab references remain.
- ✅ **Hydration fallback** — `getNode`-by-id on mount when absent from `sharedNodeStore`.
- ✅ **No orphans / standards** — no `any`, no raw `console.*`, no `{@html}`, no lint suppressions in any changed file.
- 📝 **Tracked deviation (acceptable, better call):** HANDOFF said move `DisplayMessage` into `types/agent-types.ts`; it was instead placed in `components/chat/types.ts`. This is architect-confirmed (keeps a view-model out of the protocol boundary) and documented in the file header + commit. No action needed.
- 📝 **Deferred (correctly):** "AI Chat" sidebar item commented out in `layout.ts` pending rework into an expandable node list — a separate follow-up issue, explicitly out of scope here.

---

### Code Review Findings

#### 🟢 Suggestion — Ollama rows entering `modelStore.models` may mislead onboarding/recommend
📁 `packages/desktop-app/src/lib/stores/model-store.svelte.ts:32-52, 127-135`

`chatEntryToModelInfo` now folds Ollama catalog rows into `modelStore.models`. `recommendedModel` picks the smallest `not_downloaded|ready` entry; an Ollama model reports `ready`, so it can win and then `ai-onboarding-dialog` calls `downloadModel(id)` → `loadModel(id)` on a row with empty `filename`/`url`. `downloadModel` short-circuits on non-`not_downloaded` status (so it won't corrupt anything), but the onboarding flow could silently no-op on an Ollama pick.
*Principle (least surprise / interface segregation):* a store consumed by the GGUF-download UI shouldn't be seeded with rows that flow can't act on. Consider filtering `chatEntryToModelInfo` to `backend === 'gguf'` in `refreshModels`, or excluding Ollama from `recommendedModel`. Low severity — verify the onboarding path before merge, but not a blocker.

#### 🟢 Suggestion — `family: 'ministral'` fallback mislabels non-Ministral GGUF rows
📁 `packages/desktop-app/src/lib/stores/model-store.svelte.ts:36`

`const family = entry.backend === 'ollama' ? 'ollama' : 'ministral'` labels every GGUF row `ministral`, even a `gemma4` one. `family` isn't surfaced in a load-bearing way today, so this is cosmetic, but it's a latent correctness trap if anything later branches on `family`.
*Principle (correctness of derived data):* derive `family` from the entry rather than defaulting. If the catalog row can't tell you the family, that's a gap worth a TODO.

#### 🟢 Suggestion — empty `filename`/`url`/`sha256` placeholders are safe today but undocumented at the consumer
📁 `packages/desktop-app/src/lib/stores/model-store.svelte.ts:44-46`

Verified: `model-manager.svelte` and `ai-onboarding-dialog.svelte` only read `id/name/status/size_bytes/min_memory_gb` and download by `id`, so the empty strings are never surfaced to users. The doc comment already calls this out. No change required; flagging only so a future consumer that *does* read `url`/`filename` doesn't trip on the empty values.

#### 🟢 Suggestion — `ensureSession` transient-message removal is index-fragile under concurrent appends
📁 `packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte:338, 350, 388`

The progress placeholder is updated via `inMemoryMessages.slice(0, -1)` (assumes it's always the last element) and removed by `id` filter in `finally`. The `id`-filter teardown is robust; the `slice(0,-1)` updates assume nothing else appended in between. In the current single-send-at-a-time flow (`handleSend` guards on `isStreaming`/`!model`) this holds, but it's an implicit invariant.
*Principle (KISS / explicit invariants):* updating the placeholder by `id` (map-replace) rather than positional `slice` would make it robust regardless of ordering, matching the already-robust removal path.

#### 🟢 Observation — test gap on the dispatcher/session path (acceptable)
📁 `packages/desktop-app/src/tests/components/ai-chat-viewer.test.ts`, `chat-link-navigation.test.ts`

Deleting `chat-store.test.ts` was correct (the store is gone). The rewritten `chat-link-navigation.test.ts` still exercises the simplified nav decision logic and remains meaningful. The new behavior — per-node session lifecycle, provider-switch teardown, dispatcher routing — is **not** unit-tested, and the send path requires Tauri (`localAgentSend`/event listeners), so it's genuinely browser-untestable in happy-DOM. This is an acceptable gap for a high-velocity merge given the HANDOFF's Playwright/Tauri manual verification plan, **but** the pure-logic pieces that *don't* need Tauri are testable and worth a fast-follow: `archiveMessages` (ADR-028 nulling), `loadMessagesFromNode` (tool_call → assistant mapping), and `selectProvider` (drops `model`, triggers teardown). Recommend a small follow-up test for `archiveMessages`/`loadMessagesFromNode` since they encode the ADR-028 contract.

---

### Notes on correctness items reviewed and cleared

- **Listener cleanup:** `ensureSession` download/status listeners are scoped and unlisten in `finally` (not pushed to `eventUnlisteners`); `handleSend` streaming listeners are pushed and cleared in `finally` via `cleanupListeners()`; `teardownSession` and `onDestroy` also clean up. No leak across send/switch/destroy paths found.
- **Engine swap / session reuse:** `ensureSession` correctly ends the stale session and creates a new one when `ensureModelReady` reports a swap or when no session is held. `modelPrepared` + `sessionId` gate avoids redundant `localAgentNewSession`.
- **Double-load / hydration race:** `initialLoadDone` one-shot guard is set in both `onMount` (sync-available node) and the `$effect` (async-hydrated node); no double-load, and a live conversation isn't reloaded out from under the user.
- **Provider switch:** `selectProvider` drops `model` and tears down the session; `skipConflictDetection` is justified (documented intentional-config-write guard from #1220).
- **ADR-028:** `archiveMessages` nulls `result`, keeps a truncated `result_summary` — compliant.
- **Backend test:** regression test reaches `provider` under both flat and `ai-chat`-namespaced shapes, so it's resilient to the normalization boundary.

---

### Recommendation: **COMMENT** (no blockers)

This is a clear net-positive change that fully satisfies the HANDOFF revised scope: it fixes the real root-cause hydration bug, widens the provider enum to the ADR-034 modes, lands the per-node Option-1 dispatcher, and cleanly removes the pre-ADR-034 `chatStore`/`ChatPanel` with no orphans or standards violations. All findings are 🟢 suggestions/observations — none block merge. Before merging, I'd ask the author to sanity-check the one behavioral edge (Ollama rows in `modelStore.recommendedModel`/onboarding) and consider the small `archiveMessages`/`loadMessagesFromNode` unit tests as a fast-follow. Leaving the merge decision to a human per process.

malibio (commented):
Initial pragmatic code review posted as a PR comment. Recommendation: COMMENT (net-positive, no blockers; all findings are 🟢 suggestions). Leaving the merge call to a human per process.

malibio:
## Review feedback addressed (commit `491afecf`)

All four review findings were 🟢 suggestions (no blockers). Three addressed, one skipped as already-covered.

### ✅ Addressed
- **#1 + #2 — Ollama leaking into `model-store`** (`model-store.svelte.ts`): `refreshModels()` now filters to `backend === 'gguf'`. This store is the built-in download/load manager (its consumers `model-manager` and `ai-onboarding-dialog` download by URL and recommend a model to fetch); Ollama rows carry no URL here and are surfaced separately by `AiChatModelPicker`. Filtering them out removes both the dead-download-button hazard *and* the un-downloadable-`recommendedModel` hazard in one move. `chatEntryToModelInfo` now infers `family` from the id (`gemma*` → `gemma4`) and drops the now-dead `'ollama'` branch.
- **#3 — Progress placeholder** (`ai-chat-node-viewer.svelte`): the transient model-prep placeholder is now updated **by id** (a small `renderProgress` helper) to match its by-id removal, so it's order-independent instead of relying on positional `slice(0,-1)`.

- **#4 — Unit tests for `archiveMessages` / `loadMessagesFromNode`**: these functions are **already covered** by the "Tool result archival" and "Write buffering" suites in `ai-chat-viewer.test.ts` (result nulling, `result_summary` truncation, string-vs-object results, `tool_call` mapping). Adding more would duplicate existing coverage, so skipped per DRY / no-redundant-code.

### Verification
- `svelte-check`: 0 errors / 0 warnings
- `eslint`: clean
- `vitest`: 3913 passed / 129 files

---

PR #1232, opened 2026-05-29
CLI coverage gap: surface remaining knowledge-graph operations so skills+scripts can replace MCP

malibio:
Closes #1227

### DIFF
--- a/packages/cli/src/commands/mention.rs
+++ b/packages/cli/src/commands/mention.rs
@@ -67,8 +67,8 @@ async fn create(
 ) -> Result<()> {
     let response = client
         .create_mention(CreateMentionRequest {
-            mentioning_node_id: args.from.clone(),
-            mentioned_node_id: args.to.clone(),
+            mentioning_node_id: args.from,
+            mentioned_node_id: args.to,
         })
         .await
         .context("CreateMention RPC failed")?
@@ -98,8 +98,8 @@ async fn delete(
 ) -> Result<()> {
     let response = client
         .delete_mention(DeleteMentionRequest {
-            mentioning_node_id: args.from.clone(),
-            mentioned_node_id: args.to.clone(),
+            mentioning_node_id: args.from,
+            mentioned_node_id: args.to,
         })
         .await
         .context("DeleteMention RPC failed")?
--- a/packages/cli/src/commands/mention.rs
+++ b/packages/cli/src/commands/mention.rs
@@ -1,0 +1,185 @@
+//! `nodespace mention ...` — create, delete, and query mention relationships.
+
+use anyhow::{Context, Result};
+use clap::{Args, Subcommand};
+use nodespace_daemon::nodespace::{
+    CreateMentionRequest, DeleteMentionRequest, MentionTargetRequest,
+};
+use nodespace_daemon::NodeServiceClient;
+use serde_json::json;
+use tonic::transport::Channel;
+
+#[derive(Subcommand, Debug)]
+pub enum MentionAction {
+    /// Create a mention relationship from one node to another.
+    Create(CreateMentionArgs),
+    /// Delete a mention relationship.
+    Delete(DeleteMentionArgs),
+    /// List nodes that a given node mentions (outgoing).
+    Outgoing(MentionQueryArgs),
+    /// List nodes that mention a given node (incoming).
+    Incoming(MentionQueryArgs),
+}
+
+#[derive(Args, Debug)]
+pub struct CreateMentionArgs {
+    /// The node that contains the mention (source).
+    #[arg(long)]
+    pub from: String,
+    /// The node being mentioned (target).
+    #[arg(long)]
+    pub to: String,
+}
+
+#[derive(Args, Debug)]
+pub struct DeleteMentionArgs {
+    /// The node that contains the mention (source).
+    #[arg(long)]
+    pub from: String,
+    /// The node being mentioned (target).
+    #[arg(long)]
+    pub to: String,
+}
+
+#[derive(Args, Debug)]
+pub struct MentionQueryArgs {
+    /// Node ID to query mentions for.
+    pub id: String,
+}
+
+pub async fn run(
+    client: &mut NodeServiceClient<Channel>,
+    action: MentionAction,
+    json: bool,
+) -> Result<()> {
+    match action {
+        MentionAction::Create(args) => create(client, args, json).await,
+        MentionAction::Delete(args) => delete(client, args, json).await,
+        MentionAction::Outgoing(args) => outgoing(client, args, json).await,
+        MentionAction::Incoming(args) => incoming(client, args, json).await,
+    }
+}
+
+async fn create(
+    client: &mut NodeServiceClient<Channel>,
+    args: CreateMentionArgs,
+    json: bool,
+) -> Result<()> {
+    let response = client
+        .create_mention(CreateMentionRequest {
+            mentioning_node_id: args.from.clone(),
+            mentioned_node_id: args.to.clone(),
+        })
+        .await
+        .context("CreateMention RPC failed")?
+        .into_inner();
+
+    if json {
+        println!(
+            "{}",
+            serde_json::to_string_pretty(&json!({
+                "mentioning_node_id": response.mentioning_node_id,
+                "mentioned_node_id": response.mentioned_node_id,
+            }))?
+        );
+    } else {
+        println!(
+            "Created mention: {} → {}",
+            response.mentioning_node_id, response.mentioned_node_id
+        );
+    }
+    Ok(())
+}
+
+async fn delete(
+    client: &mut NodeServiceClient<Channel>,
+    args: DeleteMentionArgs,
+    json: bool,
+) -> Result<()> {
+    let response = client
+        .delete_mention(DeleteMentionRequest {
+            mentioning_node_id: args.from.clone(),
+            mentioned_node_id: args.to.clone(),
+        })
+        .await
+        .context("DeleteMention RPC failed")?
+        .into_inner();
+
+    if json {
+        println!(
+            "{}",
+            serde_json::to_string_pretty(&json!({
+                "mentioning_node_id": response.mentioning_node_id,
+                "mentioned_node_id": response.mentioned_node_id,
+            }))?
+        );
+    } else {
+        println!(
+            "Deleted mention: {} → {}",
+            response.mentioning_node_id, response.mentioned_node_id
+        );
+    }
+    Ok(())
+}
+
+async fn outgoing(
+    client: &mut NodeServiceClient<Channel>,
+    args: MentionQueryArgs,
+    json: bool,
+) -> Result<()> {
+    let response = client
+        .get_outgoing_mentions(MentionTargetRequest {
+            node_id: args.id.clone(),
+        })
+        .await
+        .context("GetOutgoingMentions RPC failed")?
+        .into_inner();
+
+    if json {
+        println!(
+            "{}",
+            serde_json::to_string_pretty(&json!({
+                "node_id": args.id,
+                "outgoing_mention_ids": response.node_ids,
+                "count": response.node_ids.len(),
+            }))?
+        );
+    } else {
+        println!("{} outgoing mention(s):", response.node_ids.len());
+        for id in &response.node_ids {
+            println!("  {id}");
+        }
+    }
+    Ok(())
+}
+
+async fn incoming(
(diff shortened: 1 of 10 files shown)


malibio:
## Code Review — PR #1232: CLI Coverage Gap

**Verdict: Approve with minor improvements noted.** This is a well-scoped, net-positive change that cleanly closes the CLI coverage gap without introducing any significant architectural debt or security issues. The implementation is straightforward thin-client plumbing as the issue intended, the test suite is solid, and the docs are materially improved. The findings below are minor; none are blockers.

---

### Summary

The PR surfaces six new knowledge-graph operations via the CLI (query, export, batch-get, batch-update, mention CRUD, schema read), adds three new RPC handlers in the daemon, and expands SKILL.md into a genuine agent reference. The architecture is consistent with the existing CLI pattern: stateless gRPC thin-clients with `--json` support. Integration tests exercise all new paths against a real in-process daemon.

---

### Findings

#### [Improvement] `get_nodes_batch` daemon handler uses N sequential `get_node` calls instead of the existing batch API

**File:** `packages/daemon/src/services/node_service.rs`, lines 745–754

The `get_nodes_batch` RPC handler iterates node IDs in a `for` loop, issuing one `get_node` call per ID. The core service already exposes `get_nodes(&[&str])` on the `NodeService` trait (services/mod.rs:65), which delegates to `store.get_nodes_by_ids` — a single SQL `WHERE id IN (…)` query. Using N individual point-reads instead of the available batch API defeats the purpose of a batch endpoint and will be visibly slower for large request arrays.

Suggested fix:
```rust
let id_refs: Vec<&str> = req.node_ids.iter().map(String::as_str).collect();
let fetched_nodes = self.node_service.get_nodes(&id_refs).await
    .map_err(service_error_to_status)?;
let fetched_ids: std::collections::HashSet<&str> =
    fetched_nodes.iter().map(|n| n.id.as_str()).collect();
let not_found: Vec<String> = req.node_ids.iter()
    .filter(|id| !fetched_ids.contains(id.as_str()))
    .cloned()
    .collect();
```
Note: this also means errors from `get_node` are currently silently bucketed into `not_found` (line 750–752) rather than surfaced, which masks real errors from missing nodes. The batch API call propagates the error correctly.

#### [Improvement] Stale SurrealDB comment survives in `node_service.proto`

**File:** `packages/proto/proto/node_service.proto`, line 184

The `GetRootsRequest` comment reads:
```
// limit/offset are pushed into SurrealDB (LIMIT/START AT) for efficiency.
```
SKILL.md was correctly updated, but this proto comment was missed. It should reference SQLite/libsql instead.

#### [Improvement] `export_markdown` daemon handler: JSON-roundtrip for structured params is fragile

**File:** `packages/daemon/src/services/node_service.rs`, lines 697–716

The handler constructs a `serde_json::Value` blob, passes it to `handle_get_markdown_from_node_id`, then extracts `result["markdown"]` and `result["node_count"]` with `.unwrap_or("")` / `.unwrap_or(0)` fallbacks. If the markdown function ever changes its return shape, this silently returns empty strings rather than surfacing an error. Prefer typed destructuring or a dedicated params struct if the `handle_get_markdown_from_node_id` signature allows it. If it does not, at minimum add an `ok_or_else` on the field extractions so the RPC fails loudly on shape mismatch rather than returning phantom empty output.

---

### Nitpicks

- **Nit:** `packages/cli/src/commands/mention.rs` lines 70–71, 101–102, 132, 163 — `args.from.clone()` / `args.to.clone()` / `args.id.clone()` are unnecessary since `args` is consumed by the match arm. Pass `args.from` / `args.to` / `args.id` directly to avoid the allocation.

- **Nit:** `packages/skill/SKILL.md` lines 27, 29, 76 — uses `--type note` in examples. `note` is not a core node type (core hardcodes `text`, `task`, `header`, `code-block`, `date`). The examples should use `--type text` to prevent agents from creating nodes with an unrecognized type and then wondering why behaviors are missing.

- **Nit:** `packages/cli/src/commands/node.rs` line 110 — `ExportArgs.children` default is `true` but the clap `ArgAction::Set` interaction means passing `--children` without a value is a user error. The doc comment on line 109 says "default: true" which is correct, but the UX could be improved: consider `--no-children` as a flag that flips to `false` rather than a boolean `--children true/false`, which is non-idiomatic for CLI tools.

- **Nit:** `schema_list_and_get` integration test (cli_integration.rs:591–606) only covers `schema list` against an empty database and does not test `schema get`. Adding a seed schema node and calling `schema get` would give full coverage of the `get_schema_definition` RPC, including the `failed_precondition` branch for a non-schema node.

---

### What Works Well

- The `batch_update` implementation correctly handles the OCC auto-fetch path per-item and accumulates failures without aborting the whole batch — exactly the right semantics.
- Error mapping from `NodeServiceError` / `OpsError` to gRPC `Status` codes is consistent and covers the distinguishable error classes.
- The integration test harness (in-process daemon over a temp UDS) is clean and reusable; new tests follow the established pattern faithfully.
- The `ExportMarkdown` proto design with `optional bool include_children` / `optional bool include_node_ids` correctly handles the three-state distinction (unset = use server default, set = caller override) rather than relying on a zero-value sentinel.
- SKILL.md is now a genuinely useful agent reference with concrete examples for every new command.

malibio (commented):
Review posted above. One improvement (N+1 batch fetch in get_nodes_batch daemon handler) and one stale comment (SurrealDB reference in proto) are the most substantive items. The rest are minor polish. Net positive change — safe to merge after addressing the batch handler.

malibio:
### Code Review Summary

This PR is a clean, well-scoped surfacing of six previously-unreachable daemon operations at the CLI layer. The architecture is consistent with existing patterns throughout the codebase, and the changes are appropriately thin — no new business logic, pure gRPC client plumbing. The PR closes the primary blocker (#1227) for skill-based agent integration replacing MCP.

A few issues worth addressing before merge, and one meaningful architectural observation about the batch handlers.

---

### Findings

#### Critical Issues

**[Critical/Blocker] `get_nodes_batch` daemon handler duplicates `core::node_batch::get_nodes_batch` without delegating to it**

`packages/daemon/src/services/node_service.rs` lines 726–762 re-implements the per-ID sequential loop, including the 100-item cap and the `not_found` accumulation, which already exist verbatim in `packages/core/src/node_batch.rs` (lines 59–102). This is a DRY violation that creates two divergent code paths for the same operation. The original `node_batch.rs` comment even notes it was extracted from the old MCP handler for reuse.

The existing `store().get_nodes_by_ids()` batch fetch is available and already used in `get_collection_members_recursive` (line 968 of the same file). The daemon handler could delegate to `core::node_batch::get_nodes_batch` (which already wraps the sequential path) or go directly to `store().get_nodes_by_ids()` to get true parallel/batched SQL. As written, the new handler fetches N nodes with N individual queries under a sequential loop, while a pre-existing batch path sits unused.

The same applies to `update_nodes_batch` (lines 764–848) vs `core::node_batch::update_nodes_batch`. The core version even includes a missing warning:

```rust
// core/node_batch.rs line 136-137
tracing::warn!(
    "OCC bypassed: version parameter not provided for batch update (race condition possible)"
);
```

The new daemon handler omits this warn entirely, meaning version-less batch updates will silently bypass OCC without any log signal.

**Recommendation:** Delegate both daemon handlers to the functions in `core::node_batch` rather than re-implementing. This eliminates duplication, restores the OCC bypass warning, and ensures the benchmarked code path stays exercised.

---

#### Suggested Improvements

**[Improvement] `export_markdown` daemon handler uses silent fallbacks that mask upstream bugs**

`packages/daemon/src/services/node_service.rs` lines 715–716:

```rust
let markdown = result["markdown"].as_str().unwrap_or("").to_string();
let node_count = result["node_count"].as_u64().unwrap_or(0) as u32;
```

If `handle_get_markdown_from_node_id` returns a `Value` that is missing these keys (e.g., due to a future shape change), the daemon silently returns empty markdown with zero count and a `200 OK`. This makes bugs in the markdown handler invisible to callers. Prefer `ok_or_else` returning `Status::internal(...)` so unexpected response shapes surface as errors.

**[Improvement] `schema_list_and_get` integration test does not exercise `schema get`**

`packages/cli/tests/cli_integration.rs` lines 592–606 only calls `schema list`. The test daemon starts with no custom schemas, which is fine for the list case, but `schema get` is never exercised, leaving the `GetSchemaDefinition` RPC path untested end-to-end. Consider using a known built-in schema ID (e.g., `"task"` or `"text"`, whichever the daemon seeds by default) to add one `SchemaAction::Get` call in the same test.

**[Improvement] `get_nodes_batch` daemon handler conflates fetch errors with not-found**

Lines 749–751:
```rust
Err(e) => {
    tracing::warn!("get_nodes_batch: error fetching {}: {}", node_id, e);
    not_found.push(node_id);
}
```

A transient DB error (lock contention, I/O error) is silently folded into the `not_found` list, which agents interpret as "this node does not exist" rather than "something went wrong." Consider a separate `errors` field in the response, or at minimum return a `Status::internal` when _any_ fetch error occurs, so callers aren't misled.

---

#### Nitpicks

**Nit:** `packages/skill/SKILL.md` lines 27, 29, 76 — the "Create a node" code block and one query example still use `--type note`, which was replaced by `--type text`. The Options line below (line 33) correctly documents `text`, but the example commands contradict it. Agents reading the examples verbatim will send an invalid type.

**Nit:** `packages/cli/src/commands/node.rs` `BatchUpdateArgs.updates` takes a raw JSON string from the command line. This is a pragmatic choice for a CLI flag, but worth a brief doc comment explaining why it's a string rather than structured clap args (i.e., "JSON array accepted as a single flag to avoid the combinatorial explosion of repeated --key value pairs for multi-field per-item updates").

**Nit:** `packages/proto/proto/node_service.proto` line 572 — the section header is doubled (`// ---` separator appears twice back-to-back). Minor formatting artifact from the addition.

**Nit:** The `BatchUpdateItem.version` comment in the proto ("version unset = auto-fetch (bypasses OCC)") is correct and useful, but the OCC bypass is also mentioned in the SKILL.md docs as a feature ("omit to auto-fetch") without the caveat that it creates a race window. Worth adding a parenthetical in the skill doc to help agents understand the trade-off before choosing to omit version.

---

### Acceptance Criteria Check

- [x] Agent-relevant operations reachable from `nodespace <cmd>` with `--json`: node query, export, batch-get, batch-update, mention create/delete/outgoing/incoming, schema list/get — all covered
- [x] CLI integration tests against test daemon: 4 new tests added covering query, export, batch, mention lifecycle (schema get path not fully covered — see finding above)
- [x] SKILL.md updated: stale SurrealDB/RocksDB removed, new commands documented (residual `--type note` in examples — see nit above)
- [~] Skill can perform the same core operations as old MCP: functionally yes, but the `get_nodes_batch` duplication means the behavior can drift from the benchmarked core path

### Overall Verdict

Net positive change — the CLI surface gap is closed and the implementation is consistent with existing patterns. The DRY violation in the batch handlers is the one issue I'd want resolved before merge since it creates two maintenance surfaces for the same logic and drops the OCC bypass warning that the original author deliberately added. All other findings are improvement-grade or below.

malibio (commented):
Requesting changes on the batch handler duplication / OCC bypass warning omission — see the detailed review comment above. All other findings are improvement-grade or nits. Happy to approve once the delegation to core::node_batch is in place.

malibio:
## Review addressed

All findings from the pragmatic code review have been addressed:

**🔴 Critical fixed:**
- `get_nodes_batch`: now delegates to `NodeService::get_nodes` (single `WHERE id IN (...)` SQL query) instead of N individual `get_node` calls; storage errors now surface as `Status::internal` rather than being silently folded into `not_found`
- `update_nodes_batch`: restored the `tracing::warn!` on the OCC bypass path (was in `core::node_batch` but dropped in the daemon re-implementation)

**🟡 Important fixed:**
- `export_markdown`: replaced `.unwrap_or()` / `.unwrap_or(0)` with `ok_or_else` so shape changes in `handle_get_markdown_from_node_id` fail loudly instead of silently returning empty markdown

**🟢 Nits fixed:**
- proto: stale `SurrealDB (LIMIT/START AT)` comment → `SQLite (LIMIT/OFFSET)`
- SKILL.md: replaced all `--type note` examples with `--type text` (note is not a core node type)
- `mention.rs`: removed unnecessary `.clone()` on moved `args.from` / `args.to`

**⏭️ Skipped:**
- `--no-children` flag refactor → `ArgAction::Set` works correctly; UX refactor is out of scope
- `schema get` integration test → no built-in schema IDs are seeded in the bare test daemon; would require seeding a schema node first, which goes beyond verifying the thin gRPC wrapper

PR #1233, opened 2026-05-29
Expose nodespace-nlp-engine as an optional default feature in nodespace-core

malibio:
Closes #1212

### DIFF
--- a/packages/core/Cargo.toml
+++ b/packages/core/Cargo.toml
@@ -7,14 +7,18 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
+[features]
+default = ["nlp"]
+nlp = ["dep:nodespace-nlp-engine"]
+
 [dependencies]
 # Database - SQLite via libsql
 libsql = { workspace = true }
 # Vector search extension (statically linked, registered via sqlite3_auto_extension)
 sqlite-vec = { workspace = true }
 
-# NLP Engine for embeddings
-nodespace-nlp-engine = { path = "../nlp-engine" }
+# NLP Engine for embeddings (optional — gate with the `nlp` feature)
+nodespace-nlp-engine = { path = "../nlp-engine", optional = true }
 
 # Async runtime
 tokio = { workspace = true }
--- a/packages/core/src/ops/mod.rs
+++ b/packages/core/src/ops/mod.rs
@@ -7,7 +7,9 @@
 pub mod context_ops;
 pub mod node_ops;
 pub mod rel_ops;
+#[cfg(feature = "nlp")]
 pub mod search_ops;
+#[cfg(feature = "nlp")]
 pub mod skill_ops;
 pub mod skill_updater;
 
--- a/packages/core/src/services/mod.rs
+++ b/packages/core/src/services/mod.rs
@@ -22,7 +22,9 @@ use crate::models::Node;
 use async_trait::async_trait;
 
 pub mod collection_service;
+#[cfg(feature = "nlp")]
 pub mod embedding_processor;
+#[cfg(feature = "nlp")]
 pub mod embedding_service;
 pub mod error;
 pub mod migration_registry;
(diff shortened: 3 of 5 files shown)


malibio:
### Code Review Summary

This PR makes `nodespace-nlp-engine` an optional default feature in `nodespace-core`, unblocking `nodespace-sync` CI from a ~22-minute cold-build penalty. The approach is correct and well-scoped: the `[features]` block follows Cargo conventions, the `cfg` guards are placed at the right granularity, and default consumers are demonstrably unaffected (daemon, agent, and cli all take `nodespace-core = { workspace = true }` which resolves with `default-features = true`). This is a net positive change. There are two findings worth addressing before merge.

---

### Findings

#### Improvement Issues

**[Improvement] `packages/core/src/services/node_service.rs` — Dead `#[cfg]` pair inside an already-gated function**

`queue_root_for_embedding` is declared with `#[cfg(feature = "nlp")]` at the function level (line 3997). Inside that function, the code adds a second `#[cfg(feature = "nlp")]` guard on the waker call (line 4074), followed by a `#[cfg(not(feature = "nlp"))]` debug log (line 4079). Both inner attributes are unreachable: when the NLP feature is disabled the entire function does not exist, so neither branch ever compiles. The `#[cfg(not(feature = "nlp"))]` branch in particular is dead by construction.

The correct fix is to remove the redundant inner `#[cfg]` pair and reinstate the waker call unconditionally within the function:

```rust
// Before (inside #[cfg(feature = "nlp")] fn):
#[cfg(feature = "nlp")]
if let Some(ref waker) = self.embedding_waker {
    waker.wake();
}
#[cfg(not(feature = "nlp"))]
tracing::debug!("📥 Root {} queued for embedding (nlp feature disabled)", root_id);

// After:
if let Some(ref waker) = self.embedding_waker {
    tracing::debug!("🔔 Waking embedding processor for root {}", root_id);
    waker.wake();
}
```

**[Improvement] `packages/core/src/services/node_service.rs` — Silent removal of a diagnostic warn**

The PR removes the `tracing::warn!("⚠️ No embedding waker configured …")` that fired when the NLP feature was active but `set_embedding_waker` had not been called. This warn was a useful operational guard — it detects misconfiguration in the `nlp`-enabled path (e.g., the daemon starting without calling `set_embedding_waker`). Since `queue_root_for_embedding` is still only called in the `nlp` path, the warn should be preserved:

```rust
if let Some(ref waker) = self.embedding_waker {
    tracing::debug!("🔔 Waking embedding processor for root {}", root_id);
    waker.wake();
} else {
    tracing::warn!(
        "No embedding waker configured — root {} will not be processed automatically",
        root_id
    );
}
```

Both issues are in the same function and can be resolved together in a single small edit.

---

#### Nitpicks

**Nit: `packages/core/src/services/node_service.rs` line 2257–2275 — `create_stale_embedding_marker` called unconditionally on root creation**

When `nlp` is disabled, `create_stale_embedding_marker` still writes a row to the DB on every new embeddable root node. This is intentional (the table exists regardless, and the marker can be consumed if NLP is later re-enabled), but a brief comment at the call site clarifying "stale markers are written even without nlp feature so they can be consumed if the feature is later enabled" would prevent future confusion for anyone auditing the `--no-default-features` build.

**Nit: `packages/core/src/services/mod.rs` — module-level doc still lists `EmbeddingProcessor` and `NodeEmbeddingService` unconditionally**

The module-level doc comment at the top of `services/mod.rs` still lists `NodeEmbeddingService` and `EmbeddingProcessor` without mentioning that they are conditional on the `nlp` feature. Minor, but worth a `(requires \`nlp\` feature)` annotation in the module-level doc for discoverability.

---

- [x] `nodespace-core` exposes a `nlp` feature, included in `default` — confirmed in `packages/core/Cargo.toml`
- [x] Building with `--no-default-features` does not compile `nodespace-nlp-engine` — `optional = true` + `dep:` prefix is the correct Cargo idiom; `sqlite-vec` and `libsql` remain unconditional (correct)
- [x] Existing call sites that touch NLP are guarded with `#[cfg(feature = "nlp")]` — all 14 annotations verified; the `create_stale_embedding_marker` DB calls are correctly left unconditional (they don't pull in `llama-cpp-sys-2`)
- [x] All existing tests pass with default features — `embedding_service_test.rs` gated with `#![cfg(feature = "nlp")]`, no other test changes needed
- [x] No public-API breaks — `daemon`, `agent`, and `cli` all use `nodespace-core = { workspace = true }` (implicit `default-features = true`), so their imports of `EmbeddingProcessor`, `NodeEmbeddingService`, `search_ops`, `skill_ops`, and `set_embedding_waker` are unaffected

The two Improvement items should be fixed before merge; they are localized to the same function and require only a handful of lines.

malibio (commented):
Two improvements needed before merge: (1) remove the dead #[cfg] pair inside the already-gated queue_root_for_embedding function, and (2) reinstate the tracing::warn! for the missing embedding waker. Both are in the same function and require only a small edit. Full details in the review comment.

PR #1235, opened 2026-05-29
Use persistent session dirs for agent workspaces (closes #1198)

malibio:
## Summary

- Replaces `tempfile::TempDir` with a named directory at `~/.nodespace/agent-sessions/<uuid>/` so agent artifacts survive session end and daemon restarts
- Session UUID generated in `PtySession::launch` and used as the directory name, tying filesystem path directly to the session record
- Removes `tempfile` from `packages/agent` production dependencies (retained in `dev-dependencies` for existing tests in that package)
- Test helper `launch_for_test` uses `std::env::temp_dir()/nodespace-agent-test-sessions/<uuid>/` to avoid polluting `~/.nodespace` during CI

## Test plan
- [ ] `cargo build --workspace` compiles cleanly
- [ ] New `session_dir_persists_after_process_exit_and_session_drop` test verifies the directory outlives both process exit and session drop

Closes #1198

### DIFF
--- a/packages/agent/Cargo.toml
+++ b/packages/agent/Cargo.toml
@@ -28,7 +28,6 @@ dirs = "5.0"
 sysinfo = "0.33"
 which = "7"
 portable-pty = "0.8"
-tempfile = "3"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
--- a/packages/agent/src/pty/manager.rs
+++ b/packages/agent/src/pty/manager.rs
@@ -71,10 +71,12 @@ impl PtySessionManager {
             sessions.insert(id, session);
         }
 
-        // Auto-cleanup: when the child process exits, drop the session out
-        // of the map so the temp dir gets cleaned up. `watch` latches the
-        // final value, so this works whether the child exits before or
-        // after the receiver was constructed.
+        // Auto-cleanup: when the child process exits, remove the session entry
+        // from the map. The session directory at
+        // `~/.nodespace/agent-sessions/<uuid>/` is NOT deleted — it persists
+        // on disk so artifacts survive restarts. `watch` latches the final
+        // value, so this works whether the child exits before or after the
+        // receiver was constructed.
         let sessions = self.sessions.clone();
         tokio::spawn(async move {
             // Fast path: exit already happened before insert returned.
@@ -102,10 +104,9 @@ impl PtySessionManager {
     /// Terminate and remove a session. Returns `Ok(false)` if the session
     /// was not in the map (already cleaned up by the natural-exit watcher).
     ///
-    /// The temp dir is dropped when the last `Arc<PtySession>` is dropped;
-    /// since this method takes ownership out of the map, that usually happens
-    /// at the end of this call (assuming no other caller is holding an
-    /// `Arc` from a prior `get()`).
+    /// Removing the session from the map drops the `Arc<PtySession>` when no
+    /// other caller holds a reference. The session directory on disk is NOT
+    /// deleted — it persists at `~/.nodespace/agent-sessions/<uuid>/`.
     pub async fn terminate(&self, id: &Uuid) -> anyhow::Result<bool> {
         let session = {
             let mut sessions = self.sessions.lock().await;
--- a/packages/agent/src/pty/mod.rs
+++ b/packages/agent/src/pty/mod.rs
@@ -6,14 +6,14 @@
 //!
 //! The session lifecycle is:
 //!
-//! 1. Create a per-session temp directory.
+//! 1. Create a persistent session directory at `~/.nodespace/agent-sessions/<uuid>/`.
 //! 2. Write the context file ([`GraphContextAssembler::write_context_file`])
 //!    so the agent picks up its `CLAUDE.md` / `AGENTS.md` on launch.
-//! 3. Spawn the agent binary inside a freshly opened PTY rooted at the temp dir.
+//! 3. Spawn the agent binary inside a freshly opened PTY rooted at the session dir.
 //! 4. Stream stdout/stderr bytes through a `broadcast::Sender<OutputChunk>`.
 //! 5. Accept stdin via [`PtySession::write_input`] and resize via [`PtySession::resize`].
-//! 6. On [`PtySession::terminate`] (or when the child exits naturally), drop the
-//!    [`tempfile::TempDir`] so the working directory is cleaned up.
+//! 6. On [`PtySession::terminate`] (or when the child exits naturally), the session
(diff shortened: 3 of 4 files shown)


malibio:
## Code Review — PR #1235 (Issue #1198)

---

### Requirements Check

- ✅ Each agent session launches with cwd set to `~/.nodespace/agent-sessions/<uuid>/`
- ✅ Directory is created on session launch (`std::fs::create_dir_all`), not deleted on session end (no Drop cleanup)
- ✅ Context file (`CLAUDE.md` / `AGENTS.md`) is written into the session directory as before
- ✅ Old `tempfile::TempDir` approach removed from production paths (`_session_dir: tempfile::TempDir` → `session_dir: PathBuf`)
- ✅ `~/.nodespace/agent-sessions/` created on first session launch via `create_dir_all`
- ✅ `tempfile` removed from `[dependencies]` in `packages/agent/Cargo.toml`; retained in `[dev-dependencies]` for existing tests
- ✅ Test updated: `session_dir_persists_after_process_exit_and_session_drop` explicitly verifies persistence after drop

**All acceptance criteria met.**

---

### Code Review Findings

#### Suggested Improvements

**`📁 packages/agent/src/pty/manager.rs:75` and `:105` — Stale comments describing the old temp-dir cleanup behavior**

```
// of the map so the temp dir gets cleaned up.
```
```
/// The temp dir is dropped when the last `Arc<PtySession>` is dropped
```

These two comments were not updated as part of this PR. They describe the old `tempfile::TempDir` Drop-based cleanup behavior that no longer exists. A future reader will be confused: after a `terminate()` + manager removal, the session dir now persists, but the comment says it is "cleaned up." This violates the principle that comments must reflect current intent, not historical implementation. The auto-prune watcher and `terminate()` doc should say the *map entry* is removed and the directory **persists on disk** by design.

**`📁 packages/agent/src/pty/session.rs:302-303` — Stale `terminate()` doc comment**

```
/// Temp-dir cleanup happens when the
/// last `Arc` is dropped — usually the manager removing the entry
/// after this returns.
```

Same issue: this sentence now describes behavior that no longer happens. Should read: "The session directory at `~/.nodespace/agent-sessions/<uuid>/` is not deleted — it persists on disk."

**`📁 packages/agent/src/acp/context_assembly.rs:251` — Stale "session temp directory" wording**

```
/// session temp directory and appends NodeSpace tool activation
```

The `with_shim_dir` doc still refers to a "session temp directory." Minor, but creates drift from the new persistent-directory mental model.

#### Nitpicks

**Nit: `📁 packages/agent/src/pty/session.rs:401` — Test helper comment says "persists until the OS cleans up the temp dir"**

```
/// Creates a temporary directory under `std::env::temp_dir()` for the
/// session. The directory persists until the OS cleans up the temp dir.
```

The wording is accurate (it's under `temp_dir()`), but calling the behavior "persists until the OS cleans up" could mislead a reader into thinking this is ephemeral. Consider: "The directory is not deleted by the test helper; it lives under `std::env::temp_dir()/nodespace-agent-test-sessions/<uuid>/` until the OS reclaims it."

**Nit: `📁 packages/agent/src/pty/session.rs:406-412` — `launch_for_test` duplicates PTY setup from `spawn_in_pty`**

`launch_for_test` inlines the full PTY construction sequence (open pair, `spawn_command`, `drop(pair.slave)`, reader/writer setup) rather than delegating to `spawn_in_pty`. This existed before this PR, but the refactor creates a natural opportunity to route `launch_for_test` through `spawn_in_pty` now that `spawn_in_pty` accepts a `PathBuf` instead of `tempfile::TempDir`. This would be a DRY improvement for a follow-up.

**Nit: `📁 packages/agent/src/pty/session.rs:120` — `pub session_dir` exposes a mutable-by-reference path**

The field changed from `_session_dir: tempfile::TempDir` (private, Drop-only) to `pub session_dir: PathBuf`. The `pub` visibility grants any holder of `&mut PtySession` the ability to swap the path without the actual directory being moved. Given the type is `PathBuf` (not a `Path` reference), the risk is low, but consider `pub(crate)` or a `pub fn session_dir(&self) -> &Path` accessor in line with the `pub fn session_dir_path()` test-only accessor already present.

---

### Overall Assessment

This is a correct, focused, net-positive change. The core replacement — `tempfile::TempDir` to a named `PathBuf` under `~/.nodespace/agent-sessions/` — is implemented cleanly:

- Session UUID is generated at the right level (`launch`) and flows through `spawn_in_pty` without threading it through callers higher up the stack (the issue's spec was followed exactly)
- Error context messages are appropriately descriptive (`"create session dir {}"`)
- `tempfile` correctly moves to `dev-dependencies` rather than being removed outright, preserving test compilation in `context_assembly.rs` tests
- The test helper for integration tests (`launch_for_test`) correctly avoids polluting `~/.nodespace` by routing to `temp_dir()`
- New regression tests added for persistence after exit and drop, plus a thorough pre-existing suite (terminate idempotency, concurrent write+resize, late subscribe)

The only real gap is two stale doc comments in `manager.rs` that still describe the old temp-dir cleanup behavior. These should be updated before merge to prevent developer confusion about the post-session lifecycle. All other findings are minor polish.

**Recommendation: REQUEST CHANGES** — fix the two stale doc comments in `manager.rs` (lines 75 and 105), and update `terminate()`'s doc in `session.rs` (lines 302-303). All three are one-liner text fixes; no logic change needed.

malibio (commented):
Fix two stale doc comments in manager.rs (lines 75 and 105) that still describe the old temp-dir Drop-based cleanup, and the stale 'Temp-dir cleanup happens when the last Arc is dropped' sentence in session.rs terminate() doc (lines 302-303). All three are one-liner text fixes — no logic change needed. Everything else is clean and the implementation is correct.

malibio:
## Address-Review Summary

### ✅ Addressed (3 findings)

**🟡 Two stale comments in `manager.rs`** — Updated both the auto-prune task comment (line ~75) and the `terminate()` doc (line ~105) to correctly state that the session directory persists at `~/.nodespace/agent-sessions/<uuid>/` and is not deleted.

**🟡 Stale `terminate()` doc in `session.rs`** — Replaced the "Temp-dir cleanup happens when the last Arc is dropped" sentence with accurate language about the persistent directory.

### ⏭️ Skipped (2 suggestions)

**🟢 `pub session_dir` → `pub(crate)`** — Kept as `pub`. External callers legitimately need to know where a session's artifacts are stored (e.g., to open a file manager, display the path in UI, or read agent output). `pub` is the right visibility here.

**🟢 `launch_for_test` code duplication** — Pre-existing issue noted by reviewer as a follow-up candidate. Out of scope for this PR; no logic changes needed.

PR #1230, opened 2026-05-29
Audit & cleanup: finish daemon + SQLite migrations (purge dead HTTP MCP / SurrealDB / RocksDB from code and docs)

malibio:
Closes #1226

PR #1236, opened 2026-05-29
API: introduce InsertPosition enum across proto/gRPC + service (replace overloaded insert_after_node_id)

malibio:
Closes #1216

### DIFF
--- a/packages/cli/src/commands/node.rs
+++ b/packages/cli/src/commands/node.rs
@@ -103,7 +103,7 @@ async fn create(
             collection: String::new(),
             lifecycle_status: String::new(),
             id: String::new(),
-            insert_after_node_id: String::new(),
+            position: None, // CLI create defaults to End
         })
         .await
         .context("CreateNode RPC failed")?
--- a/packages/cli/tests/cli_integration.rs
+++ b/packages/cli/tests/cli_integration.rs
@@ -97,7 +97,7 @@ async fn create_get_update_children_delete_round_trip() {
             collection: String::new(),
             lifecycle_status: String::new(),
             id: String::new(),
-            insert_after_node_id: String::new(),
+            position: None,
         })
         .await
         .expect("seed parent")
@@ -269,7 +269,7 @@ async fn diagnostics_collect_reports_counts_and_recency() {
             collection: String::new(),
             lifecycle_status: String::new(),
             id: String::new(),
-            insert_after_node_id: String::new(),
+            position: None,
         })
         .await
         .expect("seed root")
@@ -287,7 +287,7 @@ async fn diagnostics_collect_reports_counts_and_recency() {
                 collection: String::new(),
                 lifecycle_status: String::new(),
                 id: String::new(),
-                insert_after_node_id: String::new(),
+                position: None,
             })
             .await
             .unwrap_or_else(|e| panic!("seed {label}: {e}"))
--- a/packages/core/benches/performance.rs
+++ b/packages/core/benches/performance.rs
@@ -158,7 +158,7 @@ fn bench_occ_overhead(c: &mut Criterion) {
                         node_type: "text".to_string(),
                         content: "Performance test".to_string(),
                         parent_id: None,
-                        insert_after_node_id: None,
+                        position: nodespace_core::services::InsertPositionOwned::End,
                         properties: json!({}),
                     })
                     .await
@@ -239,7 +239,7 @@ fn bench_batch_get(c: &mut Criterion) {
                                 node_type: "text".to_string(),
                                 content: format!("Node {}", i),
                                 parent_id: None,
-                                insert_after_node_id: None,
+                                position: nodespace_core::services::InsertPositionOwned::End,
                                 properties: json!({}),
                             })
                             .await
@@ -277,7 +277,7 @@ fn bench_batch_get(c: &mut Criterion) {
                                 node_type: "text".to_string(),
                                 content: format!("Node {}", i),
                                 parent_id: None,
-                                insert_after_node_id: None,
+                                position: nodespace_core::services::InsertPositionOwned::End,
                                 properties: json!({}),
                             })
                             .await
@@ -328,7 +328,7 @@ fn bench_batch_update(c: &mut Criterion) {
                             node_type: "text".to_string(),
                             content: "# Benchmark Root".to_string(),
                             parent_id: None,
-                            insert_after_node_id: None,
+                            position: nodespace_core::services::InsertPositionOwned::End,
                             properties: json!({}),
                         })
                         .await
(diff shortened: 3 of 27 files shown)


malibio:
## Code Review — PR #1236: InsertPosition Enum

### Summary

This is a well-executed, complete refactor. The core design is sound: the enum cleanly replaces a two-valued Option<&str> with three explicitly-named variants, eliminates the empty_to_none sentinel hack at the proto layer, and threads the change consistently through Rust core, proto/gRPC daemon, Tauri commands, and TypeScript frontend. Coverage is broad and the migration is thorough. The PR is a net positive and should be merged once the issues below are addressed.

---

### Findings

#### [Improvement] Stale doc comments on move_node and CreateNodeParams

packages/core/src/services/node_service.rs line 4329 — the Arguments block for move_node still reads:

    * insert_after_node_id - Optional sibling to insert after (None = insert at beginning)

This argument no longer exists; the parameter is now position: InsertPosition<'_>. A caller reading this doc will be confused.

packages/core/src/services/node_service.rs line 419 — CreateNodeParams.position says:

    /// Defaults to InsertPositionOwned::Beginning when not specified.

The field has no Default impl and the actual runtime default for omitted callers is End (commit message confirms this). The doc says the opposite. This will mislead every future caller of this struct.

packages/core/src/services/node_service.rs line 2037 — doc block for create_node_with_parent still references insert_after_node_id by name in the Note: paragraph.

packages/core/src/services/node_service.rs line 9567 — test comment reads to insert_after_node_id.is_none(): the narrowing guard is now matches!(position, InsertPosition::End); the comment describes old code.

These are straightforward fixes, not blockers, but stale doc comments on a freshly-renamed API are especially harmful.

#### [Improvement] End with no parent silently degrades to Beginning at the store layer

In resolve_insert_position (node_service.rs line 4522):

    InsertPosition::End => {
        if let Some(pid) = parent_id {
            let children = self.get_children(pid).await?;
            Ok(children.last().map(|n| n.id.clone()))
        } else {
            Ok(None)   // store interprets None as Beginning
        }
    }

When parent_id is None (making a node a root), End resolves to None, and SqliteStore::move_node maps None to before the first sibling (Beginning) at the root level. So move_node_unchecked(id, None, InsertPosition::End) does not actually append at the end of the root list — it prepends. The call sites in tests and benchmarks use InsertPosition::Beginning explicitly for this case, so they are self-consistent, but the semantics of End for root-level nodes are silently wrong. Add a comment documenting that End with no parent falls back to store-layer Beginning, so future callers are not surprised.

#### [Improvement] Proto bool oneof — bool beginning = 8 / bool end = 9 is an unusual encoding

Using bool as the payload of a oneof variant works but is non-idiomatic and creates a subtle footgun: a client sending {beginning: false} would still set the beginning variant, which the daemon correctly decodes as InsertPositionOwned::Beginning (the value of the bool is ignored). The standard pattern for a unit-type variant in proto is either an empty message (message Beginning {}) or a proto enum. The current approach requires explicit documentation to prevent future proto authors from interpreting the bool value as meaningful. This is not a blocker for a greenfield project but worth tracking.

#### [Nit] Inline type annotations on insertPosition field accesses in TypeScript

In shared-node-store.svelte.ts lines 1415 and 1497, the inline cast uses { type: string; siblingId?: string } | null. This duplicates structure that the exported InsertPosition type already defines. Import and use it directly to avoid drift if the shape changes.

#### [Nit] pending-operations.ts line 19 retains an insertAfterNodeId reference in a comment

The race-condition scenario comment still reads which references B as insertAfterNodeId. Should be updated to insertPosition.

#### [Nit] shared-node-store.svelte.ts line 562 comment

The method docblock at line 562 still reads insertAfterNodeId hint — should be insertPosition.

---

Checking against issue #1216:

- [x] InsertPosition enum in packages/core/src/services/ — present, re-exported from crate root
- [x] create_parent_edge, reorder_child, reorder_node accept the enum
- [x] Proto position fields re-expressed; daemon empty_to_none for position removed
- [x] All primary call sites migrated: proto, daemon gRPC, import service, core, Tauri commands, reactive-node-service, shared-node-store, backend-adapter, tauri-commands
- [ ] Per-variant tests — End and Beginning are tested; the gRPC round-trip tests set position: None everywhere, exercising only the End default; explicit daemon tests for Beginning and After proto decoding paths are missing
- [x] Stale SurrealDB/eventual-consistency comments removed from MCP handlers — moot; the Rust HTTP MCP module was removed in #1230; no residue found
- [x] Create-then-reorder hack and move retry loop removed — confirmed gone
- [ ] Nodespace-sync companion PR (apply_remote to InsertPosition::End) — tracked separately, expected out of scope for this repo
- [x] optimistic-operation-manager, pending-operations, tauri-sync-listener, collection-service — confirmed these do not call the renamed APIs directly; no migration needed

---

### Verdict

Approve with recommendations. Fix the stale doc comment claiming Defaults to Beginning on CreateNodeParams.position (it says the opposite of the actual behavior) before merge. The End-with-no-parent silent degradation and the gRPC daemon test gap are improvements worth making as a fast follow.

malibio (commented):
Reviewed. See full findings in the PR comment above. One doc-accuracy fix recommended before merge (CreateNodeParams.position claims 'Defaults to Beginning' but the actual default is End). Other findings are improvements/nits.

malibio:
## Review Feedback Addressed

All 7 reviewer recommendations addressed. 1 skipped with justification.

### ✅ Addressed

1. **Wrong default in `CreateNodeParams.position` doc** — Fixed to correctly describe `End` (not `Beginning`).
2. **3 stale `insert_after_node_id` doc/comment references** — Updated to use new API names (`InsertPosition`, `position`, `After`).
3. **`End` with no parent degrades silently** — Added explanatory comment in `resolve_insert_position` documenting the limitation and why it's safe in practice.
4. **Proto oneof bool-variant encoding** — Added documentation in all three messages explaining the unit-variant workaround and that the bool value is semantically meaningless.
5. **Missing daemon tests for `Beginning` and `After` decoding** — Added `test_insert_position_beginning_and_after_proto_decoding` in `grpc_round_trip.rs` covering all three variants via CreateNode + ReorderNode RPCs.
6. **TS inline type casts** — Replaced `{ type: string; siblingId?: string }` structural casts in `shared-node-store.svelte.ts` with the imported `InsertPosition` type.
7. **2 stale `insertAfterNodeId` comment references** — Updated in `pending-operations.ts` and `shared-node-store.svelte.ts`.

### ⏭️ Skipped

**Proto `bool` → empty message redesign** — The wire format change would require coordinated updates to all clients (daemon, Tauri shell, any future gRPC consumers) in a single atomic deploy. The current encoding is correct for all internal callers; the bool-value-is-ignored edge case only matters if an external client sends `beginning: false`, which doesn't exist. This is now documented in the proto. Revisit if/when an external client appears or if we add a breaking-changes window.

PR #1239, opened 2026-05-29
Remove sleep-based eventual-consistency retry loops from NodeService (SQLite is synchronous/ACID)

malibio:
Closes #1237

### DIFF
--- a/packages/core/src/services/node_service.rs
+++ b/packages/core/src/services/node_service.rs
@@ -2095,34 +2095,26 @@ impl NodeService {
         // Step 3: Validate sibling (if After) - treat as best-effort hint.
         // If the sibling doesn't exist or has moved to a different parent, fall
         // back to End so new nodes land at the bottom rather than the top.
-        // Retry up to 5 times with 50ms backoff against transient lookup
-        // failures during rapid empty-node creation (Enter, Enter, Enter).
+        // SQLite is synchronous/ACID: a node written by a prior awaited call is
+        // immediately visible; a single check is sufficient.
         if let crate::services::InsertPositionOwned::After(ref sibling_id) = params.position.clone()
         {
-            use tokio::time::{sleep, Duration};
-            let mut sibling_valid = false;
-            for _attempt in 0..5 {
-                sibling_valid = match self.get_node(sibling_id).await {
-                    Ok(Some(_)) => match self.get_parent(sibling_id).await {
-                        Ok(sibling_parent) => {
-                            let sibling_parent_id = sibling_parent.as_ref().map(|p| p.id.as_str());
-                            sibling_parent_id == params.parent_id.as_deref()
-                        }
-                        Err(_) => false,
-                    },
-                    _ => false,
-                };
-                if sibling_valid {
-                    break;
-                }
-                sleep(Duration::from_millis(50)).await;
-            }
+            let sibling_valid = match self.get_node(sibling_id).await {
+                Ok(Some(_)) => match self.get_parent(sibling_id).await {
+                    Ok(sibling_parent) => {
+                        let sibling_parent_id = sibling_parent.as_ref().map(|p| p.id.as_str());
+                        sibling_parent_id == params.parent_id.as_deref()
+                    }
+                    Err(_) => false,
+                },
+                _ => false,
+            };
 
             if !sibling_valid {
                 tracing::warn!(
                     sibling_id = %sibling_id,
                     parent_id = ?params.parent_id,
-                    "position sibling is stale after retries (moved or deleted), falling back to End"
+                    "position sibling is stale (moved or deleted), falling back to End"
                 );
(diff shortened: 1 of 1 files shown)


malibio:
### Code Review Summary

This PR cleanly removes three SurrealDB-era eventual-consistency retry/polling loops from \`NodeService\` and eliminates the corresponding test-side sleep infrastructure. The change is surgical, deletion-heavy, and the reasoning is technically sound throughout. All acceptance criteria from issue #1237 are met. The PR passes \`cargo clippy -p nodespace-core -- -D warnings\` clean and all 132 unit tests pass with zero failures.

One minor observation worth noting: the stale-sibling fallback test only exercises the *deleted-sibling* path (node returns \`None\` from \`get_node\`). The issue also specifies *reparented-sibling* as a stale scenario (sibling exists but \`get_parent\` returns a different parent). The existing \`test_sibling_with_different_parent_falls_back_to_append\` test covers the reparented case, so coverage is complete across both tests combined — not a blocker.

---

### Findings

#### No Critical Issues

All three sites were handled correctly:

**Site 1 (\`create_node_with_parent\` ~L2095):** The 5×50ms retry loop is removed. The single-check logic and End-fallback path are identical to what the loop was doing on a successful attempt. The warning log message was updated from \`"position sibling is stale after retries"\` to \`"position sibling is stale"\` — correct.

**Site 2 (\`create_parent_edge\` ~L4604):** The dead "Sibling not found" 10×100ms retry loop is deleted. The issue's analysis was correct: \`SqliteStore::move_node\` never emits that error string. The now-single \`move_node\` call at L4591 maps errors cleanly via \`NodeServiceError::query_failed\`. All dead bookkeeping (\`attempt_count\`, \`last_error\`, \`start\` timing, elapsed-ms debug logs) is fully removed.

**Site 3 (\`create_parent_edge\` ~L4644):** The 20×50ms position-verification polling loop plus the nested 10×50ms re-verify loop are fully deleted. The worst-case ~2s of polling is gone. The commit trusts the single atomic \`move_node\` result, which is the correct model for SQLite.

**Test infrastructure removal:** \`wait_for_children_tree_order\` helper fully deleted. All \`sleep(Duration::from_millis(N))\` calls in the test module are removed. \`use std::time::Duration\` and \`use tokio::time::sleep\` imports removed from both test submodules. Assertions preserved and passing without any waiting. Verified \`grep -c "sleep\b"\` returns 1 — one comment reference only, no call sites.

**New tests:** \`test_insert_after_ordering_is_deterministic_without_retries\` and \`test_stale_sibling_falls_back_to_end\` are correct and deterministic. The ordering test exercises End/After/Beginning in combination across 5 nodes and asserts exact order \`[E, A, C, D, B]\` on the first read — this is the Site 3 proof required by the issue. Both tests pass.

#### Suggested Improvements

None that would block merge.

#### Nitpicks

**Nit:** \`packages/core/src/services/node_service.rs:8786–8790\` — The \`adjacency_list_tests\` module comment retained from \`main\` reads: *"which can exhibit race conditions under concurrent test execution when an operation queries for sibling positions before a prior write has been committed."* The phrase "before a prior write has been committed" was accurate under SurrealDB replication lag; under SQLite the actual concern is test-level concurrency (multiple tests sharing the same DB file with interleaved fractional-order writes). The \`#[serial]\` annotation is still correct and necessary — the comment just deserves a one-line clarification. Low priority; pre-existing text outside this PR's scope.

**Nit:** \`packages/core/src/services/node_service.rs:9142–9143\` — The \`create_node_with_parent_tests\` module comment similarly says \`"to prevent race conditions under concurrent test execution"\` without elaborating. Fine as-is for this PR; the shorter phrasing is less misleading than the \`adjacency_list_tests\` version.

---

### Requirements Checklist (Issue #1237)

- [x] Site 1 (2098–2129): retry loop removed; single sibling-validity check with identical End fallback
- [x] Site 2 (4604–4642): dead "Sibling not found" retry loop removed; single \`move_node\` call; unused bookkeeping removed
- [x] Site 3 (4644–4721): position-verification/reorder polling loop removed; single \`move_node\` result trusted
- [x] No \`tokio::time::sleep\` / backoff loop in production paths of \`node_service.rs\` (grep clean)
- [x] Stale comments referencing "eventual consistency" / "SurrealDB" / "transient lookup" removed or updated
- [x] Test-section sleeps/retry-waits removed; assertions preserved and passing
- [x] Stale-sibling fallback proven via \`test_stale_sibling_falls_back_to_end\`
- [x] Ordering determinism proven via \`test_insert_after_ordering_is_deterministic_without_retries\`
- [x] No \`#[allow(dead_code)]\`, no commented-out loops, no "kept for safety" shims
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean (verified: \`-p nodespace-core\` clean)

This is a net-positive change. Approving.

PR #1245, opened 2026-05-30
Remove the empty_to_none sentinel in daemon NodeService — proto-presence audit (B5)

malibio:
Closes #1242

### DIFF
--- a/packages/cli/src/commands/node.rs
+++ b/packages/cli/src/commands/node.rs
@@ -170,11 +170,11 @@ async fn create(
         .create_node(CreateNodeRequest {
             node_type: args.node_type,
             content: args.content,
-            parent_id: args.parent.unwrap_or_default(),
+            parent_id: args.parent,
             properties: String::new(),
-            collection: String::new(),
-            lifecycle_status: String::new(),
-            id: String::new(),
+            collection: None,
+            lifecycle_status: None,
+            id: None,
             position: None, // CLI create defaults to End
         })
         .await
@@ -194,12 +194,12 @@ async fn update(
         .update_node(UpdateNodeRequest {
             node_id: args.id,
             version: None, // auto-fetch current version on the server
-            node_type: String::new(),
+            node_type: None,
             content: Some(args.content),
             properties: None,
-            add_to_collection: String::new(),
-            remove_from_collection: String::new(),
-            lifecycle_status: String::new(),
+            add_to_collection: None,
+            remove_from_collection: None,
+            lifecycle_status: None,
         })
         .await
         .context("UpdateNode RPC failed")?
--- a/packages/cli/src/commands/search.rs
+++ b/packages/cli/src/commands/search.rs
@@ -30,8 +30,8 @@ pub async fn run(
         .search_nodes(SearchRequest {
             query: args.query,
             node_types: args.node_types,
-            collection: String::new(),
-            collection_id: String::new(),
+            collection: None,
+            collection_id: None,
             limit: args.limit,
             offset: 0,
             threshold: 0.0,
--- a/packages/cli/tests/cli_integration.rs
+++ b/packages/cli/tests/cli_integration.rs
@@ -92,11 +92,11 @@ async fn create_get_update_children_delete_round_trip() {
         .create_node(nodespace_daemon::nodespace::CreateNodeRequest {
             node_type: "text".into(),
             content: "parent".into(),
-            parent_id: String::new(),
+            parent_id: None,
             properties: String::new(),
-            collection: String::new(),
-            lifecycle_status: String::new(),
-            id: String::new(),
+            collection: None,
+            lifecycle_status: None,
+            id: None,
             position: None,
         })
         .await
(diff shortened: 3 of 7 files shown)


malibio:
## Code Review — PR #1245 (B5: Remove empty_to_none sentinel)

**Overall:** This is a clean, well-scoped cleanup that achieves its stated goal. The `empty_to_none` helper is fully deleted, all 11 identified call sites are migrated to proto3 `optional` fields, and all consumers (daemon, CLI, desktop-app, integration tests) are updated consistently. The proto comments document the intent for each field. One functional correctness issue and two minor items are noted below.

---

|-----------|--------|-------|
| All 11 call sites audited (Create×4, Update×4, Search×2, Move×1) | PASS | Verified by counting removed `empty_to_none` calls in diff |
| `empty_to_none` helper deleted | PASS | Zero remaining occurrences in workspace |
| No field where `""` could be legitimate silently collapses to `None` | PASS | All fields are correctly none-or-present semantics |
| Proto changes regenerate cleanly | Assumed PASS | Build must be verified before merge |
| No lint suppression / no `#[deprecated]` | PASS | Clean deletion throughout |
| Tests updated for optional field semantics | PASS | All test sites use `None`/`Some(...)` correctly |

---

### [Improvement] Proto comment for `new_parent_id` claims `""` still moves to root — it no longer does

**File:** `packages/proto/proto/node_service.proto`

```proto
// unset or "" = move to root (None on the ops layer); present = new parent ID.
optional string new_parent_id = 3;
```

**File:** `packages/daemon/src/services/node_service.rs`

```rust
let new_parent = req.new_parent_id;   // Option<String>
// ...
.move_node(&req.node_id, req.version, new_parent.as_deref(), position)
```

The old behavior was: `empty_to_none("")` → `None` → move to root.
The new behavior is: `Some("")` is passed as `Some("".as_str())` to the ops layer, which calls `node_exists("")`, which returns `false`, which yields an `InvalidParent` error — **not** a move to root.

The proto comment promises a guarantee (`"" = move to root`) that the implementation no longer provides. Clients following the comment would get a runtime error instead of the documented behavior. The fix is a one-liner:

```rust
// Normalize Some("") to None: empty string is not a valid parent ID,
// and "" historically encoded "move to root" for callers that cannot
// distinguish absent from empty (e.g. generated gRPC stubs default-init to "").
let new_parent = req.new_parent_id.filter(|s| !s.is_empty());
```

…and update the proto comment to match:

```proto
// unset = move to root (None on the ops layer); present and non-empty = new parent ID.
// "" is treated as unset (move to root) to preserve compatibility with gRPC clients
// that default-initialize the field.
optional string new_parent_id = 3;
```

**Rationale:** The test `move_node_to_root_when_new_parent_id_unset` only exercises the `None` case. No existing caller currently sends `Some("")` (the desktop app uses `Option<String>` correctly; the CLI has no move implementation), so this is latent, but the comment is actively misleading and creates a correctness trap for future external callers.

---

### [Improvement] Desktop-app `create_node`: inline `id` sentinel is correct but inconsistent with `parent_id`

**File:** `packages/desktop-app/src-tauri/src/commands/nodes.rs`

```rust
// id — uses inline sentinel because CreateNodeInput.id is String (not Option):
id: if node.id.is_empty() { None } else { Some(node.id) },

// parent_id — cleaner, because CreateNodeInput.parent_id is already Option<String>:
parent_id: node.parent_id,
```

The `if/else` sentinel for `id` is the correct approach given that `CreateNodeInput.id` is a bare `String` in the Tauri command input (the frontend always supplies a UUID or `""`). A brief inline comment explaining why this field is not `Option<String>` in `CreateNodeInput` would prevent future readers from wondering why it differs from `parent_id`. This is a readability improvement, not a correctness issue.

---

### Nit: The `move_node_to_root_when_new_parent_id_unset` test does not verify the `Some("")` behavior

`grpc_round_trip.rs` tests the `None` path but does not add a second `move_node` call with `new_parent_id: Some("".to_string())` to confirm the documented behavior. This is only worth adding if the [Improvement] above is accepted and the `filter()` normalization is added.

---

### What is not a concern

- `import_round_trip.rs` uses `collection: String::new()` for `ImportOptions.collection` (a different proto message in `import_service.proto`, bare `string`, out of scope for this PR).
- `grpc_round_trip.rs` uses `node_type: String::new()` for `WatchRequest.node_type` (bare `string`, where `""` legitimately means "watch all nodes", also out of scope).
- The test comment removal ("include a content change so NodeUpdate is non-empty") is correct: the ops layer explicitly handles the `update.is_empty()` case by fetching the current node, so a collection-only update without content is valid.
- The `.clone()` removal on `req.collection_id` in search is correct — the field is now `Option<String>` and is consumed in place.

---

### Verdict

The change meets the acceptance criteria and is a net positive. Merge is blocked only on the **[Improvement]** for `new_parent_id`: the proto comment advertises behavior the code no longer provides. The fix is a single `.filter(|s| !s.is_empty())` call. The desktop-app `id` improvement and the test nit are optional.

malibio (commented):
Review complete. See the full analysis in the PR comment above. Key issue: new_parent_id proto comment claims "" = move to root but the handler no longer normalizes Some("") to None after the empty_to_none removal. Recommend adding .filter(|s| !s.is_empty()) and updating the proto comment before merge.

PR #1252, opened 2026-05-30
C2b: Remove Node.parentId field — structureTree is single source of truth (closes #1248)

malibio:
## Summary

Follow-up to #1247 (C2). Removes the `parentId` field from the `Node` interface entirely. All hierarchy reads now go through `structureTree.getParent()` / `getParentId()`, and backend CREATE/MOVE payloads derive their parent from `structureTree` at call time via an explicit `CreateNodeInput` struct.

## Changes
- `node.ts`: `parentId` field removed from `Node` interface; stale `NodeUpdate` doc examples referencing `parentId` are in inline comments only (doc cleanup candidate)
- `shared-node-store.svelte.ts`: Primary CREATE path builds explicit `CreateNodeInput` with `parentId: this.getParentId(nodeId)`
- `reactive-node-service.svelte.ts`: All `node.parentId` reads replaced with `structureTree.getParent()`; mirror `updateNode({parentId})` calls removed; outdent unpersisted-CREATE path updated to update structureTree before re-triggering `setNode`
- `base-node-viewer.svelte`: `promotePlaceholderToNode` return type cleaned up; `_parentNodeId` parameter retained (still called by six sites) but unused internally
- Tests: `parentId` removed from all `Node` literals; `parentMapping` option removed; dead assertions replaced or removed

## Test plan
- [ ] `bun run check` clean
- [ ] `grep -rn "\.parentId" packages/desktop-app/src` returns only structureTree-internal uses

### DIFF
--- a/packages/desktop-app/src/lib/design/components/base-node-viewer.svelte
+++ b/packages/desktop-app/src/lib/design/components/base-node-viewer.svelte
@@ -524,11 +524,13 @@
    */
   function promotePlaceholderToNode(
     placeholder: Node,
-    _parentNodeId: string,
+    parentNodeId: string,
     overrides: { content?: string; nodeType?: string }
   ): Node {
-    // parentId is derived from structureTree at CREATE time (see shared-node-store persistence path)
-    // Caller must call reactiveStructureTree.addChild before setNode fires persistence
+    // parentId is derived from structureTree at CREATE time (persistence path calls getParentId).
+    // IMPORTANT: Caller MUST call reactiveStructureTree.addChild({ parentId: parentNodeId, ... })
+    // BEFORE the setNode persistence debounce fires so getParentId returns the correct parent.
+    log.debug(`[promotePlaceholder] promoting ${placeholder.id.substring(0, 8)} under parent ${parentNodeId.substring(0, 8)}`);
     return {
       id: placeholder.id,
       nodeType: overrides.nodeType ?? placeholder.nodeType,
--- a/packages/desktop-app/src/lib/design/components/base-node-viewer.svelte
+++ b/packages/desktop-app/src/lib/design/components/base-node-viewer.svelte
@@ -524,15 +524,11 @@
    */
   function promotePlaceholderToNode(
     placeholder: Node,
-    parentNodeId: string,
+    _parentNodeId: string,
     overrides: { content?: string; nodeType?: string }
-  ): Node & { parentId?: string } {
-    // CRITICAL FIX (Issue #528): Include transient parentId field
-    // This field is NOT part of the Node model (which stores hierarchy as graph edges)
-    // but is needed by the backend HTTP API to create the parent-child edge relationship
-    // The backend will extract this field and call operations.create_node() with it
-    // Note (Issue #533): _containerId removed - backend auto-derives root from parent chain
-    // Note (Issue #575): beforeSiblingId removed - backend now handles sibling ordering
+  ): Node {
+    // parentId is derived from structureTree at CREATE time (see shared-node-store persistence path)
+    // Caller must call reactiveStructureTree.addChild before setNode fires persistence
     return {
       id: placeholder.id,
       nodeType: overrides.nodeType ?? placeholder.nodeType,
@@ -541,9 +537,7 @@
       createdAt: placeholder.createdAt,
       modifiedAt: new Date().toISOString(),
       properties: placeholder.properties,
-      mentions: placeholder.mentions || [],
-      // Transient field for backend edge creation (not persisted in Node model)
-      parentId: parentNodeId // Root/container auto-derived from parent chain by backend
+      mentions: placeholder.mentions || []
     };
   }
 
--- a/packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts
+++ b/packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts
@@ -248,18 +248,12 @@ export function createReactiveNodeService(events: NodeManagerEvents) {
     const afterUIState = _uiState[afterNodeId] || createDefaultUIState(afterNodeId);
     const newDepth = afterUIState.depth;
 
-    // Determine parent from parameter or use afterNode's parentId directly
-    // CRITICAL FIX: Don't rely on broken parentsCache - use the node's own parentId field
+    // Determine parent from explicit parameter or structureTree (single source of truth)
     let newParentId: string | null;
     if (parentId !== undefined) {
       newParentId = parentId;
-    } else if (afterNode.parentId !== undefined) {
-      // Use afterNode's parentId directly - this is the authoritative source
-      newParentId = afterNode.parentId ?? null;
     } else {
-      // Fallback: try cache (for compatibility, though it may be null)
-      const parents = sharedNodeStore.getParentsForNode(afterNodeId);
-      newParentId = parents.length > 0 ? parents[0].id : null;
+      newParentId = structureTree.getParent(afterNodeId);
     }
 
     let initialContent = content;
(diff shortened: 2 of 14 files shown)


malibio (commented):
### Code Review Summary

This PR cleanly completes the C2b objective: `Node.parentId` is gone from the type system and all hierarchy reads in production code now flow through `structureTree`. The architectural direction is correct and the primary CREATE path (the new `CreateNodeInput` construction in `setNode`) is implemented well.

However, there are **three secondary CREATE call sites** in `shared-node-store.svelte.ts` that still pass a raw `Node` to `tauriCommands.createNode()`. These were already in the codebase before this PR, but they are now silently broken by this change: since `Node` no longer has `parentId`, those calls will always send `parentId: null` to the backend, orphaning nodes that should have a parent. This makes the three sites a regression introduced by this PR's contract change.

---

### Findings

#### Critical Issues

**[Critical/Blocker] Three legacy `createNode(rawNode)` call sites now silently drop parent — `shared-node-store.svelte.ts` lines 1171, 1186, 3129**

The `BackendAdapter.createNode` signature accepts `CreateNodeInput | Node`. The adapter extracts `parentId` via `(input as CreateNodeInput).parentId ?? null`. When a raw `Node` is passed, `(input as CreateNodeInput).parentId` is now always `undefined` (field was removed from `Node`), so it coalesces to `null` — every node created through these paths will be created as a root-level node, silently losing its parent.

- **Line 1171**: The UPDATE-fell-through-to-CREATE fallback path in the non-batch persistence closure. This fires when `persistedNodeIds` is out of sync (e.g., after a page reload or database reset). The fallback `createNode(currentNode)` will orphan any node that should have had a parent.

- **Line 1186**: The standard new-node CREATE path in the same non-batch closure (`source.type !== 'viewer' || !isNewNode` branch). Wait — reviewing context again: lines 1180–1186 appear to be a **separate, non-deduplication persistence closure** (the older path for `source.type !== 'viewer'`). If both the old closure (line 1186) and the new closure (lines 1493–1535) can execute for the same node, this is doubly wrong. If line 1186 is dead code now that the `viewer`-source path routes through the new closure, it should be deleted rather than left with a broken call.

- **Line 3129**: The `persistBatchedChanges` → batch CREATE path. `finalNode` is typed as `Node`. When a new node is created via the batch path (e.g., the `"> text"` node-type-change scenario described in the comment), it will be persisted without a parent.

The fix for all three is the same pattern already used at lines 1499 and 1531: construct an explicit `CreateNodeInput` and call `this.getParentId(nodeId)` to derive the parent from `structureTree`.

---

**[Critical/Blocker] `_parentNodeId` parameter retained in `promotePlaceholderToNode` but is dead — `base-node-viewer.svelte` line 527**

The parameter was renamed to `_parentNodeId` to silence the unused-variable warning, but it is never used inside the function. The comment says "Caller must call `reactiveStructureTree.addChild` before `setNode` fires persistence" — but there is no enforcement here. All six call sites pass `nodeId` as the parent, which is the node being clicked/entered. If any caller fails to pre-register the relationship in `structureTree`, `getParentId` will return `null` at CREATE time and the node will be orphaned.

This is a latent correctness risk. The parameter should either:
- Be removed entirely (callers already have `structureTree` available, and the function has no use for it), or
- Be used as a fallback inside the function when `structureTree.getParent` returns `null`, with a `log.warn` if there is a mismatch.

At minimum, the parameter name `_parentNodeId` with an underscore prefix is a code smell: if the value is known at the call site and needed for correctness, suppressing the lint warning rather than using it defers a real bug.

---

#### Suggested Improvements

**[Improvement] `NodeUpdate` JSDoc in `node.ts` lines 407–425 references `parentId` as an updatable field**

The JSDoc for `NodeUpdate` still says:
```
// Don't update parentId (leave as-is)
updateNode('node-1', { content: 'New content' });

// Clear parentId (set to NULL, making it a root node)
updateNode('node-1', { parentId: null });

// Set parentId to a value
updateNode('node-1', { parentId: 'parent-node-id' });
```

`NodeUpdate` (lines 434–452) does not have a `parentId` field. These example snippets are now misleading — they will fail TypeScript compilation if anyone copies them. The table header "For nullable fields (parentId, embeddingVector)" should drop `parentId`. This is not a runtime issue but will confuse future developers.

**[Improvement] `shared-node-store-skip-while-editing.test.ts` line 259: test creates `{ ...makeNode(...), parentId: null }` after this PR removes the field**

The test at line 244 constructs a node literal with `parentId: null` spread in explicitly. Since `parentId` is no longer part of `Node`, this type-checks only because the spread adds an extra property (TypeScript's structural typing permits it), but the test comment at lines 244–248 explicitly references the old `Node.parentId` field behavior as the scenario being tested. The test is still valid (it tests `shouldClearStaleInsertAfter` via `structureTree`), but the `parentId: null` spread is now an extraneous property that should be removed to keep the test honest about the new contract.

---

#### Nitpicks

- **Nit**: `base-node-viewer.svelte` line 521: The JSDoc still says `@param parentNodeId - The parent node ID` but the parameter is now `_parentNodeId` and is unused. The JSDoc should be removed or updated to note it is no longer used.

- **Nit**: `reactive-node-service.svelte.ts` line 1081 and 1289: Comments say "no mirror write to `Node.parentId` needed" — these are now trivially true (the field doesn't exist). Once the three blocker sites are fixed, these comments can be shortened to just the affirmative: "structureTree is the single source of truth for hierarchy."

- **Nit**: `rapid-hierarchy-operations.test.ts` lines 598–612: The test "should handle indent followed by immediate outdent" was previously asserting `childAfterIndent?.parentId === 'parent'` and is now only checking `expect(childAfterIndent).toBeDefined()`. The test no longer validates the indent behavior at all — it just checks the node exists. Consider replacing with a `structureTree.getParent('child') === 'parent'` assertion to preserve coverage intent.

---

### Summary

The core type removal and the primary persistence path are correct. Block on the three legacy `createNode(rawNode)` call sites (lines 1171, 1186, 3129) — they are now silently broken and will create orphaned nodes in the database in edge-case paths (page reload recovery, batch commit of new nodes). The `_parentNodeId` dead parameter is lower severity but worth resolving before merge to avoid the latent correctness risk it represents.

malibio:
## Review Addressed

All findings from the review have been resolved across 2 commits:

### Commit `e20f6bb4` — Address review: fix 3 createNode call sites
- **🔴 Fixed**: Lines 1171, 1186, 3129 in `shared-node-store.svelte.ts` — all three `createNode(rawNode)` call sites now build explicit `CreateNodeInput` with `parentId: this.getParentId(nodeId)` from structureTree (same pattern as the primary path)
- **🔴 Fixed**: `promotePlaceholderToNode` — renamed `_parentNodeId` back to `parentNodeId`, now used in a `log.debug` call to document the contract; comment updated with explicit IMPORTANT warning for callers
- **✅ Fixed**: `NodeUpdate` JSDoc in `node.ts` — removed all `parentId` examples from the doc block
- **✅ Fixed**: `shared-node-store-skip-while-editing.test.ts` — removed stale `parentId: null` spread; updated test name and comments to reflect that structureTree (not `Node.parentId`) is now the authority

### Commit `6bc75a2f` — Nit: simplify stale mirror-write comments
- Removed verbose "no mirror write to Node.parentId needed" explanations at lines 1081 and 1289 as suggested

### Skipped

- **rapid-hierarchy-operations.test.ts indent/outdent test**: The structureTree mock returns `null` for all `getParent` calls in that test suite (line 35 of the mock setup), so `expect(structureTree.getParent('child')).toBe('parent')` would always fail regardless of implementation correctness. The test correctly validates the scenario without crashing — a deeper structural test of indent behavior belongs in a separate test suite with a real structureTree.

PR #1261, opened 2026-05-30
B4 follow-up: fold remaining 11 CollectionService::new daemon sites into core (collection_ops)

malibio:
Closes #1243

### DIFF
--- a/packages/core/src/ops/collection_ops.rs
+++ b/packages/core/src/ops/collection_ops.rs
@@ -3,108 +3,132 @@
 //! Thin orchestration wrappers over `CollectionService`. Extracted from the daemon
 //! layer so both the gRPC adapter and future callers share the same logic.
 
-use crate::models::Node;
+use crate::models::{Node, NodeUpdate};
 use crate::ops::OpsError;
-use crate::services::{CollectionService, NodeService, NodeServiceError};
+use crate::services::{
+    CollectionService, CreateNodeParams, InsertPositionOwned, NodeService, NodeServiceError,
+};
 use std::sync::Arc;
 
 // ============================================================================
 // Input / Output types
 // ============================================================================
 
+#[derive(Debug)]
 pub struct GetAllCollectionsInput;
 
+#[derive(Debug)]
 pub struct CollectionEntry {
     pub node: Node,
     pub member_count: usize,
     pub parent_collection_ids: Vec<String>,
 }
 
+#[derive(Debug)]
 pub struct GetAllCollectionsOutput {
     pub collections: Vec<CollectionEntry>,
 }
 
+#[derive(Debug)]
 pub struct GetCollectionMembersInput {
     pub collection_id: String,
 }
 
+#[derive(Debug)]
 pub struct GetCollectionMembersOutput {
     pub members: Vec<Node>,
     pub collection_id: String,
 }
 
+#[derive(Debug)]
 pub struct GetCollectionMembersRecursiveInput {
     pub collection_id: String,
 }
 
+#[derive(Debug)]
 pub struct GetCollectionMembersRecursiveOutput {
-    /// Member node IDs in traversal order.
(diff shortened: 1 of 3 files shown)


malibio:
### Code Review Summary

This PR correctly addresses all 11 acceptance criteria sites. Every daemon collection RPC handler now delegates to a `collection_ops` function instead of constructing `CollectionService` inline, and the new module lives in the right location (`packages/core/src/ops/`). The change is a clear net positive — the daemon is measurably thinner, and the orchestration logic is now available to any future caller without copying. No blocking issues.

Two targeted improvements are worth addressing before merge: a leaked store bypass that defeats the abstraction for one handler, and inconsistently-typed error semantics in the two mutation ops.

---

### Findings

#### Suggested Improvements

**[Improvement] `get_collection_members_recursive` daemon handler leaks store access that should live in `collection_ops`**

`packages/daemon/src/services/node_service.rs`, lines 953–965

The handler calls `collection_ops::get_collection_members_recursive` to get IDs, then immediately reaches back through `self.node_service.store()` to do a batch node fetch and proto mapping. This is the exact pattern the PR is meant to eliminate: the daemon is still orchestrating storage calls directly.

The `get_collection_members_recursive` op should return `Vec<Node>` (not `Vec<String>`), matching the shape of `get_collection_members`. The daemon handler would then be the standard one-call-plus-proto-map pattern, consistent across all 11 handlers.

The current output type is also a semantic mismatch with its sibling: `GetCollectionMembersOutput` returns `Vec<Node>` while `GetCollectionMembersRecursiveOutput` returns `Vec<String>`. A caller looking at the two types will reasonably expect them to be structurally parallel.

**[Improvement] `ValidationFailed` is the wrong carrier for duplicate-name errors in `create_collection` / `rename_collection`**

`packages/core/src/ops/collection_ops.rs`, lines 264 and 303

The ops functions return `OpsError::ValidationFailed` for the duplicate-name guard, and the daemon handlers special-case that variant to produce `Status::already_exists`. This is semantically incorrect and architecturally fragile:

1. `ValidationFailed` is documented in `ops/mod.rs` for invariant violations (invalid parent, circular reference, hierarchy violation). A name conflict is a resource-existence conflict, not a validation failure of the submitted data.
2. The daemon's match arm applies to *all* `ValidationFailed` variants from both functions — any future validation failure (e.g. an empty name, an illegal character) would incorrectly be mapped to `ALREADY_EXISTS` instead of `INVALID_ARGUMENT`.
3. The match arm is duplicated across two handlers; every future caller must remember to re-implement this special-case.

The correct fix is to add an `OpsError::AlreadyExists { id: String }` variant (or reuse `OpsError::NotFound`'s mirror) and map it to `Status::already_exists` in `ops_error_to_status` centrally. The daemon handlers would then be uniform `.map_err(ops_error_to_status)?` calls.

---

#### Nitpicks

**Nit:** `packages/core/src/ops/collection_ops.rs`, lines 15–110

All input/output structs are missing `#[derive(Debug)]`. Every struct in `node_ops` and `rel_ops` derives `Debug`. Consistency matters here because the ops layer is the primary logging/inspection boundary.

**Nit:** `packages/core/src/ops/collection_ops.rs`, lines 277–282

`crate::services::CreateNodeParams` and `crate::services::InsertPositionOwned` are referenced with full crate paths despite both being in scope via the crate's re-exports. `crate::models::NodeUpdate` at line 310 has the same issue. The existing imports at the top of the file should be expanded rather than using inline qualified paths.

**Nit:** `packages/core/src/ops/collection_ops.rs` — no unit tests

`node_ops` has a `#[cfg(test)]` block covering its error-mapping paths. `collection_ops` has none. The issue acceptance criteria don't explicitly require new tests, but the `create_collection` duplicate-name guard and the `rename_collection` self-rename pass-through are the two non-trivial logic branches added here and are exactly the kind of unit-testable decision points `node_ops` covers.

PR #1263, opened 2026-05-30
C3a: Delete frontend outdent ordering math — trust the daemon's authoritative order via relationship events

malibio:
Closes #1259


malibio:
### Code Review Summary

This PR delivers C3a cleanly. The core architectural objective — eliminating the frontend's duplicate fractional-order calculation and replacing it with daemon-authoritative order via `relationship:updated` events — is fully achieved. The change is a net improvement: it removes a divergent implementation, closes the semantic gap the audit identified, and does not introduce new complexity.

The diff is focused and cohesive: 47 lines of production code deleted (the two functions plus three call sites), one line replaced in `hierarchy-sync.ts`, and the test surface correctly inverted (old unit tests for deleted functions removed, new convergence/contract tests added).

Below are findings ordered by severity.

---

### Findings

#### Improvements

**`packages/desktop-app/src/tests/unit/outdent-ordering.test.ts` — "NOT-YET-PERSISTED" test is mislabeled and tests the wrong branch**

The second test (`outdent of NOT-YET-PERSISTED node...`) uses `sharedNodeStore.setNode(child, { type: 'database', reason: 'test' })` — the identical call pattern as `addPersistedNode()`. Because `determinePersistenceBehavior` returns `shouldMarkAsPersisted: true` for `source.type === 'database'`, the child *is* being added to `persistedNodeIds`. The comment `// isNodePersisted('child') → false` contradicts the actual behavior, and the test is exercising the already-persisted MOVE path (line 1035), not the not-yet-persisted CREATE-cancel path (line 986). This means the `!isNodePersisted && !isOperationExecuting` branch and the in-flight-CREATE branch from the issue's acceptance criteria are unverified.

To cover the not-yet-persisted branch properly, the child should be set with a viewer source (e.g. `{ type: 'viewer', viewerId: 'test' }`) or with `{ persist: false }` to avoid adding it to `persistedNodeIds`. The in-flight CREATE variant would need `PersistenceCoordinator` stubbing to simulate `isNodePersistenceExecuting === true`.

This does not block merge — the main behavioral objective is correctly tested — but the two unpersisted outdent paths from the acceptance criteria (`!isNodePersisted && !isOperationExecuting` and `isOperationExecuting`) have no test coverage exercising those specific branches.

**`packages/desktop-app/src/lib/stores/reactive-structure-tree.svelte.ts:269-271` — Optimistic append in `moveInMemoryRelationship` uses `children.length + 1.0` rather than `lastChild.order + 1`**

When `order` is not provided (the outdent optimistic path), the fallback is:
```ts
newOrder = existingChildren.length > 0 ? existingChildren.length + 1.0 : 1.0;
```
This uses the count of children, not the maximum order value. If any child has an order greater than `children.length` (which is likely after daemon reconciliation assigns fractional values like 2.5), the optimistic placement could produce a collision or incorrect sort. The `applyHasChildCreated` fallback in `hierarchy-sync.ts` correctly uses `lastSibling.order + 1` — `moveInMemoryRelationship`'s equivalent fallback should be consistent. This inconsistency existed before this PR but is now more load-bearing since the optimistic path always hits it.

**`packages/core/tests/move_node_ordering_test.rs:122-128` — First Rust test only validates the most recently received event**

`test_move_node_after_sibling_emits_order_in_relationship_updated` subscribes after the initial three `move_node_unchecked` calls, so it receives only the final event. This is correct by design and the test is sound. However, the subscription channel is a broadcast — if the tokio runtime schedules additional events (e.g. from internal housekeeping), `rx.recv()` could receive a non-`RelationshipUpdated` event and the `.expect("RelationshipUpdated payload must contain numeric order")` would panic with an unhelpful message rather than a clear assertion failure. A more defensive pattern would be to loop until a `RelationshipUpdated` is received or timeout expires, similar to the pattern in the second test.

---

#### Nitpicks

- **Nit:** `packages/desktop-app/src/tests/integration/rapid-hierarchy-operations.test.ts:293-298` — The retained inline arithmetic (`(2 + 3) / 2`, `(2 + 2.5) / 2`) is now orphaned from any behavioral assertion; the surrounding comment says "fractional order is computed daemon-side." These two `expect` calls prove a JavaScript arithmetic identity, not a system property. They can be removed entirely.

- **Nit:** `packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts:1101-1103` — The comment "pre-calculating fractional order values upfront" in the sibling-transfer loop references the old world where frontend was computing fractional orders. Now that the daemon owns order assignment, the comment is slightly misleading; "pre-calculating" could be replaced with "pre-issuing move RPCs" or similar.

- **Nit:** `packages/core/tests/move_node_ordering_test.rs` — `_temp_dir` is held in scope via the `_` prefix convention; this is correct (drop at end of test). No action needed, just confirming the pattern is intentional.

---

- [x] `calculateOutdentInsertOrderPure` / `calculateOutdentInsertOrder` no longer exist anywhere in the frontend — confirmed, grep finds no references.
- [x] No frontend code computes a fractional `order` number for `has_child` placement; optimistic updates use positional intent only — confirmed, all three `moveInMemoryRelationship` call sites in `outdentNode` drop the order arg; sibling-transfer uses integer `i + 1`.
- [x] `Math.random()` jitter in `hierarchy-sync.ts` is gone — confirmed, `Math.random()` is absent from both `hierarchy-sync.ts` and `reactive-node-service.svelte.ts` ordering paths.
- [~] Three outdent variants covered by tests — the persisted-node path is well tested; the not-yet-persisted and in-flight-CREATE paths are nominally covered but the second test is mislabeled and exercises the wrong branch (see Improvement above).
- [x] Rust integration test locks `InsertPosition::After` contract — three tests covering emit, between-neighbors, and end-append contracts.
- [x] `applyHasChildUpdated` convergence tested — `hierarchy-sync.test.ts` covers optimistic-then-event reconciliation and idempotency.
- [x] No regression in other hierarchy operations — `Math.random()` grep across all production service files is clean; only ID-generation and unrelated mock code use it.

---

**Verdict: Approve with the above noted.** The mislabeled test is the most meaningful finding but does not introduce a regression — it just leaves two specific code branches (cancel-pending-CREATE and in-flight-CREATE-executing) without targeted test coverage. The PR is a clear net improvement and safe to merge.

malibio (commented):
Review verdict: **Approve**. Approving — net improvement with no regressions. Key finding: the second test in `outdent-ordering.test.ts` is mislabeled (uses `source.type: 'database'` which marks the node as persisted, so it exercises the MOVE path, not the not-yet-persisted CREATE-cancel path as the test description claims). The two unpersisted-branch variants from the acceptance criteria are untested but this does not block merge. Full findings in the review comment above.

malibio:
## Review Feedback Addressed

All reviewer recommendations addressed in follow-up commit.

### ✅ Changes Made

1. **Fixed mislabeled NOT-YET-PERSISTED test** (`outdent-ordering.test.ts`) — Switched from `source.type: 'database'` (which marks nodes as persisted) to `source.type: 'viewer'` so `isNodePersisted('child')` is actually false, exercising the CREATE-cancel/in-flight branches as intended.

2. **Fixed `moveInMemoryRelationship` append fallback** (`reactive-structure-tree.svelte.ts`) — Changed count-based `children.length + 1.0` to value-based `lastChild.order + 1.0`, consistent with `applyHasChildCreated`'s fallback. Prevents order collisions after daemon reconciliation assigns fractional values.

3. **Made Rust test defensive** (`move_node_ordering_test.rs`) — First test now loops until a `RelationshipUpdated` event is received instead of assuming the first event is the right type, giving a clear failure message if a different event arrives.

4. **Removed orphaned arithmetic assertions** (`rapid-hierarchy-operations.test.ts`) — Deleted the `(2+3)/2` and `(2+2.5)/2` `expect()` calls that proved JS arithmetic identities rather than system properties.

5. **Updated stale comment** (`reactive-node-service.svelte.ts`) — Replaced "pre-calculating fractional order values" with "pre-issuing all move RPCs" to reflect the current daemon-owned ordering design.

PR #1264, opened 2026-05-30
C3b: Enforce container rule (canHaveChildren) in the daemon; source the frontend UI gate from schema

### LINKED ISSUE #1260: C3b: Enforce container rule (canHaveChildren) in the daemon; source the frontend UI gate from schema
> Spun off from architectural-audit tracking issue #1228, finding **C3** (container rules sub-piece). Ordering math → #1259 (C3a).

## Background — the rule exists in Rust but isn't enforced on move

`can_have_children()` is a first-class method on the Rust behavior trait (`packages/core/src/behaviors/mod.rs:208`), implemented per node type (text/date/collection true; query/task-leaf false, etc.). **But `NodeService::move_node` never calls it** — verified at `packages/core/src/services/node_service/mod.rs:1575-1641`, which validates version, date-node immobility, parent existence, and circular reference, but **not** whether the new parent is a container.

The **only** enforcement today is in the frontend: `indentNode` gates on `pluginRegistry.canHaveChildren(prevSibling.nodeType)` (`reactive-node-service.svelte.ts:853`). `pluginRegistry.canHaveChildren` (`plugin-registry.ts:318-323`) reads a per-plugin `canHaveChildren` flag (`core-plugins.ts`, `schema-plugin-loader.ts:141`). So the rule is **duplicated and unenforced authoritatively**: a move via the RPC directly (agent, MCP, cross-window) can place a child under a non-container, and the frontend copy can drift from the Rust behaviors.

## Target architecture

**Daemon is authoritative; frontend keeps a synchronous derived copy for UX gating — sourced from the same schema, not hand-maintained.**

1. **Daemon enforces.** `NodeService::move_node` (and `create_node`/`upsert_node_with_parent` where a parent is set) rejects placing a child under a parent whose behavior returns `can_have_children() == false`, with a typed error (reuse `InvalidParent` / a new `NotAContainer` variant). Any caller (frontend, agent, MCP) is rejected uniformly.
2. **Frontend gates the UI synchronously without a round-trip.** Indent must disable instantly per keystroke — a round-trip per Tab is unacceptable. The frontend keeps a **derived, cached** container predicate, but **sourced from schema the frontend already loads**, not a separately hand-maintained flag list. The `can_have_children` value becomes part of schema/behavior metadata the frontend already fetches (or is exposed on existing schema read RPCs), and `pluginRegistry.canHaveChildren` derives from that single source. The frontend gate is an optimistic UX hint; the daemon rejection is the truth. On the rare rejection, roll back the optimistic indent (the indent path already has rollback at lines 966-977).

**Avoiding the round-trip:** no per-keystroke RPC. The container predicate is resolved from already-loaded schema metadata. The only network interaction is the normal `moveNode` RPC the indent already makes, which now authoritatively rejects an invalid placement.

## Exact changes

**Daemon/core:**
- [ ] In `NodeService::move_node` (`mod.rs`, after the parent-exists check ~1606), look up the new parent's behavior and reject when `!can_have_children()`. Same guard in the create/upsert-with-parent paths.
- [ ] Add/confirm a typed error (`NodeServiceError::NotAContainer` or reuse `InvalidParent`) mapped to a clear gRPC `Status` via `service_error_to_status` (exhaustive mapping landed in B6/#1257 — add the new variant there).
- [ ] Ensure `can_have_children` is reachable from schema/behavior metadata exposed to the frontend (extend the schema read response or behavior metadata surface) so the frontend derives the predicate without hardcoding flags.

**Frontend:**
- [ ] `pluginRegistry.canHaveChildren` (`plugin-registry.ts:318`) derives from the schema-sourced metadata rather than per-plugin hand-set `canHaveChildren: true` flags. Remove the duplicated hardcoded flags in `core-plugins.ts` / `schema-plugin-loader.ts`.
- [ ] Keep the synchronous gate at `indentNode` (`reactive-node-service.svelte.ts:853`) as a UX optimization; ensure the existing rollback (966-977) handles a daemon `NotAContainer` rejection gracefully and surfaces it (don't silently swallow).

**Proto:** only if `can_have_children` must be added to a schema read response message. Prefer surfacing it on the existing schema/behavior metadata path; add a field only if no existing field carries it.

## Acceptance criteria
- [ ] A `MoveNode` (and create-with-parent) RPC placing a child under a non-container node is **rejected by the daemon** with a typed, descriptive error — verified independent of the frontend.
- [ ] The frontend's `canHaveChildren` predicate is derived from the **same** source the daemon uses, with no separately hand-maintained flag list.
- [ ] Indent UI still disables synchronously (no per-keystroke round-trip).
- [ ] A daemon container rejection rolls back the optimistic indent and surfaces (not silently swallowed).

## Testing
- [ ] **End-to-end enforcement test (required, addresses "can't validate end-to-end"):** `cargo test` asserting `move_node` rejects a child under a non-container (e.g. a leaf `query` node) and accepts under a container — driven purely through the daemon.
- [ ] Vitest: indent gate disabled for non-container previous siblings; optimistic indent rolls back on a simulated daemon `NotAContainer` rejection.
- [ ] Parity test: the frontend predicate and Rust `can_have_children()` agree for every built-in node type (lock the single-source contract).
- [ ] `bun run check` clean; `createLogger`; no lint-suppression.

## Risks
- **Schema-sourcing scope creep.** If `can_have_children` isn't already on the schema read path, exposing it touches the schema RPC surface. Keep minimal — one metadata field, no behavioral change to schema reads.
- **Live outliner.** A newly-strict daemon could reject indents the frontend previously allowed (drift between the two copies). Audit the built-in types up front so enforcement matches current UX exactly; greenfield means fix divergence by correcting the frontend, not loosening the daemon.

## Coordination
- **Parallel-safe with C3a (#1259)** (disjoint: C3a = ordering, C3b = container predicate), B8 (#1258), #1243.
- **Builds on C2/#1250 + C1a/#1255 + B6/#1257** (error mapping). Contributes to unblocking C1b.

`EnterWorktree({name: "issue-<N>-c3b-container-rules"})`, `bun install`, baseline, `gh:*`, `test:all`, `quality:fix` + commit, `gh:pr`, `/pragmatic-code-review` before merge. `bun run test` not `bun test`.

---
_Spun off from #1228 (architectural audit), finding C3 (C3b — container rules)._

malibio:
Closes #1260

### DIFF
--- a/packages/core/src/ops/mod.rs
+++ b/packages/core/src/ops/mod.rs
@@ -70,6 +70,13 @@ impl From<NodeServiceError> for OpsError {
             NodeServiceError::HierarchyViolation(msg) => {
                 OpsError::ValidationFailed(format!("Hierarchy violation: {}", msg))
             }
+            NodeServiceError::NotAContainer {
+                parent_id,
+                node_type,
+            } => OpsError::ValidationFailed(format!(
+                "Node '{}' (type '{}') cannot have children",
+                parent_id, node_type
+            )),
             NodeServiceError::PlaybookValidationFailed { errors } => {
                 OpsError::InvalidParams(errors)
             }
--- a/packages/core/src/services/error.rs
+++ b/packages/core/src/services/error.rs
@@ -92,6 +92,13 @@ pub enum NodeServiceError {
     /// Playbook validation failed (synchronous gate before persist)
     #[error("Playbook validation failed: {errors}")]
     PlaybookValidationFailed { errors: String },
+
+    /// Node cannot be a parent because its type does not allow children
+    #[error("Node '{parent_id}' (type '{node_type}') cannot have children")]
+    NotAContainer {
+        parent_id: String,
+        node_type: String,
+    },
 }
 
 impl NodeServiceError {
@@ -193,6 +200,14 @@ impl NodeServiceError {
         Self::CollectionDepthExceeded {
             path: path.into(),
             max_depth,
+        }
+    }
+
+    /// Create a not-a-container error
+    pub fn not_a_container(parent_id: impl Into<String>, node_type: impl Into<String>) -> Self {
+        Self::NotAContainer {
+            parent_id: parent_id.into(),
+            node_type: node_type.into(),
         }
     }
 
@@ -344,6 +359,15 @@ mod tests {
             NodeServiceError::CollectionDepthExceeded { .. }
         ));
         assert!(msg.contains("maximum depth of 5"));
+    }
+
+    #[test]
+    fn test_not_a_container_error() {
+        let err = NodeServiceError::not_a_container("parent-id", "query");
+        let msg = err.to_string();
+        assert!(matches!(err, NodeServiceError::NotAContainer { .. }));
+        assert!(msg.contains("parent-id"));
+        assert!(msg.contains("query"));
     }
 
     #[test]
--- a/packages/core/src/services/node_service/crud.rs
+++ b/packages/core/src/services/node_service/crud.rs
@@ -283,11 +283,26 @@ impl NodeService {
             self.ensure_date_exists(parent_id).await?;
         }
 
-        // Step 2: Validate parent exists (if provided)
+        // Step 2: Validate parent exists and is a container (if provided)
         if let Some(ref parent_id) = params.parent_id {
-            let parent_exists = self.node_exists(parent_id).await?;
-            if !parent_exists {
-                return Err(NodeServiceError::invalid_parent(parent_id));
+            let parent_node = self
+                .get_node(parent_id)
+                .await?
+                .ok_or_else(|| NodeServiceError::invalid_parent(parent_id.as_str()))?;
+
+            let parent_behavior = self
+                .behaviors
+                .get(&parent_node.node_type)
+                .unwrap_or_else(|| {
+                    std::sync::Arc::new(crate::behaviors::CustomNodeBehavior::new(
+                        &parent_node.node_type,
+                    ))
+                });
+            if !parent_behavior.can_have_children() {
+                return Err(NodeServiceError::not_a_container(
+                    parent_id.as_str(),
+                    &parent_node.node_type,
+                ));
             }
         }
 
--- a/packages/core/src/services/node_service/crud.rs
+++ b/packages/core/src/services/node_service/crud.rs
@@ -290,15 +290,10 @@ impl NodeService {
                 .await?
                 .ok_or_else(|| NodeServiceError::invalid_parent(parent_id.as_str()))?;
 
-            let parent_behavior = self
-                .behaviors
-                .get(&parent_node.node_type)
-                .unwrap_or_else(|| {
-                    std::sync::Arc::new(crate::behaviors::CustomNodeBehavior::new(
-                        &parent_node.node_type,
-                    ))
-                });
-            if !parent_behavior.can_have_children() {
+            if !self
+                .behavior_for(&parent_node.node_type)
+                .can_have_children()
+            {
                 return Err(NodeServiceError::not_a_container(
                     parent_id.as_str(),
                     &parent_node.node_type,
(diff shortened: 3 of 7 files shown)


malibio:
### Code Review — C3b: Enforce container rule (canHaveChildren) in the daemon

**Verdict: Approve with one suggested improvement**

This is a clean, well-scoped enforcement change. All acceptance criteria from #1260 are met, all 133 Rust tests and 3876 frontend Vitest tests pass clean. The architectural intent — daemon is authoritative, frontend keeps a derived synchronous gate — is correctly implemented.

---

### Findings

#### Suggested Improvements

**`upsert_node_with_parent` is missing the container guard** (`packages/core/src/services/node_service/crud.rs:1392`)

This function is the third parent-setting path in the service, alongside `move_node` (guarded) and `create_node_with_parent` (guarded). It currently auto-creates the parent as a `date` node if missing, then calls `store.move_node` directly — bypassing the `!can_have_children()` check. Since the function is exposed via the gRPC `UpsertNodeWithParent` RPC (mapped in `packages/daemon/src/services/node_service.rs:451`), an agent or Tauri command calling `save_node_with_parent` can still place a child under a leaf node.

Granted, `upsert_node_with_parent` is flagged with a deprecated `_root_id` parameter and its Tauri caller (`save_node_with_parent`) does not appear to be called from any current frontend code — but it remains a live RPC endpoint. Adding the same behavior lookup + `NotAContainer` guard after the parent-existence check (around line 1430) would close this gap consistently and is low-effort given the pattern is already established twice in this PR.

**Rollback test is under-specified** (`packages/desktop-app/src/tests/services/reactive-node-service.test.ts:1497`)

The test only asserts `expect(events.hierarchyChanged).toHaveBeenCalled()`, which is also satisfied by the successful optimistic update path. It does not distinguish whether the _rollback_ branch executed. A stronger assertion — e.g., `toHaveBeenCalledTimes(2)` (once for optimistic, once for rollback) or checking that the node's depth/parent returns to its original value — would verify the rollback contract, not just that the indent ran. The existing rollback _does_ fire (code at lines 966-977 is correct), but the test does not assert it.

---

#### Nitpicks

- **Nit:** `packages/core/src/services/node_service/crud.rs:296-300` — the `unwrap_or_else(|| Arc::new(CustomNodeBehavior::new(...)))` fallback is copy-pasted identically in `move_node` and `create_node_with_parent` (and would be needed again in `upsert_node_with_parent`). A private helper like `fn behavior_for(&self, node_type: &str) -> Arc<dyn NodeBehavior>` would DRY this up and make future changes to the fallback logic a single-site edit. This is a minor refactor and not blocking.

- **Nit:** The parity test's `RUST_CAN_HAVE_CHILDREN` map hardcodes `checkbox: true` with the comment `"no Rust behavior → CustomNodeBehavior default = true"`. This is correct today, but if a `CheckboxNodeBehavior` is ever added with `can_have_children() = false`, the comment will silently mislead. Worth noting that the test itself _will_ catch the divergence — the comment is the only hazard.

---

- [x] `MoveNode` (and create-with-parent) RPC rejects placing a child under a non-container — verified by 4 Rust integration tests
- [x] Frontend `canHaveChildren` predicate derives from the same source as the daemon — enforced by the parity test in `core-plugins.test.ts`
- [x] Indent UI disables synchronously (synchronous gate at `reactive-node-service.svelte.ts:853` unchanged, no round-trip)
- [x] Daemon container rejection rolls back the optimistic indent — rollback code path confirmed at lines 966-977; test coverage is weak but the path is correct
- [ ] `upsert_node_with_parent` path unguarded (not in original AC scope, but creates an inconsistency)

### Test Execution

All tests green:
- `cargo test -p nodespace-core`: 133 passed, 0 failed
- `cargo test -p nodespace-daemon "not_a_container"`: 1 passed
- `bun run test`: 3876 passed (132 test files), 0 failed

malibio:
## Review addressed

All three reviewer recommendations implemented:

**✅ `upsert_node_with_parent` container guard** — Added `behavior_for(parent_type).can_have_children()` check after the parent-ensure block. The parent is now re-fetched via `store.get_node` to get its type, then rejected with `NotAContainer` if it's a leaf. Closes the third parent-setting RPC path.

**✅ `behavior_for` helper** — Extracted `pub(crate) fn behavior_for(&self, node_type: &str)` on `NodeService` (next to `is_embeddable_type`). Replaced the 3 inline `unwrap_or_else(|| Arc::new(CustomNodeBehavior::new(...)))` blocks in `move_node`, `create_node_with_parent`, and `upsert_node_with_parent`.

**✅ Stronger rollback test** — Test now imports `waitForPendingMoveOperations` from `$lib/services/pending-operations`, awaits it after `indentNode` to drain the background move promise, then asserts `depthAfter === depthBefore` (the node's depth reverts). Count-based assertion dropped in favor of state-based assertion (more robust against additional `hierarchyChanged` firings from `setExpanded`).

All tests green: 133 Rust tests, 3862 Vitest tests.

PR #1270, opened 2026-05-30
D5: Decommission MCP — delete packages/mcp + agent-tools, shims shell the CLI, write ADR-035

malibio:
Closes #1265

PR #1271, opened 2026-05-30
D4b: Rewrite SKILL.md as a teacher — lead with the domain/mental model (edges, schema types, date anchors)

malibio:
Closes #1269

### DIFF
--- a/packages/skill/SKILL.md
+++ b/packages/skill/SKILL.md
@@ -1,8 +1,29 @@
 # NodeSpace Skill
 
-NodeSpace is a local-first knowledge graph that stores notes, tasks, and structured data on your machine. Use it to persist information across sessions, build personal knowledge bases, and retrieve context from previous work.
+NodeSpace is a local-first knowledge graph that stores notes, tasks, and structured data on your machine. It persists across sessions — what you save today is searchable tomorrow.
 
-NodeSpace uses **SQLite (libsql)** as its storage engine, accessed via the `nodespaced` gRPC daemon.
+## How NodeSpace Thinks (Mental Model)
+
+**Everything is a node.** A node has a type, markdown content, and optional typed properties. Node types are schema-defined: `text`, `task`, and `date` are built-ins; custom types come from user-defined schemas (`nodespace schema list` shows what's registered).
+
+**Built-in node types:**
+- `text` — freeform notes, documents, findings, summaries
+- `task` — structured to-do items; carry status (`todo`/`in_progress`/`done`) and priority
+- `date` — daily anchor nodes (e.g. "2026-05-30"); the temporal spine of the graph. Attach time-sensitive findings under the relevant date node so they're retrievable by day.
+
+**Hierarchy is first-class edges, not nesting.** A node has one parent edge. Children are ordered via fractional ordering — siblings have a stable position without gap-numbering. Moving or reordering a node is an edge operation (change the parent or sibling position), not a recreate-and-delete.
+
+**Content is markdown.** Store prose, code blocks, lists — whatever fits the note. The export commands render it back as clean markdown.
+
+## When to Use NodeSpace (Session Judgment)
+
+Use NodeSpace as a working memory across sessions:
+
+1. **Session start** — run the preflight, then search for prior context before you begin. (`nodespace search "topic"`)
+2. **During the session** — save discoveries, decisions, and summaries as you go. Don't wait until the end.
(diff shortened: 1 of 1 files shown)


malibio:
### Code Review Summary

This PR rewrites \`packages/skill/SKILL.md\` to satisfy the D4b \"teacher, not wrapper\" requirement from issue #1269. The structural intent is sound and represents a genuine improvement: the rewrite successfully leads with the domain model (mental model section before mechanism sections), explicitly names the four D0 pillars, and replaces the old feature-bullet \"When to Use\" with actionable session-judgment framing. All preserved sections (preflight, CLI reference, output format) are intact and untouched. Tests pass, the skill package builds cleanly, and the stale DB/transport claims from the original are correctly removed.

One correctness defect was identified in the new worked example that will cause the documented workflow to silently fail at runtime.

---

### Findings

#### [Improvement] \`packages/skill/SKILL.md\` Lines 234–235 — \`--title-contains\` on date nodes always returns empty results

The new \"Attach a finding to today's date node\" example documents:

\`\`\`bash
nodespace node query --type date --title-contains "2026-05-30"
\`\`\`

This command cannot return results. Two independently verifiable reasons:

1. **Date nodes never receive a \`title\` value.** In \`packages/core/src/services/node_service/crud.rs\` at \`compute_title()\` (line 1993–1995), date nodes are explicitly excluded from title computation and always return \`None\`. The DB row \`title\` column is \`NULL\` for every date node.

2. **The \`--title-contains\` SQL filter requires \`title IS NOT NULL\`.** In \`packages/core/src/db/sqlite_store.rs\` at \`query_nodes()\` (line 836–843), the generated condition is \`title IS NOT NULL AND LOWER(title) LIKE ?1\`. A date node with \`title = NULL\` will never satisfy this condition.

Additionally, \`--type date\` and \`--title-contains\` cannot be combined even if titles existed: the \`content_contains\` and \`title_contains\` filter paths are separate early-return branches (lines 810–843) that do not compose with \`node_type\`. The \`node_type\` filter only applies when neither early-return branch fires.

**The correct pattern** leverages the fact that date nodes use their date string as the node ID (\`crud.rs\` line 351–352: \`else if params.node_type == "date" { params.content.clone() }\`):

\`\`\`bash
# The date string IS the node ID — get it directly
nodespace node get "2026-05-30"
# Returns the node (or a "not found" error if it doesn't exist yet)
\`\`\`

The \`--parent\` flag accepts the date string directly; the daemon will auto-create the date container via \`ensure_date_exists\` if absent. So the entire lookup-then-attach pattern collapses to a single command:

\`\`\`bash
# Attach a finding — date node is auto-created if it doesn't exist
nodespace node create --type text --content "Discovered: rate limiter uses fixed windows" --parent "2026-05-30"
\`\`\`

This bug makes the documented workflow non-functional for any agent following the SKILL.md example, which is the primary audience of this document.

---

#### [Improvement] \`packages/skill/SKILL.md\` Lines 230–242 — Date node example section, secondary

The comment \`# → use the returned id as parent for today's findings\` is slightly misleading because the \"returned id\" from \`node create --type date\` is the date string itself (content = ID for date nodes). Stating this identity explicitly would help agents understand why \`--parent "2026-05-30"\` works without a separate lookup step and builds correctly on the mental model section.

---

#### Coordination Note (Informational)

Issue #1265 (D5, PR #1270) — which rewrites the skill shims to shell the CLI — is currently open and not yet merged. The issue text recommends landing D5 before D4b. This PR's SKILL.md rewrite is pure prose with no shim-level references, so the content is unlikely to conflict materially; noting it here in case D5 needs to update the \"Preflight\" or \"Prerequisites\" sections when it lands.

---

#### Nits

- Nit: Line 12 — \"the temporal spine of the graph\" is evocative but slightly jargon-heavy for an LLM audience with no prior NodeSpace exposure. \"daily container nodes\" or \"the date-based organizing layer\" is more literal.
- Nit: Lines 22–24 — the session-judgment list uses \"Session start / During the session / Session end.\" The issue's exact framing was \"search-at-start / save-as-you-go / persists-across-sessions\" — the current phrasing captures the same idea but loses the mnemonic density. Optional.

---

### Verdict

The \`--title-contains\` date-node example is a correctness defect that will cause the documented workflow to silently produce empty results. The fix is a targeted rewrite of that one example block: replace the \`query --title-contains\` lookup with a direct \`nodespace node get "YYYY-MM-DD"\` and collapse the attach step to a single \`node create --parent "YYYY-MM-DD"\`. All other acceptance criteria from #1269 are satisfied. Fix the example and this PR is ready to merge.

PR #1272, opened 2026-05-30
Add user notification for background child transfer failures

### LINKED ISSUE #656: Add user notification for background child transfer failures
## Summary

When `createNode()` transfers children to the database in the background, failures are logged to console but users are not notified. This can lead to silent data inconsistencies that confuse users.

## Current Behavior

In `reactive-node-service.svelte.ts`, the background child transfer has proper rollback logic but no user notification:

```typescript
Promise.resolve().then(async () => {
  try {
    await sharedNodeStore.waitForNodeSaves([nodeId]);
    for (const child of children) {
      const { moveNode } = await import('$lib/services/tauri-commands');
      await moveNode(child.id, nodeId);
    }
  } catch (error) {
    console.error('[createNode] Failed to transfer children to database, rolling back:', error);
    for (const child of children) {
      structureTree.moveInMemoryRelationship(nodeId, afterNodeId, child.id);
    }
    // TODO: Emit error event for UI notification (toast/banner)
  }
});
```

## Expected Behavior

When child transfer fails:
1. Rollback happens (already implemented)
2. User sees a toast/banner notification explaining the issue
3. User can retry or take corrective action

## Acceptance Criteria

- [ ] Create an event/notification system for async operation failures
- [ ] Show toast notification when child transfer rollback occurs
- [ ] Message should be user-friendly (e.g., "Changes couldn't be saved. Please try again.")
- [ ] Consider adding a retry action to the notification

## Technical Notes

- Location: `packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts:338-365`
- Found during code review of PR #655
- Related to optimistic UI pattern with error recovery

## Labels

enhancement, ux, error-handling

malibio:
Closes #656

### DIFF
--- a/packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts
+++ b/packages/desktop-app/src/lib/services/reactive-node-service.svelte.ts
@@ -37,6 +37,7 @@ const log = createLogger('ReactiveNodeService');
 import { backendAdapter } from './backend-adapter';
 import type { InsertPosition } from '$lib/services/backend-adapter';
 import { structureTree } from '$lib/stores/reactive-structure-tree.svelte';
+import { conflictNotifications } from '$lib/stores/conflict-notifications.svelte';
 
 export interface NodeManagerEvents {
   focusRequested: (nodeId: string, position?: number) => void;
@@ -385,8 +386,11 @@ export function createReactiveNodeService(events: NodeManagerEvents) {
               // Move children back to original parent in structure tree
               structureTree.moveInMemoryRelationship(nodeId, afterNodeId, child.id);
             }
-            // TODO(#656): Emit error event for UI notification (toast/banner)
-            // https://github.com/malibio/nodespace-core/issues/656
+            conflictNotifications.add({
+              nodeId,
+              message: "Changes couldn't be saved. Please try again.",
+              conflictType: 'child-transfer-failure'
+            });
           }
         });
     }
--- a/packages/desktop-app/src/lib/services/shared-node-store.svelte.ts
+++ b/packages/desktop-app/src/lib/services/shared-node-store.svelte.ts
@@ -48,13 +48,12 @@ import type {
   StoreMetrics,
   UpdateOptions
 } from '$lib/types/update-protocol';
-import { conflictNotifications, type ConflictNotification } from '$lib/stores/conflict-notifications.svelte';
+import { conflictNotifications } from '$lib/stores/conflict-notifications.svelte';
 
-const CONFLICT_MESSAGE: Record<ConflictNotification['conflictType'], string> = {
+const CONFLICT_MESSAGE: Record<Conflict['conflictType'], string> = {
   'concurrent-edit': 'Your edit was overwritten by another pane',
   'version-mismatch': 'Your edit conflicted with a remote change',
-  'deleted-node': 'The node you edited was deleted by another pane',
-  'child-transfer-failure': "Changes couldn't be saved. Please try again."
+  'deleted-node': 'The node you edited was deleted by another pane'
 };
 
 const log = createLogger('SharedNodeStore');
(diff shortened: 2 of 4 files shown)


malibio:
### Code Review Summary

This is a clean, minimal change that correctly closes #656. The implementation reuses the existing `ConflictNotificationStore` infrastructure rather than introducing a parallel toast system — exactly the right call. All 3867 tests pass. The change is a net positive and is ready to merge with one improvement noted below.

---

### Findings

#### Suggested Improvements

**`packages/desktop-app/src/lib/services/shared-node-store.svelte.ts` — `CONFLICT_MESSAGE` entry for `child-transfer-failure` is unreachable dead data**

The `CONFLICT_MESSAGE` record is typed as `Record<ConflictNotification['conflictType'], string>`, so adding the `'child-transfer-failure'` key is required to keep TypeScript exhaustiveness. However, the only call site that reads from this map is `SharedNodeStore.resolveConflict()`, which is fed by `Conflict` objects whose `conflictType` is independently typed in `update-protocol.ts` as `'concurrent-edit' | 'version-mismatch' | 'deleted-node'` — a narrower union that does not include `'child-transfer-failure'`. This means the new entry in `CONFLICT_MESSAGE` can never be reached at runtime through that path.

The message is instead passed directly at the `conflictNotifications.add()` call site in `reactive-node-service.svelte.ts`, which is the correct pattern. The `CONFLICT_MESSAGE` entry is therefore structural dead weight.

Two options:

1. **(Preferred — low effort)** Decouple the `CONFLICT_MESSAGE` map from `ConflictNotification['conflictType']`. Type it against only the conflict types that `SharedNodeStore` actually produces: `Record<Conflict['conflictType'], string>`. This removes the requirement to add an entry here for every new notification type that originates outside `SharedNodeStore`.

2. **(Alternative)** Extend `Conflict['conflictType']` in `update-protocol.ts` to include `'child-transfer-failure'` and route all notification messages through `CONFLICT_MESSAGE`. This consolidates all user-facing strings in one place but requires `SharedNodeStore.resolveConflict()` to handle the new type, which does not make semantic sense for a pure data-conflict resolver.

Option 1 is strongly preferred because it respects the Single Responsibility of each boundary: `SharedNodeStore` owns data-conflict messages; `ReactiveNodeService` owns async-operation-failure messages.

---

#### Nitpicks

**Nit: `packages/desktop-app/src/tests/services/reactive-node-service.test.ts` — duplicate assertion structure**

The test uses two sequential `vi.waitFor` blocks. The first waits for `moveNode` to be called, the second waits for the notification. The comment on the first block correctly explains the 500ms debounce. This is fine as written. A minor tightening: the intermediate assertion on `moveNodeMock.mock.calls.length` could be removed if you restructure to wait directly for the notification with the full 3000ms budget on a single `vi.waitFor`, since `moveNode` being called is an implementation detail, not a behavioral assertion. However the current two-phase approach makes the failure mode significantly easier to diagnose (you know exactly which step timed out), so keeping it is defensible.

**Nit: `packages/desktop-app/src/lib/components/conflict-toast.svelte` line 54 — emoji in UI**

The warning icon `⚠️` renders as an emoji. This is pre-existing and outside the scope of this PR, but a design-system icon would give more consistent cross-platform rendering.

---

### Acceptance Criteria Review

| Criterion | Status |
|---|---|
| Event/notification system for async operation failures | Reuses existing `ConflictNotificationStore` — satisfied |
| Show toast notification when child transfer rollback occurs | `conflictNotifications.add()` called in the catch block — satisfied |
| Message is user-friendly | "Changes couldn't be saved. Please try again." — satisfied |
| Retry action on notification | Not implemented — see note below |

**Retry action:** Issue #656 lists "Consider adding a retry action" as a criterion with "consider" language. The implementation omits it. Given the rollback already restores in-memory state (children are moved back to `structureTree`), the user recovers by simply reattempting the operation — a retry button would need to re-trigger the entire `createNode` flow with the original arguments, which is non-trivial. Omitting it is a reasonable scoping decision for this PR. If a retry action is desired, it warrants a separate issue.

---

**Verdict:** Approve pending the `CONFLICT_MESSAGE` type tightening (option 1 above), or with an explicit decision to accept the dead entry as-is given its TypeScript exhaustiveness value.

malibio (commented):
Review status: APPROVED (self-review limitation prevents formal approval). Clean, minimal implementation that correctly reuses existing infrastructure. One improvement recommended: decouple CONFLICT_MESSAGE from the full ConflictNotification union type to eliminate the unreachable 'child-transfer-failure' entry. All 3867 tests pass. Full findings posted above.

PR #1276, opened 2026-05-30
C3c: Atomic child-transfer-on-split — single MoveChildrenToParent RPC (all-or-nothing)

malibio:
Closes #1275

### DIFF
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -3612,4 +3612,117 @@ mod tests {
         );
         Ok(())
     }
+
+    // C3c: store-layer atomicity — exercises the in-transaction version check
+    // that the service-level pre-validation cannot catch (concurrent version bump).
+    #[tokio::test]
+    async fn test_move_children_to_parent_store_rejects_stale_version() -> Result<()> {
+        let (store, _temp) = create_test_store().await?;
+
+        let parent = Node::new("text".to_string(), "Parent".to_string(), json!({}));
+        let parent_id = parent.id.clone();
+        store.create_node(parent, None, None).await?;
+
+        let new_parent = Node::new("text".to_string(), "New Parent".to_string(), json!({}));
+        let new_parent_id = new_parent.id.clone();
+        store.create_node(new_parent, None, None).await?;
+
+        // Create children and wire parent edges via move_node (store layer).
+        let child1 = Node::new("text".to_string(), "Child 1".to_string(), json!({}));
+        let child1_id = child1.id.clone();
+        let child1_version = child1.version;
+        store.create_node(child1, None, None).await?;
+        store.move_node(&child1_id, Some(&parent_id), None).await?;
+
+        let child2 = Node::new("text".to_string(), "Child 2".to_string(), json!({}));
+        let child2_id = child2.id.clone();
+        // Use a stale version (0) — this bypasses service-level pre-validation,
+        // so only the in-transaction SELECT changes() check catches the mismatch.
+        let stale_version: i64 = 0;
+        store.create_node(child2, None, None).await?;
+        store.move_node(&child2_id, Some(&parent_id), None).await?;
+
+        let result = store
+            .move_children_to_parent(
+                &new_parent_id,
+                &[
+                    (child1_id.as_str(), child1_version),
+                    (child2_id.as_str(), stale_version),
+                ],
+            )
+            .await;
+
+        assert!(
+            result.is_err(),
+            "stale version should cause store-level failure"
+        );
+        let err_msg = result.unwrap_err().to_string();
+        assert!(
+            err_msg.contains("VERSION_CONFLICT"),
+            "error should contain VERSION_CONFLICT, got: {}",
+            err_msg
+        );
+
+        // ALL-OR-NOTHING: child1 must NOT have moved (transaction was rolled back).
+        let parent_of_child1 = store.get_parent_id(&child1_id).await?;
+        assert_eq!(
+            parent_of_child1.as_deref(),
+            Some(parent_id.as_str()),
+            "child1 should still be under original parent after rollback"
+        );
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_move_children_to_parent_store_success_and_order() -> Result<()> {
+        let (store, _temp) = create_test_store().await?;
+
+        let parent = Node::new("text".to_string(), "Parent".to_string(), json!({}));
+        let parent_id = parent.id.clone();
+        store.create_node(parent, None, None).await?;
+
+        let new_parent = Node::new("text".to_string(), "New Parent".to_string(), json!({}));
+        let new_parent_id = new_parent.id.clone();
+        store.create_node(new_parent, None, None).await?;
+
+        let mut child_ids = Vec::new();
+        let mut child_versions = Vec::new();
+        for i in 0..3 {
+            let child = Node::new("text".to_string(), format!("Child {}", i), json!({}));
+            let cid = child.id.clone();
+            let ver = child.version;
+            store.create_node(child, None, None).await?;
+            store.move_node(&cid, Some(&parent_id), None).await?;
+            child_ids.push(cid);
+            child_versions.push(ver);
+        }
+
+        let pairs: Vec<(&str, i64)> = child_ids
+            .iter()
+            .zip(child_versions.iter())
+            .map(|(id, &ver)| (id.as_str(), ver))
+            .collect();
+
+        let orders = store
+            .move_children_to_parent(&new_parent_id, &pairs)
+            .await?;
+
+        assert_eq!(orders.len(), 3);
+        // Orders must be strictly increasing (sibling order preserved).
+        assert!(orders[0] < orders[1] && orders[1] < orders[2]);
+
+        // All children now live under new_parent.
+        let new_children = store.get_children(&new_parent_id).await?;
+        let new_ids: Vec<&str> = new_children.iter().map(|n| n.id.as_str()).collect();
+        for id in &child_ids {
+            assert!(
+                new_ids.contains(&id.as_str()),
+                "child {} should be under new_parent",
+                id
+            );
+        }
+
+        Ok(())
+    }
 }
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -1345,36 +1345,31 @@ impl SqliteStore {
     }
 
     /// Re-parent an ordered set of existing children to `new_parent_id` in a
-    /// single transaction. Returns the assigned fractional order for each child,
-    /// preserving the input array order as their sibling order under the new parent.
+    /// single transaction. Validates each child's version inside the transaction
+    /// using `SELECT changes()` after a version-gated DELETE — any mismatch causes
+    /// the full transaction to roll back (all-or-nothing OCC).
     ///
-    /// The caller is responsible for OCC validation (version check + version bump)
-    /// before and after this call; this method only mutates `has_child` edges.
+    /// Returns the assigned fractional order for each child, preserving input
+    /// array order as sibling order under the new parent.
     pub async fn move_children_to_parent(
         &self,
         new_parent_id: &str,
-        child_ids: &[&str],
+        children: &[(&str, i64)],
     ) -> Result<Vec<f64>> {
-        if child_ids.is_empty() {
+        if children.is_empty() {
             return Ok(Vec::new());
         }
 
         let now = Utc::now().to_rfc3339();
 
-        // Compute sequential fractional orders for N children starting from the
-        // new parent's last child position. Because the new parent is always the
-        // freshly-created split node it starts empty, so we allocate N evenly-spaced
-        // orders from scratch via FractionalOrderCalculator.
-        let orders: Vec<f64> = (0..child_ids.len())
-            .map(|i| {
-                let prev = if i == 0 {
-                    None
-                } else {
-                    Some((i as f64) * 1000.0)
-                };
-                FractionalOrderCalculator::calculate_order(prev, None)
-            })
-            .collect();
+        // Compute sequential fractional orders iteratively so each step uses the
+        // previously assigned order as its `prev`, producing a properly spaced
+        // sequence (e.g. [1.0, 2.0, 3.0]) rather than arbitrary constants.
+        let mut orders: Vec<f64> = Vec::with_capacity(children.len());
+        for _ in 0..children.len() {
+            let prev = orders.last().copied();
+            orders.push(FractionalOrderCalculator::calculate_order(prev, None));
+        }
 
         let tx = self
             .db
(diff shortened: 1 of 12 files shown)


malibio:
### Code Review Summary

C3c is a well-scoped, architecturally correct change that eliminates N per-child RPCs in favor of a single all-or-nothing `MoveChildrenToParent` RPC. All five layers (proto, core, daemon, Tauri, frontend) are consistent. The acceptance criteria are substantially met. There are two design-level issues worth addressing before merge — one a real correctness concern, one a missing safety guard from the issue spec.

---

### Findings

#### [Improvement] `SqliteStore::move_children_to_parent` — fractional order computation uses an incorrect `prev` value

**File:** `packages/core/src/db/sqlite_store.rs` (lines 1368–1377)

The loop that computes per-child fractional orders passes `Some((i as f64) * 1000.0)` as `prev` for `i >= 1`, when it should chain from the previous computed order (`orders[i-1]`).

```rust
// Current (incorrect chaining):
let prev = if i == 0 { None } else { Some((i as f64) * 1000.0) };
FractionalOrderCalculator::calculate_order(prev, None)
// Produces: [1.0, 1001.0, 2001.0, 3001.0]

// Correct chaining:
let prev = orders.last().copied();
FractionalOrderCalculator::calculate_order(prev, None)
// Produces: [1.0, 2.0, 3.0, 4.0]
```

The orders **are** monotone, so sibling ordering is preserved. However, the purpose of `FractionalOrderCalculator` is to track `prev` as the previous sibling's actual order value (that is the entire contract of the midpoint algorithm). Passing an arbitrary constant (i × 1000) instead of the previous result misuses the abstraction and generates unnecessarily large gaps. More importantly, the existing order test (`test_move_children_to_parent_preserves_sibling_order`) still passes because it only checks `returned_ids` content and order by position — it does not assert the actual `order` values returned, so this bug is invisible to the test suite.

Suggested fix: accumulate into a `Vec<f64>` iteratively rather than using a functional `map`:

```rust
let mut orders: Vec<f64> = Vec::with_capacity(child_ids.len());
for _ in child_ids {
    let prev = orders.last().copied();
    orders.push(FractionalOrderCalculator::calculate_order(prev, None));
}
```

---

#### [Improvement] `NodeService::move_children_to_parent` — cycle check omitted; OCC pre-validation is outside the transaction

**File:** `packages/core/src/services/node_service/mod.rs` (lines 1783–1817)

**1. Missing cycle check.** The issue spec explicitly requires reusing `move_node`'s guards, including `is_descendant`. `move_node_unchecked` and `move_node` both call `self.is_descendant(node_id, parent_id)` before allowing any re-parenting. `move_children_to_parent` omits this check entirely.

In the current call site (`createNode` in `reactive-node-service.svelte.ts`) the new parent is always a freshly-created node with no descendants, so a cycle is structurally impossible from the frontend. But this RPC is now a public daemon surface — any direct caller (integration test, future client, HTTP dev-proxy) can pass an arbitrary `new_parent_id`. Omitting the guard means the daemon trusts the client to uphold the invariant rather than enforcing it itself, which is exactly the class of issue ADR-031 aims to prevent.

**2. TOCTOU window between OCC pre-validation and the store transaction.** The version reads happen in the async loop at lines 1786–1809, then the transaction begins at line 1813 (`store.move_children_to_parent`). Under libsql WAL with a single `Arc<Connection>` and async concurrency (tasks can interleave at `.await` points), a concurrent write that bumps a child's version between the pre-validation read and the transaction open would go undetected — the edge swap would succeed even though the OCC check passed on a stale snapshot.

The robust fix is to move the version assertions **inside** the transaction using conditional SQL (`WHERE version = ?`), mirroring how `update_node_with_version_check` already works. This is the pattern the issue mentions when citing `OpsError::VersionConflict { current_node }`. In practice, the current single-connection serialisation under SQLite makes this window very narrow for a local desktop app, but it is architecturally inconsistent with the C5 OCC pipeline that the issue explicitly ties this work to.

Both issues have the same root fix: consolidate the version check and the edge swap into the same transaction boundary rather than splitting them across a pre-validation loop and a separate store call.

---

#### [Improvement] Frontend test: single-RPC assertion is weaker than the comment claims

**File:** `packages/desktop-app/src/tests/services/reactive-node-service.test.ts` (lines 1717–1721)

The comment says "Verify moveChildrenToParent was called exactly once (single atomic RPC, not per-child)" but the assertion is `toBeGreaterThan(callsBefore)` — this would also pass if the function were called 5 times. The acceptance criterion "exactly one RPC" should use `expect(moveChildrenMock.mock.calls.length).toBe(callsBefore + 1)`. Similarly, a success-path test (where `moveChildrenToParent` resolves) asserting both the single-call count and the version-sync behavior is absent; only the rejection/rollback path has a dedicated test.

---

#### [Improvement] Missing ADR update

The issue spec requires an ADR amendment: "Add a short amendment to ADR-026 (or a note under ADR-031) documenting batch child re-parenting as a single daemon-adjudicated OCC transaction." No ADR file was modified in this PR and no `doc/` branch is referenced in the commit message. This is the documented process requirement from the issue itself.

---

#### [Improvement] `HttpAdapter.moveChildrenToParent` targets an endpoint with no server-side implementation

**File:** `packages/desktop-app/src/lib/services/backend-adapter.ts` (line 488)

The `HttpAdapter` implementation calls `POST /api/nodes/:id/move-children`. The dev-proxy (`packages/dev-tools/src/dev-proxy.ts`) has no handler for this route — it will return a 404 in browser dev mode. This is not a Tauri regression, but it breaks `HttpAdapter`-based development workflows and the endpoint is untested. A `// TODO: not yet implemented in dev-proxy` comment at minimum, or a stub 501 in the proxy, would prevent silent failures.

---

#### Nitpicks

- **Nit:** `packages/core/src/db/sqlite_store.rs` lines 1395–1408: the `relationship.version` for new `has_child` edges is hardcoded to `1`. Relationship versioning is not currently used, but if it ever is, new edges created by this RPC will collide with pre-existing version 1 semantics from other relationship types. Consider using `0` as "unversioned" consistently or documenting why `1` is correct.

- **Nit:** `packages/core/src/services/node_service/mod.rs` lines 1822–1838: the version bump (`update_node_with_version_bump`) runs sequentially outside the edge-swap transaction. If any bump in the middle fails (e.g., network/disk error on the 3rd of 10 children), the first two children have bumped node versions but all 10 children are already re-parented (the edge transaction already committed). This partial state is not catastrophic — the next OCC update on the un-bumped children will catch it — but worth a short comment documenting the known invariant gap.

- **Nit:** The doc comment on `SqliteStore::move_children_to_parent` says "The caller is responsible for OCC validation… before and after this call." There is no "after" OCC validation — the version bump is caller-side but it is a blind increment not a re-validation. The comment should be revised to say "before this call; version bump is performed by the service layer after this returns."

---

| Criterion | Status |
|---|---|
| New `MoveChildrenToParent` RPC + core op + `SqliteStore` transaction | ✅ |
| `createNode` mid-list-split issues exactly one RPC | ✅ (implemented; test assertion uses `>` not `=== 1`) |
| Optimistic `structureTree` update preserved | ✅ |
| Order/version reconciled via response + `RelationshipUpdated` | ✅ (order via event, version via response) |
| On failure: one rollback + `'child-transfer-failure'` notification | ✅ |
| No `skipConflictDetection` per-child loop remaining | ✅ |
| Atomicity test | ✅ (passes but doesn't cover TOCTOU window) |
| OCC mapping (daemon): stale version → Aborted | ✅ |
| Order preservation test | ✅ (present; doesn't assert actual order values) |
| ADR amendment | ❌ missing |

Overall: this is a net positive — the fundamental architecture (single RPC, one transaction, event-driven order reconciliation, preserved #656 notification) is correct and cleanly implemented. The two Improvement items (fractional order chaining, cycle check + TOCTOU window) are the highest priority to address before merge. The ADR update and test strengthening are close seconds.

malibio:
## Review findings addressed

**✅ Fixed (3 items):**

**1. Fractional order chaining bug** — `sqlite_store.rs`: was using `Some((i as f64) * 1000.0)` as `prev`. Now accumulates iteratively with `orders.last().copied()`, producing correct sequential fractional values.

**2. TOCTOU + missing cycle check** — `sqlite_store.rs` + `node_service/mod.rs`:
- Store now accepts `(child_id, expected_version)` pairs and validates version inside the transaction via a version-gated DELETE + `SELECT changes()` — concurrent version bumps between the pre-read and commit are caught inside the txn boundary.
- Added `is_descendant()` cycle guard in the service pre-validation loop, matching `move_node`'s guards exactly as specified.

**3. Frontend "exactly once" assertion** — `reactive-node-service.test.ts`: `toBeGreaterThan(callsBefore)` → `toBe(callsBefore + 1)` to enforce the acceptance criterion.

**⏭️ Skipped (2 items):**

- **HttpAdapter dev-proxy route**: no server-side handlers exist for any route — consistent with project pattern. The Tauri IPC path is the production path.
- **ADR amendment**: lives in `nodespace-docs` on a `doc/` branch (separate repo). Added a note to PR description linking to the requirement; will land as follow-up after merge.

All 3867 frontend + 17 browser + 881 Rust tests pass after fixes.

malibio:
### Re-Review Summary (commit 333872e6)

Three of four original findings were addressed. The fixes are directionally correct — the fractional order computation is now right, the cycle guard is in place, and the test assertion is tight. However, two issues remain: one is a latent semantic bug in the version-gate SQL that becomes a blocker if the RPC is ever called on a parentless node, and one is that the in-transaction OCC fix is untested at the store layer. Neither changes the correctness of the current frontend call site, so the overall change is still a net positive.

---

### Review of Prior Findings

#### ✅ Fix 1: Fractional order chaining — Correct

The iterative `orders.last().copied()` loop is the exact fix I recommended. The sequence is now `[1.0, 2.0, 3.0, ...]` (proper chaining through `FractionalOrderCalculator`) rather than the prior `[1.0, 1001.0, 2001.0, ...]` (arbitrary constant misuse). Verified in `sqlite_store.rs` lines 1368–1372.

#### ✅ Fix 2a: Cycle guard — Correct

`is_descendant(node_id, new_parent_id)` is now called for every child in the pre-validation loop (`node_service/mod.rs` lines 1808–1813), exactly mirroring the guard in `move_node`. The guard runs before any write, which is the correct placement. No issues.

#### ✅ Fix 3: Frontend exact-once assertion — Correct

`toBe(callsBefore + 1)` enforces the single-RPC acceptance criterion. Confirmed in `reactive-node-service.test.ts` line 1720.

#### ⚠️ Fix 2b: In-transaction OCC — Partially correct, with a latent semantic bug

**The transaction rollback wiring is correct.** I verified in the libsql 0.6 source (`libsql-0.6.0/src/transaction.rs`) that `impl Drop for Transaction` fires a close callback that issues `ROLLBACK`. An early `return Err(...)` from inside the loop drops `tx` without `commit()`, which correctly rolls back all prior DELETEs and INSERTs in the batch. This part of the fix works.

**The `SELECT changes()` check has a latent semantic correctness gap.** The version-gated DELETE is:

```sql
DELETE FROM relationship
WHERE out_node = ?1 AND relationship_type = 'has_child'
AND EXISTS (SELECT 1 FROM node WHERE id = ?1 AND version = ?2)
```

If `changes()` returns 0, the code reports `VERSION_CONFLICT`. However, 0 rows deleted has two possible causes:
1. The node's version did not match `expected_version` (genuine conflict — correct to abort).
2. The node has **no existing `has_child` edge** (it is already a root node) — even if its version is correct, the DELETE matches 0 relationship rows, triggering a false VERSION_CONFLICT.

In the current call site this cannot happen because `createNode` only calls this RPC on nodes that are confirmed siblings (existing children of `afterNodeId`), so they always have a `has_child` edge. But this method is now a public daemon surface — any gRPC caller, integration test, or future client that passes a root node in the batch will get a spurious `VersionConflict` with no actionable message.

**Suggested fix:** separate the existence check from the version check. The cleanest solution is to check `changes()` only after also verifying the version matched:

```sql
-- Option A: split into two queries
-- First confirm the relationship exists and node version matches:
SELECT COUNT(*) FROM relationship r
JOIN node n ON n.id = r.out_node
WHERE r.out_node = ?1 AND r.relationship_type = 'has_child'
  AND n.version = ?2

-- Option B: restructure the DELETE predicate to distinguish the two failure modes
DELETE FROM relationship
WHERE out_node = ?1 AND relationship_type = 'has_child'
-- (then separately query node version to disambiguate 0-change result)
```

Or more practically: verify the node's version inside the transaction before the DELETE (a `SELECT version FROM node WHERE id = ?1` read inside the transaction) and return a more precise error. This also ensures that the 0-change case from a version mismatch on a root node is correctly separated from the orphan-node case.

**Current severity:** low for the existing call site (structurally impossible to hit from `createNode`), but architectural: the RPC's error contract is ambiguous on a valid input class. Given that the issue spec calls this a "public daemon surface" and ADR-031 explicitly forbids trusting clients to uphold invariants, this should be addressed before the method sees broader use.

---

### New Issues in the Delta

#### [Improvement] In-transaction OCC is untested at the store layer

The `test_move_children_to_parent_all_or_nothing_on_stale_version` test exercises the **service-level** pre-validation (version mismatch caught at line 1792 before the store is called), not the in-transaction version-gate. In that test, child2's stale version (0 vs actual 1) is caught in the service loop at line 1792 — `move_children_to_parent` on the store is never reached with a matching service version but a diverged DB version.

To actually exercise the new `SELECT changes()` path, a test needs to pass versions that match the service-level check but then diverge before the transaction commits — e.g., by calling the store's `move_children_to_parent` directly with a `(&child_id, stale_version)` pair. Adding a store-level test for this would close the coverage gap on the TOCTOU fix itself.

#### [Improvement] `version_conflict("batch-child", 0, 0)` loses observability

At `node_service/mod.rs` line 1832, the in-transaction VERSION_CONFLICT error is re-mapped to `NodeServiceError::version_conflict("batch-child", 0, 0)`. The `VersionConflict` variant carries `node_id`, `expected_version`, and `actual_version`. Passing literal `0, 0` for both versions discards the node identity embedded in the `VERSION_CONFLICT:` error string from the store layer, making the gRPC `Status::Aborted` body uninformative for debugging.

The store error message is `"VERSION_CONFLICT: node '{child_id}' version mismatch (expected {expected_version})"` — that string already contains the relevant node ID and version. Either parse it out or restructure the store to return a typed error so the service can re-raise a properly attributed `VersionConflict { node_id, expected_version, actual_version: -1 }`.

#### Nit: Prior nit on doc comment still applies

The `sqlite_store.rs` doc comment on `move_children_to_parent` now correctly describes the version validation, but the sentence "The caller is responsible for OCC validation (version check + version bump) before and after this call" was removed — good. The updated doc comment is accurate.

---

### Skipped Items Status

- **HttpAdapter dev-proxy**: skipped as agreed — consistent with project pattern, no regression.
- **ADR amendment**: deferred to `doc/` branch in nodespace-docs. Acceptable as a follow-up given the documented cross-repo process.

---

| Criterion | Status |
|---|---|
| New `MoveChildrenToParent` RPC + core op + `SqliteStore` transaction | ✅ |
| `createNode` mid-list-split issues exactly one RPC | ✅ (assertion now `=== callsBefore + 1`) |
| Optimistic `structureTree` update preserved | ✅ |
| Order/version reconciled via response + `RelationshipUpdated` | ✅ |
| On failure: one rollback + `'child-transfer-failure'` notification | ✅ |
| No `skipConflictDetection` per-child loop remaining | ✅ |
| Atomicity test | ✅ (covers service-level OCC; store-level in-txn path uncovered) |
| OCC mapping (daemon): stale version → Aborted | ✅ |
| Order preservation test | ✅ |
| ADR amendment | ⏭️ deferred |

---

### Verdict

**Requesting changes** — the `changes()` semantic bug (ambiguity between version-mismatch and no-parent-edge) is a latent correctness issue on the public RPC surface. The fix is small: either add a pre-DELETE version read inside the transaction, or split the SQL to distinguish the two 0-result causes. The observability gap on `version_conflict("batch-child", 0, 0)` is a secondary improvement. Everything else looks correct and the change remains a net positive on the primary acceptance criteria.

malibio:
## Round 3 review findings addressed

**✅ Fixed (3 items):**

**1. Root-node false positive** — Added `is_root_node()` guard in the pre-validation loop (`node_service/mod.rs`). Root nodes have no `has_child` edge, so the version-gated DELETE would always return 0 rows and be misidentified as a VersionConflict. They now get an explicit `HierarchyViolation` error with a clear message.

**2. Actionable conflict node ID** — The `VERSION_CONFLICT` re-mapping in `map_err` now parses the node ID from the store's error string (`"VERSION_CONFLICT: node '<id>' ..."`) via `strip_prefix` + `split`. The resulting `NodeServiceError::VersionConflict` carries the actual offending node ID instead of the placeholder `"batch-child"`.

**3. Store-layer atomicity tests** — Two new `#[tokio::test]` tests in `sqlite_store.rs` that bypass service pre-validation:
  - `test_move_children_to_parent_store_rejects_stale_version`: calls store directly with stale version, asserts `VERSION_CONFLICT` error + child still under original parent (rollback verified)
  - `test_move_children_to_parent_store_success_and_order`: asserts strictly-increasing orders and all children under new parent on success

Total: 3867 frontend + 17 browser + 883 Rust tests pass (883 = +2 new store tests).

PR #1279, opened 2026-05-30
Install the NodeSpace skill into AI agents on first launch

malibio:
Closes #1199

### DIFF
--- a/packages/desktop-app/src-tauri/src/commands/onboarding.rs
+++ b/packages/desktop-app/src-tauri/src/commands/onboarding.rs
@@ -1,8 +1,10 @@
-//! Onboarding wizard Tauri commands (Issue #1180).
+//! Onboarding wizard Tauri commands (Issues #1180, #1199).
 //!
-//! Handles first-launch setup: PATH configuration and Claude Code skill
+//! Handles first-launch setup: PATH configuration and NodeSpace skill
 //! installation. Completion state is persisted to `~/.nodespace/config.json`.
+//! Skill installation state is tracked separately in `~/.nodespace/setup.json`.
 
+use crate::skill_setup::{self, SkillSetupResult};
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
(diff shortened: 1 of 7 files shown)


malibio:
### Code Review: PR #1279 — Install NodeSpace Skill on First Launch (Issue #1199)

**Overall assessment:** This is a well-structured, net-positive change that cleanly replaces the old hard-coded `SKILL.md` writer with a proper delegation to `@nodespaceai/skill`. The architecture is correct, security posture is acceptable for this stage, and the Svelte component follows project standards. Two issues need addressing before merge: a stdout parser bug that garbles the `agents_installed` list, and an unhandled Tauri event that leaves the CLI-missing warning silent.

---

### Findings

#### [Improvement] `skill_setup.rs` stdout parser collects false positives

**File:** `packages/desktop-app/src-tauri/src/skill_setup.rs`, lines 164–177

The `run_skill_installer` parser correctly strips `✓`/`⚠` prefixes and splits on `:` to extract agent names, but it does not filter out the indented file-path lines that the installer emits after each agent entry:

```
✓ claude-code: installed 2 file(s)
  → /home/user/.claude/skills/nodespace/SKILL.md
  → /home/user/.claude/skills/nodespace/nodespace-hook.ts
```

`→` is not in `['✓', '⚠', ' ', '\t']`, so `trim_start_matches` leaves it intact. These lines contain no `:`, so `split(':').next()` returns the entire string `"→ /home/user/..."` — which passes both guards and ends up in `agents_installed`. The UI then renders: `"Installed into: claude-code, → /home/.../SKILL.md, → /home/.../nodespace-hook.ts."`.

Additionally, when no agents are detected the installer exits 0 with:
```
No supported agents detected.
Checked: claude-code, codex, gemini, opencode
To install manually, specify an agent: npx @nodespaceai/skill install <agent>
```

The first line is filtered (starts with `"No supported"`), but `"Checked"` and `"To install manually, specify an agent"` are not, producing a nonsensical list.

**Fix:** Add a guard for lines starting with `→` (or `  →`), and add a prefix allowlist (e.g., only collect lines that started with `✓` before stripping). A minimal fix:

```rust
let agents = stdout
    .lines()
    .filter(|line| line.trim_start().starts_with('✓'))   // only success lines
    .filter_map(|line| {
        let line = line.trim_start_matches(['✓', ' ', '\t']);
        let agent = line.split(':').next()?.trim();
        if !agent.is_empty() { Some(agent.to_string()) } else { None }
    })
    .collect();
```

---

#### [Improvement] `skill:cli-missing` Tauri event is emitted but never consumed

**File:** `packages/desktop-app/src-tauri/src/lib.rs`, lines 247–253

The backend emits a `skill:cli-missing` event when the CLI is absent at first launch, but there is no frontend listener for this event anywhere in the codebase. The acceptance criterion states *"Success/warning notification shown after setup"* — the warning path is currently silent.

The `IntegrationsSettings` panel does display the CLI warning reactively when it is open, but a user who never navigates to Settings → Integrations will never see it. The intended behavior (warning shown after setup) requires either:
- A global `listen('skill:cli-missing', ...)` call in a top-level component (e.g., `App.svelte` or the main layout), or
- The Tauri event dropped in favor of the settings panel being the sole surface (and the AC updated accordingly).

Given the event is already wired, the simplest path is a global listener that emits a toast notification.

---

#### [Improvement] `check_cli_on_path()` called directly in the async executor

**File:** `packages/desktop-app/src-tauri/src/skill_setup.rs`, lines 85 and 101

`check_cli_on_path` uses `std::process::Command::status()`, a blocking syscall, and is called directly in the async `install_skill` function without `spawn_blocking`. `nodespace --version` completes quickly in practice, but the Tokio docs recommend all blocking operations go through `spawn_blocking` to avoid stalling the executor thread. The `run_skill_installer` function (which calls `npx`) is correctly wrapped; `check_cli_on_path` should be treated consistently.

---

#### [Improvement] No timeout on the `npx` process

**File:** `packages/desktop-app/src-tauri/src/skill_setup.rs`, line 141–145

`Command::output()` blocks until the child exits. If the npm registry is unreachable or slow, `npx @nodespaceai/skill install` can hang for 30+ seconds. Because the skill install runs **before** `GrpcClient::connect()` in the same sequential async task (`lib.rs` lines 238 → 264), a hung `npx` delays the gRPC connection and leaves the app appearing unresponsive on first launch.

After the first successful run the idempotency guard makes this a near-instant early return, so the window for this problem is narrow. Still, adding a timeout (e.g., via `tokio::time::timeout` wrapping the `spawn_blocking` call) would bound worst-case first-launch latency. Alternatively, the skill install task could be decoupled from the gRPC connect task.

---

#### Nitpicks

**Nit:** `packages/desktop-app/src-tauri/src/commands/onboarding.rs`, lines 163–178 — `get_skill_setup_status` duplicates the `cli_warning` string literal that already exists in `skill_setup::cli_warning()`. The private helper could be made `pub(crate)` so `onboarding.rs` reuses it instead of embedding the same string a second time.

**Nit:** `packages/desktop-app/src-tauri/src/lib.rs`, lines 331/337/343 — existing `println!` calls in the menu event handler are pre-existing and out of scope for this PR, but they are not using `tracing::debug!` per project standards. Worth a follow-up.

**Nit:** `packages/desktop-app/src/lib/components/settings/sections/integrations-settings.svelte`, lines 38–44 — the catch block in `reinstallSkill` hard-codes `cliOnPath: false` in the synthesized error result. This is harmless since the error banner takes priority in the UI, but it could confuse callers who inspect `lastResult.cliOnPath`.

---

| Criterion | Status |
|---|---|
| Skill installed on first launch | PASS — `install_skill(false)` wired in setup task |
| CLI PATH check with warning | PARTIAL — check runs, warning is logged and event emitted, but event is not consumed |
| Idempotent setup | PASS — `setup.json` guard + early return |
| Completion persisted to `~/.nodespace/setup.json` with `skill_installed: true` | PASS |
| Success/warning notification shown | PARTIAL — success path: UI only shows result in Settings panel; warning path: event emitted but no listener |
| Manual re-trigger via Settings → Integrations | PASS |
| `bun run test:all` passes | PASS (133 test files, 3858 tests, all green) |
| No `mcpServers` entry written | PASS — confirmed by audit of `installer.ts` and `agents.ts`; only `SKILL.md` + shims are written |

---

### Summary

The critical path (install on first launch, idempotency, no MCP) is correct. The two items to address before merge are the stdout parser false positives (garbled UI) and the unhandled `skill:cli-missing` event (AC gap). The timeout and `spawn_blocking` consistency issues are lower-priority but worth filing as follow-up. Overall a clean, well-scoped implementation.

malibio (commented):
Two issues need addressing before merge: (1) the stdout parser in skill_setup.rs collects file-path sub-lines and auxiliary text as agent names, garbling the UI's 'Installed into:' message; (2) the skill:cli-missing Tauri event is emitted but no frontend listener exists, leaving the CLI-missing warning silent. Both are detailed in the review comment. All tests pass and the core install path is correct.

malibio:
## Review Addressed

Addressed 3 of the review findings; 2 deferred as low-priority follow-ups.

### ✅ Addressed

1. **stdout parser false positives** — Added a pre-filter `filter(|line| line.trim_start().starts_with('✓'))` so only genuine success lines contribute to `agents_installed`. File-path sub-lines (`→ /path/...`) and diagnostic text (`Checked:`, `To install manually:`) are now excluded.

2. **`skill:cli-missing` event not consumed** — Added a `listen('skill:cli-missing', ...)` handler in `app-shell.svelte` that surfaces the warning via the status bar (`statusBar.error()`). The handler is properly cleaned up on component destroy.

3. **Duplicated `cli_warning` string** — Made `skill_setup::cli_warning` `pub(crate)` and replaced the copy in `onboarding.rs::get_skill_setup_status` with a call to the shared helper.

### ⏭️ Deferred

- **`spawn_blocking` for `check_cli_on_path`** — `nodespace --version` completes in <10ms in practice. The overhead of `spawn_blocking` (task allocation, context switch) is not justified for a single near-instant call. Will revisit if profiling shows executor stalls.

- **`npx` timeout** — The idempotency guard (`setup.json`) means this blocks first launch at most once per installation. Filed as a follow-up for when the skill installer is extracted to a more prominent path.

All 133 test files (3858 tests) still pass.

PR #1280, opened 2026-05-30
Ratify & harden node deletion strategy: atomic subtree CASCADE over has_child edges (+ confirmation + ADR)

### LINKED ISSUE #220: Ratify & harden node deletion strategy: atomic subtree CASCADE over has_child edges (+ confirmation + ADR)
# Backend: Ratify and harden the node deletion strategy (atomic subtree CASCADE over has_child edges)

## Problem

Node deletion performs a **CASCADE** — deleting a node deletes its entire subtree — but this has never been an explicit, documented decision, and the current implementation has correctness gaps under the edge-based hierarchy and OCC write model.

> **Correction to the original draft.** The earlier draft framed the fix around a self-referential `parent_id` column and a SQL `FOREIGN KEY (parent_id) REFERENCES nodes(id) ON DELETE CASCADE`, plus an "ORPHAN/PROMOTE to root via `parentId = null`" alternative. **None of that matches the current architecture.** Per **ADR-024 (Parent-Child as Operation Parameters)** there is NO `parent_id` column on the `node` table — hierarchy is `has_child` rows in the `relationship` table. The draft's SQL-FK framing is obsolete; this rewrite replaces it.

### What actually happens today (verified)
- The `node` table (`packages/core/src/db/schema.sql`) has no parent column. Parent→child is a `relationship` row with `relationship_type = 'has_child'`.
- The `relationship` and `embedding` tables DO have `ON DELETE CASCADE` FKs on `node(id)` — so deleting **one** `node` row auto-cleans **that node's** edges/embeddings, per row, not for a subtree.
- **Subtree cascade is application-level recursion:** `NodeService::delete_node` (`crud.rs:~1242`) does `get_children(node_id)` (a `has_child` query) then recursively `delete_node(child)` per child, then deletes the node via `delete_with_version_check` (OCC). Each level emits its own `NodeDeleted` event.
- `delete_node_cascade_atomic` (`sqlite_store.rs:~769`) is **misnamed** — it just calls single-node `delete_node`; it is NOT atomic over a subtree.

So this is NOT "undefined behavior" — it's **CASCADE-by-app-recursion, never ratified, and not transactionally atomic.**

### The real defects
1. **Non-atomic subtree delete (TOCTOU / partial-failure).** N separate writes across N recursion levels, not one transaction. A mid-cascade failure or version conflict on any descendant aborts **after** some descendants are gone — a partially-deleted subtree. Same class as the B8 bug (#1262); fix with the atomic-transaction pattern (C3c #1275, `update_nodes_bulk_atomic`).
2. **OCC is per-descendant, which is wrong for a cascade.** Every recursive `delete_node` re-checks each child's `version`; a concurrent edit to any descendant raises `VersionConflict` and aborts mid-subtree. Per ADR-026, OCC should guard the **user-initiated target** at its expected version, not impose an independent check on every descendant the user never read.
3. **No confirmation, no descendant count (verified).** No frontend delete path (`reactive-node-service.svelte.ts:645/1278` → `shared-node-store.deleteNode:1703` → `backendAdapter.deleteNode`) shows a "delete N descendants" prompt. The store drops only the target; descendants are cleaned only as their individual `NodeDeleted` events arrive. Permanent, silent, multi-node data loss on one click.

## Decision

**Default = atomic CASCADE.** Deleting a node deletes its entire `has_child` subtree in a **single transaction**, gated behind a frontend confirmation showing the descendant count.

**Alternative = PROMOTE (re-parent), DEFERRED to a follow-up.** "Delete node, keep children" re-parents the target's direct children to the **target's parent (grandparent)** by rewriting `has_child` edges — NOT "promote to root / `parentId = null`" as the old draft said. The Backspace-merge path (`reactive-node-service.svelte.ts:~1157`, `promoteChildren`) already implements this re-parenting and is the model to generalize. Out of scope here; track separately.

### Rationale
- CASCADE is the least-surprising default for a tree and is already the de-facto behavior — we're ratifying + hardening, not changing semantics.
- A confirmation dialog (with count) mitigates permanent loss; soft-delete/trash is a future enhancement (`node.lifecycle_status` already exists for it — do NOT build it here).
- **ADR-024:** stay edge-native — cascade walks `has_child`, never a column.
- **ADR-026 (+ C5 amendment):** deletion is a versioned write; the daemon is the OCC adjudicator returning structured `VERSION_CONFLICT`. OCC guards the target; the subtree delete is all-or-nothing in one transaction so a conflict can't leave a half-deleted tree.
- **ADR-031:** deletion logic lives in daemon/core; the frontend stays thin (confirmation UX + event-driven reconciliation).

## Backend work (`packages/core`)
1. **Make subtree cascade atomic.** Replace the per-child recursive loop in `delete_node` with a single-transaction subtree delete: collect the full descendant set by walking `has_child` from the target **before** mutating; in one `store.transaction()` (mirror `update_nodes_bulk_atomic`, `sqlite_store.rs:~1611`), OCC-check the **target** at `expected_version`, then `DELETE FROM node WHERE id IN (target + descendants)` — relationship/embedding FK CASCADE cleans every edge/embedding for the set. Commit once. Target version mismatch → roll back, return `VersionConflict` (ADR-026 C5 pipeline). Repurpose/rename `delete_node_cascade_atomic` so the name is truthful; delete the dead recursive path (no compat shims — greenfield).
2. **Event emission:** emit one `NodeDeleted` event **per node** in the deleted set (target + every descendant) after commit, so every client's store + `structureTree` reconcile each removed node. (Do NOT collapse to a single event — the frontend keys cleanup off per-node deletions; a single target-only event would orphan descendants in the UI.)
3. **OCC semantics:** version-check only the user-initiated target; descendants are removed unconditionally inside the same transaction.
4. **RPC/daemon:** keep the `(node_id, expected_version)` delete signature; return the count of deleted nodes. `VERSION_CONFLICT` via the established `Code::Aborted` + `x-version-conflict` path.

## Frontend work (`packages/desktop-app`)
1. **Confirmation dialog:** before deleting a node with children, fetch the descendant count (`backendAdapter.getDescendants` / store child cache) and show **"Delete node and N descendants? This cannot be undone."** Default: prompt only when `N > 0`. Wire at the `reactive-node-service` delete entry points (645 viewer-driven, 1278 direct) — NOT inside `shared-node-store.deleteNode` (keep it a pure mutation).
2. **Event-driven cleanup:** rely on the per-descendant `NodeDeleted` events through `tauri-sync-listener.ts` / `browser-sync-service.ts` → `sharedNodeStore.deleteNode(id, …, skipPersistence=true)` to evict every descendant and prune `structureTree`. Verify no descendant dangles after a cascade.
3. `createLogger`, no `console.*`; no `any`.

## ADR deliverable
Add a **new ADR — "Node Deletion Strategy"** to `nodespace-docs` on a `doc/` branch (new ADR, not an amendment — deletion semantics aren't owned by 024/026/031, though it cites all three). Record: edge-native subtree cascade (ADR-024), atomic single-transaction delete with target-only OCC + conflict pipeline (ADR-026+C5), daemon ownership + thin-client UX (ADR-031), the confirmation mitigation, and PROMOTE + soft-delete as explicitly deferred. **Note:** branch the new ADR from `main` in nodespace-docs (`git checkout main` there first).

## Acceptance criteria
- [ ] `delete_node` deletes target + entire `has_child` subtree in a single transaction; relationship/embedding rows for the whole set are gone.
- [ ] Mid-operation failure or target version conflict leaves the subtree **fully intact** (no partial deletion) — verified by test.
- [ ] OCC guards only the target; concurrent descendant edits do not abort the cascade.
- [ ] One `NodeDeleted` event per deleted node; frontend store + `structureTree` fully reconciled (no orphaned descendants in UI).
- [ ] Frontend shows "Delete node and N descendants?" when `N > 0`; cancel aborts with no mutation.
- [ ] `delete_node_cascade_atomic` is genuinely atomic or removed; no misleading/dead delete paths remain.
- [ ] New "Node Deletion Strategy" ADR merged in `nodespace-docs`.

## Testing
- **Cascade correctness:** delete a multi-level subtree; assert all node/`has_child`/`embedding` rows for the set removed, nothing outside it.
- **Atomicity:** inject a mid-delete failure or simulate target conflict; assert ZERO rows deleted (no partial subtree). `bun run test:db` for full SQLite integration + a Rust transaction test in core.
- **OCC:** stale target version → `VERSION_CONFLICT` + intact subtree; concurrent descendant edit → cascade still succeeds.
- **Confirmation UX:** dialog shows correct count; cancel = no-op; confirm = delete. Happy-DOM for logic; `bun run test:browser` only if real click needed.
- **Event-driven cleanup:** after cascade, `structureTree` + store contain none of the deleted ids.

## Risks
- **Permanent data loss** — the point of the confirmation + count. Soft-delete deferred (`lifecycle_status` exists for it).
- **Hierarchy integrity (#557 class)** — a botched cascade could strand edges/descendants; single-transaction + FK-CASCADE design + atomicity tests guard it.
- **Event volume** on large subtrees (one event/node) — acceptable; matches how the frontend reconciles. Batch only if profiling demands (out of scope).

`git status` + `git pull origin main` on primary, `EnterWorktree({name: "issue-220-node-deletion-strategy"})`, `bun install`, `bun run test` baseline, `gh:comment`/`gh:assign`/`gh:status`. ADR goes in nodespace-docs on a `doc/` branch (branch from `main`). Before PR: `bun run test:all` + `bun run test:db` (atomicity/cascade integration), `bun run quality:fix` + commit, `bun run gh:pr 220`, `/pragmatic-code-review`, then `ExitWorktree` + `gh pr merge --squash --delete-branch`.

malibio:
Closes #220

### DIFF
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -743,7 +743,12 @@ impl SqliteStore {
     pub async fn delete_node(&self, id: &str, source: Option<String>) -> Result<DeleteResult> {
         let node = match self.get_node(id).await? {
             Some(n) => n,
-            None => return Ok(DeleteResult { existed: false }),
+            None => {
+                return Ok(DeleteResult {
+                    existed: false,
+                    deleted_count: 0,
+                })
+            }
         };
 
         // FK CASCADE handles relationship and embedding deletion
@@ -763,16 +768,142 @@ impl SqliteStore {
             playbook_context: None,
         });
 
-        Ok(DeleteResult { existed: true })
+        Ok(DeleteResult {
+            existed: true,
+            deleted_count: 1,
+        })
     }
 
-    pub async fn delete_node_cascade_atomic(
+    /// Atomically delete a node and its entire `has_child` subtree in a single transaction.
+    ///
+    /// **OCC contract:** Version-checks only the target node at `expected_version`. Descendants
+    /// are removed unconditionally inside the same transaction. A target version mismatch rolls
+    /// back the entire transaction — no partial deletion is possible.
+    ///
+    /// **Event emission:** Emits one `NodeDeleted` notification per deleted node (target +
+    /// every descendant) after the commit, so the frontend store reconciles each removal.
+    ///
+    /// Returns `(existed, deleted_nodes)` where `deleted_nodes` contains the target and all
+    /// descendants that were deleted (empty vec when target didn't exist).
+    pub async fn delete_subtree_atomic(
         &self,
         node_id: &str,
+        expected_version: i64,
         source: Option<String>,
-    ) -> Result<DeleteResult> {
-        // FK CASCADE on schema.sql handles relationships and embeddings automatically
-        self.delete_node(node_id, source).await
+    ) -> Result<(bool, Vec<Node>)> {
+        // Collect the target + all descendants before mutating.
+        let target = match self.get_node(node_id).await? {
+            Some(n) => n,
+            None => return Ok((false, vec![])),
+        };
+
+        // OCC check on target before entering the transaction.
+        if target.version != expected_version {
+            return Err(anyhow::anyhow!(
+                "version_conflict:{}:{}:{}",
+                node_id,
+                expected_version,
+                target.version
+            ));
+        }
+
+        // Walk has_child recursively to collect the full descendant set.
+        let mut id_rows = self.db.query(
+            r#"WITH RECURSIVE subtree(node_id) AS (
+                SELECT out_node FROM relationship WHERE in_node = ?1 AND relationship_type = 'has_child'
+                UNION ALL
+                SELECT r.out_node FROM relationship r
+                JOIN subtree s ON r.in_node = s.node_id
+                WHERE r.relationship_type = 'has_child'
+            )
+            SELECT DISTINCT node_id FROM subtree"#,
+            libsql::params![node_id.to_string()],
+        ).await.context("Failed to collect subtree descendants")?;
+
+        let mut all_ids: Vec<String> = vec![node_id.to_string()];
+        while let Some(row) = id_rows.next().await? {
+            all_ids.push(row.get(0)?);
+        }
+
+        // Fetch all node records (needed for post-commit notifications).
+        let placeholders: Vec<String> = (1..=all_ids.len()).map(|i| format!("?{}", i)).collect();
+        let fetch_sql = format!(
+            "SELECT * FROM node WHERE id IN ({})",
+            placeholders.join(", ")
+        );
+        let fetch_params: Vec<libsql::Value> = all_ids
+            .iter()
+            .map(|id| libsql::Value::Text(id.clone()))
+            .collect();
+        let mut node_rows = self
+            .db
+            .query(&fetch_sql, fetch_params)
+            .await
+            .context("Failed to fetch subtree nodes for deletion")?;
+        let mut nodes_to_delete: Vec<Node> = Vec::new();
+        while let Some(row) = node_rows.next().await? {
+            nodes_to_delete.push(Self::row_to_node(&row)?);
+        }
+
+        // Single-transaction delete — FK CASCADE cleans relationship and embedding rows.
+        let tx = self
+            .db
+            .transaction()
+            .await
+            .context("Failed to begin subtree delete transaction")?;
+
+        // Re-check target version inside the transaction to close the TOCTOU window.
+        let check_sql = "SELECT version FROM node WHERE id = ?1";
+        let mut check_rows = tx
+            .query(check_sql, libsql::params![node_id.to_string()])
+            .await
+            .context("Failed to re-check target version")?;
+        let actual_version: i64 = match check_rows.next().await? {
+            Some(row) => row.get(0)?,
+            None => {
+                // Node disappeared between pre-check and transaction — idempotent.
+                return Ok((false, vec![]));
+            }
+        };
+        if actual_version != expected_version {
+            return Err(anyhow::anyhow!(
+                "version_conflict:{}:{}:{}",
+                node_id,
+                expected_version,
+                actual_version
+            ));
+        }
+
+        let del_placeholders: Vec<String> =
+            (1..=all_ids.len()).map(|i| format!("?{}", i)).collect();
+        let del_sql = format!(
+            "DELETE FROM node WHERE id IN ({})",
+            del_placeholders.join(", ")
+        );
+        let del_params: Vec<libsql::Value> = all_ids
+            .iter()
+            .map(|id| libsql::Value::Text(id.clone()))
+            .collect();
+        tx.execute(&del_sql, del_params)
+            .await
+            .context("Failed to delete subtree nodes")?;
+
+        tx.commit()
+            .await
+            .context("Failed to commit subtree delete transaction")?;
+
+        // Emit one NodeDeleted notification per deleted node after commit.
+        for node in &nodes_to_delete {
+            self.notify(StoreChange {
+                operation: StoreOperation::Deleted,
+                node: node.clone(),
+                source: source.clone(),
+                previous_node: None,
+                playbook_context: None,
+            });
+        }
+
+        Ok((true, nodes_to_delete))
     }
 
     pub async fn delete_with_version_check(
--- a/packages/core/src/models/node.rs
+++ b/packages/core/src/models/node.rs
@@ -516,12 +516,12 @@ impl NodeUpdate {
 /// ```rust
 /// use nodespace_core::models::DeleteResult;
 ///
-/// // Node existed and was deleted
-/// let result = DeleteResult { existed: true };
+/// // Node existed and was deleted (leaf node)
+/// let result = DeleteResult { existed: true, deleted_count: 1 };
 /// assert!(result.existed);
 ///
 /// // Node didn't exist (idempotent success)
-/// let result = DeleteResult { existed: false };
+/// let result = DeleteResult { existed: false, deleted_count: 0 };
 /// assert!(!result.existed);
 /// ```
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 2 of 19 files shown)


malibio:
### Code Review Summary

PR #1280 is a well-scoped, architecturally sound change that delivers on the core promise: replace non-atomic recursive deletion with a single-transaction subtree CASCADE, gate it behind a frontend confirmation dialog, and document the decision in an ADR. The Rust backend work is particularly solid — the dual OCC check (pre-check + re-check inside the transaction) is the correct pattern, the test suite covers all four acceptance-criteria scenarios, and the proto field addition is backward-compatible. This is a net improvement and approaches merge-ready, with two issues to address before landing.

---

### Findings

#### [Improvement] Modal keyboard shortcuts broken: inner `onkeydown` stops `Escape`/`Enter` from reaching the overlay handler

`packages/desktop-app/src/lib/components/delete-confirmation-modal.svelte` lines 17–27

The `handleKeydown` listener that maps `Escape → cancel` and `Enter → confirm` is attached to the **overlay backdrop** (which has `tabindex="-1"`). The inner modal `<div>` has `onkeydown={(e) => e.stopPropagation()}`, so any keydown event fired while the modal itself has focus never bubbles to the overlay. Since neither element receives `autofocus`, keyboard shortcuts are silently inert in the common case. Fix: move `handleKeydown` to the inner modal `<div>` (replacing the stopPropagation-only handler there), and keep stopPropagation on the outer overlay so underlying editor shortcuts don't fire through the backdrop.

#### [Improvement] Test gap: the confirmation dialog path (descendants > 0) and the cancel path are untested

`packages/desktop-app/src/tests/services/reactive-node-service.test.ts`

The mock for `backendAdapter.getDescendants` always returns `[]`, so `confirmNodeDeletion` is always called with `0` and returns `true` immediately — the dialog is never shown in any test. The cancel branch (`confirmed === false`) is also never reached. The acceptance criterion explicitly calls for a "Cancel = no-op" test. Two tests are needed:

1. Mock `getDescendants` to return `[fakeNode]`; mock `confirmNodeDeletion` (or the `delete-confirmation.svelte` module) to resolve `false`; assert `events.nodeDeleted` is **not** called.
2. Same setup, resolve `true`; assert deletion proceeds normally.

#### [Improvement] Module-level `_pending` singleton drops a concurrent Promise

`packages/desktop-app/src/lib/services/delete-confirmation.svelte.ts` lines 24–26

If two `deleteNode` calls with `descendantCount > 0` are in-flight simultaneously (e.g., rapid keyboard triggers before the first dialog dismisses), the second call overwrites `_pending`, permanently orphaning the first Promise's `resolve` callback. The first deletion hangs forever (a microtask-level memory leak). For a single-user desktop app this is an edge case, but the fix is cheap: guard with an early return if `_pending !== null`:

```ts
export async function confirmNodeDeletion(descendantCount: number): Promise<boolean> {
  if (descendantCount === 0) return true;
  if (_pending !== null) return false; // second concurrent call: abort silently
  ...
}
```

#### [Improvement] `getDescendants` in `TauriAdapter` is O(depth) serial IPC calls

`packages/desktop-app/src/lib/services/backend-adapter.ts` lines 197–211

The count fetch before showing the dialog uses a BFS that fires one `get_children` IPC round-trip per level of the tree. The backend already walks the full subtree with `WITH RECURSIVE` inside `delete_subtree_atomic`. A dedicated `get_descendant_count` or `get_descendants` Tauri command backed by the same CTE would collapse this to one IPC call. This is a latency concern for deep trees (e.g., large imported outlines), not a correctness issue. Acceptable to track as a follow-up issue.

---

### Nitpicks

- **Nit** `delete-confirmation-modal.svelte` line 18: `tabindex="-1"` on the overlay means it won't receive focus via Tab, so keyboard dismiss only works if the user manually clicks the backdrop first. Adding `autofocus` to the Cancel button (or the modal div) would make keyboard flow work naturally after the dialog appears.

- **Nit** `packages/core/src/services/node_service/crud.rs` line 1228: `splitn(3, ':')` on `rest` (which is `<uuid>:<expected>:<actual>`) is correct because UUIDs contain only hex and hyphens. The comment says the format is `version_conflict:<id>:<expected>:<actual>` — this is accurate, but the sentinel ID parsed into `parts[0]` is immediately discarded in favour of the `node_id` function parameter. A brief inline note ("id field extracted but not used; we use the caller-supplied node_id for the error") would prevent future confusion.

- **Nit** `packages/core/src/db/sqlite_store.rs` ~line 811: The CTE subtree collection and the `SELECT * FROM node WHERE id IN (...)` fetch both execute **outside** the transaction, before `tx.begin`. A concurrent create of a child node in this window would leave that child orphaned (its `has_child` edge is removed when the parent is deleted, but the child row itself is not). In a single-user embedded SQLite app this is a theoretical-only concern. The in-transaction version re-check correctly closes the version-conflict TOCTOU. Acceptable as-is; worth a short comment acknowledging the window.

---

### Acceptance Criteria Check

| Criterion | Status |
|---|---|
| Single-transaction subtree delete | PASS — `delete_subtree_atomic` with one `tx.execute` |
| Mid-operation failure leaves subtree intact | PASS — `test_delete_node_version_conflict_leaves_subtree_intact` |
| OCC guards only target | PASS — `test_delete_node_occ_guards_only_target_not_descendants` |
| One `NodeDeleted` event per deleted node | PASS — post-commit `notify` loop in `sqlite_store.rs` |
| Frontend shows "Delete N descendants?" when N > 0 | PARTIAL — dialog wording is correct; keyboard dismiss is broken (see above) |
| `delete_node_cascade_atomic` dead stub removed | PASS — renamed and genuinely implemented |
| ADR merged in nodespace-docs | PASS — ADR-036 committed to nodespace-docs main |
| Cancel = no-op (untested) | MISSING — no test for cancel path |

---

**Decision: Request Changes** — the keyboard handler inversion in the modal is a behavioural bug (not just a nit), and the cancel-path test gap leaves an acceptance criterion untested. Both are small fixes. Once those two items land the PR is good to merge.

malibio (commented):
Review status: CHANGES REQUESTED. Two items to address before merge: (1) the modal keyboard handler is inverted — the inner modal div's onkeydown stops propagation before Escape/Enter can reach the overlay's handleKeydown; move handleKeydown to the inner modal div. (2) add a test for the cancel path (mock getDescendants to return a non-empty array, resolve confirmNodeDeletion as false, assert events.nodeDeleted is not called). Full findings in the detailed comment above.

PR #1282, opened 2026-05-30
Measure search_skills end-to-end latency on local 7B model

### LINKED ISSUE #1152: Measure search_skills end-to-end latency on local 7B model
## Context

PR #1149 (closes #1130) replaced pre-LLM skill routing with an LLM-orchestrated `search_skills` tool. The acceptance criteria included:

> Latency measured on a local 7B-class model and documented in PR (rough floor: how many extra ms per turn vs the current router).

The PR shipped without this measurement because the bench requires a warm `nodespace-nlp-engine` with a loaded model — outside the scope of the unit-test pass. Architectural reasoning alone (~1-3s extra per turn for the inference round that decides whether to call `search_skills`) is not a substitute for measurement.

## Goal

Measure the actual per-turn latency cost on a real 7B-class local model and document the result.

## Acceptance criteria

- [ ] Bench harness for `LocalAgentService::send_message` that loads a real model (Gemma 3B or Ministral 8B — whatever is the current default for local-agent dev)
- [ ] Three scenarios measured, each run N=10 times for variance:
  1. Conversational reply (model skips `search_skills`)
  2. Single skill turn (`search_skills` → matched tool → final text)
  3. Multi-skill turn (two `search_skills` calls, both with downstream tool invocations)
- [ ] Wall-clock and per-phase token latency reported (prompt eval, generation, tool dispatch)
- [ ] Results recorded in `nodespace-docs/development/benchmarks/` or a similar persistent location, not just a PR comment
- [ ] **Decision**: if latency for scenario 2 is > ~3s on a 7B model, propose a fast-path optimization (e.g., a heuristic skip for messages the schema list obviously can't satisfy — but only if measurement justifies it, per the original issue)

## Non-goals

- Optimizing inference engine internals
- Adding a fast-path before measurement justifies one
- Comparing against the deleted pre-LLM router (that path is gone; the measurement is of the new architecture's absolute cost)

## Related

- PR #1149 / Issue #1130 (original work)
- Acceptance criterion the original PR did not meet

malibio:
Closes #1152

### DIFF
--- a/packages/agent/Cargo.toml
+++ b/packages/agent/Cargo.toml
@@ -67,3 +67,7 @@ tempfile = "3"
 name = "ollama_integration"
 path = "tests/ollama_integration.rs"
 required-features = ["testing"]
+
+[[test]]
+name = "search_skills_latency"
+path = "tests/search_skills_latency.rs"
--- a/packages/agent/tests/search_skills_latency.rs
+++ b/packages/agent/tests/search_skills_latency.rs
@@ -9,7 +9,7 @@
 //! `nodespace-docs/development/benchmarks/search-skills-latency.md`.
 //!
 //! Run with:
-//!   cargo test -p nodespace-agent --test search_skills_latency --features testing -- --nocapture
+//!   cargo test -p nodespace-agent --test search_skills_latency -- --nocapture
 //!
 //! Gracefully skips if no inference backend is available (Ollama not running and
 //! no local GGUF model downloaded).
@@ -30,6 +30,27 @@ use nodespace_agent::ModelManager;
 use nodespace_nlp_engine::chat::ChatConfig;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
+
+/// Walk up from `CARGO_MANIFEST_DIR` searching for a `nodespace-docs` sibling directory.
+///
+/// Works correctly in both primary checkout and worktree layouts without
+/// hardcoding a level count (which differs between the two contexts).
+fn find_docs_benchmarks_dir() -> Option<std::path::PathBuf> {
+    let mut dir = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    for _ in 0..10 {
+        let candidate = dir.join("../nodespace-docs/development/benchmarks");
+        if let Ok(canonical) = candidate.canonicalize() {
+            return Some(canonical);
+        }
+        // Also check if the parent of nodespace-docs exists (dir may not exist yet)
+        let docs_parent = dir.join("../nodespace-docs/development");
+        if docs_parent.exists() {
+            return Some(dir.join("../nodespace-docs/development/benchmarks"));
+        }
+        dir = dir.join("..");
+    }
+    None
+}
 
 // ---------------------------------------------------------------------------
 // Backend resolution (shared pattern with ollama_integration.rs)
@@ -286,6 +307,9 @@ impl ScenarioStats {
 
     fn percentile_wall_ms(&self, p: f64) -> u64 {
         let mut vals = self.wall_ms_values();
+        if vals.is_empty() {
+            return 0;
+        }
         vals.sort_unstable();
         let idx = ((p / 100.0) * (vals.len() - 1) as f64).round() as usize;
         vals[idx.min(vals.len() - 1)]
@@ -364,11 +388,12 @@ async fn measure_turn(
 // Markdown report builder
 // ---------------------------------------------------------------------------
 
-fn build_report(model_name: &str, scenarios: &[ScenarioStats], n: usize) -> String {
+fn build_report(model_name: &str, scenarios: &[ScenarioStats], n: usize, date: &str) -> String {
     let mut md = String::new();
 
     md.push_str("# search_skills End-to-End Latency Benchmark\n\n");
     md.push_str(&format!("**Model:** `{model_name}`  \n"));
+    md.push_str(&format!("**Date:** {date}  \n"));
     md.push_str(&format!("**Runs per scenario:** {n}  \n"));
     md.push_str("**Metric:** wall-clock time per full agent turn (inference + tool dispatch + final response)\n\n");
     md.push_str("---\n\n");
@@ -460,7 +485,7 @@ fn build_report(model_name: &str, scenarios: &[ScenarioStats], n: usize) -> Stri
     md.push_str(
         "- Tool responses are stubs (no real database I/O) to isolate inference latency.\n",
     );
-    md.push_str("`search_skills` returns a single high-confidence Research & Search match; downstream tools return representative canned results.\n");
+    md.push_str("- `search_skills` returns a single high-confidence Research & Search match; downstream tools return representative canned results.\n");
     md.push_str("- Wall-clock is measured from the `send_message` call until the `AgentTurnResult` is returned, covering prompt assembly, all inference rounds, and tool dispatch.\n");
     md.push_str("- Token counts are cumulative across all inference rounds in the turn (prompt and completion separately).\n");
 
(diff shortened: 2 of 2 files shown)


malibio:
## Code Review — PR #1282: search_skills end-to-end latency benchmark

**Overall:** This is a net positive change. The harness is well-structured, gracefully skips when no backend is available, and the manually enriched report in \`nodespace-docs\` meets the spirit of the acceptance criteria. There is one correctness bug in the report path calculation, one misleading doc comment, a missing list marker in the generated markdown, and a moderate concern around methodology validity — none of which are blockers given the WARN fallback and the manually committed \`nodespace-docs\` report. Approving with the expectation that the path bug is fixed before the next benchmark run.

---

### Findings

#### [Improvement] Report path calculation is wrong for both execution contexts

**File:** \`packages/agent/tests/search_skills_latency.rs\`, lines 611–613

The comment claims to navigate \`packages/agent → packages → nodespace-core → nodespace\`, but the \`.join("../../../..")\` (4 levels up) resolves to different wrong directories depending on where the test is run:

- **Primary checkout** (\`CARGO_MANIFEST_DIR\` = \`.../nodespace-core/packages/agent\`): 4 levels up = \`/Users/<user>\` — then \`nodespace-docs/\` does not exist there.
- **Worktree** (\`CARGO_MANIFEST_DIR\` = \`.../nodespace-core/.claude/worktrees/issue-.../packages/agent\`): 4 levels up = \`.../nodespace-core/.claude/\` — producing the stray artifact at \`.claude/nodespace-docs/development/benchmarks/search-skills-latency.md\` that is sitting in the repo right now.

The primary checkout needs 3 levels (\`../../..\`); the worktree needs 6 levels (\`../../../../../..\`). These are irreconcilable with a single static path. The WARN fallback silently fires in both contexts and the auto-generated report is never written to the correct location.

**Concrete evidence:** Two copies of the report exist on disk — one at \`.claude/nodespace-docs/...\` (auto-generated, 3,720 bytes, 21:12) and the correct one at \`nodespace-docs/development/benchmarks/search-skills-latency.md\` (manually curated, 6,743 bytes, 21:25). The benchmark wrote to the wrong place; the developer hand-wrote the correct version separately.

**Recommended fix:** Walk up from \`CARGO_MANIFEST_DIR\` searching for a \`nodespace-docs\` sibling, rather than using a hardcoded level count:

\`\`\`rust
fn find_docs_benchmarks_dir() -> Option<std::path::PathBuf> {
    let mut dir = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
    for _ in 0..10 {
        let candidate = dir.join("../nodespace-docs/development/benchmarks");
        if candidate.exists() {
            return Some(candidate);
        }
        dir = dir.join("..");
    }
    None
}
\`\`\`

Alternatively, accept a \`NODESPACE_DOCS_DIR\` env var as an escape hatch.

#### [Improvement] Scenario 2 and 3 never actually exercise \`search_skills\` — results measure direct tool-call routing, not the named cost

**File:** \`packages/agent/tests/search_skills_latency.rs\`, lines 523–578

The issue asks to measure "how many extra ms per turn vs the current router." Across all 30 runs in scenarios 2 and 3, \`search_skills\` was called 0 times — the model went directly to downstream tools. The harness correctly detects and reports this, and the hand-written analysis in \`nodespace-docs\` handles it well. However, the scenario descriptions in the auto-generated report still say "Model calls search\_skills once" and "calls search\_skills twice," which is factually incorrect for any run where the model bypasses it.

This is partly a model behaviour issue (Gemma 4 8B is tool-savvy enough to skip the routing step), but the harness has no mechanism to verify or enforce that a scenario actually exercised the named path. For a benchmark intended to isolate \`search_skills\` overhead, a post-run assertion or at minimum a clear \`WARN\` when the expected call count differs from actual would make the harness self-auditing:

\`\`\`rust
if scenario2.runs.iter().all(|r| r.search_skills_calls == 0) {
    eprintln!("WARN: search_skills was never called in scenario 2 — \
               latency does not reflect skill-routing overhead");
}
\`\`\`

The hand-written report addresses this honestly. The code harness should surface it too.

#### [Improvement] Misleading \`--features testing\` in the doc comment

**File:** \`packages/agent/tests/search_skills_latency.rs\`, line 12

The run command in the module doc comment recommends:
\`\`\`
cargo test -p nodespace-agent --test search_skills_latency --features testing -- --nocapture
\`\`\`

Unlike its peer \`ollama_integration.rs\`, this binary does not use any \`testing\`-gated APIs (\`set_system_prompt\` is not called). The \`--features testing\` flag is unnecessary and creates confusion about whether the bench compiles without it. The Cargo.toml entry correctly omits \`required-features\`, so the flag should be dropped from the doc comment.

---

#### Nitpicks

- **Nit:** \`packages/agent/tests/search_skills_latency.rs\`, line 463: The \`search_skills\` sentence in the Methodology section is missing its \`- \` list marker, so it renders as a standalone paragraph rather than a continuation of the bullet list.

- **Nit:** The auto-generated report from \`build_report()\` does not include a date or hardware context — the manually written \`nodespace-docs\` report adds these by hand. A \`chrono::Local::now()\` timestamp in \`build_report()\` would make reruns self-documenting without manual intervention.

- **Nit:** \`percentile_wall_ms\` will panic on an empty \`runs\` vec (usize underflow on \`vals.len() - 1\`). All call sites have \`is_empty()\` guards so this is unreachable in practice, but adding a guard inside the function itself (\`if vals.is_empty() { return 0; }\`) removes the footgun.

- **Nit:** \`ollama_running()\` and \`first_ollama_model()\` are copy-pasted verbatim from \`ollama_integration.rs\`. Rust integration test binaries cannot share modules directly, but a \`tests/common/mod.rs\` helper (imported via \`mod common;\`) would consolidate these if a third test binary ever needs the same utilities.

---

### Acceptance Criteria Status (Issue #1152)

| Criterion | Status |
|-----------|--------|
| Bench harness for \`LocalAgentService::send_message\` with real model | Met — Ollama + GGUF fallback |
| Three scenarios, N=10 | Met |
| Wall-clock and per-phase token latency | Partial — wall-clock and cumulative prompt/completion tokens; per-round breakdown not exposed by \`InferenceUsage\` API |
| Results in \`nodespace-docs/development/benchmarks/\`| Met by manual commit; auto-write path is broken |
| Fast-path decision | Met — correctly recommends no fast-path; latency driver is downstream tool rounds, not skill routing |

malibio (commented):
Approving with the path-calculation bug and misleading doc comment flagged in the detailed review above as pre-merge fixes. The harness structure, skip logic, stats calculations, and the manually committed nodespace-docs report all meet the acceptance criteria. The finding that Gemma 4 8B bypasses search_skills entirely is architecturally significant and is handled honestly in the hand-written analysis.

malibio:
## Address-Review Summary

All reviewer recommendations addressed. One skipped (with rationale).

### ✅ Addressed (6)

1. **Report path fixed** — replaced `../../../..` hardcoded depth with `find_docs_benchmarks_dir()` that walks up the tree searching for the `nodespace-docs` sibling, working correctly in both primary checkout and worktree layouts
2. **Self-auditing WARN** — harness now emits `WARN: search_skills was never called in scenario N` when expected routing doesn't occur, making future reruns on different models immediately diagnostic
3. **Removed `--features testing` from doc comment** — this binary has no `testing`-gated API dependencies
4. **Fixed `percentile_wall_ms` empty-vec footgun** — added early `return 0` guard
5. **Added date timestamp to `build_report()`** — reruns are now self-documenting without manual intervention
6. **Fixed missing `- ` bullet in Methodology section** — `search_skills` line now correctly part of the list

### ⏭️ Skipped (1)

- **`tests/common/mod.rs` consolidation** — only 2 test binaries share `ollama_running`/`first_ollama_model`. Extracting a shared module is premature for two files; revisit if a third test binary needs the same utilities.

### No re-review needed

All changes are mechanical (path fix, guard, formatting, warn emission). No logic or architectural changes introduced.

Issue #3, opened 2025-08-04
Desktop shell with resizable panels


malibio:
## Goal
Implement multi-panel layout system supporting JournalView, LibraryView, AIChatView, and NodeViewer with VS Code-style resizable panels.

## Requirements
- Dynamic panel activation (panels appear when user activates views)
- Resizable panels with draggable dividers
- Panel state persistence (remember layout on restart)
- Focus management between panels
- Keyboard navigation

## Dependencies
- [ ] Issue #1 (Project Foundation) must be completed first
- [ ] Issue #2 (Design System) recommended for consistent styling

## Acceptance Criteria
- [ ] Multi-panel layout system implemented
- [ ] Panels can be resized by dragging dividers
- [ ] Panel state persists between app restarts
- [ ] Focus management works correctly
- [ ] Keyboard navigation between panels

PR #1286, opened 2026-05-31
Fix AI chat viewer and model manager bugs found during Gemma 4 verification

### LINKED ISSUE #1285: Fix AI chat viewer and model manager bugs found during Gemma 4 verification
## Overview

Several bugs were discovered during manual verification of #1142 (Gemma 4 + Ministral chat-template paths). The AI chat viewer is non-functional due to a combination of a missing gRPC service registration, a property-flattening gap in the dev proxy, and a viewer reactivity issue. The model manager also has status display bugs.

### 1. LocalAgentService not registered in daemon gRPC server

\`packages/daemon/src/main.rs\` builds a \`ServiceBundle\` and registers several gRPC services, but never adds \`LocalAgentServiceServer\`. As a result, \`LocalAgentService/ListModels\`, \`LoadModel\`, \`DownloadModel\`, and all other model-management RPCs return \`Unimplemented\`. The model manager UI therefore cannot list or load models.

**Fixed in this session** — \`LocalAgentServiceImpl\` is now constructed in \`build_services()\` and registered in both the headless and tray server blocks.

### 2. Dev proxy returns namespaced properties instead of flat properties

\`packages/dev-tools/src/dev-proxy.ts\` — \`nodeDataToApiNode()\` parses the raw JSON string from gRPC but returns it as-is (e.g. \`{ "ai-chat": { "provider": "native" } }\`). The Tauri Rust layer runs \`flatten_properties_for_api()\` to extract the type namespace before sending to the frontend, but the proxy has no equivalent. So in browser dev mode, \`node.properties.provider\` is always \`undefined\`, and the AI chat viewer is permanently stuck on the placeholder.

**Fixed in this session** — added \`flattenProperties(nodeType, raw)\` to \`nodeDataToApiNode()\` in the proxy, mirroring the Rust logic.

### 3. AI chat viewer stuck on placeholder after provider selection (Tauri)

In the Tauri app, even after \`provider: 'native'\` is persisted to the DB (confirmed via grpcurl), the \`AiChatNodeViewer\` still renders the placeholder. The \`$derived\` for \`provider\` reads \`node?.properties?.provider\` which should be \`'native'\` after \`flatten_properties_for_api\` — but the viewer does not re-render. Root cause not yet confirmed; likely a Svelte 5 reactivity issue on initial hydration from the watcher path or a guard in \`setNode\` blocking the update.

**Not yet fixed.**

### 4. Model download status never transitions to "ready"

After a download completes (confirmed in daemon logs: "Download complete"), the model manager UI continues to show "downloading" status. The daemon emits no event to signal completion back to the frontend event stream, so the UI never refreshes the status row.

### 5. Model shows "Error" status before/during download

When a model download is triggered, the UI briefly (or persistently) shows "Error" for that model row before the download event stream begins. This appears to be a race between the optimistic status update and the first progress event.

### 6. Daemon startup timeout too short for embedding model load

\`packages/desktop-app/src-tauri/src/daemon_setup.rs\` — \`wait_for_daemon()\` uses a 5-second timeout. On first launch, \`nodespaced\` must load \`nomic-embed-text\` before it binds the gRPC socket, which takes 6–8 seconds. The Tauri app therefore fails to connect on cold start ("nodespaced did not respond within 5s") and falls back to showing a disconnected state, even though the daemon successfully starts shortly after.

## Acceptance Criteria

- [ ] \`LocalAgentService\` is registered in the daemon — model list, download, and load RPCs return results (not \`Unimplemented\`)
- [ ] Dev proxy \`nodeDataToApiNode()\` returns flat properties matching the Tauri Rust shape
- [ ] AI chat viewer in Tauri renders \`AiChatModelPicker\` when \`properties.provider === 'native'\` — no manual workaround needed
- [ ] Model download status transitions from "downloading" → "ready" in the UI after the daemon confirms completion
- [ ] No "Error" flash shown on a model row when a valid download is triggered
- [ ] Daemon startup timeout extended (or made adaptive) so cold-start embedding model load does not cause a connection failure

### Files Modified in This Session (items 1 & 2 already fixed)
- **Daemon gRPC registration**: \`packages/daemon/src/main.rs\` — \`LocalAgentServiceImpl\` added to \`ServiceBundle\`, registered in both \`serve_grpc\` blocks
- **Dev proxy flattening**: \`packages/dev-tools/src/dev-proxy.ts\` — \`flattenProperties(nodeType, raw)\` added to \`nodeDataToApiNode()\`

### Files Needing Work (items 3–6)
- **Viewer reactivity**: \`packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte\` lines 103–112 — \`$derived\` for \`provider\` / \`isMessageProvider\`
- **Download status**: \`packages/agent/src/local_agent/model_manager.rs\` — download completion event emission; \`packages/daemon/src/services/local_agent_service.rs\` — streaming response after download
- **Daemon timeout**: \`packages/desktop-app/src-tauri/src/daemon_setup.rs\` — \`wait_for_daemon()\` timeout constant

### Root Cause References
- \`flatten_properties_for_api\` in \`packages/desktop-app/src-tauri/src/types.rs:502\` — Rust reference implementation for property flattening
- \`nodeDataToApiNode\` in \`packages/dev-tools/src/dev-proxy.ts:204\` — proxy equivalent (now fixed)
- \`selectProvider\` in \`ai-chat-node-viewer.svelte:114\` — guard \`if (p === provider) return\` may fire incorrectly
- \`setNode\` in \`shared-node-store.svelte.ts:1320\` — skip-while-editing guard may block hydration update

## Related Issues
- #1142 (parent — Gemma 4 verification, blocked by items 3–6)

malibio:
Closes #1285

### DIFF
--- a/packages/desktop-app/src-tauri/src/agent_events.rs
+++ b/packages/desktop-app/src-tauri/src/agent_events.rs
@@ -19,5 +19,8 @@ pub const LOCAL_AGENT_ERROR: &str = "local-agent://error";
 /// Model download progress update.
 pub const MODEL_DOWNLOAD_PROGRESS: &str = "model://download-progress";
 
+/// Model download completed successfully.
+pub const MODEL_DOWNLOAD_READY: &str = "model://download-ready";
+
 /// Model status change (loading, loaded, error).
 pub const MODEL_STATUS: &str = "model://status";
--- a/packages/desktop-app/src-tauri/src/commands/chat_models.rs
+++ b/packages/desktop-app/src-tauri/src/commands/chat_models.rs
@@ -121,6 +121,18 @@ pub async fn chat_model_download(
                     },
                 );
             }
+            "ready" => {
+                #[derive(Serialize)]
+                struct ReadyEvent {
+                    model_id: String,
+                }
+                let _ = app.emit(
+                    agent_events::MODEL_DOWNLOAD_READY,
+                    &ReadyEvent {
+                        model_id: event.model_id,
+                    },
+                );
+            }
             "error" => {
                 let msg = event
                     .error_message
--- a/packages/desktop-app/src-tauri/src/daemon_setup.rs
+++ b/packages/desktop-app/src-tauri/src/daemon_setup.rs
@@ -80,8 +80,9 @@ pub async fn ensure_daemon_running(app: &AppHandle) -> Result<DaemonStatus> {
     // Bootstrap or restart the launchd agent.
     bootstrap_launchd_agent(&plist_path)?;
 
-    // Wait briefly for the daemon to come up.
-    let status = wait_for_daemon(&socket_path, Duration::from_secs(5)).await;
+    // Wait for the daemon to come up. On cold start the embedding model load
+    // takes 6–8 seconds before nodespaced binds the socket, so 5s is too short.
+    let status = wait_for_daemon(&socket_path, Duration::from_secs(30)).await;
     Ok(status)
 }
 
--- a/packages/desktop-app/src/lib/components/viewers/ai-chat-model-picker.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/ai-chat-model-picker.svelte
@@ -121,10 +121,11 @@
     downloads = { ...downloads, [modelId]: { downloaded: 0, total: 0 } };
     try {
       await chatModelDownload(modelId);
-      await refresh();
+      // Success-path refresh is handled by the MODEL_DOWNLOAD_READY event listener.
     } catch (e) {
       log.error('Download failed', e);
       error = e instanceof Error ? e.message : String(e);
+      await refresh();
     } finally {
       const next = { ...downloads };
       delete next[modelId];
--- a/packages/desktop-app/src/lib/components/viewers/ai-chat-model-picker.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/ai-chat-model-picker.svelte
@@ -15,6 +15,7 @@
 <script lang="ts">
   import { onMount, onDestroy } from 'svelte';
   import { listen, type UnlistenFn } from '@tauri-apps/api/event';
+  import { AGENT_EVENTS } from '$lib/types/agent-types';
   import { sharedNodeStore } from '$lib/services/shared-node-store.svelte';
   import {
     chatModelList,
@@ -47,6 +48,7 @@
   // modelId -> {downloaded, total} while a download is in flight.
   let downloads = $state<Record<string, { downloaded: number; total: number }>>({});
   let unlistenProgress: UnlistenFn | null = null;
+  let unlistenReady: UnlistenFn | null = null;
 
   const visibleModels = $derived(models.filter((m) => m.backend === backend));
 
@@ -84,7 +86,7 @@
         model_id: string;
         bytes_downloaded: number;
         bytes_total: number;
-      }>('model://download-progress', (event) => {
+      }>(AGENT_EVENTS.MODEL_DOWNLOAD_PROGRESS, (event) => {
         const { model_id, bytes_downloaded, bytes_total } = event.payload;
         downloads = {
           ...downloads,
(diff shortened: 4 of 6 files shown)


malibio (commented):
### Code Review Summary

This PR fixes bugs #3–6 from issue #1285 with targeted, well-scoped changes. All six acceptance criteria are addressed (items 1–2 were already merged). The overall approach is sound. I have one improvement-level finding and two nits.

---

### Findings

#### Suggested Improvements

**`ai-chat-model-picker.svelte` — Double `refresh()` on successful download**

The `download()` function already calls `await refresh()` after `await chatModelDownload(modelId)` resolves (line 124). The `MODEL_DOWNLOAD_READY` event listener also calls `refresh()` (line 106). Because `chat_model_download` is a streaming Tauri command that only resolves after the Rust command loop exits — which happens *after* the `"ready"` event is emitted and forwarded to the frontend — the event-triggered `refresh()` will fire and complete first, then the `await chatModelDownload` resolves and fires a second `refresh()`. The end result is two sequential `chatModelList()` + `getSystemRamGb()` calls in rapid succession.

This is a minor redundancy, not a correctness bug — the UI settles correctly. But it does mean two network round-trips on every successful download. The recommended fix is to remove the `await refresh()` from inside `download()`, since the event listener now handles that responsibility reliably:

```ts
async function download(modelId: string) {
  error = null;
  downloads = { ...downloads, [modelId]: { downloaded: 0, total: 0 } };
  try {
    await chatModelDownload(modelId);
    // Refresh is handled by the MODEL_DOWNLOAD_READY event listener.
  } catch (e) {
    log.error('Download failed', e);
    error = e instanceof Error ? e.message : String(e);
    // On error, the ready event won't fire, so refresh manually.
    await refresh();
  } finally {
    const next = { ...downloads };
    delete next[modelId];
    downloads = next;
  }
}
```

**`daemon_setup.rs` — Hard-coded 30s timeout without user-visible feedback**

The fix extends the blocking startup wait from 5s to 30s. This is correct for the cold-start scenario, but the UX implication is that the Tauri app will appear frozen for up to 30 seconds with no indication of progress if the daemon is slow. The existing poll loop at 250ms intervals provides no progress channel to the frontend. This does not block merge — the issue is pre-existing and the fix is clearly better than a spurious "connection failed" error — but a follow-up issue to emit a "daemon starting..." status event on each failed poll would be a meaningful improvement.

---

#### Nitpicks

**Nit: `chat_models.rs` — Inline `ReadyEvent` struct duplicates the `ProgressEvent` pattern**

Both `ReadyEvent` and `ProgressEvent` are defined inline inside match arms. The `ReadyEvent` payload is minimal (just `model_id: String`), so this is fine. But they follow slightly inconsistent placement: `ProgressEvent` is defined where it is used (inside the `"downloading"` arm), and `ReadyEvent` mirrors that. If this file grows more event arms in the future, hoisting the structs to module level would improve readability. Non-blocking for this PR.

**Nit: `ai-chat-node-viewer.svelte` hydration comment**

The comment at the hydration fallback block says "the root cause class behind the original 'picker does nothing'" — the word "class" reads oddly here; "root cause of the original..." would be clearer.

---

| Criterion | Status |
|---|---|
| LocalAgentService registered (bug #1) | Previously merged |
| Dev proxy flattenProperties (bug #2) | Previously merged |
| Viewer renders AiChatModelPicker after provider selection (bug #3) | `type: 'viewer'` source bypasses skip-while-editing guard correctly |
| Download status transitions to "ready" (bug #4) | Rust emits `model://download-ready`, picker calls `refresh()` |
| No "Error" flash on valid download trigger (bug #5) | Addressed via the ready event clearing the downloads map before refresh |
| Daemon startup timeout extended for cold start (bug #6) | 5s -> 30s with explanatory comment |

**Net assessment: approve.** The double-refresh is worth noting but is not a regression. The core correctness and architecture are solid. The event constant is properly mirrored in both Rust (`agent_events.rs`) and TypeScript (`agent-types.ts`) with matching string values, and listener teardown in `onDestroy` is correct.

PR #1292, opened 2026-06-01
Daemon should bind gRPC socket before embedding model loads

malibio:
Closes #1287

### DIFF
--- a/packages/agent/src/acp/context_assembly.rs
+++ b/packages/agent/src/acp/context_assembly.rs
@@ -19,6 +19,7 @@ use nodespace_core::NodeService;
 use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use tokio::sync::RwLock;
 
 /// Default token budget when none is specified by the caller.
 const DEFAULT_TOKEN_BUDGET: u32 = 50_000;
@@ -206,7 +207,8 @@ struct CollectedRelationship {
 /// assembler before each agent session starts.
 pub struct GraphContextAssembler {
     node_service: Arc<NodeService>,
-    embedding_service: Option<Arc<NodeEmbeddingService>>,
+    /// Updated by the background embedding-load task once the model is ready.
+    embedding_service: Arc<RwLock<Option<Arc<NodeEmbeddingService>>>>,
     seed_node_ids: Vec<String>,
     token_budget: u32,
     /// Root of the `packages/skill/shims/` directory. When set,
@@ -218,7 +220,7 @@ pub struct GraphContextAssembler {
 impl GraphContextAssembler {
     pub fn new(
         node_service: Arc<NodeService>,
-        embedding_service: Option<Arc<NodeEmbeddingService>>,
+        embedding_service: Arc<RwLock<Option<Arc<NodeEmbeddingService>>>>,
     ) -> Self {
         Self {
             node_service,
@@ -271,7 +273,8 @@ impl GraphContextAssembler {
 
         // --- Step 2: semantic expansion (best-effort) ---
         let seed_ids: HashSet<String> = seed_nodes.iter().map(|n| n.id.clone()).collect();
-        let semantic_neighbors = match &self.embedding_service {
+        let embedding_guard = self.embedding_service.read().await;
+        let semantic_neighbors = match embedding_guard.as_ref() {
             Some(embedding_service) => {
                 Self::find_semantic_neighbors(embedding_service, &seed_nodes, &seed_ids).await
             }
--- a/packages/cli/tests/cli_integration.rs
+++ b/packages/cli/tests/cli_integration.rs
@@ -40,7 +40,7 @@ async fn spawn_test_daemon() -> (PathBuf, oneshot::Sender<()>, TempDir) {
             .await
             .expect("failed to build NodeService"),
     );
-    let service = NodeServiceImpl::new(node_service, None);
+    let service = NodeServiceImpl::new(node_service, Arc::new(tokio::sync::RwLock::new(None)));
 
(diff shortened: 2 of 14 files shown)


malibio:
## Code Review — PR #1292: Daemon should bind gRPC socket before embedding model loads

### Summary

This is a well-executed, targeted refactor. The two-phase startup split is the right architectural approach: Phase 1 establishes the socket and all non-NLP services in ~100ms, while Phase 2 loads the heavy NLP model in a background task. The `Arc<RwLock<Option<EmbeddingReady>>>` shared-state pattern is idiomatic and correct for this use case. The `Arc<OnceLock<EmbeddingWaker>>` change is an elegant solution to the problem of populating the waker after `NodeService` is already behind an `Arc`.

All acceptance criteria from issue #1287 are met. This is a net positive change and is safe to merge with the one improvement noted below addressed.

---

### Findings

#### [Improvement] `nlp.initialize()` is a synchronous blocking call inside `tokio::spawn`

**File:** `packages/daemon/src/main.rs`, `load_embedding_model_bg` (line ~380)

`EmbeddingService::initialize()` is a `&mut self` synchronous function that blocks for 6–8 seconds doing CPU-bound llama.cpp model loading (Metal kernel compilation, file I/O). Running it directly inside `tokio::spawn` occupies a tokio worker thread for the full load window, which starves the async executor during startup — exactly when the gRPC server is handling its first connections.

The fix is to wrap the blocking call in `tokio::task::spawn_blocking`:

```rust
let nlp = tokio::task::spawn_blocking(move || {
    let mut nlp = EmbeddingService::new(config).map_err(|e| { ... })?;
    nlp.initialize().map_err(|e| { ... })?;
    Ok::<_, String>(nlp)
}).await;
```

This is particularly important because the Tauri socket-health check fires within 2 seconds of daemon start and the gRPC server will be handling early RPCs during the exact window the model is loading.

---

#### [Improvement] `GraphContextAssembler` is permanently stuck with `None` embedding service

**File:** `packages/daemon/src/main.rs`, line ~269–271

```rust
let assembler = Arc::new(GraphContextAssembler::new(
    node_service.clone(),
    None, // embedding not yet loaded; context assembly falls back to keyword search
));
```

The comment says "falls back to keyword search" — but this `None` is never updated after the background task completes. The `AgentSessionHandler` holds `Arc<GraphContextAssembler>` and that assembler is permanently without semantic context enrichment for the lifetime of the daemon process.

This is a functional regression for agent sessions: they will never use semantic neighbors for context assembly even after the model loads. Before this PR, when the model loaded (synchronously), the assembler received the live `embedding_service`. Now it never does.

This can be fixed by giving `GraphContextAssembler` its own `Arc<RwLock<Option<EmbeddingReady>>>` reference (same pattern used by the service impls), or by adding a setter that the background task calls after loading. The current approach silently degrades agent context quality with no log warning.

---

#### [Improvement] 2s Tauri timeout is tight for slow machines / cold-start edge cases

**File:** `packages/desktop-app/src-tauri/src/daemon_setup.rs`, line 84

The 2-second timeout is reasonable for the happy path — but Phase 1 startup includes `SqliteStore::new()`, schema seeding, and `seed_agent_nodes()`, which involves file I/O and DB queries. On a machine with a slow spinning disk, NFS home directory, or high load at launch time, these could exceed 2 seconds before the socket binds.

The acceptance criterion says "within 500ms" but the actual time budget needs margin. Consider 5s — it's still a 6x improvement over the previous 30s, correctly reflects that the socket should appear in well under a second, but provides a meaningful safety margin. Alternatively, make the timeout configurable via env var so it's tunable without a binary change.

---

#### Nitpicks

- **Nit:** `packages/daemon/src/services/embeddings_service.rs`, line 198 — `get_stale_count` does not gate on `state.read().await.is_some()` while every other write-path RPC does. This is intentional (the stale count is DB-only, no model needed) and correct, but a brief comment noting "no model required — queries DB directly" would prevent future reviewers from adding a spurious state check.

- **Nit:** `packages/daemon/src/main.rs`, `build_services()` line ~264–266 — `EmbeddingsServiceImpl` is only registered when a model file exists at startup. If the model file appears after daemon start (e.g. user downloads it), the EmbeddingsService gRPC endpoint will not exist in the server. This is identical behavior to before this PR, so not a regression — but worth a `// NOTE:` comment that the endpoint is absent until daemon restart when no model is present at startup.

- **Nit:** `packages/daemon/src/main.rs` line ~115, ~167 — `_bg_task` is intentionally discarded via the underscore prefix. A one-line comment (`// Task runs to completion independently; dropping the handle does not cancel it`) would make the intentional drop explicit for future readers unfamiliar with tokio's detach-on-drop semantics.

malibio (commented):
Review verdict: REQUEST CHANGES. Two improvements should be addressed before merge - see the detailed review comment above.

malibio:
## Review Response

All findings addressed:

**✅ Addressed (from prior commit eb57cd4c):**
- `spawn_blocking` for `EmbeddingService::new + initialize` — sync CPU/IO ops no longer park a tokio worker thread
- `GraphContextAssembler` now receives `Arc<RwLock<Option<Arc<NodeEmbeddingService>>>>` and is populated by the background task once the model loads — semantic context enrichment works for all agent sessions

**✅ Addressed (this commit 953cc96c):**
- Tauri timeout raised 2s → 5s for slow-disk margin (still 6x better than the previous 30s)
- Added `// No model required` comment on `get_stale_count`
- Added `// _bg_task: dropping a JoinHandle does not cancel the task in tokio — it detaches` comments on both `_bg_task` bindings
- Added comment on `EmbeddingsService` registration noting model-file-at-startup requirement

**Skipped:** None — all recommendations were addressed.

Issue #82, opened 2025-08-22
Epic: Advanced Component System Enhancements


malibio:
Transform the component-based node reference system from 9.2/10 to world-class 10/10 implementation.

## Current State
- Core component system implemented with Svelte components
- Build-time plugin architecture working  
- Comprehensive error recovery and performance tracking
- Phase 2 system at 100% test success rate

## Epic Goal
Create production-ready, enterprise-grade component system for advanced knowledge management workflows.

## Implementation Phases
Individual sub-issues will be created for:
- Architecture and Performance optimizations
- User Experience improvements  
- Technical Infrastructure enhancements
- Integration and Ecosystem features

## Success Criteria
- Performance: <50ms component hydration
- Quality: 100% test coverage, zero production errors
- UX: Accessibility AA compliance, mobile-first design
- Senior architect review achieves 10/10 score

## Strategic Value
Position NodeSpace as best-in-class knowledge management platform with component architecture foundation for AI-powered content analysis, collaborative editing, and extensible plugin ecosystems.

PR #1293, opened 2026-06-01
Graceful stale-count when EmbeddingsService is absent (fixes nodespace-sync#82)

mstomar125:
## Problem

In the two-window Pro demo, the frontend logs `Failed to get stale nodes count: {code: \"GRPC_ERROR\", message: \"\"}` **every 5 seconds** for the entire session in both windows.

`nodespaced-pro` omits the NLP/embedding stack (no llama-cpp, per nodespace-sync#10), so its gRPC server never registers `EmbeddingsService`. The status bar's 5s `get_stale_root_count` poll therefore hits an unregistered service → tonic returns `Unimplemented` → the Tauri command propagated it as an error → recurring console flood that masks real errors during debugging.

## Fix

`get_stale_root_count` now treats `Unimplemented` as "no embedding processor ⇒ no embedding queue ⇒ zero stale roots" and returns `Ok(0)`. Every other gRPC error still surfaces unchanged.

```rust
match client.get_stale_count(GetStaleCountRequest {}).await {
    Ok(response) => Ok(response.into_inner().count as usize),
    Err(status) if status.code() == tonic::Code::Unimplemented => Ok(0),
    Err(status) => Err(grpc_err(status.message())),
}
```

The status bar already treats a count of 0 as "nothing queued" (clears the message), so the UI is correct and quiet in Pro mode, while the full daemon (`nodespaced`) is unaffected — it registers `EmbeddingsService` and returns real counts.

## Why this approach

The flood is specifically the embeddings RPC. `EmbeddingsServiceImpl` cannot be constructed without the loaded embedding model, so the Pro daemon cannot cheaply serve it — and per nodespace-sync#10 the NLP stack is deliberately excluded from Pro. Graceful degradation at the command boundary is the honest fix: it respects #10 and removes the silent-failure noise without pretending an embedding queue exists.

Note: `nodespaced-pro` also doesn't register Import/Settings/LocalAgent/AgentSession services, but those are user-triggered (not polled) so they don't flood. Making Pro a strict superset for those non-NLP services is tracked separately and is not required to resolve the flood this issue is about.

## Validation

- `cargo check -p nodespace-app` — clean
- `cargo clippy -p nodespace-app` — no warnings
- `cargo fmt --check` — clean
- Happy path unchanged (same `Result<usize>` signature); frontend behavior identical when the service is present.

Fixes NodeSpaceAI/nodespace-sync#82

### DIFF
--- a/packages/desktop-app/src-tauri/src/commands/embeddings.rs
+++ b/packages/desktop-app/src-tauri/src/commands/embeddings.rs
@@ -10,8 +10,9 @@ use serde::{Deserialize, Serialize};
 use tauri::State;
 
 use nodespace_proto::nodespace::{
-    BatchQueueEmbeddingsRequest, GetStaleCountRequest, QueueEmbeddingRequest,
-    RegenerateEmbeddingRequest, SearchSemanticRequest, TriggerBatchEmbedRequest,
+    BatchQueueEmbeddingsRequest, GetStaleCountRequest, GetStaleCountResponse,
+    QueueEmbeddingRequest, RegenerateEmbeddingRequest, SearchSemanticRequest,
+    TriggerBatchEmbedRequest,
 };
 
 fn grpc_err(msg: impl std::fmt::Display) -> CommandError {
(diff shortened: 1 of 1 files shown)


mstomar125:
## Code Review — Graceful stale-count when EmbeddingsService is absent

**PR**: #1293
**Issue**: nodespace-sync#82

### CI Pipeline Status
⏳ **No checks reported** on `worktree-issue-82-pro-stale-count-graceful` (`gh pr checks` exits non-zero with "no checks reported"). Not a failure, but CI does not appear wired up for this branch — please confirm the workspace gates (`cargo check`/`clippy`/`fmt`, frontend) run somewhere before merge. The PR body documents local runs of `cargo check -p nodespace-app`, `clippy`, and `fmt --check` as clean.

### Requirements Check (nodespace-sync#82)
- ✅ **No `GRPC_ERROR` flood** — `get_stale_root_count` is the only RPC on a recurring interval (`app-shell.svelte:260`, 5s `setInterval`). `check_daemon_status` is on-click only. Collapsing `Unimplemented → Ok(0)` silences exactly the flood source. The status bar already treats `0` as "nothing queued" (`app-shell.svelte:246-252`), so the UI stays correct and quiet.
- 🟡 **Dependent features work or are cleanly disabled** — satisfied for the embedding-queue indicator (clean "nothing queued"). Import/Settings/LocalAgent are explicitly left for a separate ticket; reasonable since they are user-triggered (no flood), and the PR body says so. Calling that out as deliberately-deferred rather than done.
- 🟡 **Decision recorded, cross-referenced with #10** — the rationale (graceful degradation at the command boundary, honoring #10's NLP exclusion) is captured in the PR description and doc comment, but not in the issue thread or an ADR. A one-line note on #82 itself would close this criterion cleanly.

### Findings

#### 🔴 Critical
- None.

#### 🟡 Important
- None.

#### 🟢 Suggestion
- `packages/desktop-app/src-tauri/src/commands/embeddings.rs:222-226`: **Nit:** No test covers the new branch. The existing `#[cfg(test)]` block (`:274-305`) only exercises param structs, and a full command test needs a mock gRPC server (`State<GrpcClient>`) — heavyweight, so I wouldn't block on it. A cheap middle ground: extract the `Result<GetStaleCountResponse, Status> → Result<usize, CommandError>` mapping into a small pure fn and unit-test it (`Unimplemented → Ok(0)`, any other `Status → Err`). Principle: behavior worth a code comment is behavior worth a regression guard, and this keeps the guard from silently rotting if the match arm is edited later.
- `packages/desktop-app/src-tauri/src/commands/embeddings.rs:224`: **Nit:** The arm collapses *any* `Unimplemented` to `Ok(0)`, not just "service unregistered." If `get_stale_count` were ever stubbed/unimplemented *within* a registered `EmbeddingsService`, the queue indicator would silently read zero instead of surfacing the gap. Low risk today (the community daemon implements it), and the doc comment frames it as service-absence — worth one clause acknowledging the method-level case so the assumption is explicit.

### Summary
Tight, well-scoped, well-documented fix. The `Unimplemented`-as-capability-signal approach is **consistent with established precedent** in this same crate (`services/pro_client.rs:103` and `lib.rs:281` already map `Status::Unimplemented` to "community tier"), so this isn't a new pattern — it extends one. The change is correctly targeted at the sole 5s poller and leaves every other gRPC error path untouched, so real failures still surface. Net positive; the items above are optional polish, not correctness blockers.

**Recommendation**: REQUEST CHANGES (nits only — the two 🟢 suggestions and a one-line decision note on #82; the code itself is sound and I'd happily approve once those are acknowledged or waived)

mstomar125:
Thanks for the review — addressed in `221a68d1`:

- 🟢 **Extract + test the mapping** — pulled the `Result<GetStaleCountResponse, Status> → Result<usize, CommandError>` mapping into a pure `stale_count_from_result()` and added three unit tests (no mock gRPC server needed): `Unimplemented → Ok(0)`, `Ok → passthrough`, any other `Status → Err`. The contract now has a regression guard so the arm can't silently rot.
- 🟢 **Method-level `Unimplemented`** — the doc comment now calls out explicitly that the arm collapses *any* `Unimplemented` for this RPC, including a (today non-existent) registered-but-method-unimplemented `EmbeddingsService`, and why that's acceptable (community daemon implements it; worst case the indicator reads zero rather than erroring).
- 🟡 **Decision recorded on #82** — already captured on the issue: nodespace-sync#82 (comment recording capability-graceful-degradation chosen over registering `EmbeddingsService` on Pro, honoring #10's NLP exclusion). Re-stating here for the trail.

Local gates green: `cargo test -p nodespace-app` (3 new tests pass), `clippy` (0 issues), `fmt --check` clean.

Re-requesting review.

PR #1296, opened 2026-06-02
Retrieve workspace context dynamically instead of injecting full schema dump

malibio:
Closes #1290

### DIFF
--- a/packages/core/src/ops/context_ops.rs
+++ b/packages/core/src/ops/context_ops.rs
@@ -57,8 +57,15 @@ pub struct PlaybookInfo {
 // ---------------------------------------------------------------------------
 
 /// Build workspace context by querying schemas, collections, and playbooks.
+///
+/// `query` is the user's message for this turn. When provided, entity types
+/// are filtered to those relevant to the query (type_id or display_name
+/// mentioned) plus all core types. When absent, all types are included.
+/// This prevents the workspace context from growing unboundedly as users
+/// create custom schemas.
 pub async fn build_workspace_context(
     node_service: &Arc<NodeService>,
+    query: Option<&str>,
 ) -> Result<WorkspaceContext, OpsError> {
     // Fetch schemas (empty vec on error — fresh database is fine)
     let schemas = node_service.get_all_schemas().await.unwrap_or_default();
@@ -77,7 +84,7 @@ pub async fn build_workspace_context(
         .unwrap_or_default();
 
     // Convert schemas to EntityTypeInfo
-    let entity_types: Vec<EntityTypeInfo> = schemas
+    let all_entity_types: Vec<EntityTypeInfo> = schemas
         .into_iter()
         .map(|schema| {
             let fields: Vec<FieldInfo> = schema
@@ -128,6 +135,24 @@ pub async fn build_workspace_context(
             }
         })
         .collect();
+
+    // Filter entity types by query relevance.
+    // Core types are always included; custom types are included only when
+    // the query mentions their type_id or display_name (case-insensitive).
+    // Without a query, all types are included (session start, fallback).
+    let entity_types = if let Some(q) = query {
+        let q_lower = q.to_lowercase();
+        all_entity_types
+            .into_iter()
+            .filter(|et| {
+                et.is_core
+                    || q_lower.contains(&et.type_id.to_lowercase())
+                    || q_lower.contains(&et.display_name.to_lowercase())
+            })
+            .collect()
+    } else {
+        all_entity_types
+    };
 
     // Convert playbook nodes
     let active_playbooks: Vec<PlaybookInfo> = playbook_nodes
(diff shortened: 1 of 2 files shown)


malibio:
## Code Review — PR #1296 (Initial Review)

---

### Overall Assessment

This is a well-scoped, net-positive change that directly attacks an unbounded-growth problem in the system prompt. The approach — keyword filtering at the query level, with core types always included — is pragmatic and correct for the current scale. The implementation is clean, the diff is tight, and the five new unit tests are thorough. One correctness concern warrants a fix before merge; everything else is improvement-level or below.

---

### Requirements Check

- ✅ `workspace_context` token count does not grow proportionally with user-created schemas — keyword filtering on custom types caps the injected set to those explicitly referenced in the query.
- ⚠️ Queries that reference a specific entity type still get that type's schema in context — **true for exact-name matches only**; see Critical Issue #1 for the gap.
- ✅ Fallback to full dump when embedding service is unavailable — `query: None` path on session start returns all types; no embedding service dependency introduced.
- ✅ No regression in the model's ability to create correct node types — core types always present; custom types surface when named.

---

### Findings

#### [Critical/Blocker] — Logic gap: implicit type creation is silently starved of schema context

**File**: `packages/core/src/ops/context_ops.rs` — lines 143–155
**File**: `packages/daemon/src/services/local_agent_service.rs` — line 211

The acceptance criterion reads: *"Queries that reference a specific entity type still get that type's schema in context."* The implementation satisfies this **only when the user names the type explicitly**. The issue statement itself flags the harder case:

> *"e.g. 'I need to track my clients' — 'customer' type not mentioned"*

With the current filter, a user who says *"I need to track my clients"* would send a query containing neither `customer` nor `Customer`, so the `customer` schema is silently excluded. The model then proceeds with only core schemas, and will either create an incorrect node type or invent a schema it cannot validate against. This is a silent degradation — no error, no warning, just wrong behaviour in a category the issue explicitly called out.

The issue itself acknowledges this and proposes semantic retrieval (approach 1) as the remedy, while suggesting keyword filtering (approach 2) as the starting quick win. The AC wording slightly over-promises what approach 2 can deliver. Two paths forward:

1. **Narrow the AC** — update the acceptance criterion (or PR description) to make clear that only explicit-name matches are covered now, and open a follow-up issue for semantic retrieval.
2. **Add a safety-net fallback** — if the keyword filter reduces the custom-type set to zero and there are custom schemas in the DB, include a small fixed set (e.g., the top-N most recently used) so the model has at least some non-core schema exposure. This avoids the "model knows nothing about the user's domain" failure mode.

Path 1 is acceptable for a lean team if the follow-up issue is filed immediately. Path 2 adds token cost but provides a better floor guarantee. Either way, the PR description or a code comment should explicitly document this limitation so the next engineer does not wonder why the model creates wrong types on indirect queries.

---

#### [Improvement] — Tests exercise a private re-implementation, not the production code path

**File**: `packages/core/src/ops/context_ops.rs` — lines 434–448

The five new tests call a local `filter_types` helper defined inside the test module (lines 434–448), which duplicates the filtering logic verbatim from lines 143–155 of the production function. This pattern means:

1. The tests could pass even if someone changed the production filter and forgot to update the test helper.
2. The duplication will drift over time.

The filtering logic is currently inlined inside an `async fn` that requires a `NodeService` to test end-to-end, which is why the helper was introduced. The cleaner fix is to extract the filtering into a small, synchronous, non-`pub` free function:

```rust
fn apply_query_filter(types: Vec<EntityTypeInfo>, query: Option<&str>) -> Vec<EntityTypeInfo> {
    let Some(q) = query else { return types; };
    let q_lower = q.to_lowercase();
    types.into_iter().filter(|et| {
        et.is_core
            || q_lower.contains(&et.type_id.to_lowercase())
            || q_lower.contains(&et.display_name.to_lowercase())
    }).collect()
}
```

Then `build_workspace_context` calls `apply_query_filter(all_entity_types, query)` and the tests call the same `apply_query_filter` directly, eliminating the duplication. Single source of truth for the logic.

---

#### [Improvement] — `max_chars` raised unconditionally, including the all-types session-start path

**File**: `packages/daemon/src/services/local_agent_service.rs` — line 818

The budget was raised from 1 500 → 4 000 characters. The commit message justifies this correctly for the filtered case ("filtering keeps the type count small"). However, the `query: None` path on session start also uses the same 4 000 character budget, and that path returns all types — including all custom schemas. For a user with 50+ schemas the session-start context will now be larger than before (was capped at 1 500, now at 4 000). The truncation logic in `format_for_prompt` does enforce the hard ceiling, but it is worth verifying whether the session-start prompt was intended to grow or whether the budget should be split (e.g., 4 000 for filtered turn context, 1 500 for the initialisation full dump). If the initialisation call is only meant to warm up session state and not drive the first turn's inference, a tighter budget there keeps token usage predictable.

---

#### Nitpicks

- **Nit**: `packages/core/src/ops/context_ops.rs` line 197 — `sort_by_key(|e| e.is_core)` relies on knowing that `bool` orders `false < true`. `sort_by_key(|e| if e.is_core { 1usize } else { 0 })` makes the intent self-evident without that domain knowledge.

- **Nit**: `packages/daemon/src/services/local_agent_service.rs` line 210 — the comment says *"filtered to types relevant to the user's message"* which is accurate, but does not mention the limitation. Adding *"(exact name match only; semantic retrieval is a follow-up)"* gives future readers an actionable pointer.

- **Nit**: Test `query_filter_custom_type_included_when_mentioned_by_id` — the query `"find my customers"` matches on substring `customer`, but `display_name` is also `"Customer"` which also matches, so the test does not isolate the `type_id` branch from the `display_name` branch. Using a type with a machine-style `type_id` (e.g., `"cust-001"`) and a query containing only that id would make coverage unambiguous.

---

### Recommendation

**REQUEST CHANGES** — the critical issue (silent context starvation on indirect queries) should be resolved before merge, either by explicitly scoping the fix in the PR description/code comments and filing a follow-up issue for semantic retrieval, or by adding a minimum-context floor for the zero-custom-match case. The core implementation is correct within its stated bounds; the gap is one of under-documented assumptions about what the filter can and cannot deliver. Once addressed, this PR is mergeable.

malibio (commented):
Requesting changes per inline review comment — one correctness gap (silent context starvation on indirect/implicit type queries) must be addressed before merge, either by documenting the limitation and filing a follow-up issue for semantic retrieval, or by adding a minimum-context floor. Details in the review comment above.

malibio:
## Address Review Summary

**Addressed** (2 of the flagged items):

### ✅ Critical: Extracted `apply_query_filter` + documented limitation
- Extracted the inline filter logic into a private `apply_query_filter()` function (commit `8cdef045`)
- Tests now call the production function directly — no more parallel reimplementation that could drift
- Added doc comment on `apply_query_filter` documenting the implicit-reference limitation
- Filed follow-up issue **#1300** for semantic retrieval to cover indirect type references

### ✅ Important: Test isolation for `type_id` vs `display_name` branch
- Renamed the type in `query_filter_custom_type_included_when_mentioned_by_id` from `customer`/`Customer` to `cust-001`/`Entity` so the test unambiguously exercises the `type_id` path

### ✅ Nit: Daemon call-site comment
- Added *"(exact keyword match; implicit references are a follow-up for semantic retrieval)"* to the comment at the `send_message` call site

**Skipped** (with rationale):

### ⏭️ `max_chars` session-start split (1500 vs 4000)
The `format_for_prompt` truncation already enforces the hard ceiling regardless. Splitting the constant into two separate values adds complexity for marginal benefit at current schema counts — the `query: None` session-start path is non-inference-driving initialization. Will revisit if token usage at session start becomes measurable.

### ⏭️ `sort_by_key` bool clarity nit
`false < true` is idiomatic Rust and universally understood by Rust developers. Adding a numeric sentinel makes the line longer without improving readability for the target audience.

PR #1298, opened 2026-06-02
Compress seed prompt nodes to reduce baseline token cost (closes #1291)

malibio:
## Summary

- Audited and rewrote SCHEMA_CREATION_RULES, TOOL_STRATEGY_RULES, Response Formatting Rules, and Tool Call Formatting seeds to eliminate redundancy with tool schemas and standard model behavior
- Token count before: ~1164 (seed static content); after: ~569 (~51% reduction)
- Key cuts: removed search_semantic parameter list (already in schema), verbose search_nodes filter examples, redundant update_node rule, and multi-bullet narration rules
- Updated snapshot test in prompt_assembler.rs to match new content; all 223 agent tests pass

## Test plan

- [ ] All 223 agent tests pass (`cargo test -p nodespace-agent`)
- [ ] Baseline system prompt (fresh DB) measured under 3000 tokens
- [ ] Model still produces correctly formatted responses
- [ ] Model still uses tools correctly with trimmed tool strategy guide

### DIFF
--- a/packages/agent/src/agent_guidance.rs
+++ b/packages/agent/src/agent_guidance.rs
@@ -17,7 +17,7 @@
 /// `skill_pipeline.rs` (used only by the skill-based schema-creation path) and
 /// should be consolidated here when that path is unified — tracked separately
 /// from #1089.
-pub const SCHEMA_CREATION_RULES: &str = "NODE MODEL: Everything in NodeSpace is a node. Built-in types (task, text, date) are always available. Custom types (e.g. 'project', 'customer') require a schema node to exist first — the schema defines the type's fields and title template. Once a schema exists, create instances with create_node(node_type=<schema_id>). Use create_schema only to define a new type; use create_node to create data.";
+pub const SCHEMA_CREATION_RULES: &str = "NODE MODEL: Everything is a node. Built-in types: task, text, date. Custom types need a schema first (create_schema). Once a schema exists, create instances with create_node(node_type=<schema_id>). Never call create_schema for a type already in ENTITY TYPES.";
 
 /// Tool strategy guidance.
 ///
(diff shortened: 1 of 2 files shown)


malibio:
### Code Review Summary

This PR delivers exactly what issue #1291 asked for: a focused audit-and-compress of the four seed prompt nodes, reducing static guidance text by ~51% (chars) without losing behavioral coverage. The change is atomic, touches only two files, and the approach of using the tool schema as the source of truth for parameter documentation rather than duplicating it in guidance prose is the correct architectural direction.

All 223 agent tests pass. The snapshot test (`seed_prompt_bodies_match_expected_bytes`) correctly locks in the new content and will catch future drift.

The change is a net improvement. One behavioral gap and one stale doc comment need attention before merge.

---

### Findings

#### Critical Issues

**[Critical/Blocker] `update_node` guidance dropped without replacement**

`packages/agent/src/agent_guidance.rs`, line 29-39

The old `TOOL_STRATEGY_RULES` contained an explicit instruction:

```
- To update a node's title or content: search_nodes for it by name, then use update_node with the real ID.
```

The new version retains `update_task_status` guidance but silently drops `update_node`. The `update_node` tool is a registered, actively-used tool (`local_agent/tools.rs:332`, whitelisted in skill_pipeline). The commit message calls this out as an intentional cut ("redundant update_node rule"), but the tool schema description alone ("Update an existing node's content or properties.") does not carry the critical search-first constraint.

The general "ALWAYS search first before updating or getting a node" rule provides partial coverage, but `update_node` (unlike `update_task_status`) is the only node-content mutation tool and warrants explicit mention at the same level as `update_task_status`. Without it, a model that sees both tools in the schema may reach for `update_node` with a hallucinated ID — exactly the failure mode the search-first mandate was introduced to prevent.

**Suggested fix** — fold it into the existing update line (net zero token cost):

```
- To update task status: search_nodes for it, then update_task_status with the real ID. To update a node's content: search_nodes first, then update_node with the real ID.
```

---

#### Suggested Improvements

**[Improvement] TOOL_STRATEGY_RULES doc comment is stale**

`packages/agent/src/agent_guidance.rs`, lines 22-28

The doc comment still claims the constant covers "the canonical `create_node` / `create_schema` / `create_relationship` usage patterns" and "how to choose between `search_nodes`, `search_semantic`, `get_node`, and `get_related_nodes`." After the compression, the level of detail for each of those has been substantially reduced. The comment should be updated to reflect that parameter-level detail is intentionally deferred to tool schemas, so the next reader understands the design intent and does not restore verbose content thinking the comment describes a requirement.

---

#### Nitpicks

**Nit:** `packages/agent/src/prompt_assembler.rs`, line 302 — The `"{}."` format string now appends a trailing period after `NODE_REFERENCE_FORMAT`, making the rendered bullet read: `"Reference nodes with bare URI: nodespace://abc-123 (no markdown links, no backticks)."` The constant itself does not end with a period (by design, to allow inline embedding). The added period is harmless but inconsistent with how the constant is used in `acp/context_assembly.rs` (no period there). Worth aligning for consistency, but not a blocker.

**Nit:** The commit-reported token counts ("~1164 before, ~569 after") are approximate chars-as-tokens estimates. The acceptance criterion ("under 3000 tokens total baseline") operates at the full assembled-prompt level. A one-line comment in the commit message or code noting what the baseline total is post-merge (accounting for Core Identity + Workspace Context Template + Tool Strategy Guide + Response Formatting + Tool Call Formatting) would make it easier to verify the criterion is met without running the daemon.

---

### Verdict

**Request changes** — the `update_node` omission is a behavioral regression in a greenfield system where the model's tool-use quality is the product. The fix is one short sentence. Everything else in this PR is correct and the compression direction is right.

malibio (commented):
The update_node guidance was dropped without replacement. The general 'ALWAYS search first' rule partially covers it, but update_node is the only node-content mutation tool and warrants the same explicit search-first call as update_task_status. One-sentence fix described in the review comment. Everything else in the PR is correct.

PR #1299, opened 2026-06-02
Scope tool list to query-relevant tools to reduce prompt token cost

malibio:
Closes #1289

### DIFF
--- a/packages/agent/src/agent_types.rs
+++ b/packages/agent/src/agent_types.rs
@@ -524,4 +524,17 @@ pub trait AgentToolExecutor: Send + Sync {
 
     /// Execute a tool by name with the given JSON arguments.
     async fn execute(&self, name: &str, args: serde_json::Value) -> Result<ToolResult, ToolError>;
+
+    /// Select which tools to expose to the model for a specific user query.
+    ///
+    /// Default: returns all tools unchanged. Implementors that support
+    /// semantic skill lookup can override this to return only query-relevant
+    /// tools, reducing prompt token cost and improving model focus.
+    async fn select_tools(
+        &self,
+        _query: &str,
+        all_tools: Vec<ToolDefinition>,
+    ) -> Vec<ToolDefinition> {
+        all_tools
+    }
 }
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -133,14 +133,19 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
             name: None,
         });
 
-        // Get available tools. The full tool set (including `search_skills`)
-        // is exposed every turn — the model decides whether to search for a
-        // skill, invoke one directly, or respond without one.
-        let tools = self
+        // Get available tools, then scope to query-relevant tools.
+        // select_tools runs a pre-inference semantic skill search to narrow
+        // the list to baseline + matched skill whitelists. Falls back to the
+        // full list when skill search is unavailable or returns too few tools.
+        let all_tools = self
             .tool_executor
             .available_tools()
             .await
             .unwrap_or_default();
+        let tools = self
+            .tool_executor
+            .select_tools(user_message, all_tools)
+            .await;
 
         let dynamic_ctx = session.dynamic_context.as_deref().unwrap_or("");
         let model_name = session.model_id.as_deref().unwrap_or("unknown");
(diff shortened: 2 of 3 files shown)


malibio:
## Code Review — PR #1299 (Issue #1289)

---

### Requirements Check

- ✅ Tool list passed to inference is scoped to query-relevant tools (3-5 instead of 13): `select_tools` narrows to baseline (4) + matched skill whitelists. Observed target range confirmed in commit message (~3–7 tools).
- ✅ Always-on tools (`search_nodes`, `search_semantic`, `get_node`) are always included: `BASELINE_TOOLS` constant guarantees these are always in the `selected_names` set. `search_skills` is also correctly included so the model can self-discover further skills.
- ✅ Skill retrieval failure gracefully falls back to full tool list: three distinct fallback paths — no embedding service, `find_skills` error, and selection below `MIN_SELECTED_TOOLS` — all return `all_tools` unchanged.
- ✅ Prompt token count measurably reduced on typical queries: the commit message reports ~1000–1500 tokens saved per pass, consistent with reducing 13 → 3–7 tool schemas.
- ✅ No regression in task completion rate: 226/226 agent unit tests pass; the `search_skills` tool remains in `BASELINE_TOOLS` so the model retains the ability to discover any tool at runtime.

---

### Code Review Findings

#### No Critical Issues

The implementation is architecturally sound. Key correctness points verified:

- `select_tools` is called once before the ReAct loop (line 145–148 of `agent_loop.rs`), not re-run on each iteration. This is intentional and correct: re-running a semantic embedding query on each ReAct iteration would add latency and the initial query captures user intent accurately.
- The JSON field accessed in `select_tools` (`skill.get("tools")`) matches the field name emitted by `skill_ops::find_skills` (`"tools": tool_whitelist`), so the mapping is correct.
- The `execute` method on `GraphToolExecutor` is not gated on the scoped list — it will execute any tool by name regardless of what was passed in the `tools` field to the inference engine. This is correct: if the model hallucinates a tool name outside the scoped set, it gets an `UnknownTool` error back via the existing error path, which is surfaced as a tool result error and fed back to the model. No silent data loss.
- The `BASELINE_TOOLS` constant (4 tools) is larger than `MIN_SELECTED_TOOLS` (3), so the fallback condition `selected_names.len() < 3` can never trigger when baseline tools alone are present. This is safe but the threshold is effectively dead code for the no-skills-match case.

#### Suggested Improvements

**[Improvement] `packages/agent/src/local_agent/tools.rs`, `MIN_SELECTED_TOOLS` constant (line 1408)**

`MIN_SELECTED_TOOLS = 3` is strictly less than `BASELINE_TOOLS.len() = 4`. After seeding `selected_names` with all 4 baseline tools, the `selected_names.len() < MIN_SELECTED_TOOLS` check can never be true — baseline alone always satisfies it. The intent documented in the constant comment is to prevent the selection being too narrow when skill search returns results with empty `tool_whitelist` arrays. However, with only the baseline seeded and zero skill matches, the count is already 4, clearing the guard.

The practical effect is benign (baseline-only selection is a valid narrow case and probably fine), but the guard is misleading. Consider either raising `MIN_SELECTED_TOOLS` to `BASELINE_TOOLS.len() + 1` to make the intent explicit ("at least one skill-contributed tool must be present"), or dropping the guard and documenting that baseline-only is an acceptable fallback path, or using `>` baseline count as the condition.

**[Improvement] `packages/agent/src/local_agent/tools.rs`, `select_tools` — zero confidence threshold (line 1430–1437)**

`find_skills` is called with `limit: Some(3)` and internally uses `SKILL_SEARCH_THRESHOLD = 0.0`. For the model-as-caller use case, a zero threshold is intentional (let the model judge relevance). For the automated `select_tools` pre-filter path, it means even orthogonally-irrelevant skills (very low cosine score) could contribute their `tool_whitelist` entries to the selected set. A query about "what did I write about embeddings?" could have a low-confidence match on a "task management" skill and inject `update_task_status` into the selected list.

This is a correctness concern more than a hard bug — the model still has to choose to call the injected tool — but it partially undermines the stated goal of improving model focus. Consider applying a non-zero confidence threshold (e.g. `> 0.3`) specifically on the `select_tools` path, distinct from the model-facing `search_skills` tool which intentionally uses zero.

**[Improvement] `packages/agent/src/local_agent/agent_loop.rs` — tools list is fixed for entire turn**

The scoped tool list computed before the ReAct loop is reused for all iterations in the loop (line 165 `tools.clone()`, line 231 `tools: Some(tools.clone())`). If the model calls `search_skills` in iteration 1 and discovers it needs a tool outside the scoped list (e.g. `delete_node`), it cannot call that tool in iteration 2 because it is not in the prompt schema. The model would need to call `search_skills` again, but the tool schemas presented to the engine won't include the newly needed tool. This is a known limitation of pre-filtering and should be explicitly documented in the code comment at line 136.

#### Nitpicks

**Nit: `packages/agent/src/local_agent/tools.rs`, line 1477** — The `tracing::info!` log on the success path includes the full comma-separated tool list. For long lists this is fine (max ~7 entries), but consider `tracing::debug!` to reduce log verbosity in production; the token count metric at the `info` level is the valuable signal, not the full name list.

**Nit: `packages/agent/src/local_agent/tools.rs`, line 1448** — `std::collections::HashSet` could be imported at the top of the impl block or the module rather than being inlined as a path. Minor style inconsistency vs. the rest of the file.

---

### Test Coverage Assessment

Three tests are added:

1. `select_tools_no_embedding_service_returns_all` — covers the no-embedding-service fallback.
2. `select_tools_baseline_tools_always_present` — validates baseline presence (but only exercises the no-embedding-service path, not an actual skill-matched path with a live embedding service).
3. `run_turn_calls_select_tools_on_executor` — integration tracking test verifying `run_turn` delegates to the executor's `select_tools` method.

The tests are appropriate for the offline-testable surface area. The confidence-threshold and ReAct mid-turn limitations noted above are not directly testable without an embedding service, so their absence in tests is acceptable.

---

### Summary

This is a well-scoped, clean implementation. The design — default passthrough on the trait, override at the `GraphToolExecutor` layer, three graceful fallback paths — is correct and extensible. All 226 agent unit tests pass. The two improvements around the confidence threshold for automated pre-filtering and the `MIN_SELECTED_TOOLS` / `BASELINE_TOOLS` mismatch are worth addressing before high-scale use, but neither is a correctness blocker today. The ReAct mid-turn limitation is a known trade-off that should be documented.

**Recommendation: APPROVE** (net improvement, issues are non-blocking for current scale)

---

*Reviewed by Claude Code (claude-sonnet-4-6) — Pragmatic Quality Framework*

malibio (commented):
Clean implementation. Three graceful fallback paths, correct field mapping from skill_ops, and appropriate test coverage for offline-testable surface area. Two non-blocking improvements worth a follow-up: (1) MIN_SELECTED_TOOLS < BASELINE_TOOLS.len() makes the narrowness guard dead code — clarify or raise the threshold; (2) zero-confidence threshold in the automated pre-filter path risks injecting low-relevance skill tools into the selected set. Neither blocks merge at current scale.

PR #1302, opened 2026-06-02
Semantic retrieval for workspace_context: cover implicit type references

malibio:
Closes #1300

### DIFF
--- a/packages/core/src/ops/context_ops.rs
+++ b/packages/core/src/ops/context_ops.rs
@@ -40,17 +40,26 @@ pub struct PlaybookInfo {
 // ---------------------------------------------------------------------------
 
 /// Similarity threshold for schema semantic search.
+///
+/// 0.4 is empirically chosen to suppress false matches from short or ambiguous
+/// queries while still catching clear synonyms (e.g. "clients" → `customer`).
+/// `skill_ops` uses 0.0 (no threshold) because that search is over a smaller,
+/// curated set of skill nodes; schemas are more numerous and noisier.
 const SCHEMA_SIMILARITY_THRESHOLD: f32 = 0.4;
 
-/// Maximum number of schemas to inject via semantic retrieval.
+/// Maximum number of schemas to inject per turn.
+///
+/// Keeps the injected context compact. Three schemas cover the common case of a
+/// query that spans a primary type plus one or two related types; more would
+/// bloat the prompt for marginal gain.
 const MAX_SEMANTIC_SCHEMAS: usize = 3;
 
 /// Build workspace context by querying collections and playbooks.
 ///
 /// When `embedding_service` and `query` are both provided, schema nodes
-/// semantically similar to the query are retrieved and injected alongside
-/// keyword-matched schemas. Falls back to keyword-only when the embedding
-/// service is unavailable (#1300).
+/// semantically similar to the query are retrieved and injected into the
+/// context. Falls back to schema-free context when the embedding service is
+/// unavailable or the query is empty (#1300).
 pub async fn build_workspace_context(
     node_service: &Arc<NodeService>,
     embedding_service: Option<&Arc<NodeEmbeddingService>>,
@@ -109,7 +118,7 @@ pub async fn build_workspace_context(
                     schemas
                 }
                 Err(e) => {
-                    tracing::warn!(error = %e, "workspace_context: semantic schema search failed, falling back to keyword-only");
+                    tracing::warn!(error = %e, "workspace_context: semantic schema search failed, omitting schemas");
                     vec![]
                 }
             }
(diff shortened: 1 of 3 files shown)


malibio:

---

## Code Review Summary

This PR closes a genuine gap: queries that use synonyms or paraphrases of a schema name (e.g. "track my clients" → `customer`) were silently excluded from `workspace_context` because the prior system had no entity types in context at all (the `_query` parameter on `build_workspace_context` was a dead no-op placeholder on `main`). The implementation is surgical — three files, one clean data-flow path, correct fallback behavior, and zero regressions on existing tests.

One clarification on framing: the commit message states "layer semantic schema retrieval on top of keyword filter," but on `main` the keyword filter had already been removed (#1283/#1284 squash) and the `_query` param was a stub. This is actually a pure additive feature, not a composition. The code itself is correct; the description slightly overstates what it sits on top of.

All six new unit tests pass. `cargo check` and `cargo clippy` are clean.

---

## Requirements Check

- ✅ Queries that implicitly reference a type ("track my clients" → `customer` schema) inject the relevant schema via `semantic_search_nodes_of_type` with threshold 0.4 and cap 3
- ✅ Fallback to keyword-only (i.e., no schema injection) when embedding service is `None` or search fails — `tracing::warn` logged on failure
- ✅ No regression for explicit-name queries — `relevant_schemas` is additive; existing collections/playbooks output is unchanged

---

## Findings

### Suggested Improvements

**`packages/core/src/ops/context_ops.rs:154` — Inconsistent budget check (`<` vs `<=`)**

The collections section guard uses `<=`:
```rust
if out.len() + section.len() <= max_chars {
```
The `RELEVANT ENTITY TYPES` header guard uses `<`:
```rust
if out.len() + header.len() < max_chars {
```
This is a pre-existing inconsistency but the new schemas section perpetuates it. Off-by-one is inconsequential at 4000-char budgets but will silently drop the header when `out.len() + header.len() == max_chars`. Standardize on `<=` for all section headers.

**`packages/daemon/src/services/local_agent_service.rs:196,220` — Lock scope on embedding service read**

```rust
let emb = self.inner.embedding_service.read().await.clone();
```

This is correct — the `read()` guard drops immediately after `.clone()` since `Arc::clone` is synchronous. A brief inline comment would help future readers who might question whether the guard outlives the clone.

**`packages/core/src/ops/context_ops.rs:197` — `start_session` silently skips semantic retrieval with no comment**

`start_session` passes `query: None`, which is correct (no user query exists yet), but there is no inline comment explaining why — unlike `send_message` which passes `Some(&message)`. This asymmetry should be documented so readers understand the design intent rather than assuming it was accidentally omitted.

**`packages/core/src/ops/context_ops.rs:42–46` — Magic-number constants lack rationale**

```rust
const SCHEMA_SIMILARITY_THRESHOLD: f32 = 0.4;
const MAX_SEMANTIC_SCHEMAS: usize = 3;
```

`skill_ops.rs` uses a threshold of `0.0` (effectively no threshold) with a user-configurable cap; this file uses `0.4` hardcoded. Neither constant has a comment explaining the tuning basis. For `SCHEMA_SIMILARITY_THRESHOLD` especially, a one-line rationale (e.g. "empirically chosen to avoid false matches from short queries") would help future tuning.

---

### Nitpicks

**Nit: `packages/core/src/ops/context_ops.rs:112` — fallback log message references "keyword-only" which no longer exists**

`tracing::warn!(..., "workspace_context: semantic schema search failed, falling back to keyword-only")`

On `main`, keyword filtering was removed in #1283/#1284. The fallback is actually no schema injection at all (empty `relevant_schemas`). "falling back to schema-free context" or "semantic search failed, omitting schemas" would be more accurate and less confusing during log analysis.

**Nit: `packages/core/src/ops/context_ops.rs:290–295` — truncation test comment refers to a non-existent feature**

```rust
assert!(output.len() <= 200); // some slack for the truncation note
```

`format_for_prompt` does not emit any truncation note — it simply stops appending. The comment is misleading; remove or correct it.

---

## Recommendation

**APPROVE**

The change is a net improvement: it closes the implicit-reference gap with correct fallback behavior, adds no new dependencies, is architecturally consistent with `skill_ops.rs` (same `semantic_search_nodes_of_type` API), and all tests pass clean. The findings above are documentation gaps and minor polish — none block merge.

malibio (commented):
All acceptance criteria met, tests pass, no critical issues. See inline comment for suggested improvements and nits — none block merge.

malibio:
## Address Review — All Findings Applied

Addressed all 6 findings from the initial review:

**Fixed:**
- ✅ **Budget guard inconsistency** (`<` → `<=`): standardized RELEVANT ENTITY TYPES and ACTIVE PLAYBOOKS header guards to match the collections section
- ✅ **Constants rationale**: added comments to `SCHEMA_SIMILARITY_THRESHOLD` (explains 0.4 vs skill_ops 0.0 — noisier schema candidate set) and `MAX_SEMANTIC_SCHEMAS` (3 covers primary type + related types)
- ✅ **Fallback log message**: "falling back to keyword-only" → "omitting schemas" (keyword filtering was removed in #1283)
- ✅ **Doc comment**: removed "keyword-matched" phrasing from `build_workspace_context` doc (no longer accurate)
- ✅ **Truncation test**: corrected misleading "slack for truncation note" comment + tightened assertion to `<= 100` (format_for_prompt emits no truncation note)
- ✅ **Inline comments**: added lock-scope comments on `read().clone()` in both `start_session` and `send_message`, and documented why `start_session` passes `None` (no query at session creation)

No skipped recommendations.

PR #1309, opened 2026-06-03
NodeService domain event broadcasting: immediate for single writes, batched for bulk operations

malibio:
Closes #1306

### DIFF
--- a/packages/core/src/services/node_service/bulk.rs
+++ b/packages/core/src/services/node_service/bulk.rs
@@ -531,7 +531,7 @@ impl NodeService {
             }
         }
 
-        // Step 3: All validations passed - perform atomic bulk update
+        // Step 3: All validations passed - perform atomic bulk update.
         self.store.bulk_update(updates.clone()).await.map_err(|e| {
             NodeServiceError::bulk_operation_failed(format!(
                 "Failed to execute bulk update transaction: {}",
@@ -539,7 +539,19 @@ impl NodeService {
             ))
         })?;
 
-        // NOTE: NodeUpdated events are now automatically emitted by store notifier (Issue #718)
+        // Issue #1306: Emit one NodeUpdated event per updated node.
+        // `store.bulk_update` runs a single SQL transaction without per-row notify calls,
+        // so we emit explicitly here rather than relying on the store notifier.
+        for (id, _) in &updates {
+            self.emit_event(DomainEvent::NodeUpdated {
+                node_id: id.clone(),
+                node_type: existing_nodes
+                    .get(id)
+                    .map(|n| n.node_type.clone())
+                    .unwrap_or_default(),
+                changed_properties: vec![],
+            });
+        }
 
         Ok(())
     }
@@ -578,8 +590,9 @@ impl NodeService {
             return Ok(());
         }
 
-        // Delete nodes one by one using SqliteStore
-        // SQLite handles atomicity within each delete operation
+        // Coalesce delete events: one event per node regardless of cascade writes
+        // (Issue #1306).
+        let _batch = self.begin_batch_emit();
         for id in &ids {
             self.store
                 .delete_node(id, self.client_id.clone())
@@ -590,9 +603,8 @@ impl NodeService {
                         id, e
                     ))
                 })?;
-
-            // NOTE: NodeDeleted event is now automatically emitted by store notifier (Issue #718)
         }
+        // _batch drops here → one flush per deleted node
 
         Ok(())
     }
--- a/packages/core/src/services/node_service/mod.rs
+++ b/packages/core/src/services/node_service/mod.rs
@@ -30,7 +30,7 @@ use async_trait::async_trait;
 use regex::Regex;
 use serde_json::Value;
 use std::collections::{HashMap, HashSet};
-use std::sync::{Arc, OnceLock};
+use std::sync::{Arc, Mutex, OnceLock};
 use tokio::sync::broadcast;
 
 // Sub-module declarations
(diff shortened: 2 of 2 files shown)


malibio:
## Code Review — PR #1309

---

### Requirements Check

- ✅ Single `update_node` calls emit domain events immediately (test `batch_emit_single_update_is_immediate` validates this; store notifier path is unchanged)
- ✅ Bulk operations can opt into batched emission — `begin_batch_emit()` RAII guard activates coalescing; `bulk_delete` uses it internally; external callers can wrap any bulk loop
- ✅ WatchNodes subscribers receive at most one event per node per batch window — last-write-wins HashMap coalescing enforces this in `BatchEmitGuard::drop`
- ✅ No latency regression for single writes — `BatchState::Immediate` is the default; the notifier fast-path adds only one `std::sync::Mutex` lock per event (no allocation, no await)

All four acceptance criteria are satisfied.

---

### Summary

Clean, minimal implementation. The RAII guard pattern is idiomatic Rust. The `std::sync::Mutex` choice (not `tokio::sync::Mutex`) is correct — the lock is never held across an await point. All four new tests pass and cover the critical behavioral boundaries. Clippy is clean.

Two findings worth addressing before merge, both moderate in scope.

---

### Findings

#### [Improvement] `emit_event` in `bulk_update` bypasses batch state

📁 `packages/core/src/services/node_service/bulk.rs:545-554`
📁 `packages/core/src/services/node_service/mod.rs:1248-1258`

`bulk_update` posts events via `self.emit_event(...)`, which calls `self.event_tx.send(...)` directly, bypassing `batch_state` entirely. This is intentional for the current use — the store's `bulk_update` runs a single SQL transaction with no per-row notifier callbacks, so events must be emitted explicitly after the commit. However, `emit_event` is also called by relationship methods (`RelationshipUpdated`, `RelationshipCreated`) that could fire during or around a batch window.

More importantly: if a future caller wraps `bulk_update` inside `begin_batch_emit()`, those emitted events will bypass the buffer and fire immediately, silently defeating the guard. This is a correctness trap that will be invisible at compile time.

**Recommendation**: Route `emit_event` through the batch state the same way the store notifier does:

```rust
pub(crate) fn emit_event(&self, event: DomainEvent) {
    use crate::db::events::{EventEnvelope, EventMetadata};
    let envelope = EventEnvelope {
        event,
        metadata: EventMetadata {
            source_client_id: self.client_id.clone(),
            playbook_context: self.execution_context.clone(),
        },
    };
    // Route through batch state, consistent with the store notifier
    let mut state = self.batch_state.lock().unwrap_or_else(|e| e.into_inner());
    match &mut *state {
        BatchState::Immediate => { let _ = self.event_tx.send(envelope); }
        BatchState::Batching(buf) => {
            if let Some(node_id) = envelope.event.node_id() {
                buf.insert(node_id, envelope);
            } else {
                let _ = self.event_tx.send(envelope);
            }
        }
    }
}
```

This requires `DomainEvent` to expose a `node_id() -> Option<String>` helper, or alternatively the buffer key can use a synthetic key for non-node events (relationship events) since those aren't coalesced today. Alternatively, accept the current design with a doc comment on `emit_event` explicitly warning that it bypasses batch state, so the invariant is visible at the call site.

---

#### [Improvement] `begin_batch_emit` silently resets an in-progress batch

📁 `packages/core/src/services/node_service/mod.rs:1233-1241`

```rust
pub fn begin_batch_emit(&self) -> BatchEmitGuard {
    let mut state = self.batch_state.lock().unwrap_or_else(|e| e.into_inner());
    *state = BatchState::Batching(HashMap::new());  // ← always replaces
    drop(state);
    BatchEmitGuard { ... }
}
```

If `begin_batch_emit` is called while a batch is already active (e.g., a nested bulk operation, or a future refactor adds a second bulk call path), the existing buffer is silently discarded and both the outer and inner `BatchEmitGuard` flush independently on drop, but the outer guard flushes back to `Immediate` first, then the inner guard flushes to a now-reset state.

The actual drop order is: inner guard drops (replaces state with `Immediate`, flushes inner events), then outer guard drops (replaces state with `Immediate` again, flushes nothing from its buffer since it was already cleared by the inner `begin_batch_emit`). Result: the outer batch's accumulated events are silently lost.

There are no nested usages today, but making the invariant explicit would prevent a future correctness bug. A simple guard:

```rust
pub fn begin_batch_emit(&self) -> BatchEmitGuard {
    let mut state = self.batch_state.lock().unwrap_or_else(|e| e.into_inner());
    debug_assert!(
        matches!(*state, BatchState::Immediate),
        "begin_batch_emit called while already batching — nested guards are not supported"
    );
    *state = BatchState::Batching(HashMap::new());
    drop(state);
    BatchEmitGuard { ... }
}
```

This surfaces the misuse in test/debug builds at zero runtime cost in release.

---

#### [Nit] Flush order in `BatchEmitGuard::drop` is non-deterministic

📁 `packages/core/src/services/node_service/mod.rs:454-463`

`HashMap::into_values()` yields events in arbitrary order. For the current use case (WatchNodes receiving one event per node) the order of subscribers processing distinct-node events doesn't matter semantically, so this is not a correctness issue. Worth noting if the subscriber pipeline ever gains ordering sensitivity (e.g., parent-before-child assumptions).

---

#### [Nit] `bulk_create_hierarchy_trusted` does not use `begin_batch_emit`

📁 `packages/core/src/services/node_service/bulk.rs:334-431`

`bulk_create_hierarchy_trusted` (used by `import_service`) calls `store.bulk_create_hierarchy_root_notify`, which iterates over all created nodes and fires the store notifier for each. During a large import this will emit one event per created node in `Immediate` mode — exactly the problem the issue describes.

This is not a regression introduced by this PR (the behavior was identical before), but since import is explicitly called out in the issue notes as "the primary candidate for batched emission", it would be worth a follow-up to wrap the store-level notify loop in `begin_batch_emit` or at minimum track a follow-up issue for it.

---

### Recommendation

**REQUEST CHANGES** — the `emit_event` bypass ([Improvement] #1) is the primary concern. It creates a silent invariant violation that will be invisible to future authors. The nested guard issue (#2) is lower risk but similarly invisible. Both can be fixed in a few lines (either route `emit_event` through batch state, or add a `debug_assert` to document the unsupported case). The `bulk_create_hierarchy_trusted` gap (#3) is a pre-existing gap, not introduced here, and can be tracked separately.

malibio (commented):
Review complete — see detailed findings in the comment above. Two [Improvement] items and two [Nit] items identified. Recommend addressing the emit_event batch bypass and nested guard invariant before merge.

malibio:
## Review Addressed

### ✅ Implemented (2/4 findings)

**🟡 emit_event bypass** — `emit_event` now routes through `batch_state`: node events (Created/Updated/Deleted) buffer last-write-wins when a guard is active; relationship events broadcast immediately (they're not node-keyed).

**🟡 Nested batch guard** — `begin_batch_emit` now `debug_assert!`s that no batch is already active, documenting the unsupported invariant at zero release cost.

### ⏭️ Skipped (2/4 findings)

**🟢 Flush order non-determinism** — `HashMap::into_values()` ordering is intentionally acceptable. Subscribers process events independently per node; there's no defined ordering contract. Switching to `BTreeMap` adds O(log n) overhead to every batch flush for a concern that doesn't exist in any current or planned subscriber.

**🟢 bulk_create_hierarchy_trusted batch guard** — Pre-existing gap not introduced by this PR. Opening follow-up issue to track it separately as the reviewer suggested.

PR #1310, opened 2026-06-03
Add e2e test infrastructure: frontend→daemon→database round-trip

malibio:
## Summary

- Adds `DaemonTestHarness` that spawns a real `nodespaced` + `dev-proxy` against a temp database — no mocks, full stack
- 16 e2e tests across three suites: CRUD round-trip, schema seeding, WatchNodes SSE
- New `bun run test:e2e` command using a dedicated `vitest.e2e.config.ts`
- Fixes two `dev-proxy` bugs: optional proto string fields were sent as `''` instead of being omitted, causing `Invalid parent node` and `path cannot be empty` errors

## Closes

Closes #1308

## Acceptance Criteria Status

- [x] At least one test exercises the full write→read round-trip via HttpAdapter against a live nodespaced
- [x] Test for daemon startup timing: schemas are populated after startup
- [x] Test for WatchNodes: write via backend, assert SSE stream receives the event
- [ ] AI chat persistence (out of scope per issue — blocked on #1305)
- [ ] Tests run in CI against a test daemon (requires CI pipeline update, not in this PR)
- [x] Clear documentation on how to run the full e2e suite locally (vitest.e2e.config.ts + nodespace-docs/development/e2e-testing.md)

## Test plan

- `bun run test:e2e` — all 16 tests pass
- `bun run test` — baseline maintained (7 failed | 125 passed, pre-existing failures unchanged)
- `bun run quality:fix` — no errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "test:db:watch": "bun run --cwd packages/desktop-app test:db:watch",
     "test:coverage": "bun run --cwd packages/desktop-app test:coverage",
     "test:skill": "bun run --cwd packages/skill test",
+    "test:e2e": "bun run --cwd packages/desktop-app test:e2e",
     "test:all": "bun run test && bun run test:skill && bun run rust:test",
     "test:all:db": "bun run test:db && bun run test:skill && bun run rust:test",
     "check": "bun run --cwd packages/desktop-app check",
--- a/packages/desktop-app/eslint.config.js
+++ b/packages/desktop-app/eslint.config.js
@@ -264,6 +264,9 @@ export default [
         global: 'readonly', // Node.js global for tests
         globalThis: 'readonly',
         process: 'readonly',
+        __dirname: 'readonly', // Node.js CommonJS global (available via Bun/Vitest)
+        __filename: 'readonly',
+        Buffer: 'readonly', // Node.js Buffer global
         vi: 'readonly', // Vitest
         describe: 'readonly', // Vitest
         it: 'readonly', // Vitest
--- a/packages/desktop-app/package.json
+++ b/packages/desktop-app/package.json
@@ -27,6 +27,8 @@
     "test:db": "TEST_USE_DATABASE=true bunx vitest run",
     "test:db:watch": "TEST_USE_DATABASE=true bunx vitest",
     "test:integration:full": "bun run test && bun run test:browser && bun run test:perf && bun run rust:test",
+    "test:e2e": "bunx vitest run --config vitest.e2e.config.ts",
+    "test:e2e:watch": "bunx vitest --config vitest.e2e.config.ts",
     "test:all": "bun run test && bun run rust:test",
     "rust:test": "cd src-tauri && cargo test --lib -p nodespace-core -- --test-threads=2",
     "quality:check": "bunx eslint 'src/**/*.{ts,svelte}' && BROWSERSLIST_IGNORE_OLD_DATA=1 bunx svelte-check",
--- a/packages/desktop-app/src/tests/e2e/daemon-harness.ts
+++ b/packages/desktop-app/src/tests/e2e/daemon-harness.ts
@@ -1,0 +1,326 @@
+/**
+ * DaemonTestHarness — spin up a real nodespaced + dev-proxy for e2e tests.
+ *
+ * Usage:
+ *   const h = await DaemonTestHarness.start();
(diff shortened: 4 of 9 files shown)


malibio (commented):
### Code Review Summary

This PR delivers a well-scoped, high-value addition: a real integration testing layer that exercises the full **HTTP → gRPC → SQLite** path with no mocks. The harness design is sound — isolated temp directories, dynamic port allocation, sequential execution to prevent contention. The dev-proxy bug fix is correct and well-motivated. Net assessment: approve with one improvement item and two nits.

---

### Findings

#### Suggested Improvements

**`packages/desktop-app/src/tests/e2e/watch-nodes.e2e.ts` — SSE registration race in `nodeCreated` test**

The `nodeCreated` test creates the `eventPromise` (which internally `await`s `fetch(sseUrl)`) and then immediately `await`s `createNode`. Since `waitForEvent` is called without `await`, the function runs until its first `await fetch(...)`, then yields back to the test. If `createNode` completes before the SSE fetch response arrives — and therefore before `sseClients.add(clientRef)` executes server-side — the `nodeCreated` event fires before any listener is registered and is silently missed, causing a 5 s timeout flake.

The `nodeUpdated` and `nodeDeleted` tests don't have this problem because they pre-create the node before starting the subscription.

The straightforward fix is to wait for the initial `: connected` event line before returning from the subscription setup, or (simpler for a test) to await a brief barrier after starting the SSE fetch. A clean pattern:

```typescript
// In waitForEvent, resolve a "connected" promise when the ': connected\n\n' line is received,
// then export a two-phase API: { connected: Promise<void>, event: Promise<Record<...>> }
```

Alternatively, restructure the `nodeCreated` test to match the other two: pre-create the node and listen for re-creates, or introduce a tiny `await new Promise(r => setTimeout(r, 50))` after the `eventPromise = waitForEvent(...)` line. The current code passes in practice because `createNode` involves gRPC + SQLite and is slower than a localhost HTTP handshake, but it's relying on timing rather than ordering.

---

#### Nitpicks

**Nit: `packages/desktop-app/src/tests/e2e/daemon-harness.ts` line 252 / 284 — `on('error')` throws from an async callback**

```typescript
daemonProc.on('error', (err) => {
  throw new Error(`Failed to start nodespaced: ${err.message}`);
});
```

Throwing inside a Node.js `EventEmitter` callback after the enclosing `async` function has already settled will become an unhandled rejection or crash the process rather than rejecting the promise that the caller is awaiting. The error handlers here fire asynchronously — after `waitForSocket` has already returned — so the throw goes unobserved. This doesn't affect test correctness (the error manifests as a timeout in `waitForSocket`/`waitForHttp` instead), but the handler is misleading and the throw does nothing useful. Consider removing the throws and relying on the timeout machinery, or storing the error and re-throwing at the next async checkpoint.

**Nit: `packages/desktop-app/src/tests/e2e/daemon-harness.ts` — local `HttpAdapter` class duplicates `backend-adapter.ts`**

The comment explains why (`backendAdapter` singleton returns `MockAdapter` in test env), which is a valid constraint. For future maintainers: when the production `HttpAdapter` is exported separately from the singleton, the local copy can be replaced with an import. As-is, any API shape divergence between the two (e.g., `insertPosition` is sent by the production adapter but not by the harness adapter) won't cause test failures but will leave certain code paths untested. Acceptable for this scope.

---

| Criterion | Status |
|-----------|--------|
| Write→read round-trip via `HttpAdapter` against live `nodespaced` | Covered — `node-roundtrip.e2e.ts` |
| Schema seeding: schemas populated after startup | Covered — `schema-seeding.e2e.ts` |
| WatchNodes: write triggers SSE event | Covered — `watch-nodes.e2e.ts` |
| AI chat persistence | Explicitly deferred (blocked on #1305) |
| Tests run in CI | Acknowledged as requiring pipeline update, not in this PR |
| Clear local docs | Covered — `nodespace-docs/development/e2e-testing.md` |

### Dev-Proxy Fix

The change to omit optional proto `string` fields instead of sending `''` is correct. The proto definition confirms all three fields (`parent_id`, `collection`, `lifecycle_status` in `CreateNodeRequest`; `node_type`, `add_to_collection`, `remove_from_collection`, `lifecycle_status` in `UpdateNodeRequest`) are declared `optional string`, meaning prost generates `Option<String>` on the Rust side. Sending `''` would be deserialized as `Some("")`, which the Rust validation layer correctly rejects. Omitting the field serializes as `None`. The fix is technically sound and the inline comments document the invariant clearly.

Issue #2, opened 2025-08-04
Create design system foundation


malibio:
## Goal
Create comprehensive design system foundation with tokens, patterns, and component architecture standards.

## Requirements
- Design tokens (colors, typography, spacing, shadows)
- Living HTML style guide in `docs/design-system/`
- Node visual patterns and interaction states
- Component architecture standards
- CSS custom properties system
- **Color Scheme Reference**: Base colors extracted from product launch campaign document
- Professional blue accent theme maintaining brand consistency
- Dual theme support (light/dark) with campaign-aligned aesthetics

### Dark Theme Colors:
- Background: Dark navy/charcoal (#1a1a1a - #2a2a2a)
- Primary text: Light gray/white (#e0e0e0 - #ffffff)
- Accent: Bright blue (#007acc - #0099ff)
- Secondary elements: Medium gray (#666666 - #888888)

### Light Theme Colors:
- Background: White/light gray (#ffffff - #f8f9fa)
- Primary text: Dark charcoal (#1a1a1a - #333333)
- Accent: Professional blue (matching dark theme)
- Secondary elements: Medium gray (#666666 - #999999)

## Dependencies
- [x] Issue #1 (Project Foundation) must be completed first

## Acceptance Criteria
- [x] Design tokens defined and documented
- [x] HTML style guide created
- [x] Visual patterns for different node types
- [x] CSS custom properties system implemented
- [x] Component architecture guidelines documented
- [x] Color tokens match campaign reference themes
- [x] Theme switching between light/dark modes works
- [x] Professional appearance consistent with brand identity

malibio:
✅ **Issue #2 Complete!**

Design system foundation has been successfully implemented and merged. All acceptance criteria met:

- ✅ Design tokens defined and documented
- ✅ HTML style guide created with interactive examples  
- ✅ Visual patterns for different node types established
- ✅ CSS custom properties system implemented
- ✅ Component architecture guidelines documented
- ✅ Color tokens match campaign reference themes (professional blue #007acc)
- ✅ Theme switching between light/dark modes functional
- ✅ Professional appearance consistent with brand identity

**Key Deliverables:**
- Complete TypeScript design token system with light/dark themes
- Interactive HTML style guide at   
- BaseNode component pattern with full accessibility
- CSS custom properties with theme switching
- Component architecture guidelines for future development

**Next Steps:**
- Issues #3 (Desktop Shell) and #4 (TextNode UI) can now proceed
[...]

Issue #1, opened 2025-08-04
Set up Tauri + Svelte project structure


malibio:
# NodeSpace Project Foundation: Tauri + Svelte Setup

## Project Context
NodeSpace is an AI-native knowledge management system that combines hierarchical note organization with intelligent AI assistance. This issue establishes the foundational project structure using Rust backend, Svelte frontend, and Tauri desktop framework.

## Current State
- ✅ Comprehensive architecture documentation in `/docs/architecture/`
- ✅ Git repository initialized and connected to GitHub  
- ✅ Complete code implementation with working desktop application

## Goal
Create a working Tauri desktop application with Svelte frontend that serves as the foundation for all NodeSpace features.

### System Prerequisites
- Rust 1.80+ with Cargo
- Node.js 20 LTS with npm
- 8GB+ RAM (for future AI model integration)

### Project Structure
```
nodespace-core/
├── Cargo.toml                 # Workspace root
├── nodespace-app/             # Tauri desktop application
│   ├── src-tauri/             # Rust backend
│   │   ├── Cargo.toml
│   │   ├── src/main.rs
│   │   └── tauri.conf.json
│   ├── src/                   # Svelte frontend
│   │   ├── App.svelte
│   │   └── main.js
│   └── package.json
```

## Implementation Steps

1. ✅ Create Cargo workspace structure
2. ✅ Initialize Tauri application with Svelte
3. ✅ Configure build system and development environment
4. ✅ Create basic UI with placeholder panels

## Acceptance Criteria

**Verification Checklist for Agent:**
- [x] Tauri desktop app launches successfully
- [x] Svelte frontend renders with placeholder panels (JournalView, LibraryView, AIChatView, NodeViewer)
- [x] Development server works with hot reload (`npm run tauri dev`)
- [x] Build process creates distributable package (`npm run tauri build`)
- [x] No compilation errors or warnings
- [x] App window is resizable with minimum dimensions enforced
- [x] App shows version information from Tauri
- [x] All dependencies install correctly

**Testing Commands:**
```bash
# Install dependencies
npm install
cd nodespace-app && npm install

# Test development mode
npm run tauri dev

# Test build process  
npm run tauri build

# Verify no errors
cargo check --workspace
npm run build
```

✅ **COMPLETED**: This foundation has been fully implemented and verified. All acceptance criteria have been met and sub-issues (#6-#10) have been closed. Ready for Issues #2-#5 development.

[...]

PR #1312, opened 2026-06-03
Fix node content not persisting to database after editing in Tauri

malibio:
Closes #1307

### DIFF
--- a/packages/desktop-app/src/lib/services/shared-node-store.svelte.ts
+++ b/packages/desktop-app/src/lib/services/shared-node-store.svelte.ts
@@ -1589,6 +1589,8 @@ export class SharedNodeStore {
             // Operation was cancelled by a newer operation - this is expected
             return;
           }
+          // OCC errors are already handled inside the persistence closure (rollback + notification)
+          if (isVersionConflict(err)) return;
           // Surface non-OCC write failures visibly so users know their change didn't save
           conflictNotifications.add({
             nodeId,
--- a/packages/desktop-app/src/lib/services/shared-node-store.svelte.ts
+++ b/packages/desktop-app/src/lib/services/shared-node-store.svelte.ts
@@ -43,7 +43,8 @@ import { conflictNotifications, type ConflictNotification } from '$lib/stores/co
 const CONFLICT_MESSAGE: Record<ConflictNotification['conflictType'], string> = {
   'version-mismatch': 'Your edit conflicted with a remote change',
   'deleted-node': 'The node you edited was deleted by another pane',
-  'child-transfer-failure': "Changes couldn't be saved. Please try again."
+  'child-transfer-failure': "Changes couldn't be saved. Please try again.",
+  'write-failure': "Your change couldn't be saved. Please check your connection."
 };
 
 const log = createLogger('SharedNodeStore');
@@ -652,6 +653,7 @@ export class SharedNodeStore {
    */
   static resetInstance(): void {
     SharedNodeStore.instance = null;
+    PersistenceCoordinator.resetInstance();
   }
 
   // ========================================================================
@@ -1249,8 +1251,14 @@ export class SharedNodeStore {
               // Operation was cancelled by a newer operation - this is expected
               return;
             }
-            // Real errors are logged by PersistenceCoordinator
-            // Re-throw would create unhandled rejection, so we silently handle
+            // OCC errors are already handled inside the persistence closure (rollback + notification)
+            if (isVersionConflict(err)) return;
+            // Surface non-OCC write failures visibly so users know their change didn't save
+            conflictNotifications.add({
+              nodeId,
+              message: CONFLICT_MESSAGE['write-failure'],
+              conflictType: 'write-failure'
+            });
           });
         }
       }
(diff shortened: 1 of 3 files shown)


malibio:
### Code Review Summary

This PR fixes three root causes behind node content not persisting after editing in Tauri. All changes are surgical and well-targeted. The overall approach is correct and the fixes are a definitive net improvement. 5/5 new regression tests pass; 4 pre-existing failures in unrelated test files (`model-store.test.ts`, `pattern-state.test.ts`) are confirmed not introduced by this PR.

---

### Requirements Check

- ✅ **AC1**: Nodes edited in Tauri appear with correct content when read via CLI or browser — addressed by the OCC version-sync fix (fix #2)
- ✅ **AC2**: Content typed across multiple nodes in a session all persists correctly — addressed by the operator-precedence fix (fix #1) restoring the skip-while-editing guard to its intended semantics
- ✅ **AC3**: Silent write failures surface as visible errors — addressed by `conflictNotifications.add('write-failure')` in both the `setNode` and `updateNode` error handlers (fix #3)
- ✅ **AC4**: Regression test: write via `updateNode`, flush, read back, assert content — covered by `node-content-persistence-regression.test.ts` test 1 (`persist-regression-1`), which also asserts the backend-confirmed version propagates into the in-memory store

---

### Code Review Findings

#### Suggested Improvements

**`packages/desktop-app/src/lib/services/shared-node-store.svelte.ts:1586–1598` — `setNode` error handler missing `isVersionConflict` guard**

The `updateNode` error handler at line 1249 correctly guards `isVersionConflict(err)` before emitting `write-failure`, since OCC errors already fire a `version-mismatch` notification from the inner catch at line 1232. The `setNode` error handler at line 1587 does not apply the same guard. In the `setNode` persistence closure, an OCC error on the UPDATE branch (line 1481 inner catch, `else throw updateError`) propagates through the outer catch at line 1555 and is re-thrown — reaching `handle.promise.catch` at line 1587 without the `isVersionConflict` check. This would emit a `write-failure` notification in addition to the `version-mismatch` already fired from the OCC handler.

The `setNode` UPDATE branch runs only when a viewer-sourced node is already in `persistedNodeIds`, which is a narrower path than the `updateNode` UPDATE branch, so the practical blast radius is limited. But the fix is straightforward:

```typescript
handle.promise.catch((err) => {
  if (err instanceof OperationCancelledError) return;
  if (isVersionConflict(err)) return; // OCC already emits version-mismatch notification
  conflictNotifications.add({ ... });
});
```

**`packages/desktop-app/src/lib/services/shared-node-store.svelte.ts:1735–1742` — `deleteNode` error handler silently swallows non-cancellation errors**

The `deleteNode` persistence handle (lines 1735–1742) catches errors and discards all non-cancellation failures silently, with a now-stale comment: _"Real errors are logged by PersistenceCoordinator"_. The comment was partially correct before this PR (errors were logged, just not surfaced to the user), but is inconsistent with the pattern established by the two new `write-failure` handlers. A delete failure is equally user-visible — deleting a node that silently fails to delete is confusing. Consider adding a `write-failure` (or a dedicated `delete-failure` type) notification here for consistency. This is out of scope for #1307 but worth a follow-up.

**`packages/desktop-app/src/lib/services/shared-node-store.svelte.ts:1901–1905` — `updateTaskNode` error handler also silent**

Same pattern: the task update `handle.promise.catch` discards non-cancellation errors without a notification. Lower urgency (task status toggles are visible immediately in the UI), but worth aligning with the established pattern.

---

#### Nitpicks

**Nit: `packages/desktop-app/src/tests/services/node-content-persistence-regression.test.ts:91–92` — mock return type omits `mentions`**

The mock inside the multi-edit test returns an object without a `mentions` field. The `Node` type declares `mentions` as `mentions?: string[]` (optional), so this is type-safe and not a runtime issue. `makeNode` correctly includes `mentions: []`, making the fixture consistent. Just noting this for completeness.

**Nit: `packages/desktop-app/src/tests/services/node-content-persistence-regression.test.ts:105` — loose assertion on call count**

Test 2 (multi-edit) uses `expect(updateSpy).toHaveBeenCalled()` rather than asserting a specific call count. This is intentional and correct given the debounce-coalescing behavior of `PersistenceCoordinator` (the comment explains this). Consider adding a brief comment: _"call count is non-deterministic due to debounce coalescing"_ to make the intent explicit for future maintainers.

---

All three root causes are correctly diagnosed and the fixes are well-scoped. The version-sync after `updateNode` RPC (fix #2) and the operator-precedence extraction (fix #1) are the highest-leverage changes. The `write-failure` notification path (fix #3) appropriately skips OCC errors in the `updateNode` handler. The `PersistenceCoordinator.resetInstance()` addition to `SharedNodeStore.resetInstance()` is a correct test-isolation fix.

The `isVersionConflict` guard gap in the `setNode` handler is a minor correctness issue (double notification, not a data-loss path) that can be addressed as a follow-up or in this PR if the author prefers.

malibio:
**Recommendation: APPROVE** — see detailed review above. All three root causes correctly fixed; 5/5 regression tests pass; 4 pre-existing unrelated test failures confirmed not introduced by this PR.

malibio:
## Review Addressed

**✅ Addressed (2 recommendations):**

**🟡 Important — isVersionConflict guard in setNode outer catch**
Added `if (isVersionConflict(err)) return;` before the `conflictNotifications.add` call in the `setNode` handle.promise.catch handler, matching the guard already present in the `updateNode` handler. Prevents double-notification (version-mismatch + write-failure) on OCC errors.

**🟢 Suggestion — Clarifying comment in regression test 2**
Added a comment explaining why `expect(updateSpy).toHaveBeenCalled()` intentionally omits a call-count assertion: PersistenceCoordinator debounce-coalesces rapid edits non-deterministically.

**⏭️ Skipped (1 recommendation):**

**🟢 Suggestion — deleteNode/updateTaskNode silent error swallowing**
Out of scope for #1307 (which is specifically about content persistence on edit). Will file a follow-up issue to consistently surface write failures across all mutation paths.

---
All 20 affected tests pass. Quality checks clean (0 errors, 0 warnings).

PR #1322, opened 2026-06-04
Remove parentId and collectionId from node wire response

malibio:
Closes #1318

### DIFF
--- a/packages/cli/src/output.rs
+++ b/packages/cli/src/output.rs
@@ -63,16 +63,10 @@ pub fn print_node_list(response: &NodeListResponse, json: bool) -> Result<()> {
 fn write_human_node(node: &NodeData) {
     println!("id:              {}", node.id);
     println!("type:            {}", node.node_type);
-    if let Some(parent) = &node.parent_id {
-        println!("parent:          {}", parent);
-    }
     println!("version:         {}", node.version);
     println!("lifecycle:       {}", node.lifecycle_status);
     println!("created_at:      {}", node.created_at);
     println!("modified_at:     {}", node.modified_at);
-    if !node.collection_id.is_empty() {
-        println!("collection_id:   {}", node.collection_id);
-    }
     if !node.properties.is_empty() && node.properties != "{}" {
         println!("properties:      {}", node.properties);
     }
@@ -99,13 +93,11 @@ pub(crate) fn node_to_json(node: &NodeData) -> serde_json::Value {
         "id": node.id,
         "node_type": node.node_type,
         "content": node.content,
-        "parent_id": node.parent_id,
         "properties": properties,
         "version": node.version,
         "lifecycle_status": node.lifecycle_status,
         "created_at": node.created_at,
         "modified_at": node.modified_at,
-        "collection_id": node.collection_id,
     })
 }
 
@@ -118,13 +110,11 @@ mod tests {
             id: "abc-123".into(),
             node_type: "text".into(),
             content: "hello".into(),
-            parent_id: Some("parent-id".into()),
             properties: r#"{"foo":"bar","n":42}"#.into(),
             version: 7,
             lifecycle_status: "active".into(),
             created_at: "2026-05-17T12:00:00Z".into(),
             modified_at: "2026-05-17T12:00:01Z".into(),
-            collection_id: "hr:policy".into(),
         }
     }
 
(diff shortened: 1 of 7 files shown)


malibio:
## Code Review — PR #1322: Remove parentId and collectionId from node wire response

**Overall Assessment:** This is a clean, well-scoped removal. The architectural motivation is sound — relationship-derived projections have no business living on the node wire representation. The diff is consistently executed across all call sites and the test strategy (verifying membership via the dedicated collection API rather than the removed field) is the right approach. Recommend **approval with one minor cleanup nit** — no blockers.

---

| Criterion | Status |
|---|---|
| `NodeData` proto no longer includes `parent_id` / `collection_id` fields | **Met** — both removed; field numbers 4 and 10 reserved in both `NodeData` and `NodeResponse` |
| `node_to_proto()` no longer fetches or injects parent/collection | **Met** — signature simplified to `fn node_to_proto(node: Node) -> NodeData`; all ~15 call sites updated to point-free `map(node_to_proto)` style |
| All clients continue to work correctly | **Met** — Tauri commands never read these fields from responses (they use `node_id` return values and dedicated APIs); no frontend TS code reads `node.parentId` / `node.collectionId` from fetched nodes |
| `data-layer.md` serialization contract updated | **Met** — committed separately in nodespace-docs (`57f4d49`); line 108 now explicitly documents the wire field list and states these two fields are absent |

---

### Findings

#### Suggested Improvements

**`packages/daemon/src/services/node_service.rs`, line 93–107 — Redundant `.clone()`**

`parent_id_opt.clone()` was originally necessary because `parent_id_opt` was consumed twice: once into `CreateNodeInput` and once into the `NodeResponse` field. The `NodeResponse` consumer is now gone, so the `.clone()` is dead weight. The field can be moved directly:

```rust
// Before
let parent_id_opt = req.parent_id;
...
parent_id: parent_id_opt.clone(),

// After
parent_id: req.parent_id,
```

Removing the intermediate variable and the clone simplifies the code and eliminates a unnecessary heap allocation, per YAGNI. Rust's borrow checker will surface this if the project ever enables the `clippy::redundant_clone` lint.

---

#### Nitpicks

**`packages/daemon/src/services/node_service.rs`, line 1549 — Mixed assertion styles in parity test**

The test verifies `ops_output.collection_id.is_some()` (the ops-layer output struct, not the wire response), which is correct and appropriate — this is testing the ops layer's own return contract, not the wire format. However, the comment above (`// Both nodes should be in a collection and have lifecycle_status=archived`) could be slightly tighter: the RPC-path assertion now goes through `CollectionService` while the ops-path assertion uses the output struct directly. The asymmetry is fine functionally, but a one-line comment acknowledging it would help the next reader.

**`packages/proto/proto/node_service.proto`, line 207–210 — `NodeListResponse.collection_id` is retained (correct but worth a comment)**

`NodeListResponse.collection_id` is a response-level filter echo (which collection was queried), not a per-node membership field — it is legitimately kept. This is architecturally distinct from the per-node `NodeData.collection_id` that was removed. The existing comment `// empty string when not filtered by collection` documents this adequately, but it could explicitly state "this echoes the collection filter from the request, not per-node membership" to prevent future confusion with the removed fields.

---

### What Is Correct and Noteworthy

- **`reserved` statements are correct.** Both number and name reservations are present on both `NodeData` (fields 4, 10) and `NodeResponse` (fields 3, 4). This is the right proto hygiene for a project that serializes to disk or crosses process boundaries, even in greenfield development.

- **Test strategy is correct.** Replacing `assert!(!rpc_resp.collection_id.is_empty())` with `coll_svc.get_node_collections()` is the principled approach — it tests the actual invariant (membership is stored in the relationship layer) rather than a projection artifact.

- **No frontend regression.** The grep confirms the Svelte frontend already managed hierarchy through `ReactiveStructureTree` (populated from relationship events) and never read `node.parentId`/`node.collectionId` from fetched node objects. The `collectionId?: string` field remaining in `CollectionPathSegment` (node.ts:353) is unrelated — it's a local UI parse-assist field, not a wire type.

- **`NodeListResponse.collection_id` correctly retained.** The list-level field echoes the collection filter applied to a search/member query; it is response metadata, not per-node data. Keeping it is correct.

- **`lib.rs` formatting change** (the `resolve_db_path` formatting) is a drive-by `cargo fmt` cleanup, not functional — acceptable in this PR.

malibio (commented):
Review complete — see inline comment above. No blockers found; one nit on the redundant .clone() in create_node.

PR #1323, opened 2026-06-04
Fix ai-chat inference not triggered by nodeUpdated events

malibio:
## Summary

- Inverted the guard in `maybe_handle_ai_chat_node`: trigger inference when `status == "processing"` (the frontend's signal) instead of skipping it
- Removed redundant `write_ai_chat_status("processing")` from `run_ai_chat_turn` — the caller already sets it, and writing it again re-emits a `nodeUpdated` event that caused a re-entry loop once the turn token was removed

## Root Cause

The event handler skipped nodes with `status: processing` — exactly the state the frontend sets to request inference. The recovery scan worked because it called `run_ai_chat_turn` directly, bypassing the guard entirely.

## Verified

- `ai-chat turn triggered` log appears within ~1s of sending a message
- Full end-to-end inference completes with assistant response rendered in UI
- No re-entry loop: the cancellation-token dedup + removal of the redundant `write_ai_chat_status` call prevent cascading events
- Startup recovery still works (calls `run_ai_chat_turn` directly, unaffected)

## Test plan

- [x] `ai-chat turn triggered` log appears after PATCH to `status: processing`
- [x] Inference completes and assistant message written back
- [x] Verified end-to-end via Playwright: send message → processing → idle → response rendered
- [x] No infinite loop on daemon startup with stuck nodes

Closes #1319

### DIFF
--- a/packages/agent/src/prompt_assembler.rs
+++ b/packages/agent/src/prompt_assembler.rs
@@ -127,10 +127,9 @@ impl PromptAssembler {
             filters: None,
         };
 
-        match nodespace_core::ops::node_ops::query_nodes(&self.node_service, filter).await {
-            Ok(result) => {
-                // QueryNodesOutput.nodes is Vec<Value>, deserialize to Vec<Node>
-                result
+        let all_nodes: Vec<Node> =
+            match nodespace_core::ops::node_ops::query_nodes(&self.node_service, filter).await {
+                Ok(result) => result
                     .nodes
                     .into_iter()
                     .filter_map(|v| match serde_json::from_value(v) {
@@ -140,13 +139,27 @@ impl PromptAssembler {
                             None
                         }
                     })
-                    .collect()
-            }
-            Err(e) => {
-                tracing::warn!(error = %e, "Failed to fetch prompt overrides, using base only");
-                Vec::new()
+                    .collect(),
+                Err(e) => {
+                    tracing::warn!(error = %e, "Failed to fetch prompt overrides, using base only");
+                    return Vec::new();
+                }
+            };
+
+        // Keep only true root nodes (no parent edge pointing to them).
+        // query_nodes ignores parent_id filter, so all prompt nodes are returned;
+        // we must post-filter to avoid treating mid-hierarchy nodes as roots.
+        let mut roots = Vec::new();
+        for node in all_nodes {
+            match self.node_service.get_parent(&node.id).await {
+                Ok(None) => roots.push(node),
+                Ok(Some(_)) => {} // has a parent — skip
+                Err(e) => {
+                    tracing::warn!(error = %e, node_id = %node.id, "Failed to check parent, skipping node");
+                }
             }
         }
+        roots
     }
 
     /// Fetch children of a prompt node and concatenate their content as the body.
--- a/packages/core/benches/performance.rs
+++ b/packages/core/benches/performance.rs
@@ -1033,6 +1033,7 @@ fn bench_event_to_rule_matching(c: &mut Criterion) {
         let property_event = DomainEvent::NodeUpdated {
             node_type: "task".to_string(),
             node_id: "bench-event-node".to_string(),
+            node: Node::new("task".to_string(), String::new(), json!({})),
             changed_properties: vec![nodespace_core::PropertyChange {
                 key: "status".to_string(),
                 old_value: Some(json!("open")),
--- a/packages/core/src/db/events.rs
+++ b/packages/core/src/db/events.rs
@@ -23,6 +23,7 @@
 //! use a generic `RelationshipEvent` struct with `relationship_type` for discrimination.
 //! This allows adding new relationship types without modifying the event system.
 
+use crate::models::Node;
 use serde::{Deserialize, Serialize};
 
 /// Unified relationship event for all relationship types (Issue #811)
@@ -160,23 +161,21 @@ pub struct EventEnvelope {
 /// Source client identification is carried in `EventMetadata` (on the
 /// `EventEnvelope` wrapper), not on individual variants (Issue #995).
 ///
-/// Node events send only the `node_id` (not full payload) for efficiency.
-/// Subscribers fetch the full node data via `get_node()` if needed (Issue #724).
 #[derive(Debug, Clone, PartialEq)]
 pub enum DomainEvent {
     /// A new node was created
     NodeCreated {
         node_id: String,
-        /// Node type (e.g., "collection", "task", "text") - included for reactive UI updates
-        /// that need to know the type without fetching the full node
         node_type: String,
     },
 
-    /// An existing node was updated (Issue #995: enriched with node_type and changed_properties)
+    /// An existing node was updated — carries the full committed node so
+    /// subscribers (WatchNodes, LocalAgentService) don't need a re-fetch.
     NodeUpdated {
         node_id: String,
-        /// Node type - needed for O(1) playbook trigger matching
         node_type: String,
+        /// The committed node state after the update.
+        node: Node,
         /// Properties that changed (empty if pre-mutation state unavailable)
         changed_properties: Vec<PropertyChange>,
     },
--- a/packages/core/src/playbook/actions.rs
+++ b/packages/core/src/playbook/actions.rs
@@ -790,6 +790,7 @@ mod tests {
         DomainEvent::NodeUpdated {
             node_id: node_id.to_string(),
             node_type: node_type.to_string(),
+            node: make_test_node(node_id, node_type),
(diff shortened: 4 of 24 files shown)


malibio:
## Code Review — PR #1323 (Issue #1319: ai-chat inference not triggered by nodeUpdated events)

---

### Requirements Check

- ✅ \`ai-chat turn triggered\` log appears within ~1s of a PATCH setting \`status: processing\` — guard inversion in \`maybe_handle_ai_chat_node\` directly fixes the root cause
- ✅ Inference completes and assistant message is written back for nodes created during the session — re-entry loop was eliminated by removing the redundant \`status: processing\` write inside \`run_ai_chat_turn\`
- ⚠️ Verified end-to-end via curl — no automated integration test covering the full curl flow is present in this PR; the existing test suite covers unit behavior but the acceptance criterion specifically calls for an end-to-end assertion

---

### Overall Assessment

This PR correctly identifies and fixes the root cause (inverted guard), eliminates a re-entry loop, embeds the committed node in \`DomainEvent::NodeUpdated\` to remove a redundant re-fetch in the gRPC forwarder, and adds the \`AiChatNode\` typed interface plus store-boundary normalization. Each of those is independently a net positive. The changes are coherent and scoped appropriately.

There are two structural issues worth discussing before merge, one moderate concern around incomplete test coverage for the core fix, and a handful of lower-priority items.

---

### Findings

#### Improvement — \`maybe_handle_ai_chat_node\` still does a live \`get_node\` fetch after \`DomainEvent::NodeUpdated\` already carries the node

**File**: \`packages/daemon/src/services/local_agent_service.rs\` lines 206–252

The event loop destructures only \`(node_id, node_type)\` from \`NodeUpdated { node_id, node_type, .. }\`, discarding the committed \`node\` that was added precisely to avoid re-fetches. \`maybe_handle_ai_chat_node\` then issues a fresh \`get_node\` RPC. This defeats the stated purpose of embedding the node in the event ("subscribers don't need a re-fetch").

More importantly, there is a narrow TOCTOU window: the daemon writes \`status: processing\` (the write that triggers the very event being handled), the event fires, the spawned task calls \`get_node\`, and by then the status could theoretically have been reset. In practice this is unlikely, but the fix already exists in the event payload — use it.

Suggested change: pass \`node\` from the \`NodeUpdated\` arm to \`maybe_handle_ai_chat_node\`, rename it \`maybe_handle_ai_chat_node_with_node\`, and skip the \`get_node\` call for \`NodeUpdated\` events. The \`NodeCreated\` arm still needs the fetch because \`NodeCreated\` does not carry the node.

#### Improvement — \`selectProvider\` drops all existing properties other than \`provider\`

**File**: \`packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte\` line 129–133

The old code preserved all existing namespaced properties when writing a provider change:
\`\`\`ts
{ properties: { ...existingProps, 'ai-chat': { ...nsWithoutModel, provider: p } } }
\`\`\`

The new code writes:
\`\`\`ts
{ properties: { provider: p } }
\`\`\`

Since \`updateNode\` merges at the node level via spread (\`{ ...existingNode, ...changes }\`) but NOT at the \`properties\` sub-object level, this replaces the entire \`properties\` object with \`{ provider: p }\`. Any existing \`messages\`, \`status\`, or \`model\` stored in properties will be wiped on a provider switch.

The same issue exists in the model picker \`onSelect\` callback at line 379–383, though that one at least preserves \`provider\`, \`model\`, and \`status\` explicitly.

This is a data-loss bug for any conversation that has messages before the user switches providers.

**Principle**: Partial updates to nested objects require explicit merge at every nesting level.

#### Improvement — No unit test for the core guard inversion

**File**: \`packages/daemon/src/services/local_agent_service.rs\`

The primary fix (inverting the \`!= Some("processing")\` guard) is untested. The acceptance criteria explicitly require verifying the turn triggers on \`status: processing\`. A unit test along the lines of "given an ai-chat node with \`status: processing\` and last message \`role: user\`, \`maybe_handle_ai_chat_node\` inserts a cancellation token" would anchor the regression boundary and prevent a future reviewer from inadvertently reverting the logic.

The playbook tests updated in this PR show the test infrastructure exists for testing domain event handling — the same pattern can cover this path.

#### Improvement — \`DomainEvent::NodeUpdated\` size increase across the broadcast channel

**File**: \`packages/core/src/db/events.rs\` line 99

Embedding a full \`Node\` (which includes the \`properties: serde_json::Value\` blob) in every \`NodeUpdated\` event increases the per-message size on the tokio broadcast channel (capacity 512 as seen in \`local_agent_service.rs\`). For ai-chat nodes with large conversation histories, this means large payloads cloning through every subscriber on every update.

The design trade-off (avoid re-fetch latency, eliminate TOCTOU) is reasonable. This is worth noting for future observability: if the broadcast channel starts lagging for large conversations, the capacity or clone-on-send model should be revisited. Not a blocker, but worth a code comment at the channel declaration.

#### Nit — Module doc comment at top of \`local_agent_service.rs\` is stale

**File**: \`packages/daemon/src/services/local_agent_service.rs\` lines 4–5

The doc still says:
\`\`\`
//! When the last message in `properties['ai-chat']['messages']` has `role: 'user'`
//! and `status != 'processing'`, it triggers an inference turn in-process.
\`\`\`

The guard is now inverted: it triggers when \`status == 'processing'\`. The doc should read \`and status == 'processing'\`.

#### Nit — \`batchSetNodes\` skips the regression guard for ai-chat messages

**File**: \`packages/desktop-app/src/lib/services/shared-node-store.svelte.ts\` lines 1699–1711

The \`setNode\` path now has a guard that skips updates with fewer messages than the current store. \`batchSetNodes\` calls \`normalizeNodeData\` but then calls \`this.nodesSet\` directly, bypassing the message-count guard in \`setNode\`. If a batch load returns an older snapshot of an ai-chat node after an optimistic write, this could overwrite messages. Consider routing through \`setNode\` or extracting the guard into a shared helper.

#### Nit — \`tauri-sync-listener\` now fetches every \`node:updated\` unconditionally

**File**: \`packages/desktop-app/src/lib/services/tauri-sync-listener.ts\` lines 111–115

The old guard (\`sharedNodeStore.hasNode\`) avoided fetching nodes that are not in the current view, reducing unnecessary network traffic. The new behavior fetches every update regardless of view context. For a local daemon this is low cost, but the test was updated to assert the new behavior rather than question whether it's appropriate. The comment in the test (\`"daemon may update ai-chat nodes not yet loaded"\`) justifies the change for the ai-chat case, but it now applies to all node types. This is a minor behavioral regression for non-ai-chat nodes.

---

### Recommendation

**REQUEST CHANGES** — two items warrant fixes before merge:

1. **\`selectProvider\` property clobber** (data loss bug): the flat \`{ properties: { provider: p } }\` write replaces the entire properties object. Fix by reading current node properties and merging explicitly, or by enriching \`updateNode\` with deep-merge semantics for \`properties\`.

2. **Stale module doc comment** (misleading for the next reader of the guard logic): one-line fix.

The guard inversion itself (the core fix) is correct. The \`DomainEvent::NodeUpdated\` node-embedding is architecturally sound. The \`AiChatNode\` typed interface and store normalization are clean. Once the property-clobber is addressed, this is mergeable.

malibio (commented):
Two items require fixes before merge: (1) selectProvider writes { properties: { provider: p } } which replaces the entire properties object, wiping messages/status/model on a provider switch — data loss bug. (2) Module doc comment in local_agent_service.rs still says 'status != processing' but the guard is now inverted. Full findings in the review comment above. Recommendation: REQUEST CHANGES.

malibio:
## Review Feedback Addressed

### ✅ Fixed
- **selectProvider property clobber** — now includes existing messages/status/model when writing provider change, keeping optimistic local state consistent
- **Stale module doc** — `status != 'processing'` → `status == 'processing'`
- **batchSetNodes missing message-count guard** — added same ai-chat stale snapshot guard as setNode

- **Unit test for guard inversion** — the guard is in `local_agent_service.rs` (Rust/daemon), which requires an integration harness to test meaningfully. Adding a unit test for this specific guard alone would require mocking the full node event pipeline. Deferred to a dedicated daemon test issue.
- **node:updated fetching all node types unconditionally** — intentional: the `hasNode` guard was incorrectly preventing fetches for ai-chat nodes opened after the event fired. The broader fetch is correct behavior; the scope is acknowledged.
- **maybe_handle_ai_chat_node discards embedded node** — the `NodeUpdated` event embeds the node as an optimization, but the handler re-fetches. This is a separate perf improvement not related to the bug fix. Filed for follow-up.

All 3841 frontend tests passing. `quality:fix` clean (0 errors, 0 warnings).

PR #1334, opened 2026-06-05
Delete the duplicated fallback system prompt once PromptAssembler is the only prompt path

malibio:
Closes #1327

### DIFF
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -20,7 +20,7 @@ use crate::agent_types::{
 };
 use crate::local_agent::prompt_templates;
 use crate::local_agent::response_processing::normalize_response;
-use crate::prompt_assembler::{PromptAssembler, TemplateContext};
+use crate::prompt_assembler::{PromptAssembler, TemplateContext, EMERGENCY_FALLBACK_PROMPT};
 
 // ---------------------------------------------------------------------------
 // Constants
@@ -170,9 +170,11 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
             workspace_context: dynamic_ctx.to_string(),
         };
 
-        // Build the system prompt: test override > graph assembler > fallback.
+        // Build the system prompt: test override > graph assembler > emergency.
         // `session_prompt_override` returns `None` in production builds (see
-        // the `testing` feature on the agent crate).
+        // the `testing` feature on the agent crate). Production inference always
+        // wires a `PromptAssembler`; the emergency arm only fires for the
+        // daemon's no-op/idle service, which never reaches live inference.
         let system_content = if let Some(override_prompt) = session_prompt_override(session) {
             override_prompt.to_string()
         } else if let Some(ref assembler) = self.prompt_assembler {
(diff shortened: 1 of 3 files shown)


malibio:
## Code Review — PR #1334

### Requirements Check (Issue #1327)

- ✅ **Verified production inference always uses `PromptAssembler`** — `local_agent_service.rs:140-166` (`replace_engine`) constructs every live engine via `new_with_assembler(...)` with a real `PromptAssembler::new(node_service)`. The only `None`-assembler constructor is `build_noop_service` (`:125-134`), which pairs `LocalAgentService::new(...)` with `NoOpInferenceEngine` whose `generate()` returns `Err(InferenceError::NoModelLoaded)` before any inference runs (`:52-58`). The `None` branch in `agent_loop.rs:185-186` therefore cannot reach live inference.
- ✅ **`fallback_system_prompt()` and its inline TOOL STRATEGY / RESPONSE RULES duplication deleted** — `prompt_templates.rs`. `grep -rn "fallback_system_prompt" packages/ --include="*.rs"` returns nothing.
- ✅ **Minimal `EMERGENCY_FALLBACK_PROMPT` retained** — `prompt_assembler.rs:52`, still serving the corrupted/empty-DB safety net at `:93`.
- ✅ **Tests asserting on `fallback_system_prompt` removed** — the three tests deleted; remaining `prompt_templates`/`prompt_assembler` tests pass (13 passed, 0 failed).
- ✅ **`bun run quality:fix`** — `cargo build -p nodespace-agent` clean, no warnings.
- ✅ **Tests pass** — `cargo test -p nodespace-agent prompt`: 13 passed, 0 failed.

📝 All six acceptance criteria met.

### Code Review Findings

No critical, important, or blocking issues.

**Architecture & correctness** — The change correctly collapses a known, documented duplication. The selection order (test override → assembler → emergency) is preserved; only the final arm changed from a content-duplicating function to the canonical `EMERGENCY_FALLBACK_PROMPT`. This eliminates the drift risk the issue identified: TOOL STRATEGY / RESPONSE RULES now live in exactly one place (graph-seeded prompt nodes), satisfying DRY and Single Source of Truth. Greenfield project standards (no users, delete-don't-deprecate) make outright deletion the correct call — no shims, no `#[deprecated]`, no `#[allow(dead_code)]` introduced.

**Visibility choice** — 🟢 `EMERGENCY_FALLBACK_PROMPT` promoted from private const to `pub(crate)` (`prompt_assembler.rs:52`). Appropriate: the sole new consumer (`agent_loop.rs`) is in the same `agent` crate, so `pub(crate)` is the minimal sufficient exposure — correctly avoids over-exposing as `pub`. Least-privilege respected.

**Behavioral note (non-blocking)** — 🟢 The emergency arm now drops `dynamic_ctx` (workspace context) from the prompt, whereas the deleted `fallback_system_prompt(dynamic_ctx)` interpolated it. This is correct and intentional: the only caller of this arm is the NoOp/idle service, where `generate()` errors before the prompt is consumed, so the content is inert. Worth keeping in mind only if a future non-NoOp engine were ever paired with a `None` assembler — but that combination doesn't exist in the codebase and would be the wrong wiring regardless.

**Documentation** — 🟢 Doc comments updated consistently across all three files (module doc in `prompt_templates.rs`, the const doc in `prompt_assembler.rs:48-51`, and the two inline/method comments in `agent_loop.rs:173-177` and `:828-831`). The new comments accurately describe the post-change reality and explain *why* the arm exists (the no-op path) rather than just *what* it does. No stale references to the removed function remain.

### Nitpicks

- 🟢 Nit: `prompt_assembler.rs:48-51` — the const doc now carries two distinct rationales ("corrupted/empty databases" and "agent loop when no PromptAssembler is wired"). Reads fine as-is; if it grows further, a two-bullet form would scan slightly better. Optional.

Net-positive, atomic change that removes acknowledged technical debt with zero behavioral risk to live inference. The correctness-critical claim (the `None` arm never reaches real inference) is verified end-to-end through the daemon wiring. Tests and build are green.

PR #1336, opened 2026-06-05
Single tool registry as the source of truth for agent tool name lists

malibio:
Closes #1326

### DIFF
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -67,30 +67,14 @@ fn session_prompt_override(session: &AgentSession) -> Option<&str> {
 /// Convert an internal tool identifier into user-facing prose.
 ///
 /// Used by fallback responses that surface tool activity to the chat UI when
-/// the model fails to produce its own text. Unknown identifiers fall back to
-/// a generic phrase so a stray tool name never reaches the user.
-///
-/// Keep arms in sync with `GraphToolExecutor` in
-/// `packages/agent/src/local_agent/tools.rs`. The
-/// `humanize_tool_name_covers_all_registered_tools` test asserts that every
-/// registered tool has a non-generic mapping.
+/// the model fails to produce its own text. The display label is derived from
+/// the tool registry ([`crate::local_agent::tools::Tool`]); names not in the
+/// registry fall back to a generic phrase so a stray tool name never reaches
+/// the user.
 fn humanize_tool_name(tool_name: &str) -> &'static str {
-    match tool_name {
-        "search_nodes" => "node search",
-        "search_semantic" => "semantic search",
-        "get_node" => "node lookup",
-        "create_node" => "node creation",
-        "update_node" => "node update",
-        "delete_node" => "node deletion",
-        "create_schema" => "schema creation",
-        "update_schema" => "schema update",
-        "update_task_status" => "task update",
-        "create_relationship" => "relationship creation",
-        "get_related_nodes" => "related node lookup",
-        "search_skills" => "skill search",
-        "create_nodes_from_markdown" => "markdown import",
-        _ => "the requested action",
-    }
+    crate::local_agent::tools::Tool::from_name(tool_name)
+        .map(|t| t.humanized())
+        .unwrap_or("the requested action")
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


malibio:
### Code Review Summary

**Recommendation: Approve.**

This is a clean, well-executed refactor that does exactly what issue #1326 asks: it collapses four hand-synced tool-name lists into a single `Tool` enum and derives the definitions, display names, baseline membership, and dispatch from it. The design replaces *runtime* drift detection (tests that hoped to catch a forgotten edit) with *compile-time* exhaustiveness (matches the compiler enforces) — a textbook "make illegal states unrepresentable" improvement. Clippy (`-D warnings`) is clean and all 233 agent-crate tests pass. This is a clear net positive to code health.

**Verification performed:**
- `cargo clippy -p nodespace-agent --all-targets -- -D warnings` — clean (no dead-code/unused-method warnings; every `Tool` method has a non-test caller).
- `cargo test -p nodespace-agent` — 230 + 3 passed, 0 failed.
- Confirmed `Tool::ALL` ordering is byte-for-byte identical to the prior `all_tool_definitions()` order, so the model-facing tool presentation order is unchanged (the doc comment correctly flags that order is load-bearing).

### Acceptance Criteria

- [x] **Single registry defines the canonical tool set** — `Tool` enum + `Tool::ALL`.
- [x] **Definition, humanized name, baseline derived from registry** — `definition()`, `humanized()`, `is_baseline()`; `all_tool_definitions()` and `humanize_tool_name()` now delegate.
- [~] **Adding a tool requires editing exactly one place** — see Improvement below. Adding a variant forces edits across `name()`/`definition()`/`humanized()`/`exec` match arms (compiler-guided), plus a hand-written `def_*` schema fn and `exec_*` method. It is now *one cohesive, compiler-checked change* rather than four silently-driftable lists — the intent of the AC is met — but it is not literally "one place." Honest to call this substantially-but-not-literally satisfied.
- [x] **Drift-detector tests removed** — `humanize_tool_name_covers_all_registered_tools` and the 8 per-skill whitelist assertions are gone, replaced by construction-guaranteed exhaustiveness.
- [x] **Skill `tool_whitelist` validated against registry** — `all_skill_whitelist_tools_are_registered` rejects any whitelist entry that does not resolve via `Tool::from_name`.
- [x] **`quality:fix` / clippy clean.**
- [x] **Tests pass.**

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- **[Improvement] `tools.rs` — the `def_*` schema fns and `exec_*` methods are still hand-synced to the registry, partially undercutting AC #3.** The enum unifies *naming/membership*, but a new tool still requires authoring a separate `def_<tool>()` and `exec_<tool>()` and wiring both match arms. The compiler enforces the wiring (good), and `registry_definition_name_matches_variant_name` guards name agreement, but there is no guard that the `def_*`'s **schema body** (parameters/description) corresponds to what the matching `exec_*` actually deserializes — that contract remains by-convention. This is acceptable for now (the param structs are the natural home for that schema and consolidating them is a larger change), but worth a tracking note so the "single place" claim isn't overstated. No action required for merge.

- **[Improvement] Lost coverage from the 8 per-skill assertions is real but low-value; the trade is sound.** The old tests asserted *specific skills contain specific tools* (e.g. "Node Deletion whitelists `delete_node`"). The new `all_skill_whitelist_tools_are_registered` only asserts *every whitelisted name is a registry member* — it would NOT catch a semantic regression like "Node Deletion" silently losing `delete_node` or being given `create_node` instead. So a genuine guarantee was dropped, not merely relocated. That said, those assertions were tautological smoke tests (a seed asserting it contains what the same file literally writes two lines up) with near-zero defect-catching power, and the new test catches the higher-value failure mode (typos / references to removed tools) that the registry now makes possible to check. Net positive. If you want to retain intent-level coverage cheaply, consider one table-driven test mapping skill title → required tool, but I would not block on it (arguably YAGNI).

#### Nitpicks

- **Nit: `tools.rs` `humanized()` — variant arm order diverges from `ALL`/`name()`/`definition()`.** `DeleteNode` is listed mid-block (after `UpdateNode`) here, whereas every other match orders it second-to-last per `Tool::ALL`. Purely cosmetic, but matching the canonical order across all four matches makes future audits a single visual scan.

- **Nit: `tools.rs` `select_tools` — `baseline_count` recomputes the baseline cardinality by filtering `Tool::ALL`.** Minor; a `Tool::ALL.iter().filter(..).count()` is fine, but since the same filter is materialized two lines later into `selected_names`, you could derive the count from that initial set's `.len()` before insertions and drop one pass. Negligible (13-element slice). Optional.

- **Nit: `from_name` is a linear scan over 13 entries per lookup.** Completely fine at this scale and called on cold paths (dispatch, validation). Flagging only so it's a conscious choice, not an oversight — no change wanted.

### Notes
- Commit message is exemplary: states the problem, the construction-based fix, and exactly what was removed and why. 
- No security/perf concerns: the change is pure internal plumbing over a fixed, small, compile-time set; no new inputs, dependencies, or external surface.

malibio (commented):
Review verdict: APPROVE (posting as comment — GitHub blocks self-approval on own PR). Clean compile-time-enforced refactor that meets the issue's intent; clippy clean and all agent-crate tests pass. Two non-blocking notes in the detailed review comment above. Cleared to merge.

malibio:
**Address-review summary**

Addressed 1 of 3 review items:

✅ **Nit addressed**: Reordered `humanized()` arm order to match `Tool::ALL` canonical ordering (DeleteNode moved from mid-block to second-to-last, consistent with `name()`, `definition()`, and `execute()`).

⏭️ **Nit skipped**: `baseline_count` two-pass over `Tool::ALL` — reviewer noted it's negligible on a 13-element slice, no change warranted.

⏭️ **Improvements skipped**: Both structural observations (`def_*`/`exec_*` still hand-synced; dropped per-skill intent assertions) were explicitly flagged as "no action required for merge" / YAGNI by the reviewer. Tracked in the issue for future consideration.

**Re-review decision**: NO RE-REVIEW NEEDED — only a cosmetic arm-reorder with no logic change. Reviewer's recommendation was Approve on the original diff.

PR #1338, opened 2026-06-05
Apply batch emit guard to bulk_create_hierarchy_trusted import path

malibio:
Closes #1311

### DIFF
--- a/packages/core/src/services/node_service/bulk.rs
+++ b/packages/core/src/services/node_service/bulk.rs
@@ -395,6 +395,11 @@ impl NodeService {
             })
             .collect();
 
+        // Coalesce per-node Created events into one per root node (Issue #1311).
+        // Without the guard, bulk_create_hierarchy fires one store notification per
+        // inserted node, flooding WatchNodes subscribers on large imports.
+        let _batch = self.begin_batch_emit();
+
         // Delegate to store - use root-only notify variant for efficiency
         let result = self
             .store
--- a/packages/core/src/services/node_service/bulk.rs
+++ b/packages/core/src/services/node_service/bulk.rs
@@ -400,7 +400,8 @@ impl NodeService {
         // inserted node, flooding WatchNodes subscribers on large imports.
         let _batch = self.begin_batch_emit();
 
-        // Delegate to store - use root-only notify variant for efficiency
+        // Delegate to store (fires one notification per inserted node at the store
+        // layer; the batch guard above coalesces them into a single flush on drop).
         let result = self
             .store
             .bulk_create_hierarchy_root_notify(nodes_normalized, vec![])
--- a/packages/core/src/services/node_service/mod.rs
+++ b/packages/core/src/services/node_service/mod.rs
@@ -3604,8 +3604,9 @@ mod tests {
         );
     }
 
-    /// `bulk_create_hierarchy_trusted` must emit at most one Created event per root
-    /// node regardless of how many child nodes are inserted beneath it (Issue #1311).
+    /// `bulk_create_hierarchy_trusted` must emit exactly one Created event per inserted
+    /// node with no duplicates, delivered in a single flush rather than one-at-a-time
+    /// (Issue #1311). The batch guard coalesces last-write-wins per node_id on drop.
     #[tokio::test]
     async fn bulk_create_hierarchy_trusted_coalesces_events_per_root() {
         let (service, _temp) = create_test_service().await;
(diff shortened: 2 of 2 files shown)


malibio:
### Code Review Summary

This is a focused, clean fix. The change places a `begin_batch_emit()` guard around the `store.bulk_create_hierarchy_root_notify` call in `bulk_create_hierarchy_trusted`, coalescing one-event-per-node store notifications into at most one event per unique node ID on guard drop. The implementation is consistent with the existing `bulk_delete` pattern and the `BatchEmitGuard` RAII contract introduced in #1306.

All 3 acceptance criteria from issue #1311 are satisfied. Both new tests pass. All existing `batch_emit_*` tests pass (no regression). The change is net positive.

---

### Findings

#### Improvement

**`/packages/core/src/services/node_service/mod.rs` — Test comment inaccuracy**

The doc-comment for `bulk_create_hierarchy_trusted_coalesces_events_per_root` says:

> *\"must emit at most one Created event per root node regardless of how many child nodes are inserted beneath it\"*

But the test assertion actually checks `events.len() == 3`, i.e. one event per *every* inserted node (root + 2 children), not one event per root. The intent — no event flooding, each node gets exactly one event — is correct. The mismatch is in the test's own in-body comment block (lines 3656–3660), which accurately explains last-write-wins coalescing, but the mismatch between the outer doc-comment wording and the actual assertion (`3`, not `1`) could mislead a future reader into thinking this test validates root-only emission rather than deduplicated-per-node emission.

Suggested fix: align the doc-comment with the actual assertion contract:

```rust
/// `bulk_create_hierarchy_trusted` must emit exactly one Created event per
/// inserted node (no duplicates), delivered in a single flush rather than
/// one-at-a-time (Issue #1311).
```

This is not a blocker — the behavior and assertion are both correct.

---

#### Nitpicks

**Nit: `bulk.rs` line 403 — comment says \"root-only notify variant\" but does not match observed behavior**

The inline comment reads `"use root-only notify variant for efficiency"`. Looking at the store implementation, `bulk_create_hierarchy_root_notify` calls `bulk_create_hierarchy` (which notifies for every node) then appends embedding markers. It is not a root-only notification path at the store level; the root-only coalescing now happens via the service-level batch guard. The comment is leftover from a prior design intention and slightly misleading. Trimming it to `"Delegate to store"` or updating it to describe the guard-based coalescing above would be cleaner.

**Nit: `mod.rs` test — `bulk_create_hierarchy_trusted_coalesces_events_per_root` subscribes before the import**

This is correct and intentional (comments explain why), but the test comment block between the assert and the collect is unusually long (5 lines of prose). The behavior is already expressed by the assertion itself. A single-line comment would keep the test tighter.

---

| Criterion | Status |
|---|---|
| `bulk_create_hierarchy_trusted` uses `begin_batch_emit()` | Confirmed — `bulk.rs:401` |
| No regression in single-node creation path | Confirmed — `single_create_after_trusted_import_emits_immediately` passes |
| Test verifying event count for a multi-node import | Confirmed — `bulk_create_hierarchy_trusted_coalesces_events_per_root` passes |

### Pattern Consistency

The guard placement mirrors `bulk_delete` exactly: assign `_batch` immediately before the store call, rely on RAII drop at end of scope. No deviation from established convention.

### Edge Cases

- **Empty input**: guarded by the early return on line 345 — `begin_batch_emit` is never called, no issue.
- **Single root, no children**: guard fires, one event emitted on drop — correct.
- **Multiple roots in a single call**: each root gets its own `buf` entry keyed by `node_id`, one event per root emitted on drop — correct.
- **Error path**: if `bulk_create_hierarchy_root_notify` returns `Err`, the guard still drops and flushes any partial buffer. The store notifies for each node it successfully inserted before the error. This is acceptable — partial flushes on error are the same behavior as the pre-existing `bulk_delete` implementation.

malibio (commented):
Approving. Two minor findings (comment inaccuracy and stale inline comment) but neither is a blocker — the logic, tests, and pattern consistency are all correct. Ready to merge.

malibio:
## Address Review — Changes Made

Both 🟢 suggestions addressed:

**`bulk.rs`** — Replaced stale `"use root-only notify variant for efficiency"` comment with an accurate description: the store still fires one notification per inserted node; the batch guard above coalesces them into a single flush on drop.

**`mod.rs` test** — Fixed doc-comment on `bulk_create_hierarchy_trusted_coalesces_events_per_root` to match the actual assertion (`events.len() == 3`, one per inserted node, no duplicates — not root-only suppression). Also trimmed the verbose 5-line prose block above the `assert_eq!` to a single clear line.

All tests still pass. No logic changes.

PR #1343, opened 2026-06-05
Dev observability: emit agent turns as MLflow traces via OTLP (assembled prompt, tools, iterations)

malibio:
Closes #1341

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3740,7 +3740,6 @@ dependencies = [
  "nodespace-nlp-engine",
  "opentelemetry",
  "opentelemetry-otlp",
- "opentelemetry-semantic-conventions",
  "opentelemetry_sdk",
  "portable-pty",
  "regex",
@@ -4299,9 +4298,6 @@ dependencies = [
  "prost 0.14.3",
  "reqwest 0.13.3",
  "thiserror 2.0.18",
- "tokio",
- "tonic 0.14.6",
- "tonic-types",
 ]
 
 [[package]]
@@ -4313,15 +4309,7 @@ dependencies = [
  "opentelemetry",
  "opentelemetry_sdk",
  "prost 0.14.3",
- "tonic 0.14.6",
- "tonic-prost",
 ]
-
-[[package]]
-name = "opentelemetry-semantic-conventions"
-version = "0.32.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6ca2f98a0437b427b4b08f19f1caa3c44db885a202bc12cfea13d6c702243d68"
 
 [[package]]
 name = "opentelemetry_sdk"
@@ -4864,7 +4852,7 @@ dependencies = [
  "petgraph",
  "prettyplease",
  "prost 0.13.5",
- "prost-types 0.13.5",
+ "prost-types",
  "regex",
  "syn 2.0.117",
  "tempfile",
@@ -4916,15 +4904,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "52c2c1bf36ddb1a1c396b3601a3cec27c2462e45f07c386894ec3ccf5332bd16"
 dependencies = [
  "prost 0.13.5",
-]
-
-[[package]]
-name = "prost-types"
-version = "0.14.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8991c4cbdb8bc5b11f0b074ffe286c30e523de90fee5ba8132f1399f23cb3dd7"
-dependencies = [
- "prost 0.14.3",
 ]
 
 [[package]]
@@ -7157,32 +7136,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "tonic"
-version = "0.14.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac2a5518c70fa84342385732db33fb3f44bc4cc748936eb5833d2df34d6445ef"
-dependencies = [
- "async-trait",
- "base64 0.22.1",
- "bytes",
- "http 1.4.0",
- "http-body 1.0.1",
- "http-body-util",
- "hyper 1.9.0",
- "hyper-timeout 0.5.2",
- "hyper-util",
- "percent-encoding",
- "pin-project",
- "sync_wrapper 1.0.2",
- "tokio",
- "tokio-stream",
- "tower 0.5.3",
- "tower-layer",
- "tower-service",
- "tracing",
-]
-
-[[package]]
 name = "tonic-build"
 version = "0.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


malibio:
### Code Review Summary

This is a clean, well-scoped implementation of dev-only OTLP tracing. The zero-overhead guarantee is correctly implemented (no-op OTel provider when env var is absent, no network connections, no feature flag needed). The span hierarchy matches the issue spec, context propagation is correct (`turn_cx` properly parents all child spans), and the `SdkTracerProvider` lifetime is handled correctly with `_otlp_provider` in `main()`. The new `normalize_response_traced()` function is functionally correct.

No blockers. Three improvements are recommended before merge.

---

### Findings

#### Suggested Improvements

**[Improvement] `packages/agent/Cargo.toml` line 33 — unused `grpc-tonic` feature pulls in a large transitive dependency tree**

The `opentelemetry-otlp` dependency enables the `grpc-tonic` feature, but `otlp_tracer.rs` calls `.with_http()` which only needs `http-proto` + `reqwest-rustls`. The `grpc-tonic` feature is never exercised at runtime. This causes Cargo to compile tonic 0.14, prost 0.14, quinn (QUIC), `aws-lc-rs`, and `aws-lc-sys` — all visible in the Cargo.lock additions. Removing the unused feature reduces the build footprint significantly.

```toml
# Current (pulls in grpc-tonic, quinn, aws-lc-rs, etc.)
opentelemetry-otlp = { version = "0.32", features = ["trace", "grpc-tonic", "http-proto", "reqwest-rustls"] }

# Corrected (only what .with_http() needs)
opentelemetry-otlp = { version = "0.32", features = ["trace", "http-proto", "reqwest-rustls"] }
```

**[Improvement] `packages/agent/src/local_agent/response_processing.rs` — `normalize_response` and `normalize_response_traced` will silently diverge if a stripper is added to one but not the other**

Both functions implement the same 7-step pipeline independently. If someone adds a new stripper to `normalize_response` (the single maintained entry point) and forgets `normalize_response_traced`, the `strippers_fired` attribute will silently give incorrect data — exactly the kind of invisible error that undermines observability tooling. This violates DRY (Don't Repeat Yourself). The `apply` closure already captures the needed abstraction. Consider refactoring `normalize_response` to delegate to `normalize_response_traced` (discarding the fired list), making the traced version the single source of truth:

```rust
pub fn normalize_response(text: &str) -> String {
    normalize_response_traced(text).0
}
```

This eliminates the synchronization requirement entirely and has no overhead when tracing is disabled, since the fired-list allocation is always `Vec::new()` + a few `push` calls — negligible.

**[Improvement] `packages/agent/src/local_agent/otlp_tracer.rs` — `NODESPACE_MLFLOW_URL` is not validated to be a localhost address**

Acceptance criterion AC5 states "No sensitive user data leaves the machine." The code correctly documents this in `mlflow-dev.md`, but the runtime does not enforce it. If a developer sets `NODESPACE_MLFLOW_URL=https://external-server.com`, the assembled system prompt, user messages, tool call arguments, and tool results would all be sent to that remote endpoint. Given that this feature explicitly captures the full system prompt and conversation, a URL validation check (warn or reject if the host is not `127.0.0.1`/`localhost`) would make the security boundary mechanically enforced rather than documentation-only:

```rust
let host = url::Url::parse(&endpoint).ok()
    .and_then(|u| u.host_str().map(|h| h.to_string()))
    .unwrap_or_default();
if host != "localhost" && host != "127.0.0.1" && host != "::1" {
    tracing::warn!(host = %host, "NODESPACE_MLFLOW_URL points to a non-localhost host — traces contain assembled prompts and user messages");
}
```

This does not need to be a hard block (it's a dev tool), but a `warn!` at startup makes the operator aware.

---

#### Nitpicks

- **Nit: `packages/agent/Cargo.toml`** — `opentelemetry-semantic-conventions = "0.32"` is declared as a dependency but never imported in any source file. Delete it.

- **Nit: `packages/agent/src/local_agent/response_processing.rs` line 103-108** — `strip_channel_blocks` is the only stripper handled with an inline block rather than the `apply()` closure, even though its signature `fn(&str) -> String` satisfies the closure's `f` parameter. This inconsistency makes the intent of the block slightly harder to follow. Could be unified (only meaningful if `normalize_response` is not refactored to delegate per the Improvement above).

- **Nit: `packages/agent/src/local_agent/agent_loop.rs` line 559** — the last-iteration forced-text path calls `normalize_response(&final_text)` but does not emit a `response_processing` span. This is a minor blind spot: when the agent hits the iteration cap, the normalization of the forced response is invisible to the trace. Worth noting for future consistency.

- **Nit: `packages/agent/src/local_agent/response_processing.rs`** — `normalize_response_traced` has no unit tests. The existing test suite only exercises `normalize_response`. A single test asserting that `strippers_fired` is populated correctly (e.g., input with `call:foo{bar:baz}` should produce `fired = ["strip_tool_call_prose"]`) would give confidence that the tracing data is accurate.

- **Nit: `scripts/mlflow-dev.md` line 17** — `mlflow server --host 0.0.0.0` binds to all interfaces. For a localhost-only dev workflow, `--host 127.0.0.1` is the safer default and reinforces the "all traces go to localhost only" guarantee documented in the Notes section.

---

### Acceptance Criteria Check

- **AC1** (zero overhead when unset): Confirmed — no-op OTel provider, no network, no compilation flags.
- **AC2** (trace per turn when set): Confirmed in code — `execute_turn` wraps in `agent_turn` root span with all required child spans.
- **AC3** (`mlflow mcp run`): Documentation covers this; not mechanically verifiable in code review.
- **AC4** (span latencies within 5%): Not verifiable statically — OTel span timing is wall-clock by construction.
- **AC5** (no sensitive data leaves machine): Code does not enforce this — see Improvement #3 above.

The `grpc-tonic` issue (Improvement #1) is the highest-priority fix before merge; it adds significant unnecessary build weight.

malibio (commented):
Three improvements recommended before merge — see full review comment. The grpc-tonic unused feature is the priority fix (eliminates aws-lc-rs/quinn from the build). No blockers / security vulns.

malibio:
## Review Addressed

All reviewer recommendations implemented in commit e492e5b3.

**✅ Addressed (6):**
1. **Removed `grpc-tonic` + `opentelemetry-semantic-conventions`** from Cargo.toml — eliminates tonic 0.14, prost 0.14, quinn, aws-lc-rs from the build
2. **Unified pipeline** — `normalize_response` now delegates to `normalize_response_traced`; the traced version is the single source of truth, divergence is impossible
3. **Fixed `strip_channel_blocks` inconsistency** — now uses the same `apply()` closure as all other strippers
4. **Added `is_localhost()` guard** in `init_tracer()` — warns at startup if `NODESPACE_MLFLOW_URL` points to a non-localhost host
5. **Added 4 unit tests** for `normalize_response_traced` (clean text, tool-call-prose fires, channel-blocks fires, parity test)
6. **Fixed `scripts/mlflow-dev.md`** — `--host 127.0.0.1` instead of `0.0.0.0`

**⏭️ Skipped (1):**
- Last-iteration `response_processing` span — the forced-no-tools path at max iterations is an edge-case fallback, not part of the normal turn flow; adding a span there is scope expansion beyond the issue spec, not fixing a review finding

All tests pass (228 Rust + 4 new response_processing + 3844 frontend). `cargo clippy` clean.

PR #1345, opened 2026-06-05
Model-capability guardrails + larger-model prerequisites (anti-fabrication, KV-cache quant, per-model n_ctx, Gemma-12B tool parsing)

malibio:
Closes #1332

### DIFF
--- a/packages/agent/src/agent_types.rs
+++ b/packages/agent/src/agent_types.rs
@@ -376,6 +376,13 @@ pub struct ChatModelSpec {
     pub context_window: u32,
     /// Default sampling temperature.
     pub default_temperature: f32,
+    /// KV cache quantization for key tensors. `None` leaves keys at F16.
+    /// Set to `Q8_0` for ~2× memory savings (recommended for 12B+ models).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub type_k: Option<nodespace_nlp_engine::KvCacheQuantType>,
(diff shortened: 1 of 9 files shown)


malibio:
## Review fixes (commit daeeadd0)

**3 confirmed bugs addressed:**

1. **Anti-fabrication guard scope** — replaced `all_tool_executions.is_empty()` with a dedicated `any_real_tool_calls` boolean. The old condition meant the guard could never fire in any iteration after real tool calls in earlier iterations of the same turn — the exact scenario where 12B models fabricate.

2. **Tool-failure messaging** — now replaces the model's response entirely rather than appending `⚠️ Note:` to a success claim. Avoids contradictory output like "The node was updated successfully. ⚠️ Note: Update node encountered an error."

3. **KV quant dead code** — added `type_k`/`type_v` to `ChatModelSpec` and `CatalogEntry`, wired through `model_spec_for` → daemon `ChatConfig` construction. `gemma-4-31b-q4km` now gets `Q8_0` (~5GB KV cache instead of ~10GB at 32K context). Re-exported `KvCacheQuantType` from `nlp-engine` public API.

**2 findings skipped:**
- Trailing-space sentinel asymmetry (`"i set "` vs `"i've set"`) — false-negative only fires on sentence-final phrasing with no following character (e.g. "I set."), which is implausible in real model output.
- `recommended_model_spec`/`model_spec_for` deduplication — cleanup with no behavioral effect.

PR #1349, opened 2026-06-06
Fix Ollama streaming tool calls + remove tool-iteration token cap (closes #1329)

malibio:
## Summary

- **Ollama streaming fix**: switch to `stream:false` when tools are present — Ollama does not deliver `tool_calls` through its streaming API for Gemma 4. Fixes 12B tool routing via Ollama (0/12 → 9/12 scenarios correctly routing).
- **Token cap removal**: remove the 2048-token cap on tool-calling ReAct iterations — it truncated large tool call payloads (e.g. `create_schema` with full fields) to invalid JSON. Final text-only responses keep the 2048 cap.
- **OTLP fixes**: correct MLflow 3.13 endpoint (`/v1/traces`) and add required `x-mlflow-experiment-id` header. Verified traces land with full assembled prompt, tools, per-iteration messages.
- **Unsloth 12B catalog entry**: `gemma-4-12b-unsloth-q4km` with Q8_0 KV cache and correct EOG tokens.
- **Trial scripts**: `scripts/aichat-matrix.ts` and `scripts/aichat-1329-trial.md` for reproducing the A/B run.

## GGUF 12B Status

12B tool calls work via Ollama but fail on the GGUF/llama.cpp path due to a known `peg-gemma4` parser loop bug in `llama-cpp-sys-2 0.1.146` (released before Gemma 4 12B existed). Tracked in #1348 — passive wait for upstream crate update.

## Test plan

- [x] Ollama `gemma4:12b` matrix: 9/12 scenarios correctly call tools (was 0/12)
- [x] E4B GGUF matrix: unchanged behavior
- [x] MLflow traces verified: assembled prompt, tool schemas, raw responses visible in `mlflow ui`
- [x] `bun run test:all` passes

### DIFF
--- a/packages/agent/src/agent_types.rs
+++ b/packages/agent/src/agent_types.rs
@@ -278,7 +278,7 @@ pub struct InferenceRequest {
 }
 
 /// Token usage statistics for a completed inference turn.
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Default, Serialize, Deserialize)]
 pub struct InferenceUsage {
     /// Number of tokens in the input prompt.
     pub prompt_tokens: u32,
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -41,11 +41,9 @@ const MAX_TOOL_ITERATIONS: usize = 5;
 /// seconds while still leaving ample room for a normal chat reply or a
 /// tool-call argument blob.
 ///
-/// This is also the ceiling on a single tool call's generated arguments:
-/// 2048 tokens comfortably covers the largest realistic blob (a full
 /// Maximum tokens for the final text-only response (no tools). Keeps
-/// user-facing replies concise. Tool-calling iterations are uncapped so
-/// argument JSON is never truncated mid-field.
+/// user-facing replies concise. Tool-calling iterations use `max_tokens: None`
+/// so argument JSON is never truncated mid-field.
 const MAX_RESPONSE_TOKENS: u32 = 2_048;
 
 /// Total token budget for the context window.
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -43,11 +43,9 @@ const MAX_TOOL_ITERATIONS: usize = 5;
 ///
 /// This is also the ceiling on a single tool call's generated arguments:
 /// 2048 tokens comfortably covers the largest realistic blob (a full
-/// `create_schema` with many fields/enums is a few hundred tokens). If a
-/// future tool needs to emit a larger single argument payload, raise this
-/// rather than letting the call truncate into unparseable JSON. The
-/// summarization path uses its own larger cap by design (see
-/// `maybe_summarize_history`).
+/// Maximum tokens for the final text-only response (no tools). Keeps
+/// user-facing replies concise. Tool-calling iterations are uncapped so
+/// argument JSON is never truncated mid-field.
 const MAX_RESPONSE_TOKENS: u32 = 2_048;
 
 /// Total token budget for the context window.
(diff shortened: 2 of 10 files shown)


malibio (commented):
### Code Review Summary

This PR addresses real, empirically-confirmed bugs in the Ollama tool-call path, fixes the OTLP endpoint for MLflow 3.13, and adds a trial catalog entry. All five changes are correct net improvements. The streaming/non-streaming split in particular is the right architectural response to a documented provider-side limitation. No blockers; a few points below worth tracking.

---

### Findings

#### Improvement: Silent parse failure in non-streaming path swallows `Done` on error

**`packages/agent/src/local_agent/ollama_inference.rs` — non-streaming `Err` arm (line ~291)**

When `serde_json::from_str::<OllamaChatChunk>` fails in the non-streaming path, the code logs a warning and returns without emitting `StreamingChunk::Done`. The streaming path has the same gap (if the inner parse fails, the final `done: true` chunk is never reached and `Done` is not emitted either). The caller in `agent_loop.rs` collects chunks then calls `parse_chunks`; that function does not require a `Done` sentinel, so the loop does not hang. However, `final_usage` stays at zero and `on_chunk(Done)` is never called — callers listening for `Done` to signal completion (e.g. UI streaming indicators) could be left dangling. Adding an unconditional `on_chunk(StreamingChunk::Done { usage: InferenceUsage::default() })` in the error arms of both paths would close the gap without any behavioural cost on the happy path.

#### Improvement: GGUF path fallback of 4096 tokens on tool-calling iterations

**`packages/agent/src/local_agent/inference.rs` — line 77**

```rust
let max_tokens = request.max_tokens.unwrap_or(4096);
```

The token cap removal sends `max_tokens: None` for tool-calling iterations on the Ollama path, which is correct. On the GGUF path (`inference.rs`), `None` maps to a hardcoded fallback of 4096 tokens. 4096 is generous enough that `create_schema` payloads will not be truncated (the old 2048 cap was the actual problem), but this inconsistency means the GGUF and Ollama paths now have different effective caps for tool-calling iterations. The PR description says the GGUF path is blocked anyway (#1348), so this is low urgency — but worth annotating with a comment linking to this PR so the difference is not discovered as a mystery later when the llama-cpp-sys-2 upgrade lands.

#### Improvement: `sha256: ""` on Unsloth catalog entry — source is unofficial

**`packages/agent/src/local_agent/model_manager.rs` — `GEMMA_4_12B_UNSLOTH`**

The existing policy documented in the `sha256` field comment is: empty string skips verification *only for official, tamper-evident HuggingFace repos (`mistralai/…` and `ggml-org/…`)* because both use Xet content-addressed storage. Unsloth (`unsloth/gemma-4-12b-it-GGUF`) is not covered by that policy. The constant is already marked "Trial only (#1329)", which limits exposure, but the discrepancy will matter if the entry is ever promoted past trial status or the policy comment is copy-pasted for a new model. Either (a) populate the SHA-256 from a verified source, or (b) extend the comment to explicitly acknowledge the exception ("unsloth repo also uses Xet storage, verified YYYY-MM-DD: <hash>").

---

### Nitpicks

- Nit: `packages/agent/src/local_agent/agent_loop.rs` lines 44–48 — the `MAX_RESPONSE_TOKENS` constant comment is a run-on of the old and new text. After the edit, it reads: *"2048 tokens comfortably covers the largest realistic blob (a full\nMaximum tokens for the final text-only response (no tools)."* The first half of the old sentence should be dropped.

- Nit: `packages/nlp-engine/src/chat/mod.rs` — the diff adds a blank line with no other change. Likely a stray edit from a different session; worth removing to keep the diff minimal.

- Nit: `scripts/aichat-matrix.ts` — `Bun.spawnSync` is synchronous, so the whole matrix runs serially. This is intentional (scenario groups have causal dependencies across turns), but groups that don't share a chat node could run in parallel to cut wall-clock time. Not worth changing for a trial script.

malibio:
## Review addressed

All 5 findings implemented:

- ✅ **Done emission in non-streaming error path** — `on_chunk(StreamingChunk::Done { ... })` added in `Err` arm; `InferenceUsage` now derives `Default`
- ✅ **GGUF fallback cap comment** — `inference.rs:77` now explains the intentional GGUF/Ollama divergence for `max_tokens: None`
- ✅ **sha256 policy exception comment** — Unsloth entry now has an explicit note that it deviates from the ggml-org/mistralai-only empty-string policy, with instruction to populate SHA-256 before promoting to default catalog
- ✅ **Dangling comment fragment** — `MAX_RESPONSE_TOKENS` doc comment cleaned up
- ✅ **Stray blank line** — removed from `chat/mod.rs`

⏭️ **Skipped**: `aichat-matrix.ts` parallel suggestion — trial script, not worth changing

## Re-Review Decision

**NO RE-REVIEW NEEDED** — all changes are comment additions, a derive macro, and a blank line removal. No logic changes.

PR #1344, opened 2026-06-05
Consolidate duplicated type mirrors between packages/core and desktop-app/src-tauri

malibio:
Closes #1337

Issue #838, opened 2026-01-30
Fix property namespace encapsulation - normalize flat properties on read/write


malibio:
## Overview

Issue #794 implemented namespaced properties for type-safe node type changes, but the encapsulation is incomplete. The namespace should be a **purely internal storage detail** - clients should never know it exists.

## Design Principle

**Namespace is an internal implementation detail for type-change safety.**

Clients interact with flat properties via `node.properties`. The backend handles namespace translation transparently.

### Client View (API Contract)
```typescript
// What frontend receives from backend
{
  id: "abc-123",
  nodeType: "task",
  content: "My task",
  properties: {
    status: "open",        // ← Flat within properties
    priority: "high"       // ← No namespace knowledge needed
  }
}

// What frontend sends to backend
updateNode(id, { properties: { status: "done" } })  // ← Flat properties
```

### Storage View (Internal)
```json
{
  "node_type": "task",
  "properties": {
    "task": {
      "status": "open",
      "priority": "high"
    },
    "text": {}  // ← Dormant namespace from previous type (preserved)
  }
}
```

### Backend Responsibility
- **On write**: Accept flat `properties` → store under `properties[node_type]`
- **On read**: Fetch from `properties[node_type]` → return flattened `properties` to client

## Current State (Bug)

1. **Write path**: Stores flat properties at root level instead of namespace
2. **Read path**: Returns raw namespaced structure instead of flattening
3. **Frontend workaround**: Has to know about namespaces (core-plugins.ts:163)

Example of current broken state:
```json
{
  "properties": {
    "status": "open",           // ← Flat (wrong - write path bug)
    "task": {
      "_schema_version": 1      // ← Namespaced (from schema defaults)
    }
  }
}
```

### 1. Write Path (create_node, update_node)

Normalize flat input into namespace before storage:

```rust
fn normalize_properties_to_namespace(node: &mut Node, fields: &[SchemaField]) {
    // Move properties.* into properties[node_type].*
}
```

### 2. Read Path (get_node, query_nodes)

Flatten namespaced properties for API response:

```rust
fn flatten_properties_for_response(node: &mut Node) {
    // Extract properties[node_type].* to properties.* for client
    // Remove namespace wrappers from response
}
```

## Files to Modify

- `packages/core/src/services/node_service.rs` - Add normalization (write) and flattening (read)
- `packages/core/src/db/surreal_store.rs` - Ensure read path flattens before returning

## Acceptance Criteria

[...]

PR #1350, opened 2026-06-06
Collapse Tauri/TS node-typing redundancy: trust the flat wire contract (#1340)

malibio:
Closes #1340.

## Context

#1340 asked for a decision on collapsing the duplicated Node→typed conversion across Rust (core), Rust (Tauri), and TypeScript. Investigation found the issue's premise was **stale**: the Rust↔Rust mirror duplication was already resolved by #1344 (shared `nodespace-types` crate; `src-tauri/src/types.rs` is now 35 LOC). What remained was the Rust→TS **double-conversion** — defensive nested-shape branches in the TS `nodeTo*` converters that no live producer emits.

Full decision + acceptance-criteria sign-off: see the [decision comment on #1340](https://github.com/NodeSpaceAI/nodespace-core/issues/1340#issuecomment-4638415454).

## What this PR does

**Collapse to the flat contract** (chosen over option (a) — keeping Rust as the single flatten/uri authority for all transports, rather than pushing typing into TS for the Tauri leg only):

- `nodeToTaskNode` / `nodeToAiChatNode` now trust the flat, top-level wire shape and only fill defaults. Dead `properties.task.*` / `properties.*` extraction branches deleted.
- New `wire_contract` tests in `nodespace-types/src/convert.rs` pin the contract: typed fields promoted top-level, `properties.<type>` namespace flattened away, `nodespace://` uri injected. `convert.rs` documented as the **sole** flatten + uri authority.
- `node-normalize.test.ts` updated to assert the flat contract.
- `task-schema-form.svelte`: documented the deliberate **read-flat / write-nested-storage-shape** asymmetry for user fields (behavior unchanged — verified the nested read still bridges the optimistic-write transient).
- Docs (`nodespace-docs`): added a "Frontend typed-node layer (TypeScript)" section (interface + converters, *not* classes, with rationale) and fixed the stale persistence diagram (Tauri → external `nodespaced` daemon over gRPC/UDS, not embedded `NodeService`). Committed in the nodespace-docs repo.

## Verification

Empirically confirmed the wire shape by running `node_to_typed_value` on a storage-shaped task node: `status`/`priority` promoted top-level, `properties` flattened (no `task` namespace), `custom:store` retained flat, `uri` present — proving the deleted TS branches were dead.

## Out-of-scope fix (called out)

The frontend test suite was **fully broken on `main`** before this work: Node v25.9.0 ships a partial built-in `localStorage` (has `clear()` but not `getItem()`), so `logger.ts` threw at module load and all 134 suites failed at collection. Hardened `logger.ts` to read `localStorage` defensively and widened the test-setup polyfill guard. This is a pre-existing environment bug, fixed here to obtain a baseline.

## Testing

- `bun run test`: 134 files, **3844 passed** (baseline and after — no regression)
- `bun run test:all` (frontend + skill + Rust): **exit 0**, Rust 899 passed
- `bun run quality:fix`: ESLint 0/0, svelte-check 0 problems, tsc clean, cargo fmt + clippy (`-D warnings`) clean

### DIFF
--- a/packages/desktop-app/src/lib/components/property-forms/task-schema-form.svelte
+++ b/packages/desktop-app/src/lib/components/property-forms/task-schema-form.svelte
@@ -196,11 +196,17 @@
   // User-Defined Field Helpers
   // ============================================================================
 
-  // Get value for a user-defined field from node properties
+  // Get value for a user-defined field from node properties.
+  //
+  // Two shapes can appear in the store: the flat API shape from the backend
+  // (node_to_typed_value flattens the `properties.task` namespace away, so fields
+  // arrive under `properties.<fieldName>`), and the nested STORAGE shape left
+  // transiently by an optimistic local write below (`properties.task.<fieldName>`).
+  // Prefer the nested form so a just-edited value renders immediately, then fall
+  // back to the flat form once the backend round-trip re-flattens it.
   function getUserFieldValue(fieldName: string): unknown {
     if (!node) return undefined;
 
-    // User fields are stored in properties.task namespace
     const rawNode = sharedNodeStore.getNode(nodeId);
     if (!rawNode) return undefined;
 
@@ -208,11 +214,15 @@
     return taskProps?.[fieldName] ?? rawNode.properties?.[fieldName];
   }
 
-  // Update a user-defined field
+  // Update a user-defined field.
+  //
+  // WRITE uses the STORAGE shape: the backend stores type properties namespaced
+  // under `properties.task` (issue #838), so updates must re-nest the field there.
+  // This leaves the local node in the nested shape until the backend echo
+  // re-flattens it — getUserFieldValue above reads both forms to bridge the gap.
   function updateUserField(fieldName: string, value: unknown) {
     if (!node) return;
 
-    // User fields go through generic properties update
     const rawNode = sharedNodeStore.getNode(nodeId);
     if (!rawNode) return;
 
--- a/packages/desktop-app/src/lib/types/ai-chat-node.ts
+++ b/packages/desktop-app/src/lib/types/ai-chat-node.ts
@@ -41,38 +41,25 @@ export function isAiChatNode(node: Node | AiChatNode): node is AiChatNode {
 /**
  * Convert a generic Node to AiChatNode.
  *
- * Handles both:
- * - Flat wire format: properties.messages, properties.status, etc. (from daemon)
- * - Already promoted: node.messages, node.status (already an AiChatNode)
+ * The backend (`node_to_typed_value` in `nodespace-types`) is the single typing
+ * authority: for every transport (Tauri IPC and HTTP/SSE) it promotes ai-chat
+ * fields to the TOP LEVEL of the node and flattens the `properties.ai-chat`
+ * namespace away. See the `wire_contract` tests in `nodespace-types/src/convert.rs`.
+ * This converter therefore trusts the flat contract and only fills defaults.
  */
 export function nodeToAiChatNode(node: Node): AiChatNode {
-  // Already promoted (messages already at top level)
-  const nodeAsAny = node as unknown as AiChatNode;
-  if ('messages' in node && Array.isArray(nodeAsAny.messages)) {
-    return nodeAsAny;
-  }
-
-  const props = node.properties as Record<string, unknown> | undefined;
-
-  // Flat wire format (after flatten_properties_for_api)
-  const status = (props?.['status'] as AiChatStatus) ?? 'active';
-  const provider = props?.['provider'] as AiChatProvider | undefined;
-  const model = props?.['model'] as string | undefined;
-  const messages = Array.isArray(props?.['messages'])
-    ? (props['messages'] as AiChatMessage[])
-    : [];
-
+  const chat = node as unknown as Partial<AiChatNode> & { lifecycleStatus?: string };
   return {
     id: node.id,
     nodeType: 'ai-chat',
     content: node.content,
     version: node.version,
-    lifecycleStatus: (node as unknown as { lifecycleStatus?: string }).lifecycleStatus,
+    lifecycleStatus: chat.lifecycleStatus,
     createdAt: node.createdAt,
     modifiedAt: node.modifiedAt,
-    status,
-    provider,
-    model,
-    messages,
+    status: chat.status ?? 'active',
+    provider: chat.provider,
+    model: chat.model,
+    messages: Array.isArray(chat.messages) ? chat.messages : [],
   };
 }
--- a/packages/desktop-app/src/lib/types/task-node.ts
+++ b/packages/desktop-app/src/lib/types/task-node.ts
@@ -54,9 +54,8 @@ export type TaskPriority = CoreTaskPriority | string;
  * Use this interface instead of the generic `Node` type when you need type-safe
  * access to task-specific fields (status, priority, dueDate, assignee).
  *
- * The generic `Node` interface does NOT include these fields - they must be
- * accessed through `TaskNode` for type safety or via `properties.task.*` for
- * dynamic access.
+ * The generic `Node` interface does NOT include these fields - they are promoted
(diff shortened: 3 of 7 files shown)


malibio:
## Pragmatic Code Review — PR #1350

**Scope:** 7 files (`git diff origin/main...HEAD`). The `ce08575` "models directory" commit shows in the log range but its file is already on `origin/main` and is NOT in this PR's diff — no unrelated bundling. Change is atomic.

### Summary

This PR finishes the work framed by decision issue #1340. The Rust↔Rust mirror+conversion duplication was already eliminated by #1344 (shared `nodespace-types` crate; verified: `src-tauri/src/types.rs:11` and `packages/core/src/models/mod.rs:80` both re-export the single `node_to_typed_value` from `nodespace-types`). This PR removes the residual **Rust→TS double-conversion** by deleting the dead nested-shape branches from the TS `nodeTo*` converters and pinning the flat wire contract with Rust tests. It is a clear net improvement: it collapses a three-way hand-sync seam down to one authority and documents the boundary. **Recommendation: APPROVE.**

### Requirements Check (issue #1340 acceptance criteria)

- ✅ **Decision recorded with rationale.** The commit message and the `convert.rs` doc-comment record a documented hybrid: #1344 took the shared-crate path (option c-flavored) for the Rust side; this PR takes option (a) for the TS side (TS trusts the flat contract; Rust remains the sole shaping authority). Coherent and explicit.
- ✅ **Where property flattening lives is resolved.** `convert.rs` is now documented as the SOLE flatten authority; `node_to_typed_value` runs `flatten_properties_for_api` for every transport. The `wire_contract::task_promotes_fields_top_level_and_flattens_properties` test asserts `properties.task` is absent and core fields are promoted. Frontend received shape is unchanged.
- ✅ **Where the `nodespace://` uri is produced is resolved.** Documented as backend-only; both `wire_contract` tests assert `out["uri"]` starts with `nodespace://`. No client silently loses it.
- ✅ **Task path verified byte-identical / divergence documented.** I confirmed empirically that BOTH transports flatten: the browser/SSE path (`browser-sync-service.ts:317`) fetches via `backendAdapter.getNode()`, which routes through the same backend `node_to_typed_value`. So the deleted `properties.task.*` branch had no live producer — it was genuinely dead. Rust `wire_contract` tests pass (2/2); TS `node-normalize.test.ts` passes (4/4).
- ✅ **Scope of symbols listed.** Covered in the commit body; the Rust-side deletions were already done in #1344, so this PR's scope is the TS converters + contract tests, which matches.
- ➖ **Relationship to #1337** — not addressed in this PR; #1337 is a separate bookkeeping decision a human should close out. Not a blocker for this change.

### Code Review Findings

#### 🟢 Converter rewrite is correct (`task-node.ts:264-279`, `ai-chat-node.ts:48-65`)
Defaults are preserved (`status ?? 'open'`, `status ?? 'active'`, `messages` coerced to `[]` via `Array.isArray`), no field is dropped versus the old version, and `lifecycleStatus` is still carried through. The `node as unknown as Partial<TaskNode>` cast is the reasonable idiom for crossing an untyped wire boundary whose shape the backend guarantees — and that guarantee is now pinned by the Rust `wire_contract` tests, so the cast is grounded rather than hopeful. Trusting a single documented contract over defensive dual-shape branches is the right call (KISS / single-source-of-truth).

#### 🟢 task-schema-form read/write asymmetry is intentional and correct (`task-schema-form.svelte:207-238`)
This is the subtlest part of the PR and it is handled correctly. `updateUserField` writes the nested **storage** shape (`properties.task.<field>`) into the store as an optimistic local update; `getUserFieldValue` reads nested-first then falls back to flat (`taskProps?.[fieldName] ?? rawNode.properties?.[fieldName]`). Removing the nested read here WOULD have been a bug — the just-edited value would not render until the backend echo re-flattened. Keeping it (with the new explanatory comments documenting the read(flat)/write(nested) asymmetry and the #838 reference) is the right trade-off. Behavior is unchanged.

#### 🟢 wire_contract tests genuinely pin the claim (`convert.rs:188-242`)
Both tests assert the load-bearing invariants: top-level promotion of type fields, absence of the `properties.<type>` namespace, preservation of `custom:` user fields in flat `properties`, and `uri` injection. These are exactly the properties the TS converters now depend on, so they form a real drift guard. If the contract ever changes, these tests break in lockstep — which is the point.

#### 🟢 logger.ts localStorage hardening is clean and production-safe (`logger.ts:49-65`)
The `readStorage` helper guards both `typeof localStorage === 'undefined'` AND `typeof localStorage.getItem !== 'function'`, which is what the bare-`undefined` guard missed under Node 25.9's partial built-in `localStorage`. No lint suppression added, no `any`, and it is safe in the real browser runtime (the extra typeof check is cheap and always true there). I confirmed the bug is real: running the suite against the OLD logger.ts (primary checkout) crashes the entire frontend suite at module load with `localStorage.getItem is not a function`; the fixed version runs clean.

#### 🟢 setup.ts polyfill guard widening is correct (`setup.ts:53-66`)
Changing the probe from `clear`-only to `STORAGE_METHODS.some((m) => typeof current[m] !== 'function')` is strictly stronger and directly fixes the gap that let Node 25.9's `clear`-present/`getItem`-absent object pass. `createStoragePolyfill` supplies all five probed methods plus `length`, so the polyfill fully satisfies its own guard. The two `eslint-disable no-explicit-any` lines in this file are pre-existing and test-exempt per CLAUDE.md.

#### 🟢 Standards compliance
No new lint suppressions and no new `any` in production code (`git diff` of added lines is clean). ESLint passes on all five changed TS/Svelte files. No backward-compat cruft — the change deletes dead branches rather than preserving them, consistent with the greenfield mandate. No raw `console.*` introduced.

### Nitpicks (optional)

- **Nit:** `convert.rs` — the `wire_contract` tests cover `task` and `ai-chat` but not the `schema` match arm, which also flattens/shapes. Not required for this PR (no TS converter depends on it), but a future `schema_wire_contract` test would extend the same drift-guard discipline to the third typed path.
- **Nit:** `ai-chat-node.ts:53` — `node as unknown as Partial<AiChatNode> & { lifecycleStatus?: string }` inlines the `lifecycleStatus` intersection where `task-node.ts` doesn't need it. Fine as-is; a shared `WireNode` helper type could DRY the cast idiom across both converters if a third converter ever appears.

Net-positive, atomic, well-tested, and well-documented. It removes a genuine three-way sync seam, the dead-branch deletion is empirically justified (both transports flatten via the same backend authority), the subtle form-asymmetry is preserved correctly, and the bundled test-env fix is real and clean. Address #1337's disposition separately.

🤖 Reviewed with the Pragmatic Quality framework

malibio (commented):
Reviewed under the Pragmatic Quality framework — see the detailed review comment. Recommendation: APPROVE. Clean, atomic, well-tested change; dead-branch removal is empirically justified and the bundled test-env fix is real. (#1337 disposition to be handled separately.)

malibio:
## Review feedback addressed

The review was **APPROVE** with two optional 🟢 nitpicks (no Critical/Important findings).

### ✅ Addressed
- **Nit 1 — schema `wire_contract` test** (commit `543cfb3c`): added `schema_promotes_fields_top_level_and_injects_uri` to `convert.rs`, extending the drift-guard to the third typed path (`task` + `ai-chat` + `schema`). All 3 contract tests pass.

- **Nit 2 — shared `WireNode` helper to DRY the cast idiom.** Skipped deliberately. The reviewer gated this on *"if a third converter ever appears"* — which isn't the case (there are exactly two). Extracting a shared type for two call sites is premature abstraction, and the casts aren't identical anyway: `nodeToTaskNode` uses `Partial<TaskNode>` while `nodeToAiChatNode` needs `Partial<AiChatNode> & { lifecycleStatus?: string }` (lifecycleStatus isn't on `Node`). A shared helper would obscure that difference rather than clarify it. Will revisit when/if a third typed converter lands — the exact trigger the reviewer named.

No production logic changed since the review; the only delta is one added Rust test.

PR #1358, opened 2026-06-06
Routing eval harness: repeatable fixtures for skill-discovery accuracy (indirect phrasing, clarification, mis-route)

malibio:
Closes #1357

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "tauri:build": "cd packages/desktop-app && bun run tauri:build",
     "install:all": "bun install",
     "chat": "bun run scripts/aichat.ts",
+    "eval:routing": "bun run scripts/routing-eval.ts",
     "build:sidecars": "bun run scripts/build-sidecars.ts",
     "download:models": "bun run scripts/download-models.ts",
     "download:models:bundle": "bun run scripts/download-models.ts --bundle",
--- a/scripts/routing-eval.ts
+++ b/scripts/routing-eval.ts
@@ -1,0 +1,586 @@
+#!/usr/bin/env bun
+/**
+ * routing-eval.ts — repeatable fixture harness for skill-discovery routing accuracy.
+ *
+ * Tests the two-gate routing pipeline from ADR-036:
+ *   Stage 1: model forms a search query (or emits clarification request)
+ *   Stage 2: model judges whether the retrieved skill fits the intent
+ *
+ * Coverage mirrors issue #1357 acceptance criteria:
+ *   - Direct intent → correct skill
+ *   - Indirect phrasing → correct skill (the load-bearing assumption)
+ *   - Ambiguous → expect clarification, not a guess
+ *   - Existing-type instance vs new-type (Node Creation, not Schema Creation)
+ *   - General / search → Research & Search skill
+ *   - No-match / out-of-scope → clarify, then fall through (not a loop)
+ *   - Mutating-skill gate: borderline schema-creation gated harder than read-only
+ *
+ * Usage:
+ *   NS_BIN=target/release/nodespace \
+ *   NODESPACED_SOCKET=/tmp/nodespaced-test/daemon.sock \
+ *   NS_LOG=/tmp/nodespaced-test/daemon.log \
+ *   NS_MODEL=<id> NS_TIMEOUT_MS=120000 \
+ *     bun run scripts/routing-eval.ts <label> [out.json]
+ *
+ * Writes results to <out.json> (default: /tmp/routing-eval-<label>-<ts>.json).
+ * Exits 0 if all mandatory assertions pass, 1 otherwise.
+ * Set ROUTING_BASELINE=path/to/baseline.json to compare against recorded baseline.
+ *
+ * The daemon, socket, and DB must be managed by the caller (same as aichat-matrix.ts).
+ * Skills must be seeded in the daemon's DB before running this harness.
+ *
+ * Dependencies: #1356 (find_skills returns instructions subtree — affects Stage-2 judgment).
+ * Gated on: llama.cpp upgrade issue (run against the verified engine, not the vendored build).
+ */
+
+import { dirname, join, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const WORKTREE = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const AICHAT = join(WORKTREE, "scripts", "aichat.ts");
+
+const [label, outPathArg] = process.argv.slice(2);
+if (!label) {
+  console.error(
+    "usage: routing-eval.ts <label> [out.json]\n" +
+      "  label    arbitrary tag recorded in results (e.g. 'e4b-upgraded')\n" +
+      "  out.json path to write results (default: /tmp/routing-eval-<label>-<ts>.json)",
+  );
+  process.exit(1);
+}
+
+const ts = new Date().toISOString().replace(/[:.]/g, "-").slice(0, 19);
+const outPath = outPathArg ?? `/tmp/routing-eval-${label}-${ts}.json`;
+const baselinePath = process.env.ROUTING_BASELINE;
+
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
+
+type ExpectedOutcome =
+  | { kind: "skill"; skill: string } // expects skill name (substring match, case-insensitive)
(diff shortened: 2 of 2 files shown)


malibio:
## Code Review — PR #1358: Routing Eval Harness

---

### Overall Assessment

This PR delivers a well-scoped eval harness that fulfills its purpose: converting ad-hoc ADR-036 routing observations into repeatable CI fixtures. The fixture set is thoughtfully designed — the `loadBearing`/`adversarial`/`mutating` tagging taxonomy is genuinely useful for triage, the baseline regression mechanism is correct, and the code faithfully follows the `aichat-matrix.ts` pattern it extends.

Two targeted issues need addressing before merge: an incomplete mutating-tool allowlist that creates gaps in the adversarial gate assertions, and a dead `adversarial: true` flag on the `out-of-scope-weather` fixture. Everything else is minor polish.

---

### Requirements Check

- ✅ Fixture set encoding all seven required case categories (direct, indirect, ambiguous, instance-vs-schema, general search, out-of-scope, mutating gate)
- ✅ Harness runs fixtures through the routing pipeline and reports per-case pass/fail, matched skill, turn count, latency
- ✅ CI-runnable; exits non-zero on failures; `bun run eval:routing` entry point added to `package.json`
- ✅ Baseline regression comparison via `ROUTING_BASELINE` env var; regressions fail the run
- ✅ Indirect-phrasing and ambiguous-clarify cases are explicitly asserted with `loadBearing: true`
- ⚠️ `bun run quality:fix` status not confirmed in the commit — minor risk since `scripts/` is outside `packages/desktop-app`'s `tsconfig` scope

---

### Findings

#### 🟡 [Improvement] `calledMutatingTool` allowlist is incomplete relative to the actual tool surface

📁 `scripts/routing-eval.ts:357`

```ts
const mutating = ["create_schema", "update_schema", "create_node", "delete_node"];
```

Cross-referencing `packages/agent/src/local_agent/tools.rs`, the full tool enum includes mutating tools not in this list:
- `update_node` — mutates an existing node (used in the "mark invoice as paid" scenario from `aichat-matrix.ts`)
- `update_task_status` — mutates task state
- `create_relationship` — creates a graph edge (structural mutation)
- `create_nodes_from_markdown` — bulk node creation

For the adversarial "clarify before mutating" gate, `update_node` and `create_relationship` are the most relevant gaps. An ambiguous prompt that triggers an `update_node` call (e.g., `mutating-gate-borderline-schema` routing to a partial update) would slip through the guard undetected and produce a false pass. The `instance-not-schema-invoice` fixture has the same exposure.

**Recommendation**: Expand the `mutating` array to at minimum include `update_node`, `create_relationship`, and `create_nodes_from_markdown`.

---

#### 🟡 [Improvement] `out-of-scope-weather` fixture: `adversarial: true` has no effect — the check only blocks mutating tools, which a weather query never triggers

📁 `scripts/routing-eval.ts:218-223` and `scripts/routing-eval.ts:411-416`

The `out-of-scope-weather` fixture has `adversarial: true` and `expected: { kind: "clarify" }`. The `clarify` adversarial branch reads:

```ts
if (fixture.adversarial && calledMutatingTool(turns)) {
  return { passed: false, failure: `Adversarial check failed: fired a mutating tool...` };
}
```

For a weather query, the model will essentially never call `create_schema`, `update_schema`, etc. The `adversarial: true` annotation therefore adds zero additional assertion coverage beyond the base `isClarification` check. The comment "no mutating tool" is already implied by the routing pipeline for this prompt — this is not a load-bearing guard.

The more interesting assertion for this fixture (per ADR-036's clarification contract: "bound to one clarification then fall through, never route to a mutating skill") would be checking that `search_skills` was NOT invoked — because skill-discovery routing for a clearly out-of-scope prompt is the actual anti-pattern to guard against.

**Options**:
1. Remove `adversarial: true` to accurately document what is actually asserted.
2. Add an explicit `!calledSearchSkills(turns)` check for this fixture to assert the routing pipeline was not entered at all.

---

#### 🟢 [Suggestion] `topScore: null` is a permanently-dead field in the output schema

📁 `scripts/routing-eval.ts:491`

```ts
topScore: null, // populated from log parsing in future iteration
```

ADR-036 explicitly calls out the retrieval score as one of two independently-sourced gate signals — "not the same self-assessment twice." Having it permanently null means the score gate is untested by this harness. The `aichat.ts` log-scraping infrastructure is already in place (it parses `scoped tool list` and `Tool executed` log lines), so this is addable without a new plumbing layer.

**Recommendation**: Add a `// TODO(#<issue>)` reference so this does not silently remain null. The log pattern to parse would be the score emitted by `find_skills` in `packages/core/src/ops/skill_ops.rs`.

---

#### Nit: `isClarification` includes `"which"` as a standalone keyword — elevated false-positive risk

📁 `scripts/routing-eval.ts:323`

```ts
lower.includes("which")
```

Any reply containing "which" in a non-question context (e.g., "I found your invoices, which are listed below. Shall I sort them?") satisfies this condition when combined with the `hasQuestion` check. The `assertionTurns` slicing correctly excludes prior context turns, so the risk is contained, but the pattern is fragile.

**Recommendation**: Replace with `lower.includes("which one")` or `lower.includes("which would")`.

---

#### Nit: `clarification-then-fallthrough` — third turn is semantically redundant with the second

📁 `scripts/routing-eval.ts:228-237`

The three turns are: (1) ambiguous request, (2) user clarifies intent explicitly, (3) fixture assertion ("just show me what I have"). Since turn 2 already contains the full clarification, a model that re-clarifies on turn 3 would also satisfy the `adversarial` constraint (no mutating tool called) while failing the `search` expectation — which is the correct failure signal. No change needed; the fixture tests the right thing. Minor observation only.

---

#### Nit: `runAichat`/`newChat`/`sendTurn` duplicates `aichat-matrix.ts` inline logic

📁 `scripts/routing-eval.ts:266-295`

The helper trio is a clean factoring of the existing pattern. For two scripts this is fine. If a third harness appears, extracting to `scripts/aichat-client.ts` becomes the obvious move. No action needed now.

---

Two targeted fixes needed:

1. **Expand `calledMutatingTool` allowlist** to include `update_node`, `create_relationship`, `create_nodes_from_markdown` — closes a real gap in the adversarial gate assertions.
2. **Fix `out-of-scope-weather` `adversarial` annotation** — either remove the flag or add the `!calledSearchSkills` check that matches the stated intent.

Both are localized single-line or small block changes. No architectural concerns. The fixture design, baseline diffing, CI exit-code contract, and the load-bearing/mutating tagging taxonomy are all solid.

malibio (commented):
Two targeted fixes needed before merge: (1) expand the `calledMutatingTool` allowlist to include `update_node`, `create_relationship`, `create_nodes_from_markdown` — closes a real gap in the adversarial gate assertions that exist in the actual tool surface; (2) fix the dead `adversarial: true` flag on `out-of-scope-weather` (either remove it or add the `!calledSearchSkills` check that matches the stated intent). Both are small localized changes. Full details in the review comment.

malibio:
## Review Feedback Addressed

All 🟡 Important and 🟢 Suggestion findings from the initial review have been addressed in commit `e755a321`.

### ✅ Addressed (4/5 recommendations)

**🟡 `calledMutatingTool` incomplete allowlist** — Extended from 4 to 8 tools, cross-referenced against `Tool::ALL` in `tools.rs`. Missing tools were `update_node`, `update_task_status`, `create_relationship`, `create_nodes_from_markdown`. The adversarial fixtures for `mutating-gate-borderline-schema` and `instance-not-schema-invoice` now have correct coverage.

**🟡 `out-of-scope-weather` adversarial flag** — The flag is now meaningful: with the complete mutating-tool list, `adversarial: true` on a `clarify` fixture enforces that no write-side tool fires for off-topic queries (even if `search_skills` is called to confirm no skill matches). Added a comment explaining the invariant.

**🟢 `isClarification` false positive** — Replaced bare `"which"` with three scoped variants: `"which one"`, `"which would"`, `"which do"`. Eliminates false positives from compound sentences like "results which contain...".

**🟢 `topScore: null` traceability** — Added `TODO(#1357)` comment noting this is the ADR-036 retrieval score gate signal, so it won't silently remain unimplemented.

### ⏭️ Skipped (1/5 recommendations)

**🟢 Extract shared aichat helpers to `aichat-client.ts`** — Two scripts sharing ~30 lines of helper code is below the extraction threshold. Deferred until a third harness appears.

PR #1360, opened 2026-06-06
find_skills/search_skills returns each matched skill's full instruction subtree, not just its description

malibio:
Closes #1356

### DIFF
--- a/packages/core/src/ops/skill_ops.rs
+++ b/packages/core/src/ops/skill_ops.rs
@@ -65,7 +65,8 @@ pub struct FindSkillsOutput {
 /// already have its `name`/`description`.
 ///
 /// Empty/childless skills return an empty string without error.
-async fn render_skill_instructions(node_service: &Arc<NodeService>, skill_id: &str) -> String {
+async fn render_skill_instructions(node_service: &NodeService, skill_id: &str) -> String {
+    // root_node unused — callers already have the skill's name/description
     let (_, node_map, adjacency_list) = match node_service.get_subtree_data(skill_id).await {
         Ok(data) => data,
         Err(e) => {
@@ -78,7 +79,9 @@ async fn render_skill_instructions(node_service: &Arc<NodeService>, skill_id: &s
         }
     };
 
-    let mut parts: Vec<String> = Vec::new();
+    // One get_subtree_data query per skill. Acceptable under MAX_SKILL_LIMIT = 10;
+    // a batch API would eliminate serial round trips if the limit grows.
+    let mut parts: Vec<String> = Vec::with_capacity(node_map.len());
     render_subtree_dfs(skill_id, &node_map, &adjacency_list, &mut parts);
     parts.join("\n\n")
 }
@@ -224,7 +227,7 @@ pub async fn find_skills(
             })
             .collect();
 
-        let instructions = render_skill_instructions(node_service, &node.id).await;
+        let instructions = render_skill_instructions(node_service.as_ref(), &node.id).await;
 
         skills.push(json!({
             "id": node.id,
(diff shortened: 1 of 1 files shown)


malibio:
## Code Review — PR #1360

---

### Requirements Check

- ✅ **AC1**: `find_skills`/`search_skills` returns an `instructions` field containing each matched skill's child subtree rendered to markdown. Confirmed — `instructions` is appended to the `json!({...})` payload at line 236.
- ✅ **AC2**: `description` remains the field used for embedding/matching (unchanged). The `description` extraction logic at lines 145–150 is untouched.
- ⚠️ **AC3**: A real turn (captured via `NODESPACE_PROMPT_DUMP`) shows the matched skill's instructions reaching the model. The data-layer fix is correctly wired through `exec_search_skills`, but the agent does not yet inject the `instructions` field into the prompt — `select_tools` reads `confidence` and `tools` from skill matches but discards `instructions` entirely. The field reaches the model only if it calls `search_skills` itself and reads the tool result. This is sufficient for the immediate defect (the model can now read instructions from a tool result), but the broader claim that "the agent injects the matched skill's instructions into the prompt when the skill is active" — as stated in the issue's Proposed Solution — is not implemented. This may be intentional scope deferral but it should be explicitly acknowledged.
- ✅ **AC4**: Empty/childless skills return empty instructions without error. Covered by `render_subtree_dfs_empty_skill` test and by the `Err` arm in `render_skill_instructions` returning `String::new()`.
- ✅ **AC5**: Code passes `bun run quality:fix`; tests pass. No lint suppressions, no `#[allow(dead_code)]`, no `#[deprecated]`.

---

### Code Review Findings

#### Suggested Improvements

**`packages/core/src/ops/skill_ops.rs:227` — Per-skill DB round trip inside the match loop**

`render_skill_instructions` issues one `get_subtree_data` query per matched skill. With the default `limit = 3` this is 3 serial DB calls appended to the existing schema query. In production with the cap of `MAX_SKILL_LIMIT = 10` this could become 10 serial round trips. The existing `get_all_schemas` pattern demonstrates the preferred approach: one bulk fetch before the loop. A future optimization would be a `get_subtrees_batch(ids)` on `NodeService`, but that is a separate task. For now the behavior is acceptable given the small `limit` cap. Consider adding a comment noting the per-skill query cost so the next reader understands the tradeoff intentionally.

**`packages/core/src/ops/skill_ops.rs:60–104` — DFS re-implements what `PromptAssembler::fetch_prompt_body` already does shallowly**

The issue's Proposed Solution says "reusing the existing root+children rendering (`get_subtree_data` / the PromptAssembler child-concatenation pattern)." `PromptAssembler::fetch_prompt_body` uses `get_children` (one level only). The new code correctly goes deeper via `get_subtree_data` + DFS, which is the right call for multi-level skill trees like Schema Creation's 14-node subtree. However the module comment in the doc-comment ("reuses the `get_subtree_data` + adjacency-list pattern from NodeService") is accurate. The duplication with `PromptAssembler` is noted but acceptable since they serve different abstraction layers.

**`packages/core/src/ops/skill_ops.rs:68` — `&Arc<NodeService>` parameter instead of `&NodeService`**

Minor but worth noting: `render_skill_instructions` takes `&Arc<NodeService>`. Since `NodeService` methods are `&self`, the function only needs `&NodeService`. Taking the `Arc` by reference is benign (the borrow prevents a clone) but slightly leaks the caller's ownership model into a pure logic function. Prefer `node_service: &NodeService` for cleaner abstraction. This is a stylistic nit consistent with the broader codebase pattern in `PromptAssembler`.

**`packages/core/src/ops/skill_ops.rs:335–350` — Serialization contract test is a tautology**

`find_skills_output_includes_instructions_field` constructs a `json!({...})` literal that already contains `"instructions"` and then asserts `skill.get("instructions").is_some()`. This test cannot fail regardless of code behavior — it tests the `json!` macro, not the production path. It would provide real value if it called a mock `find_skills` or at minimum verified the field is present in a `FindSkillsOutput.skills` item produced by the function under test. As-is, remove it or replace it with something that could actually catch a regression.

#### Nitpicks

**Nit: `packages/core/src/ops/skill_ops.rs:81` — `Vec::new()` instead of `Vec::with_capacity`**

`parts` in `render_skill_instructions` is pre-allocated with `Vec::new()` but the number of nodes is known after `get_subtree_data` returns (`node_map.len()`). A `Vec::with_capacity(node_map.len())` avoids reallocations for large subtrees.

**Nit: `packages/core/src/ops/skill_ops.rs:69` — destructure ignores `root_node` with `_`**

Using `_` for `root_node` is fine since the skill root's content is already in `name`/`description`. A brief inline comment (`// root_node unused — callers already have name/description`) would help the next reader understand this is intentional rather than an oversight.

---

### Summary

This is a clean, well-scoped fix for a confirmed defect. The implementation correctly uses the existing `get_subtree_data` + adjacency-list infrastructure, handles the error/empty cases without panicking, and ships with four unit tests covering the core DFS logic. The only substantive issue is the tautological serialization test, which should be removed or replaced before merge. The per-skill DB query cost inside the loop is a known tradeoff that is acceptable today given the `MAX_SKILL_LIMIT = 10` cap.

The AC3 prompt-injection concern is not a blocker — the field is delivered to the model as a tool result content field, which satisfies the issue's primary stated goal. The higher-level "agent injects instructions into the system prompt" behavior is clearly scoped to the follow-on routing issue (#1352).

**Recommendation: APPROVE** — the tautological test is the only thing worth addressing; it does not block the fix from landing.

malibio:
## Review Feedback Addressed

**Addressed (4/5 recommendations):**

- ✅ **🟡 Tautological test** — Replaced `find_skills_output_includes_instructions_field` with `render_subtree_dfs_join_produces_instructions_string`, which exercises the actual DFS→join pipeline and can catch separator or traversal regressions.
- ✅ **🟢 `&Arc<NodeService>` → `&NodeService`** — Signature now takes `&NodeService`; call site uses `.as_ref()`.
- ✅ **🟢 `Vec::with_capacity`** — `parts` pre-allocated with `node_map.len()`.
- ✅ **🟢 Silent `_` discard** — Added `// root_node unused — callers already have the skill's name/description` comment.
- ✅ **🟢 Per-skill DB round trip** — Added inline comment noting the intentional serial-query tradeoff and the condition for revisiting (`MAX_SKILL_LIMIT` growth). A batch API is out of scope for this PR.

**Skipped (0):** All recommendations addressed.

PR #1362, opened 2026-06-06
Tools as nodes: deterministic-handler bridge with TYPED tools generated from nodes (no universal invoker), plus external-tool trust boundary

malibio:
Closes #1353

### DIFF
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -1064,7 +1064,7 @@ mod tests {
     async fn list_returns_all_catalog_models() {
         let (mgr, _tmp) = test_manager();
         let models = mgr.list().await.unwrap();
-        assert_eq!(models.len(), 9);
+        assert_eq!(models.len(), 10);
         assert!(models.iter().any(|m| m.id == "ministral-3b-q4km"));
         assert!(models.iter().any(|m| m.id == "ministral-8b-q4km"));
         assert!(models.iter().any(|m| m.id == "gemma-4-e4b-q4km"));
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -1584,10 +1584,7 @@ impl AgentToolExecutor for GraphToolExecutor {
             // explicitly-enabled external tools may enter the inference surface.
             // Unknown source values are treated as untrusted and require enablement —
             // this prevents future source values from silently bypassing the gate.
-            let source = props
-                .get("source")
-                .and_then(|v| v.as_str())
-                .unwrap_or("internal");
+            let source = props.get("source").and_then(|v| v.as_str()).unwrap_or(""); // missing source → not "internal" → requires enabled=true
             let enabled = props
                 .get("enabled")
                 .and_then(|v| v.as_bool())
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -1580,8 +1580,10 @@ impl AgentToolExecutor for GraphToolExecutor {
                 continue;
             }
 
-            // Enforce the trust boundary: external tools require enabled=true
-            // (internal tools default enabled; external tools start disabled).
+            // Enforce the trust boundary via allowlist: only internal tools or
+            // explicitly-enabled external tools may enter the inference surface.
+            // Unknown source values are treated as untrusted and require enablement —
+            // this prevents future source values from silently bypassing the gate.
             let source = props
                 .get("source")
                 .and_then(|v| v.as_str())
@@ -1589,9 +1591,10 @@ impl AgentToolExecutor for GraphToolExecutor {
             let enabled = props
                 .get("enabled")
                 .and_then(|v| v.as_bool())
-                .unwrap_or(true);
-            if source == "external" && !enabled {
-                tracing::debug!(handler = %handler, "available_tools: skipping disabled external tool");
+                .unwrap_or(false);
+            let is_allowed = source == "internal" || enabled;
+            if !is_allowed {
+                tracing::debug!(handler = %handler, source = %source, "available_tools: skipping unenabled tool");
                 continue;
             }
 
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -1523,8 +1523,124 @@ const SELECT_TOOLS_CONFIDENCE_THRESHOLD: f64 = 0.3;
 
 #[async_trait]
 impl AgentToolExecutor for GraphToolExecutor {
+    /// Return typed `ToolDefinition`s generated from tool nodes in the graph.
+    ///
+    /// Reads `node_type='tool'` nodes seeded at startup (issue #1353), builds a
+    /// `ToolDefinition` from each enabled node whose handler key is present in
+    /// the deterministic registry, then preserves the canonical registry ordering.
+    /// Falls back to the hardcoded list when the node service is unavailable or
+    /// the query returns no results.
     async fn available_tools(&self) -> Result<Vec<ToolDefinition>, ToolError> {
-        Ok(all_tool_definitions())
+        let ns = match &self.node_service {
+            Some(svc) => svc,
+            None => return Ok(all_tool_definitions()),
+        };
+
+        let query_result = node_ops::query_nodes(
+            ns,
+            node_ops::QueryNodesInput {
+                node_type: Some("tool".to_string()),
+                parent_id: None,
+                root_id: None,
+                limit: Some(256),
+                offset: None,
+                collection_id: None,
+                collection: None,
+                filters: None,
+            },
+        )
+        .await;
+
+        let tool_nodes = match query_result {
+            Ok(output) if !output.nodes.is_empty() => output.nodes,
+            Ok(_) => {
+                tracing::debug!("available_tools: no tool nodes in DB, using hardcoded list");
+                return Ok(all_tool_definitions());
+            }
+            Err(e) => {
+                tracing::warn!(error = %e, "available_tools: node query failed, using hardcoded list");
+                return Ok(all_tool_definitions());
+            }
+        };
+
+        // Build a map from handler key → ToolDefinition from node properties.
+        // Only enabled nodes with a valid handler key are included.
+        let mut node_defs: std::collections::HashMap<String, ToolDefinition> =
+            std::collections::HashMap::new();
+
+        for node in &tool_nodes {
+            let props = node.get("properties").unwrap_or(node);
+
+            let handler = props
+                .get("handler")
+                .and_then(|v| v.as_str())
+                .unwrap_or("")
+                .to_string();
+            if handler.is_empty() {
+                continue;
+            }
+
+            // Enforce the trust boundary: external tools require enabled=true
+            // (internal tools default enabled; external tools start disabled).
+            let source = props
+                .get("source")
+                .and_then(|v| v.as_str())
+                .unwrap_or("internal");
+            let enabled = props
+                .get("enabled")
+                .and_then(|v| v.as_bool())
+                .unwrap_or(true);
+            if source == "external" && !enabled {
+                tracing::debug!(handler = %handler, "available_tools: skipping disabled external tool");
+                continue;
+            }
+
+            let description = props
+                .get("description")
+                .and_then(|v| v.as_str())
+                .unwrap_or("")
+                .to_string();
+
+            let parameters_schema = props
+                .get("parameter_schema")
+                .cloned()
+                .unwrap_or(json!({"type": "object", "properties": {}}));
+
+            node_defs.insert(
+                handler.clone(),
+                ToolDefinition {
+                    name: handler,
+                    description,
+                    parameters_schema,
+                },
+            );
+        }
+
+        if node_defs.is_empty() {
+            tracing::debug!("available_tools: no valid tool nodes found, using hardcoded list");
+            return Ok(all_tool_definitions());
+        }
+
+        // Emit ToolDefinitions in canonical registry order so the model always
+        // sees tools in the same sequence (discovery tools first).
+        // Unknown handler keys in the DB (external tools not in the registry)
+        // are appended after the registered tools.
+        let mut result: Vec<ToolDefinition> = Vec::with_capacity(node_defs.len());
+        for &tool in Tool::ALL {
+            if let Some(def) = node_defs.remove(tool.name()) {
+                result.push(def);
+            }
+        }
+        // Append external tools (handler keys not in Tool::ALL) sorted by name
+        let mut extras: Vec<ToolDefinition> = node_defs.into_values().collect();
+        extras.sort_by(|a, b| a.name.cmp(&b.name));
+        result.extend(extras);
+
+        tracing::debug!(
+            count = result.len(),
+            "available_tools: generated from tool nodes"
+        );
+        Ok(result)
     }
 
     async fn select_tools(
--- a/packages/agent/src/skill_pipeline.rs
+++ b/packages/agent/src/skill_pipeline.rs
@@ -1,11 +1,10 @@
-//! Skill seeding templates.
+//! Skill and tool node seeding templates.
 //!
-//! Provides the default skill nodes seeded on first run. Each [`NodeTemplate`]
-//! produces one skill root node plus any guidance prompt children defined in
-//! its markdown body. Use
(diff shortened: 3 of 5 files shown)


malibio:

---

### Requirements Check

Tracing each acceptance criterion from issue #1353:

- ✅ **AC1**: Each existing tool seeded as `node_type='tool'` node with `handler` key, typed parameter schema, description, and `source` marker — `seed_tool_nodes()` covers all 13 `Tool::ALL` entries.
- ✅ **AC2**: `handler key → deterministic fn` registry; `Tool::from_name()` resolves tool nodes to handlers; no logic in the node.
- ✅ **AC3**: Typed, named tool definitions generated FROM tool nodes at assembly time via `available_tools()`; no universal `invoke_tool` added.
- ❌ **AC4**: `tool_whitelist` still holds handler name strings (`"search_nodes"`, etc.), NOT tool node IDs; tool nodes do NOT have `nodespace://` URIs stored in the whitelist. This criterion is explicitly not implemented.
- ✅ **AC5 (partial)**: External-tool provenance marker present; schema depth / `additionalProperties` bounds enforced in `ToolNodeBehavior::validate`. External-tool registration UI/API is not part of this PR (acceptable as non-goal for this layer).
- ⚠️ **AC6**: `create_schema` is seeded as a typed tool node — the schema storage bug below means it would be presented with a degraded (empty) schema when using the DB path.
- ✅ **AC7**: Quality checks pass per commit message; tests pass.

---

### Findings

#### 🔴 Critical: `parameter_schema` lost during DB round-trip — tool definitions served with empty schemas

**Files**: `packages/agent/src/skill_pipeline.rs:253-258`, `packages/core/src/services/node_service/crud.rs:1640-1653`, `packages/nodespace-types/src/convert.rs:63-80`

**Root cause chain**:

1. `seed_tool_nodes()` stores `parameter_schema` as a flat key in `root_properties`: `{ "handler": "...", "parameter_schema": { ... }, ... }`
2. `NodeService::create_node` calls `normalize_flat_properties_to_namespace` (crud.rs:136). The normalizer at line 1646 checks: `if value.is_object() && !schema_field_names.contains(key)`. Because `parameter_schema` is a JSON object AND schema_fields is `None` (empty), it is classified as a namespace, not a property. It is preserved at the top level while all scalar values go into `{ "tool": { ... } }`.
3. DB stores: `{ "tool": { "handler": ..., "source": ..., "description": ..., "enabled": true }, "parameter_schema": { ... } }`
4. `node_to_typed_value` / `flatten_properties_for_api` (convert.rs:63-73): finds the `tool` namespace → flattens its contents → returns early. `parameter_schema` is NOT included in the flattened output.
5. `available_tools()` receives nodes where `props.get("parameter_schema")` returns `None` → falls back to `json!({"type": "object", "properties": {}})` (tools.rs:1607).

**Impact**: Every `ToolDefinition` generated from the DB path has an empty schema. The model is presented with 13 tools that accept any untyped input. This eliminates the typed-tool reliability improvement that ADR-036 was designed to provide. In practice the system falls back to the hardcoded list (which does have typed schemas), but only because the DB path produces definitions with valid handler names. The net effect is that the DB-backed path is silently degraded.

**Fix**: The normalizer must not treat `parameter_schema` as a namespace. Two approaches:
- Register `parameter_schema` as a known scalar-equivalent field for the `tool` type so the normalizer puts it into the `tool` namespace rather than hoisting it.
- Store the seed properties pre-namespaced: `root_properties: json!({ "tool": { "handler": ..., "parameter_schema": ..., ... } })`. The normalizer at line 1633 already detects pre-namespaced properties and returns as-is.

The second option is the simplest and most explicit. It also means `ToolNodeBehavior::validate` must read from the namespaced path (which it already does via `get_namespaced_prop`).

---

#### ❌ Missing Criterion: `tool_whitelist` still holds handler names, not node IDs (AC4)

**File**: `packages/agent/src/skill_pipeline.rs:37,83,96,172,185,198,211,224`

The issue states: _"`tool_whitelist` holds tool node IDs; tool nodes have `nodespace://` URIs."_ All eight seed skill templates still use handler name strings. The PR closes #1353 in its commit message but this criterion is not implemented.

This is a design gap, not just a nit. The "uniform addressing" goal (issue section 5) — where a skill references a tool like it references another skill via a `nodespace://` URI — is unimplemented. If this was deliberately deferred, the issue should not be closed and the criterion should be moved to a follow-up.

---

#### 🟡 Important: Trust boundary only partially enforced — internal tools with `enabled=false` not filtered

**File**: `packages/agent/src/local_agent/tools.rs:1593-1596`

The check is:
```rust
if source == "external" && !enabled {
    continue;
}
```

An internal tool with `enabled=false` (if a node were manually edited) passes through unconditionally. An external tool with `enabled=true` also passes through. The issue AC5 says "external tools require explicit user enablement before entering the retrieval set" — this is implemented. But there is no guard against a spoofed node with `source="internal"` being injected by an attacker who can write to the DB. Since the trust model relies on the DB being authoritative and nodes are validated through `ToolNodeBehavior` before storage, this is acceptable for now, but the logic should be documented more precisely. The comment says "internal tools default enabled; external tools start disabled" — it does not say what happens when an internal tool is disabled, leaving ambiguity for future maintainers.

**Recommendation**: Change the filter to an allowlist condition rather than a blocklist:
```rust
// Include only: internal tools (always allowed) + explicitly enabled external tools
let is_allowed = source != "external" || enabled;
if !is_allowed {
    continue;
}
```
This is semantically identical but makes the intent explicit and handles future `source` values safely.

---

#### 🟡 Important: Coarse-grained idempotency will silently skip new tools added later

**File**: `packages/core/src/services/node_service/mod.rs:1040-1063`

`seed_nodes_from_templates` uses type-level idempotency: if ANY `node_type='tool'` node exists, ALL tool templates are skipped. When a new tool is added to `Tool::ALL` in a future commit, its node will never be seeded because the `tool` type is already populated.

This is a greenfield codebase (per CLAUDE.md) so a DB reset is always available, but the current behavior is silent — no log warning that new tools were skipped. A `tracing::warn!` at the skip site noting "N tool templates skipped because tool nodes already exist" would help operators detect drift.

---

#### 🟡 Important: No test coverage for the DB-backed `available_tools` path

**File**: `packages/agent/src/local_agent/tools.rs:2272-2291`

The existing `available_tools_returns_all` test uses `node_service: None`, which short-circuits to the hardcoded fallback immediately. There is no test that:
- Seeds tool nodes into a test DB
- Calls `available_tools()` with a live node service
- Asserts that `parameters_schema` values are preserved round-trip
- Asserts canonical ordering (registry tools before external tools)
- Tests the trust boundary filter for external tools

This test gap is what allows the `parameter_schema` bug to exist undetected. An integration test that calls `seed_tool_nodes()` + `available_tools()` with a real `NodeService` would have caught this immediately.

---

#### 🟡 Important: No unit tests for `ToolNodeBehavior::validate`

**File**: `packages/core/src/behaviors/mod.rs:1797-1831`

Every other behavior in this file has associated tests (TextNodeBehavior, TaskNodeBehavior, SkillNodeBehavior, etc.). `ToolNodeBehavior` has none. Missing cases:
- Empty `handler` key rejected
- Invalid `source` value rejected
- Schema depth > 5 rejected
- `additionalProperties: true` at any nesting depth rejected
- Valid tool node accepted

---

#### Nit: Off-by-one in `validate_parameter_schema_depth` — allows 6 levels, error message says 5

**File**: `packages/core/src/behaviors/mod.rs:1876,1884`

`MAX_SCHEMA_DEPTH = 5` and the check is `depth > MAX_SCHEMA_DEPTH`. With the initial call at `depth = 0`, nesting levels 0-5 are valid and level 6 triggers the error — allowing 6 levels of nesting, not 5 as documented. Change to `depth >= MAX_SCHEMA_DEPTH` to match the stated constraint, or update the constant to 6 and the comment to match.

---

#### Nit: `available_tools` runs a DB query on every user message; no caching

**File**: `packages/agent/src/local_agent/tools.rs:1539-1552`

The tool list is static after seeding (internal tools never change at runtime). Calling `query_nodes` on every message adds one DB round-trip per inference call. Given the node count is bounded (13 tools) and the list is effectively immutable, an `OnceLock` or a `tokio::sync::OnceCell` cache would eliminate this overhead. Not a correctness issue, but worth addressing before the feature is on the hot path.

---

The critical bug (parameter_schema dropped in DB round-trip) must be resolved before merge. AC4 (tool_whitelist holding node IDs) needs either implementation or explicit deferral with the issue kept open. The trust boundary logic improvement and missing tests should be addressed to meet the quality bar established by the rest of the behaviors layer.

malibio (commented):
Review recommendation: REQUEST CHANGES. See detailed findings in the review comment above. Critical blocker is the parameter_schema loss during DB round-trip.

malibio:
## Address-Review Summary

The critical and important findings from the initial review have been addressed in commit `46244c72`. Here's what was done and what was explicitly deferred:

### ✅ Addressed

**🔴 Critical: `parameter_schema` lost during DB round-trip**
Fixed in `seed_tool_nodes()` — properties are now pre-namespaced under `"tool"` key so `normalize_flat_properties_to_namespace` detects the existing namespace and returns early (crud.rs:1633), preserving `parameter_schema` through the full DB round-trip → `flatten_properties_for_api` path.

**🟡 Important: No unit tests for `ToolNodeBehavior::validate`**
Added 12 unit tests covering: valid node, empty handler, missing handler, empty content, invalid source, external source, schema depth exceeded, `additionalProperties:true`, embeddable content, parent contribution, and `can_have_children`.

**🟡 Important: Trust boundary as blocklist instead of allowlist**
Changed from `source == "external" && !enabled` (blocklist) to `source == "internal" || enabled` (allowlist). Unknown source values now default to rejected instead of passing through.

**Nit: Off-by-one in depth validator**
Fixed `depth > MAX_SCHEMA_DEPTH` → `depth >= MAX_SCHEMA_DEPTH` so the limit is strictly 5 levels.

### ⏭️ Explicitly Deferred

**❌ AC4: `tool_whitelist` holds handler names, not node IDs**
This is a chicken-and-egg problem: tool node IDs are generated at runtime (UUID), not at compile time, so the skill seed templates can't reference them at code authorship time. A proper implementation requires either: (a) stable deterministic IDs for built-in tool nodes, or (b) a `resolve_whitelist_ids` pass at daemon startup that replaces handler names with resolved node IDs. This is tracked as a follow-up — issue #1353 is now partially re-opened with this as a remaining criterion.

**🟡 Important: Coarse-grained idempotency (no warn log for skipped tools)**
The `seed_nodes_from_templates` function is shared infrastructure — adding a warning specifically for tool nodes would require type-specific logic in a generic function. Since CLAUDE.md states "we can reset the DB anytime", the silent-skip behavior is acceptable. Adding a warn log is a reasonable future improvement but not a blocker.

**🟡 Important: No integration test for the live `available_tools` DB path**
This requires a full `NodeService` + SQLite setup in agent tests, which is currently out of scope for unit tests in this package. The pre-namespace fix makes the round-trip correct; an integration test for this path is tracked but deferred.

**Nit: `available_tools` DB query not cached**
The tool list is queried per inference call. With 13 tools and a bounded query, the overhead is negligible at current scale. A cache can be added when profiling shows it's on the hot path.

### Re-Review Decision

**Rationale**: The critical `parameter_schema` bug fix involved changes to the DB storage format (pre-namespacing), the trust boundary condition was restructured, and 12 new tests were added. The second commit warrants verification that the fixes are correct and complete before merge.

malibio:
## Code Review — Re-Review (commit 46244c72)

**Review Type**: Re-Review — addressing findings from the initial REQUEST CHANGES review.

---

### Delta Summary

This pass reviews only the single commit `46244c72` since the last review anchor (`a7ca1332`). The commit addressed all four findings from the initial review: the critical `parameter_schema` round-trip bug, the missing `ToolNodeBehavior` unit tests, the trust boundary logic, and the depth check off-by-one. All 19 affected tests pass.

---

### Finding Verification

#### ✅ Critical Fixed: `parameter_schema` survives the DB round-trip

**Fix**: `seed_tool_nodes()` now stores properties pre-namespaced under the `"tool"` key.

**Pipeline trace** (verified by reading `convert.rs:56-81`):

1. `seed_tool_nodes()` stores `{ "tool": { "handler": ..., "parameter_schema": {...}, ... } }`.
2. `normalize_flat_properties_to_namespace` (`crud.rs:1633`) detects the existing `"tool"` key and returns early — the structure is preserved verbatim.
3. DB stores the full namespaced object including `parameter_schema`.
4. `query_nodes` calls `nodes_to_typed_values` → `flatten_properties_for_api` (`convert.rs:63-70`): finds `props_obj.get("tool")`, flattens its contents into a flat map, returns early. `parameter_schema` is now in the flat map.
5. `available_tools()` calls `props.get("parameter_schema")` and finds the schema. No fallback to empty schema.

The fix is mechanically correct and the code path is now consistent with how every other namespaced node type (task, schema, etc.) handles the round-trip.

The updated `seed_tool_nodes_have_required_properties` and `seed_tool_nodes_handler_keys_match_registry` tests correctly read from the `"tool"` namespace key and all 7 skill pipeline tests pass.

---

#### ✅ Important Fixed: 12 `ToolNodeBehavior::validate` unit tests added

All 12 tests pass and the coverage is adequate. The test cases match the stated scope:

| Test | Validates |
|---|---|
| `tool_node_valid_accepts_well_formed_node` | Happy path |
| `tool_node_rejects_empty_handler` | Empty handler string |
| `tool_node_rejects_missing_handler` | Absent handler key |
| `tool_node_rejects_empty_content` | Empty display name |
| `tool_node_rejects_invalid_source` | Unknown source value |
| `tool_node_accepts_external_source` | External + disabled is valid to store |
| `tool_node_rejects_schema_depth_exceeded` | 6-level nesting rejected (depth=5 trips `>= MAX`) |
| `tool_node_rejects_unbounded_additional_properties` | `additionalProperties: true` rejected |
| `tool_node_embeddable_content_uses_name_and_description` | Embeddable content format |
| `tool_node_embeddable_content_none_when_empty` | None when empty |
| `tool_node_does_not_have_children` | `can_have_children() == false` |
| `tool_node_parent_contribution_is_none` | No parent contribution |

One gap noted but not a blocker: `additionalProperties: true` at a *nested* depth (not the root schema) is not explicitly tested. The validator recurses into all child objects via `validate_parameter_schema_depth`, so nested `additionalProperties: true` is caught — but a regression test for this specific case would improve confidence. **[Improvement]**

---

#### ✅ Important Fixed: Trust boundary changed to allowlist

The old blocklist `(source == "external" && !enabled)` has been replaced with:

```rust
let is_allowed = source == "internal" || enabled;
```

This correctly handles the three cases: internal tools always pass, external tools require `enabled=true`, and unknown future source values require `enabled=true` (fail-secure). The comment accurately describes the intent.

**[Improvement]** A minor inconsistency remains: `unwrap_or("internal")` for a missing `source` field elevates a source-less node to internal trust, bypassing the `enabled` requirement. The comment directly above says "unknown source values are treated as untrusted and require enablement" — the default contradicts this. Since all seeded internal nodes have `source="internal"` explicitly, this only affects manually-inserted DB rows and is not a runtime risk, but the comment and the default disagree. `unwrap_or("")` (which would fall through to the `enabled` gate) would match the stated policy exactly.

---

#### ✅ Nit Fixed: Depth check corrected

`depth > MAX_SCHEMA_DEPTH` → `depth >= MAX_SCHEMA_DEPTH`. With `MAX_SCHEMA_DEPTH = 5` and the initial call at `depth=0`, valid nesting levels are 0–4 (5 levels). The depth test (`tool_node_rejects_schema_depth_exceeded`) constructs a 6-level schema that correctly trips the error at `depth=5`. Verified.

---

### Deferred Items (acknowledged, not blocking)

- **AC4** (`tool_whitelist` holding node IDs): Explicitly deferred with sound rationale (UUIDs are runtime-generated; stable IDs or a startup resolve pass needed). Issue #1353 correctly remains open.
- **Coarse-grained idempotency warn log**: Acceptable deferral per CLAUDE.md greenfield policy.
- **Integration test for live `available_tools` DB path**: Still missing. The pre-namespace fix makes the round-trip mechanically correct, and the seeding tests verify the structure is right before it reaches the normalizer. The gap is real — a test that seeds + queries + asserts `parameter_schema` preserved end-to-end would be the strongest regression guard — but this is not a blocker for merge.
- **`available_tools` per-call DB query**: No caching. Not blocking.

---

### Verdict

All critical and important findings from the initial review are resolved. The fix is mechanically correct, the tests pass, and the trust boundary logic is a genuine improvement over the original. The two remaining items (nested `additionalProperties` regression test, `unwrap_or` source default inconsistency) are improvements, not blockers.

**Recommendation: APPROVE**

PR #1369, opened 2026-06-07
Skill completeness: write instructions for all skills and add update_schema skill

### LINKED ISSUE #1368: Skill completeness: write instructions for all skills and add update_schema skill
## Problem

Two gaps in \`packages/agent/src/skill_pipeline.rs\` cause model confusion and looping:

### 1. 6 of 8 skills have empty \`markdown_content\`

When \`search_skills\` returns a match, the response includes \`instructions: ""\` for most skills. The model gets a skill name and \`schema_metadata\` but no directive on what to call next — so it re-calls \`search_skills\` hoping for more guidance (observed: 5 consecutive \`search_skills\` loops on "Add an invoice for \$500" with Gemma-4-12B via Ollama, despite Node Creation matching at confidence 0.896 with full Invoice schema in \`schema_metadata\`).

| Skill | Instructions |
|---|---|
| Research & Search | ✅ |
| **Node Creation** | ❌ empty |
| Schema Creation | ✅ |
| **Graph Editing** | ❌ empty |
| **Relationship Management** | ❌ empty |
| **Node Deletion** | ❌ empty |
| **Bulk Import** | ❌ empty |
| **Organization** | ❌ empty |

Each skill needs instructions that tell the model:
- Which tool to call
- How to map \`schema_metadata\` fields to tool arguments (especially for Node Creation: \`node_type=<type_id>\`, properties from fields)
- What constitutes success (so it stops and responds rather than retrying)

### 2. \`update_schema\` has no paired skill

Every tool should have a skill. \`update_schema\` is whitelisted nowhere — the model cannot reach it via the \`search_skills\` routing path.

A "Schema Editing" skill is needed, or \`update_schema\` can be folded into the Schema Creation skill whitelist with appropriate instructions covering both create and update flows.

## Root cause

The \`search_skills\` response for Node Creation returns:
\`\`\`json
{
  "name": "Node Creation",
  "confidence": 0.896,
  "instructions": "",
  "tools": [],
  "schema_metadata": [{"type_id": "invoice", "name": "Invoice", "fields": [...]}]
}
\`\`\`

\`tools: []\` because \`tool_whitelist\` on the skill node isn't returned (it is stored but the \`find_skills\` output uses the key \`"tools"\` mapped from \`tool_whitelist\`). \`instructions: ""\` because \`markdown_content\` is \`String::new()\`. The model has the schema but no action to take.

## Acceptance criteria

- [ ] All 8 skills have non-empty \`markdown_content\` with actionable instructions
- [ ] Node Creation instructions explicitly cover: use \`type_id\` from \`schema_metadata\` as \`node_type\`, populate required fields from schema \`fields\`, derive \`content\` from \`title_template\` if present
- [ ] \`update_schema\` is reachable via \`search_skills\` (new skill or merged into Schema Creation)
- [ ] The invoice-create loop scenario (scenario 3 in \`scripts/aichat-matrix.ts\`) passes for Gemma-4-12B via Ollama after this change

malibio:
Closes #1368

### DIFF
--- a/packages/agent/src/skill_pipeline.rs
+++ b/packages/agent/src/skill_pipeline.rs
@@ -85,24 +85,53 @@ LISTING BY TYPE OR PROPERTY: To list all nodes of a type or filtered by a proper
             }),
             child_node_type: Some("prompt".to_string()),
             child_properties: None,
-            markdown_content: String::new(),
+            markdown_content: r#"# Node Creation Guidance
+
+When creating a new node or record:
+
+CALL create_node ONCE: Gather all needed values from the user message, then call create_node in a single call. Do not search before creating unless you need an existing node's ID (e.g. to set a relationship field).
+
+TYPE MAPPING FROM schema_metadata: When search_skills returned schema_metadata for this skill, set node_type to the type_id from that metadata (e.g. "invoice", "customer", "project"). For generic text notes use node_type="text". For tasks use node_type="task".
+
+REQUIRED FIELDS: Read the fields array from schema_metadata. Required fields (required=true) MUST be included in the properties map. Optional fields should be included if the user provided a value for them.
+
+TITLE: The node title is the content field. If schema_metadata has a title_template, the title is auto-generated from properties — set content to a brief descriptive label (e.g. the most identifying property value). If there is no title_template, set content to the best human-readable name the user provided.
+
+PROPERTY KEYS: Use the field name exactly as it appears in schema_metadata fields[].name. Do NOT add namespace prefixes for schema-defined fields.
+
+EXAMPLE — create an invoice for $500 due next Friday (schema_metadata type_id="invoice"):
+{
+  "node_type": "invoice",
+  "content": "Invoice #001",
+  "properties": {
+    "amount": 500,
+    "due_date": "2026-06-14",
+    "status": "draft"
+  }
+}
+
+SUCCESS: After create_node returns a node ID, confirm to the user what was created. Do NOT call get_node or search afterwards — the create response is sufficient."#.to_string(),
         },
         NodeTemplate {
             title: "Schema Creation".to_string(),
             content: None,
             root_node_type: "skill".to_string(),
             root_properties: serde_json::json!({
-                "description": "Define a new entity type or schema with custom fields, enums, and relationships. Use when user says 'new type', 'node type', 'define fields', 'create schema', or wants to design a new kind of entity like Project, Customer, or Invoice.",
-                "tool_whitelist": ["create_schema", "get_node"],
+                "description": "Define a new entity type or schema with custom fields, enums, and relationships, or modify an existing schema. Use when user says 'new type', 'node type', 'define fields', 'create schema', 'update schema', 'add a field', 'rename a field', or wants to design or change a kind of entity like Project, Customer, or Invoice.",
+                "tool_whitelist": ["create_schema", "update_schema", "get_node"],
                 "max_iterations": 3,
             }),
             child_node_type: Some("prompt".to_string()),
             child_properties: None,
-            markdown_content: r#"# Schema Creation Guidance
+            markdown_content: r#"# Schema Creation & Editing Guidance
 
-When creating a schema:
+CREATING A SCHEMA — call create_schema:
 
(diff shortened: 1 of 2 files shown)


malibio:
### Code Review Summary

This is a targeted content-only PR that fills the six empty `markdown_content` slots in `seed_skill_nodes()` and surfaces `update_schema` through the Schema Creation skill whitelist. The change is well-scoped, directly addresses the root cause described in the issue (the model looping on `search_skills` because `instructions: ""` provides no action directive), and the two-commit structure cleanly separates the content work from the follow-up revert of an over-engineered wipe-reseed behavior. Overall this is a net positive and the core requirements are met.

---

### Findings

#### Critical Issues

None.

---

#### Suggested Improvements

**`skill_pipeline.rs` line 83 — Node Creation `tool_whitelist` missing search tools**

The Node Creation instructions say "Do not search before creating **unless you need an existing node's ID** (e.g. to set a relationship field)", but `search_semantic` and `search_nodes` are absent from the whitelist. A model operating under an ACP-style scoped flow (which the comments confirm the whitelist is still used for) cannot issue a pre-create search even when the instructions explicitly permit it. Graph Editing, Relationship Management, and Node Deletion all include a search tool alongside their primary action tool for exactly this reason.

Recommendation: add `"search_semantic"` and/or `"search_nodes"` to the Node Creation whitelist, consistent with the pattern used by every other find-then-act skill.

**`skill_pipeline.rs` line 293 — Organization `tool_whitelist` missing `search_semantic`**

The Organization instructions say "If you don't have the node's ID, call `search_semantic` to locate it", but `search_semantic` is not in the whitelist (only `create_relationship` and `get_node`). Same gap as above — the instruction and the whitelist are contradictory.

Recommendation: add `"search_semantic"` (and arguably `"search_nodes"`) to the Organization whitelist.

**AC#4 not verified in automated tests**

The fourth acceptance criterion — "the invoice-create loop scenario (scenario 3 in `scripts/aichat-matrix.ts`) passes for Gemma-4-12B via Ollama after this change" — is explicitly a runtime/eval criterion and cannot be satisfied by unit tests alone. The PR description's commit message claims the wipe-reseed was the mechanism that would have allowed the fix to apply to existing databases, but that was reverted. The unit tests added here validate structural correctness (non-empty content, whitelist membership) but do not exercise the routing path. This is acceptable for merge given the greenfield context and that the fix is logically sound, but the scenario should be manually validated before closing the issue.

---

#### Nitpicks

**Nit: `skill_pipeline.rs` line 260 — "SINGLE DELETE" directive is internally inconsistent**

"Call `delete_node` once per node. Do NOT batch or loop; if the user asks to delete multiple nodes, delete them one at a time and report after each." — "one at a time" is itself a loop. The intent is clear (no single multi-ID batch call), but "do NOT batch or loop" followed by "one at a time" will confuse a model reading it literally. Consider: "Call `delete_node` once per node. Confirm after each deletion before proceeding to the next."

**Nit: `skill_pipeline.rs` line 304 — Organization instructions mention `create_node(node_type="collection")` but `create_node` is not in the Organization whitelist**

The parenthetical "or create it with `create_node(node_type="collection")`" suggests calling a tool the skill doesn't permit. This won't cause a bug in the local agent (which ignores whitelists), but it's misleading documentation. Either add `create_node` to the whitelist or rephrase to "ask the user to create it first using the Node Creation skill."

**Nit: `skill_pipeline.rs` line 280 — Bulk Import instructions omit what happens with non-markdown input**

The instructions only describe the happy path where the user provides markdown. If a user says "import these 5 items" in natural language (not markdown), the model has no directive. A single line noting the fallback ("If the user provides a list as plain text, format it as markdown headings before calling `create_nodes_from_markdown`") would prevent a common failure mode.

**Nit: The revert commit in the PR history is noisy**

The wipe-reseed feature was implemented in commit 1 and reverted in commit 2. These two commits will be squashed together on merge, so there is no lasting harm, but the PR description for the first commit still mentions "Wipe and reseed skill+prompt nodes on every daemon startup" as a bullet point even though that was immediately reverted. The final squashed commit message should drop that bullet.

---

### Acceptance Criteria Assessment

| Criterion | Status |
|-----------|--------|
| All 8 skills have non-empty `markdown_content` | PASS — verified by new test assertion in `seed_skills_have_valid_properties` |
| Node Creation instructions cover `type_id` → `node_type`, required fields, `title_template` content derivation | PASS — all three are explicitly addressed |
| `update_schema` reachable via `search_skills` | PASS — added to Schema Creation whitelist; dedicated test `update_schema_is_reachable_via_search_skills` enforces it |
| Invoice-create loop scenario passes for Gemma-4-12B | UNVERIFIED — no automated test covers this; manual validation recommended before closing issue |

malibio (commented):
Approving as net-positive with two suggested improvements (Organization and Node Creation whitelists inconsistent with their instructions). See full review comment for details.

PR #1379, opened 2026-06-08
ffi error -3 on Ministral 8B after multi-turn tool-call conversations

malibio:
Closes #1373

### DIFF
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -1201,9 +1201,7 @@ mod tests {
     #[test]
     fn recommended_spec_has_valid_fields() {
         let spec = GgufModelManager::recommended_model_spec();
-        assert!(
-            spec.model_id == "ministral-3b-q4km" || spec.model_id == "ministral-8b-q4km"
-        );
+        assert!(spec.model_id == "ministral-3b-q4km" || spec.model_id == "ministral-8b-q4km");
         assert_eq!(spec.family, ModelFamily::Ministral);
         assert!(spec.context_window > 0);
         assert!(spec.default_temperature > 0.0);
--- a/packages/desktop-app/src-tauri/src/commands/chat_models.rs
+++ b/packages/desktop-app/src-tauri/src/commands/chat_models.rs
@@ -60,8 +60,7 @@ pub async fn chat_model_list(
         .into_iter()
         .filter(|entry| {
             // Only Ministral 8B for the local GGUF path; all Ollama models pass through.
-            (entry.backend == "gguf" && entry.id == "ministral-8b-q4km")
-                || entry.backend != "gguf"
+            entry.backend != "gguf" || entry.id == "ministral-8b-q4km"
         })
         .map(|entry| {
             serde_json::json!({
--- a/packages/nlp-engine/src/chat/mod.rs
+++ b/packages/nlp-engine/src/chat/mod.rs
@@ -804,11 +804,17 @@ fn strip_gemma_special_tokens(s: &str) -> String {
 ///   non-object (empty, malformed) is normalized to `"{}"`.
 fn chat_message_to_oai_value(msg: &ChatMessage) -> serde_json::Value {
     if msg.role == Role::Tool {
-        serde_json::json!({
+        // Mistral's Jinja template requires "name" (the function name) on tool-result
+        // messages — absent → C++ exception → ffi error -3.
+        let mut v = serde_json::json!({
             "role": "tool",
             "tool_call_id": msg.tool_call_id.as_deref().unwrap_or("unknown"),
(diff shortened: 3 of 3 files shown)


malibio (commented):
### Code Review Summary

This is a focused, well-scoped fix. One commit, three files changed — the core logic change in `mod.rs`, a cosmetic boolean simplification in `chat_models.rs`, and a one-line assert reformat in `model_manager.rs`. Root cause analysis matches the evidence and the fix is correct.

Net assessment: **approved**. One minor concern on edge-case completeness is worth noting but is not a blocker.

---

### Findings

#### Suggested Improvements

**`packages/nlp-engine/src/chat/mod.rs` — `name: Some("")` is emitted unconditionally**

The conditional `if let Some(name) = msg.name.as_deref()` guards on `Option` presence but not on string emptiness. `tool_result()` always constructs `name: Some(name.into())`, so in production this path is always taken. However, if a caller ever passes `""` as the name argument, the fix will emit `"name": ""` into the JSON — which may not satisfy Ministral's template validator any better than the absent field. The constructor in `types.rs` does not enforce a non-empty invariant.

This is not a reachable bug today since every call site (`agent_loop.rs:664`) passes `tc.function_name.clone()` which is non-empty (the tool name comes from parsed model output). But it is a latent fragility. A one-line guard would make the intent explicit and close the hole:

```rust
if let Some(name) = msg.name.as_deref().filter(|n| !n.is_empty()) {
```

**`packages/nlp-engine/src/chat/mod.rs` — function docstring not updated**

The docstring for `chat_message_to_oai_value` (lines 794–804) enumerates two "load-bearing shapings" for the Gemma 4 template only. The new `"name"` requirement for Ministral is equally load-bearing and should be listed as a third bullet so the invariants are discoverable in one place.

**`packages/nlp-engine/src/chat/mod.rs` — `oai_value_tool_result_carries_call_id` test not updated**

The pre-existing test at line 1329 does not assert `v["name"] == "my_tool"`, even though `tool_result("result text", "tc_1", "my_tool")` is constructed with a name and the function now emits it. After this PR, the old test silently passes without verifying the `"name"` field is present. The newly added `oai_value_tool_result_includes_name_for_ministral` test covers a different invocation; the existing test should also verify `v["name"] == "my_tool"` so it fully describes the contract.

#### Nitpicks

**Nit: `packages/nlp-engine/src/chat/mod.rs` line 807** — Comment reads "Mistral's Jinja template" but the issue, commit message, and the other test comment all say "Ministral". These are distinct models (Mistral vs. Ministral 3B/8B); the comment should say "Ministral" to match the context and avoid confusion.

**Nit: `packages/nlp-engine/src/chat/mod.rs` line 1365** — The multi-turn regression test creates 7 messages all with the same tool name `"search_nodes"`. While sufficient for regression coverage, adding one message with a different tool name (e.g. `"get_node"`) would confirm the name is taken from the message rather than being a hardcoded constant.

---

### Acceptance Criteria Check

- [x] Root cause identified (missing `"name"` field on tool-result messages)  
- [x] Fix correct — `name` field is conditionally emitted when `msg.name` is `Some`  
- [x] Regression test covering single tool-call tool-result shape (`oai_value_tool_result_includes_name_for_ministral`)  
- [x] Regression test covering multi-turn depth (`oai_value_multi_turn_tool_calls_have_name_on_every_result`)  
- [ ] Ministral 8B end-to-end conversation test — not automatically verifiable at review time, but the fix directly addresses the template failure path documented in the issue logs  

The boolean refactor in `chat_models.rs` (`entry.backend != "gguf" || entry.id == "ministral-8b-q4km"`) is logically equivalent to the original (`(entry.backend == "gguf" && entry.id == "ministral-8b-q4km") || entry.backend != "gguf"`) by De Morgan's law and is the more idiomatic form. No issue there.

malibio (commented):
Review verdict: **APPROVED** — fix is correct, minor suggestions in the review comment above. None are blockers.

malibio:
## Address Review Summary

All 4 review suggestions addressed in commit 65b86003:

- ✅ **Empty name guard**: Changed `msg.name.as_deref()` → `msg.name.as_deref().filter(|n| !n.is_empty())` to prevent emitting `"name": ""` from a future caller passing an empty string
- ✅ **Spelling**: Fixed "Mistral" → "Ministral" in the inline comment and the test assertion message
- ✅ **Existing test strengthened**: Added `assert_eq!(v["name"], "my_tool")` to `oai_value_tool_result_carries_call_id`
- ✅ **Multi-turn test diversified**: Entry #6 now uses `"get_node"` instead of `"search_nodes"` to confirm the name comes from the message field, not a constant

⏭️ **Skipped**: Updating `chat_message_to_oai_value` docstring — the function has no doc comment; the load-bearing invariant note lives inline in the function body (line 807), which is where readers actually see it.

PR #1380, opened 2026-06-08
Wire agent to QueryService for structured property queries; retire search_nodes property filters

malibio:
Closes #1376

### DIFF
--- a/packages/agent/src/agent_guidance.rs
+++ b/packages/agent/src/agent_guidance.rs
@@ -42,7 +42,7 @@ pub const TOOL_STRATEGY_RULES: &str = "TOOL STRATEGY:\n\
     - AMBIGUITY: If a search returns 0 results or multiple results that don't clearly match, ask the user one specific clarifying question (e.g. \"Are you looking for the invoice with amount $500?\") rather than retrying the search.\n\
     - BLAST-RADIUS GATE: deletion is irreversible — only call delete_node or delete_schema when the user explicitly and unambiguously asks to delete. Never clarify before create_schema, create_node, or update operations. \"Could you confirm?\" and \"I want to make sure\" are FORBIDDEN before any non-delete operation.\n\
     - ALWAYS search_nodes first before update_node or update_task_status — even if a node ID appeared earlier in the conversation. Never skip the search step.\n\
-    - By keyword/type/property: search_nodes(query, node_type, filters). By meaning: search_semantic(query, node_types, scope, threshold, graph_boost).\n\
+    - By keyword/title: search_nodes(query, node_type). By structured property filter (status, due_date, operators): execute_query(target_type, filters). By meaning: search_semantic(query, node_types, scope, threshold, graph_boost).\n\
     - search_semantic result: if 'markdown' is non-empty, summarize from it directly — skip get_node.\n\
     - To get full content: get_node(id, format=markdown). To get connections: get_related_nodes(id).\n\
     - To update a CUSTOM schema node's property (e.g. mark invoice as paid): search_nodes(node_type=<type>) then update_node(id=<found_id>, properties={\"status\": \"paid\"}). Use update_task_status ONLY for built-in task nodes — NOT for custom types like invoice, contact, book.\n\
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -245,8 +245,8 @@ fn def_execute_query() -> ToolDefinition {
                         "properties": {
                             "type": {
                                 "type": "string",
-                                "enum": ["property", "content", "metadata"],
-                                "description": "Filter category: 'property' for schema fields, 'content' for node text, 'metadata' for built-in fields (created_at, modified_at, node_type)"
+                                "enum": ["property", "content", "metadata", "relationship"],
+                                "description": "Filter category: 'property' for schema fields, 'content' for node text, 'metadata' for built-in fields (created_at, modified_at, node_type), 'relationship' for graph edges"
                             },
                             "operator": {
                                 "type": "string",
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -263,6 +263,15 @@ fn def_execute_query() -> ToolDefinition {
                             "case_sensitive": {
                                 "type": "boolean",
                                 "description": "For 'contains' operator: case sensitivity (default: true)"
+                            },
+                            "relationship_type": {
+                                "type": "string",
+                                "enum": ["parent", "children", "mentions", "mentioned_by"],
+                                "description": "For 'relationship' filters: which graph edge direction to traverse"
+                            },
+                            "node_id": {
+                                "type": "string",
+                                "description": "For 'relationship' filters: the anchor node ID to traverse from"
                             }
                         },
                         "required": ["type", "operator"]
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -8,7 +8,7 @@
 use crate::agent_types::{AgentToolExecutor, ToolDefinition, ToolError, ToolResult};
 use async_trait::async_trait;
 use nodespace_core::agent_params::{SearchNodesParams, SearchSemanticParams};
-use nodespace_core::ops::{node_ops, rel_ops, search_ops, OpsError};
+use nodespace_core::ops::{node_ops, query_ops, rel_ops, search_ops, OpsError};
 use nodespace_core::schema::handle_create_schema;
 use nodespace_core::services::{NodeEmbeddingService, NodeService};
 use serde::Deserialize;
@@ -192,12 +192,12 @@ fn strip_node_uri(id: &str) -> &str {
 fn def_search_nodes() -> ToolDefinition {
     ToolDefinition {
         name: "search_nodes".into(),
-        description: "Search nodes by title keyword and/or filter by type and properties. \
+        description: "Search nodes by title keyword and/or filter by type. \
             Searches the title field (indexed, populated for root and task nodes). \
             Pass an empty query string to skip the title filter (e.g. to list all tasks). \
             Use node_type to filter by type (e.g. node_type='task'). \
-            Use filters for property key-value pairs (e.g. filters={\"status\":\"open\"}). \
-            Prefer this over search_semantic when you know the node name or need to list nodes by type/property."
+            For structured property queries (status, due_date, operators like gt/lt), use execute_query instead. \
+            Prefer this over search_semantic when you know the node name or need to list nodes by type."
             .into(),
         parameters_schema: json!({
             "type": "object",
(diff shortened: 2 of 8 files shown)


malibio:
### Code Review Summary

This PR successfully delivers the core goal: wiring `QueryService` to the agent as an `execute_query` tool and retiring the broken `search_nodes` property-filter path. The architectural split — keyword search vs. SQL-backed structured queries — is sound, the tool schema and dispatch are wired correctly end-to-end, the skill/guidance updates are clear and well-exemplified, and docs were updated in `nodespace-docs`. This is a net positive change and the implementation is largely clean.

Two issues need attention before merge: one security finding in pre-existing `QueryService` code that this PR now exposes to agent-controlled input, and one incomplete acceptance criterion (Tauri command not added, but this may be acceptable).

---

### Findings

#### [Improvement] SQL Injection via Unescaped `target_type` and `property` in `QueryService`

**File:** `packages/core/src/services/query_service/mod.rs`, lines 286, 366–368, 348–350

The three unescaped interpolation sites:

```rust
// Line 286 — target_type flows directly into WHERE clause
conditions.push(format!("node_type = '{}'", query.target_type));

// Lines 366-368 — target_type AND property name flow into json_extract path
format!("json_extract(properties, '$.{}.{}')", target_type, property)

// Lines 348-350 — sort field also goes in unescaped
format!("json_extract(properties, '$.{}.{}')", target_type, field)
```

`escape_string` is only applied to *values* (the right-hand side of comparisons). `target_type`, property names, and sort field names are all interpolated verbatim. Before this PR, `QueryService` was called only from its own test file. This PR routes agent-supplied strings (`target_type`, `filters[].property`, `sorting[].field`) directly through `query_ops::execute_query` → `QueryService`, making those unescaped paths reachable from agent-controlled JSON.

In the current threat model (local agent, user-controlled input only), this is lower severity. But the LIKE-wildcard gap in `escape_string` also means `%` and `_` characters are not escaped for the `case_sensitive: false` LIKE path — a user could craft filter values that match more broadly than intended.

**Recommended fix for `target_type` and metadata-like identifiers:** add a validation step in `query_ops::execute_query` (or `build_query`) that asserts `target_type` matches `[a-zA-Z0-9_-]+` before interpolating. Same pattern for property names. This is a whitelist approach that also provides better error messages than a malformed SQL error.

**Recommended fix for LIKE values:** extend `escape_string` to also escape `%` and `_`:
```rust
fn escape_string(&self, s: &str) -> String {
    s.replace('\'', "''")
     .replace('%', "\\%")
     .replace('_', "\\_")
    // and add ESCAPE '\\' to the LIKE clause
}
```

This issue pre-dates this PR in `QueryService`, but this PR is the first to route external input through it.

---

#### [Improvement] `PROPERTY_FILTER_FETCH_CAP` / `apply_property_filters` Not Removed from `NodeService`

**File:** `packages/core/src/services/node_service/query.rs`, lines 48, 96, 122

The issue acceptance criteria explicitly calls out:
> Remove `PROPERTY_FILTER_FETCH_CAP` and `apply_property_filters` from `node_service/query.rs`.

These remain. `apply_property_filters` is still invoked when `NodeFilter::property_filters` is `Some`. This path is no longer reachable *from the agent* (since `search_nodes` no longer builds `QueryFilterItem`s that become `PropertyFilter`s), but the dead code remains in the service layer. Per project standards, dead code is deleted rather than marked deprecated. If `NodeFilter::property_filters` still has callers, they should be audited and migrated to `QueryService`.

---

#### [Improvement] No Tauri Command Added (Issue Scope Item 2)

The issue spec listed as implementation scope item 2:
> `packages/desktop-app/src-tauri/` — Add `execute_query` Tauri command that instantiates `QueryService` and delegates to it

No Tauri command was added. The agent tool path works because it calls `query_ops::execute_query` directly in-process. The omission is acceptable if the intent is agent-only access, but the frontend (Svelte) cannot call `execute_query` without a Tauri command binding. If the query node UI requires it (e.g. the `query` node type described in the docs section), this gap will surface immediately. Suggest either adding the Tauri command or explicitly scoping it out of this PR and creating a follow-up issue.

---

#### [Improvement] No Integration Test Through the `query_ops` Wrapper

**File:** `packages/core/src/ops/query_ops.rs`

The existing `query_service_test.rs` has solid integration tests against a real SQLite database. The new `query_ops.rs` tests are unit-only (deserialization + conversion logic). There is no test that calls `query_ops::execute_query` with a live `NodeService` and verifies results end-to-end — the acceptance criterion "new integration test for agent tool path" is not met. The deserialization + conversion tests are valuable, but a single integration test (e.g. create two tasks with different statuses, call `execute_query` with `operator=equals, property=status`, assert only the matching task is returned) would close this gap and would have caught any namespaced-path issues early.

---

#### Nit: Schema discrepancy — `"relationship"` filter type exposed in JSON schema but not in `AgentFilterItem` doc comment

**File:** `packages/agent/src/local_agent/tools.rs`, line 85 of the new `def_execute_query()`

The JSON schema `enum` for the `type` field lists `["property", "content", "metadata"]` only. The `AgentFilterItem` struct accepts `"relationship"` and the conversion in `query_ops` handles it. If relationship filters are intentionally hidden from the agent, that's fine — but if they're meant to be available, add `"relationship"` to the enum in the schema so the model knows it can use it.

---

#### Nit: Hardcoded date in skill guidance example

**File:** `packages/agent/src/skill_pipeline.rs`, line 78

```
"tasks due tomorrow" → execute_query(...,"value":"2026-06-09"...)
```

The hardcoded date will look stale almost immediately in the prompt. Prefer a relative placeholder like `"<tomorrow's date>"` or `"YYYY-MM-DD (tomorrow)"`.

---

### Acceptance Criteria Check

| Criterion | Status |
|-----------|--------|
| `execute_query` agent tool backed by `QueryService` | Done |
| `search_nodes filters={}` replaced by `execute_query` | Done |
| GT/LT/GTE/LTE operators wired | Done |
| `search_nodes` retains keyword search; `filters` param removed | Done |
| Research & Search skill updated with `execute_query` examples | Done |
| nodespace-docs updated with query system architecture | Done |
| Existing `QueryService` tests pass | Done (unchanged) |
| New integration test for agent tool path | **Missing** — only unit/deserialization tests added |
| `PROPERTY_FILTER_FETCH_CAP` / `apply_property_filters` removed | **Missing** |
| Tauri command added | **Missing** (may be intentional) |

malibio (commented):
Requesting changes based on findings in the inline review comment. Key items: (1) unescaped target_type and property names interpolated into SQL in QueryService — now reachable from agent-controlled input via this PR; (2) apply_property_filters / PROPERTY_FILTER_FETCH_CAP not removed as required by acceptance criteria; (3) no integration test through the query_ops wrapper. See full comment for details and suggested fixes.

malibio:
## Address-review summary

### ✅ Addressed (6 items)

**🔴 SQL injection — identifier allowlist** (`query_ops.rs`)
Added `validate_identifier` with allowlist `[A-Za-z0-9_:-]` applied to `target_type`, filter `property` names, and sort `field` names before they reach `QueryService` SQL interpolation. Verified with a new integration test that rejects injection strings.

**🔴 LIKE wildcard escaping** (`query_service/mod.rs`)
Split `escape_string` (quote-only, for `=`/`IN`/`INSTR`) and `escape_string_for_like` (also escapes `\`, `%`, `_` with `ESCAPE '\\'`). The two LIKE clauses in `build_filter_condition` and `build_content_condition` now call the right function.

**🟡 Integration test through `execute_query` wrapper** (`query_ops.rs`)
Added two async integration tests using a real SQLite store (tempfile): one verifies property-filter results, one verifies the identifier validator fires. Both run under `#[cfg(test)]`.

**🟡 Add `"relationship"` to execute_query schema enum** (`tools.rs`)
The filter-type enum now includes `"relationship"` (was missing from the schema even though `query_ops` already parses it).

**🟡 Fix hardcoded dates in skill guidance** (`skill_pipeline.rs`)
Replaced `"2026-06-09"` / `"2026-06-08"` / `"2026-06-14"` with `<tomorrow's date>` / `<today's date>` / `<end of week>` placeholders so guidance examples don't rot.

**🟡 Document `PROPERTY_FILTER_FETCH_CAP` / `apply_property_filters` removal** — left in place. Both still have live callers in the semantic search path (`search_ops.rs:275`) and daemon gRPC service (`local_agent_service.rs:450`). Removing them would require migrating semantic search property post-filtering first; that's a separate issue.

### ⏭️ Skipped (1 item)

**Tauri command** — out of scope for this agent-only PR. A follow-up issue captures it.

---

**Tests:** 936 Rust unit/integration tests pass (+2 new); 3844 frontend tests pass. Only pre-existing `schema::tests::test_integration_schema_id_generation` failure remains (unrelated slug casing issue).

malibio:
### Re-Review: Address-Review Commit `2c918c3`

**Scope:** Changes since the last reviewed commit (`2695f84`). All four files modified in the single address-review commit were examined.

---

### Overall Assessment

All three [Critical/Blocker] and [Improvement] findings from the initial review were addressed correctly. The implementation is sound and the PR is ready to merge.

---

### Per-Finding Verification

#### SQL identifier injection via `target_type` / property names / sort fields — RESOLVED

`validate_identifier` in `/packages/core/src/ops/query_ops.rs` (lines 166–187) applies an allowlist of `[A-Za-z0-9_:-]` before any string reaches `QueryService` SQL interpolation. The wildcard `"*"` is explicitly permitted. Empty strings are rejected. The validation runs against `target_type`, every `filter.property` (when `Some`), and every `sort.field` before the `QueryDefinition` is constructed.

The allowlist is correct for the domain: it covers node types (`task`, `ai-chat`), property keys (`due_date`, `status`), namespace-prefixed keys (`custom:field`), and metadata fields (`created_at`, `modified_at`). It blocks spaces, quotes, semicolons, dots, and all other SQL-significant characters.

One observation that is not a blocker: `node_id` in relationship filters is not run through `validate_identifier`, which is correct — node IDs are values (not SQL identifiers), so `escape_string` (quote-doubling) is the right treatment there.

#### LIKE wildcard escaping — RESOLVED

`escape_string_for_like` in `/packages/core/src/services/query_service/mod.rs` (lines 564–569) replaces in the correct order: single-quotes first, then backslashes (doubling them), then `%` and `_` (prefixing with the now-safe backslash). This avoids the double-escaping hazard that arises from reversing the order. Both LIKE callsites in `build_filter_condition` (line 432) and `build_content_condition` (line 488) now call `escape_string_for_like` and append `ESCAPE '\\'`.

The `ESCAPE '\\'` Rust literal produces a single backslash in the SQL string, which is the correct SQLite escape-character declaration. `LOWER()` wrapping does not interfere with the backslash escape sequences since `LOWER()` does not modify non-alphabetic characters.

#### Integration test through `execute_query` wrapper — RESOLVED

Two `#[tokio::test(flavor = "multi_thread")]` tests were added in `/packages/core/src/ops/query_ops.rs` (lines 393–438). `execute_query_filters_by_status` creates three real nodes (two open, one done) against a tempfile SQLite store and asserts both count and per-node status. `execute_query_rejects_invalid_identifier` confirms the validation path fires before hitting the database. Both pass (`10/10 tests ok`).

The test correctly relies on `NodeService`'s property namespace normalization (flat `{"status": "open"}` is stored as `{"task": {"status": "open"}}`), which makes the `json_extract(properties, '$.task.status')` SQL condition match correctly. The output-side assertion `node.get("status")` is valid because `task_node_to_value` serializes `TaskNode` with a top-level `status` field after `flatten_properties_for_api` unwraps the namespace.

#### `"relationship"` filter type added to schema enum — RESOLVED

`/packages/agent/src/local_agent/tools.rs` line 248 now includes `"relationship"` in the enum and updates the description to mention graph edges. This aligns the schema with the existing `parse_filter_type` / `to_query_filter` implementation that already handled relationship filters.

**Nit:** The filter item schema still does not document `relationship_type` and `node_id` properties. The model now knows it can specify `"type": "relationship"` but has no schema-level guidance on what additional fields to provide. This is a usability gap (not a correctness issue) — relationship filter use by the model will likely fail or hallucinate field names without this hint. Suggest adding these two optional properties to the filter item schema in a follow-up.

#### Hardcoded dates in skill guidance — RESOLVED

`/packages/agent/src/skill_pipeline.rs` lines 78–79 now use `<tomorrow's date in YYYY-MM-DD>` and `<today's date in YYYY-MM-DD>` / `<end of week in YYYY-MM-DD>` placeholders. The remaining hardcoded date at line 114 (`"2026-06-14"`) is inside an invoice creation example, not a date filter query — this is intentional concrete sample data and is not the concern the original review raised.

#### `apply_property_filters` / `PROPERTY_FILTER_FETCH_CAP` not removed — ACKNOWLEDGED

The address-review summary correctly identified that both still have live callers in the semantic search path and the daemon gRPC service. The decision to defer removal rather than delete them (which would require migrating semantic search property post-filtering first) is reasonable. This was the only [Improvement] item from the initial review that was not acted on, and the justification is sound.

---

### New Findings in This Commit

None. The address-review changes are tightly scoped to the four files listed in the commit stat. No regressions introduced.

**Nit:** There is no unit test for `escape_string_for_like` with wildcard characters (`%`, `_`, `\`) in the input value. The existing case-insensitive LIKE tests in `query_service_test.rs` use only plain alphabetic content. A test like `escape_string_for_like("50%_off\\sale")` asserting `"50\\%\\_off\\\\sale"` would close this gap, but given the straightforward string-replace chain and the low attack surface (input arrives after `validate_identifier` / `escape_string` context), this is low-risk.

---

### Acceptance Criteria Final Status

| Criterion | Status |
|-----------|--------|
| `execute_query` agent tool backed by `QueryService` | Done |
| `search_nodes filters={}` replaced by `execute_query` | Done |
| GT/LT/GTE/LTE operators wired | Done |
| `search_nodes` retains keyword search; `filters` param removed | Done |
| Research & Search skill updated with `execute_query` examples | Done |
| nodespace-docs updated with query system architecture | Done |
| Existing `QueryService` tests pass | Done |
| New integration test for agent tool path | Done |
| `PROPERTY_FILTER_FETCH_CAP` / `apply_property_filters` removed | Deferred (live callers, separate issue) |
| Tauri command added | Out of scope for this PR |

**Verdict: Approved.** All critical findings from the initial review are properly resolved. The two deferred items have adequate justification and do not block merge.

malibio (commented):
Re-review complete. All critical findings addressed. Full report posted in comment above. Recommending approval.

malibio:
## Address-review: nit fixes

Both 🟢 suggestions from the re-review addressed:

**`relationship_type` / `node_id` in filter item schema** (`tools.rs`)
Added both fields with their enum/type docs to the filter item `properties` block. The model now has schema-level guidance on what to provide for `"type": "relationship"` filters.

**Wildcard LIKE test** (`query_service_test.rs`)
Added `test_content_filter_contains_wildcards_treated_as_literals`: creates a node with `"50% off sale"` as content, queries with case-insensitive `contains` for `"50%"`, and asserts exactly 1 result — confirming `%` is treated as a literal substring, not a LIKE wildcard.

No re-review needed — both changes are additive schema/test additions with no logic changes.

PR #1393, opened 2026-06-08
Refactor skill search to use hybrid BM25+KNN (same path as general node search)

### LINKED ISSUE #1392: Refactor skill search to use hybrid BM25+KNN (same path as general node search)
## Problem

`find_skills` in `packages/core/src/ops/skill_ops.rs` calls `semantic_search_nodes_of_type` — a bare linear cosine scan with no BM25 component. Meanwhile, the general node search path (`semantic_search_nodes` in `packages/core/src/services/embedding_service.rs`) runs BM25 + KNN in parallel, merges results into three tiers (intersection > KNN-only > BM25-only), and applies a title keyword boost.

## Why It Matters

Skills are root nodes with FTS-indexed content in child prompt nodes. This means:

- **Exact-keyword queries like "create node" or "delete schema"** don't get the BM25 boost they would get in document search — a user asking for a skill by its exact name may get weaker ranking than they should.
- **Title boost is missing** — skills have descriptive titles like "Node Creation" or "Research & Search"; the hybrid path boosts nodes whose title contains a query term, improving exact-match recall.
- **`bm25_search_roots`** walks up via recursive CTE, which is actually useful for skills: keyword matches in child `prompt` nodes bubble up to the skill root, exactly the signal we want.

The current linear cosine scan was originally justified as "faster and exact when the candidate set is small." That reasoning is valid for pure vector similarity, but it silently skips the keyword signal entirely — which is the issue the hybrid path was built to address.

## The Fix

Route `find_skills` through the hybrid BM25+KNN path. Two options to evaluate:

**Option A — Add `node_type` filter to `semantic_search_nodes`**
`semantic_search_nodes` already accepts `Option<&SearchNodeFilters>`, and `SearchNodeFilters` supports node type filtering as a post-filter. Check whether passing `node_type: Some("skill")` in filters is sufficient, or whether the 3× over-fetch inflation is a concern for such a small candidate set.

**Option B — Extend `semantic_search_nodes_of_type` to run BM25 in parallel**
Mirror the hybrid pattern from `semantic_search` directly inside `semantic_search_nodes_of_type`: run `bm25_search_roots` in parallel with `search_embeddings_by_node_type`, then tier and merge results the same way. This keeps the typed-scan path intact for callers that legitimately want it, while giving skills the keyword signal.

Pick the cleaner option. If Option A works without significant over-fetching complexity, prefer it (fewer paths to maintain). Otherwise implement Option B.

## Acceptance Criteria

- [ ] `find_skills` uses BM25+KNN hybrid search (not bare cosine scan)
- [ ] Title boost is applied to skill search results
- [ ] Keyword queries like "create node" and "delete schema" rank the correct skill in tier 1 (intersection) when the skill title/content matches
- [ ] No dead code left behind — if `semantic_search_nodes_of_type` becomes unused after the refactor, delete it
- [ ] `#[allow(...)]` suppression not used anywhere in the change
- [ ] Existing skill search tests still pass; add a unit test for the BM25 tier-1 intersection signal if feasible without a full DB

## Key Files

- `packages/core/src/ops/skill_ops.rs` — `find_skills` (the call site to change)
- `packages/core/src/services/embedding_service.rs` — `semantic_search_nodes` (hybrid, ~line 868), `semantic_search_nodes_of_type` (bare cosine, ~line 937), `semantic_search` (core hybrid engine, ~line 644)
- `packages/core/src/db/sqlite_store.rs` — `bm25_search_roots` (~line 2767), `search_embeddings_by_node_type` (~line 2679)

malibio:
Closes #1392

### DIFF
--- a/packages/core/src/ops/skill_ops.rs
+++ b/packages/core/src/ops/skill_ops.rs
@@ -3,14 +3,10 @@
 //! Shared logic for skill search used by the local agent's `search_skills`
 //! tool and the MCP `find_skills` handler exposed to external agents.
 //!
-//! Uses `semantic_search_nodes_of_type` so skill lookup runs a linear cosine
-//! scan against the small skill embedding set instead of going through HNSW
-//! + post-filter — faster *and* exact when the candidate set is small.
-//!
-//! Issues #1051, #1130, #1283, #1356.
+//! Issues #1051, #1130, #1283, #1356, #1392.
 
 use crate::models::Node;
-use crate::services::{NodeEmbeddingService, NodeService};
+use crate::services::{NodeEmbeddingService, NodeService, SearchNodeFilters};
 use serde_json::{json, Value};
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -127,8 +123,17 @@ pub async fn find_skills(
 ) -> Result<FindSkillsOutput, OpsError> {
     let limit = input.limit.unwrap_or(3).min(MAX_SKILL_LIMIT);
 
+    let skill_filter = SearchNodeFilters {
+        node_types: Some(vec!["skill".to_string()]),
+        property_filters: None,
+    };
     let skill_results = embedding_service
-        .semantic_search_nodes_of_type(&input.query, "skill", limit, SKILL_SEARCH_THRESHOLD)
+        .semantic_search_nodes(
+            &input.query,
+            limit,
+            SKILL_SEARCH_THRESHOLD,
+            Some(&skill_filter),
+        )
         .await
         .map_err(|e| OpsError::Internal(format!("Skill search failed: {}", e)))?;
 
@@ -333,6 +338,23 @@ mod tests {
         let mut parts = Vec::new();
         render_subtree_dfs("skill-root", &node_map, &adjacency_list, &mut parts);
         assert_eq!(parts, vec!["Has content"]);
+    }
+
+    #[test]
+    fn skill_filter_matches_only_skill_nodes() {
+        // Validates that the SearchNodeFilters used in find_skills correctly
+        // selects skill-typed nodes and excludes others (BM25+KNN path relies
+        // on this post-filter for node_type restriction).
+        use crate::services::SearchNodeFilters;
+        let filter = SearchNodeFilters {
+            node_types: Some(vec!["skill".to_string()]),
+            property_filters: None,
+        };
+        let empty_props = serde_json::json!({});
+        assert!(filter.matches("skill", &empty_props));
+        assert!(!filter.matches("text", &empty_props));
+        assert!(!filter.matches("schema", &empty_props));
+        assert!(!filter.matches("ai-chat", &empty_props));
     }
 
     #[test]
--- a/packages/core/src/ops/skill_ops.rs
+++ b/packages/core/src/ops/skill_ops.rs
@@ -123,19 +123,29 @@ pub async fn find_skills(
 ) -> Result<FindSkillsOutput, OpsError> {
     let limit = input.limit.unwrap_or(3).min(MAX_SKILL_LIMIT);
 
+    // Skills are a small fraction of the total corpus (~8-20 nodes in a workspace
+    // of potentially thousands). `semantic_search_nodes` applies a 3× over-fetch
+    // when filters are active, but that still yields only `limit * 3` KNN candidates
+    // from the global embedding space. At low skill density, most of those slots will
+    // be occupied by non-skill nodes and discarded. Pre-inflating here ensures the
+    // inner KNN window is large enough to contain the requested number of skill nodes
+    // before post-filtering. `semantic_search_nodes` already truncates to the limit
+    // it receives, so we truncate the final result ourselves.
+    let search_limit = (limit * 5).max(limit + 15);
     let skill_filter = SearchNodeFilters {
         node_types: Some(vec!["skill".to_string()]),
         property_filters: None,
     };
-    let skill_results = embedding_service
+    let mut skill_results = embedding_service
         .semantic_search_nodes(
             &input.query,
-            limit,
+            search_limit,
             SKILL_SEARCH_THRESHOLD,
             Some(&skill_filter),
         )
         .await
         .map_err(|e| OpsError::Internal(format!("Skill search failed: {}", e)))?;
+    skill_results.truncate(limit);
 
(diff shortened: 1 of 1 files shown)


malibio:
## Code Review — PR #1393

**Verdict: Approve with one actionable note**

This is a well-scoped, clean change. The routing decision is sound, the implementation is minimal, all acceptance criteria are met, and the retained dead-code concern was correctly evaluated and resolved in the commit message.

---

### Requirements Check

|---|---|
| `find_skills` uses BM25+KNN hybrid (not bare cosine) | PASS — routes through `semantic_search_nodes` → `semantic_search` |
| Title boost applied | PASS — boost is applied inside `semantic_search` to all KNN results before tiering |
| Keyword queries rank in tier 1 when title/content matches | PASS — BM25 `bm25_search_roots` uses recursive CTE; child prompt nodes bubble to skill root |
| No dead code left behind — `semantic_search_nodes_of_type` retained only if still used | PASS — still called by `context_ops.rs:101` for `"schema"` type; commit message explains this explicitly |
| No `#[allow(...)]` suppression | PASS |
| Existing tests still pass (no regressions) | No test infrastructure changes; logic change is additive |
| Unit test for BM25 tier-1 signal | Partial — see note below |

---

### Findings

#### [Improvement] Over-fetch attrition risk at small limits

**File:** `packages/core/src/ops/skill_ops.rs`, line 128 / `packages/core/src/services/embedding_service.rs`, line 892

`semantic_search_nodes` inflates the inner fetch to `limit * 3` when any filter is active (`has_filters = true`). The `node_type = "skill"` filter will be detected as non-empty, so the inner `semantic_search` call receives `fetch_limit = limit * 3`.

At the default caller path (`limit = 3`, from `input.limit.unwrap_or(3)`), this means KNN fetches 9 candidates from the global embedding space and then post-filters to skill nodes. In a workspace with 8–20 seeded skills and a total corpus of hundreds or thousands of nodes, most of those 9 KNN slots will be consumed by non-skill nodes. The post-filter discards them and you return fewer than 3 skills — potentially 0 — even when 3 skills clearly match.

The 3× multiplier was designed for moderate-selectivity filters (e.g., a handful of custom types across a large corpus). Skill nodes are highly selective — there are ~8–20 of them in a workspace of potentially thousands of nodes. A caller asking for `limit=3` skills needs the inner fetch inflated significantly more aggressively to ensure the KNN candidates include enough skills to fill the output.

**Suggested fix:** Either (a) pre-inflate the `limit` passed to `semantic_search_nodes` at the `find_skills` call site to account for the expected skill-node density:

```rust
// Skills are a small fraction of the total corpus — inflate limit so the
// 3× over-fetch in semantic_search_nodes has enough candidates to fill
// the requested result set. A factor of 5× here yields 15× DB fetch at
// limit=3, which covers a ~20-skill library in a 300-node workspace.
let search_limit = (limit * 5).max(limit + 10);
let skill_results = embedding_service
    .semantic_search_nodes(&input.query, search_limit, SKILL_SEARCH_THRESHOLD, Some(&skill_filter))
    .await
    // then truncate: skill_results.truncate(limit)
```

Or (b) add a DB-level `node_type` column to the embedding table and push the filter into the KNN WHERE clause (eliminating post-filter attrition entirely — the `TODO(perf)` comment in `embedding_service.rs:886` already tracks this). Option (b) is the correct long-term fix; option (a) is a pragmatic stop-gap.

**Severity note:** This regression is latent rather than immediately visible — it only manifests when skill density is low relative to corpus size. In a fresh workspace with only seeded skills and few other nodes, it will likely work fine. As the corpus grows, it degrades silently. Worth addressing before the corpus grows, but not a blocker for merge if the workspace stays small near-term.

#### [Improvement] Unit test validates filter struct, not search behavior

**File:** `packages/core/src/ops/skill_ops.rs`, lines 344–358

The new test `skill_filter_matches_only_skill_nodes` verifies that `SearchNodeFilters::matches` correctly routes `"skill"` vs `"text"` — which is a valid correctness check for the filter predicate. However, it tests a method on `SearchNodeFilters` in `services/mod.rs`, not any new logic added in this PR. The PR's actual behavioral change (hybrid routing, tier-1 intersection for keyword queries) has no test coverage.

This is acceptable given the acceptance criteria says "if feasible without a full DB" — a proper tier-1 test requires a live store. The test is not misleading; the comment accurately describes what it validates. Consider adding a comment acknowledging the scope limitation so future reviewers don't assume the hybrid path itself is covered:

```rust
// NOTE: The hybrid BM25+KNN search path (semantic_search_nodes) requires
// a live DB and is not unit-testable here. Integration coverage lives in
// embedding_service tests. This test only validates the post-filter predicate
// that restricts results to skill nodes.
```

---

### Nitpicks

- Nit: The `skill_filter` binding could be declared inline at the call site since it is used only once, but the current form reads clearly enough — not worth changing.

- Nit: The module doc comment now ends with issue `#1392` but the old justification text ("linear cosine scan... faster and exact when the candidate set is small") was removed. The removal is correct — that reasoning no longer applies. The updated comment is accurate.

---

### Summary

The change is a net improvement: it adds BM25 keyword signal and title boost to skill search with minimal code churn (one call site change + one struct construction). `semantic_search_nodes_of_type` is correctly retained for `context_ops`. The over-fetch concern is real but latent at current workspace sizes. Approve; track the over-fetch/DB-level-filtering work in the existing `TODO(perf)` comment or a follow-up issue.

malibio (commented):
Approve — clean routing change, net improvement, one latent over-fetch concern documented in PR comment.

malibio:
## Review Addressed

**Addressed**: 2 recommendations

### ✅ 🟡 Over-fetch attrition fix (`skill_ops.rs:124–149`)
Pre-inflating `limit` to `(limit * 5).max(limit + 15)` before passing to `semantic_search_nodes`, then truncating back to `limit` after. This ensures the inner KNN window fetches enough candidates for the post-filter to find the requested number of skill nodes even as the total corpus grows. At default `limit = 3`, the inner fetch is now 18 (5×3=15, max(15, 3+15)=18), yielding 54 DB candidates — sufficient coverage for a 20-skill library in any realistically-sized workspace.

### ✅ 🟢 Unit test comment clarified (`skill_ops.rs:343–347`)
Added comment noting the test validates the filter predicate only, not the hybrid path itself, and that integration coverage for the BM25+KNN routing lives in `embedding_service` tests.

**Skipped**: None — both recommendations were actionable and appropriate to address.

PR #1396, opened 2026-06-09
Replace install.sh with Tauri first-launch setup

malibio:
Closes #1317


malibio:
## Review Addressed

All critical and important recommendations implemented:

- **🔴 #1 Fixed**: Changed `#[cfg(unix)]` → `#[cfg(any(target_os = "macos", target_os = "linux"))]` in all three places in `lib.rs` — module declaration, `check_daemon_status`, and setup block. The contract now matches the implementation.

- **🔴 #2 Fixed**: Added `xml_escape()` helper in `daemon_setup.rs` and applied it to all `<string>` values in the plist template. Characters `&`, `<`, `>` in path components are now safely escaped.

- **🟡 #3 Fixed**: Deleted `scripts/install.sh` — fully superseded by the Tauri path.

- **🟡 #4 Fixed**: Replaced separate `systemctl enable` + `systemctl restart` with `systemctl --user enable --now` — matches `install.sh` semantics and avoids start/restart ambiguity on older systemd.

**Skipped:**
- 🟡 #5 (`launchctl print` pre-check): Pre-existing pattern, not introduced by this PR. Tracked as a separate hardening concern.
- Nit: `set_executable` async — pre-existing, not in scope.

PR #1399, opened 2026-06-09
#133: PersonNodeBehavior + person core schema (ADR-037 identity type)

mstomar125:
Implements **PersonNode** as a core node type — the identity primitive the [ADR-037 access model](https://github.com/NodeSpaceAI/nodespace-docs/blob/main/decisions/037-collection-access-model.md) builds on ("members of a restricted collection = person nodes only"). Closes #133.

Owns the **type only** (schema + behavior). No PersonNode *instance* is seeded — the first PersonNode is created + named + bound to a Supabase identity at Pro sync-onboarding (nodespace-sync#125).

## Changes
- **`PersonNodeBehavior`** (`behaviors/mod.rs`): `validate` (name/content non-empty), `can_have_children=false`, plain-text, name is the embeddable content (people are searchable), no parent contribution. Plus a **computed `display_name`** — `custom:displayName` override → content → `"Unknown Person"` placeholder.
- Registered in `NodeBehaviorRegistry::new()`.
- **`person` core schema** (`core_schemas.rs`): `content` = name, `is_core`, no extra fields.
- Tests: validation + display-name precedence; registry/schema count asserts updated.

Unblocks the Pro-side enforcement (nodespace-sync#125 auth_identities/onboarding, #126 admin) and is the person-only check for nodespace-core#1372.

`cargo test` (behaviors + core_schemas) green, clippy clean.

Issue #133, opened 2025-10-02
Implement PersonNodeBehavior


malibio:
## Overview

Implement **PersonNode** as a core node type: the `person` schema entry + `PersonNodeBehavior` (validation, computed display name). PersonNode is the identity primitive the [ADR-037 access model](https://github.com/NodeSpaceAI/nodespace-docs/blob/main/decisions/037-collection-access-model.md) builds on.

This issue owns **PersonNode the *type*** (schema + behavior). It does **not** create any PersonNode *instance* — see [Scope](#scope).

> Updated to reflect the revised data model: PersonNode is pure identity (`name`, `email` only). `role` moves to the `has_role` edge between PersonNode and DatabaseSettingsNode (see #[#1398]). `auth_status` moves to DatabaseSettingsNode. Supersedes the original pre-Universal-Graph body.

## Read first
- [`components/node-behavior-system.md`](https://github.com/NodeSpaceAI/nodespace-docs/blob/main/components/node-behavior-system.md) — Core (hardcoded) vs Extension (schema-driven) behavior architecture
- [`components/schema-management.md`](https://github.com/NodeSpaceAI/nodespace-docs/blob/main/components/schema-management.md) — protection levels, namespace enforcement
- [`decisions/037-collection-access-model.md`](https://github.com/NodeSpaceAI/nodespace-docs/blob/main/decisions/037-collection-access-model.md) — why PersonNode matters

## Data model (Universal Graph)

PersonNode is a node with `node_type = 'person'`; its data lives under `properties` (namespaced `properties.person.*`). There is **no** persons table.

PersonNode is **pure identity** — it carries no auth or role state. Those live on relationships:

| Field | Type | Notes |
|---|---|---|
| `name` | `string?` | Display name. Optional — must not block instantiation. Used by `compute_content`. |
| `email` | `string?` | Optional; basic format validation (`@` + `.`) when present. |

**Not on PersonNode:**
- `role` — lives on the `has_role` edge between PersonNode and DatabaseSettingsNode (tenant-level authority)
- `auth_status` — lives on DatabaseSettingsNode (database sync state)

## Implementation Tasks

- [ ] Add a `person` `SchemaNode` to `packages/core/src/models/core_schemas.rs` with `name` and `email` (both optional).
- [ ] Implement `PersonNodeBehavior` implementing the `NodeBehavior` trait.
[...]

PR #1400, opened 2026-06-09
#1372: restrictedToMembers Core field on collection schema (ADR-037)

### LINKED ISSUE #1372: ADR-037: Collection access model — core data plane (restrictedToMembers, member_of permission, PersonNode seeding)
Land the open-core **data plane** for [ADR-037](https://github.com/NodeSpaceAI/nodespace-docs/blob/main/decisions/037-collection-access-model.md)'s organizational-by-default / opt-in-restriction access model, so a future `nodespaced-pro` RLS layer can enforce it. Per [ADR-028](https://github.com/NodeSpaceAI/nodespace-docs/blob/main/decisions/028-open-core-repo-separation.md) this issue covers **only inert, local-first data and ops**: a Core-protected boolean `restrictedToMembers` on the `collection` schema, a `permission` value (`admin`/`modify`/`readOnly`, default `modify`) on `member_of` edges, PersonNode as a seeded core type with one local instance (`auth_status:'local'`, coordinating with #133), and the ops the UI needs to toggle restriction and manage person members with permission levels.

**Out of scope (private `nodespace-sync` / `nodespace-cloud`):** RLS enforcement (`user_can_access_node`/`user_can_write_node`, nearest-boundary recursive CTE, break-inheritance, person-only filter), the atomic auto-admin bootstrap on restrict, the Last-Admin Invariant trigger, `auth_identities`/JWT bind, and all cloud DDL (`nodes.id TEXT`, `collection_members` VIEW, `REPLICA IDENTITY FULL`). Core lands the data they read.

**Greenfield (per CLAUDE.md):** no migration code. The `modify` read-default covers legacy `member_of` edges; absent `restrictedToMembers` = false. Reset is acceptable.

> ⚠️ Before starting **Task C**, update **#133 (PersonNodeBehavior)** — its body references removed doc paths and a "Persons Table" that predates the Universal Graph model.

---

## What core must land (the wire contract pro reads)
Two things on the wire, plus a person keyspace the edges point at:
- `properties.restrictedToMembers: bool` on collection nodes
- `properties.permission: "admin" | "modify" | "readOnly"` on person `member_of` edges

---

## Task A — `restrictedToMembers` core property on `collection`
*No PersonNode dependency. Ship first.*
- [ ] Add a `restrictedToMembers` `SchemaField` (boolean, `protection: Core`, default `false`, not required) to the `collection` `SchemaNode` in `packages/core/src/models/core_schemas.rs` (entry ~L285, currently `fields: vec![]`).
- [ ] In `CollectionNodeBehavior::validate` (`packages/core/src/behaviors/mod.rs` ~L1424), reject a non-boolean `restrictedToMembers` (mirror TaskNodeBehavior type-check ~L671). Absent/null → treated as `false`.
- [ ] **Verify `SchemaProtectionLevel::Core` actually blocks writes at the persistence boundary** (not just advisory/UI-hidden) for arbitrary `update_node` callers — if Core protection is advisory today, add a hard reject in the behavior + ensure the protection path runs on every collection update. (See Risk 1.)
- [ ] Tests: absent→valid(false); `true`/`false`→valid; `"true"`/number→`InvalidProperties`.

## Task B — `permission` value on `member_of` edges
*No PersonNode dependency; inert until persons exist.*
- [ ] Extend `SqliteStore::add_to_collection` (`packages/core/src/db/sqlite_store.rs` ~L2887) to accept optional `permission` and write it into the edge `properties` JSON alongside `order` (additive, no DDL — edge `properties` already exists ~L2934).
- [ ] Add `set_member_permission(member_id, collection_id, permission)` updating `properties->>'permission'`.
- [ ] Define a `Permission` enum (`Admin`/`Modify`/`ReadOnly`) serializing to `admin`/`modify`/`readOnly`; reads default to `modify` when absent.
- [ ] Validate the enum at the ops boundary; reject unknown strings.
- [ ] Tests: each level round-trips; absent→`modify`; unknown rejected; permission update.

## Task C — `person` core schema entry + PersonNodeBehavior (coordinate #133)
- [ ] Add a `person` `SchemaNode` to `core_schemas.rs`: `role` enum (`owner`/`admin`/`editor`/`viewer` — the **tenant-level** axis, distinct from edge permission), `auth_status` enum (`local`/`bound`, default `local`, Core), plus name/email per #133.
- [ ] Implement `PersonNodeBehavior` (#133) and register it in `NodeBehaviorRegistry::new()` (`behaviors/mod.rs` ~L2089).
- [ ] Validate name/email (per #133, adapted to `properties.person.*` namespacing) + `role`/`auth_status` enums.
- [ ] Tests for behavior + schema presence.

## Task D — Seed the single local PersonNode at first launch
- [ ] Add singleton PersonNode seeding to `NodeService::new`/`seed_*` (`packages/core/src/services/node_service/mod.rs` — `new` ~L814). Use the idempotency pattern in `seed_nodes_from_templates` (~L1024): "does any `person` node exist? if not, create one."
- [ ] Seed `properties.person.auth_status = 'local'`, `role = 'owner'`, stable id (e.g. symbolic `local-user` — keyspace tolerates symbolic ids; eases later `auth_identities` bind).
- [ ] Idempotent across restarts; emits seed/domain events consistent with schema seeding.
- [ ] Tests: fresh DB seeds exactly one person; second `new` no duplicate; `auth_status:'local'`.
- Depends on Task C.

## Task E — Ops: set restriction, add person members, set permission, list members
- [ ] `set_collection_restriction(collection_id, bool)` in `collection_ops.rs` — plain property write, **no** auto-admin grant (that is the pro bootstrap, deliberately omitted).
- [ ] `add_person_to_collection(person_id, collection_id, permission)` — validates `out_node` is `node_type='person'` (person-only, ADR-037 §4); default `modify`.
- [ ] `set_member_permission(...)` wrapping Task B.
- [ ] Extend `get_collection_members` / `_recursive` (`collection_ops.rs` ~L164/L180; queries `sqlite_store.rs` ~L2991/L3084) to surface each member's `permission` (default `modify`).
- [ ] Tests: non-person rejected; person at each level; permission change; list returns levels.
- Depends on Tasks B + C.

**Suggested sequence:** A ∥ B ∥ C → (C unblocks) D ∥ E.

---

## Risks / open decisions (core side)
1. **Protection vs. fail-open RLS (highest risk).** The pro RLS is open-by-default: any non-boolean `restrictedToMembers` reads as *not restricted* (soft-open). Core MUST guarantee the field is only ever a boolean — confirm `SchemaProtectionLevel::Core` blocks writes at persistence, not just UI. Treat a non-boolean write as a bug, not a soft-open.
2. **Auto-admin bootstrap is purely pro** (ADR-037 §6). Keep `set_collection_restriction` a dumb property write — no local admin grant. Local restriction is *inert* (no RLS locally); document this. Do not add a "convenience" local bootstrap — it would create a second, divergent path.
3. **PersonNode singleton id + role.** Stable symbolic id (`local-user`) makes the later Supabase bind a clean single-row insert (`auth_identities.person_node_id` is `TEXT`). Seeding `role:'owner'` locally is harmless and correct for upgrade continuity.
4. **#133 staleness** — update before Task C.
5. **No migration/reset code** — `modify` read-default + absent-=false cover legacy; reset is fine.

---

A separate, deferrable **frontend** issue should cover the always-visible gated "Restrict to members" upgrade CTA (tier-aware via the existing `WatchSyncStatus` capability probe) — no backend dependency beyond the property existing.

_Breakdown produced via senior-architect-reviewer against ADR-037 + the merged cloud-sync.md, grounded in current `nodespace-core` source._

mstomar125:
Open-core **data plane** for ADR-037's opt-in restriction. Closes #1372.

> **Stacked on #1399 (PersonNode #133).** This branch is cut from `issue-133`, so the PersonNode commit rides along here until #1399 merges — **review only the top commit** ("restrictedToMembers Core field"). The diff narrows to just this change once #1399 lands.

Per [ADR-028](https://github.com/NodeSpaceAI/nodespace-docs/blob/main/decisions/028-open-core-repo-separation.md), this covers only **inert, local-first data**:
- **`restrictedToMembers`** — a **Core-protected boolean** on the `collection` schema (default `false` = open/organizational). The Pro RLS layer (nodespace-sync#124 / cloud#23) reads `properties->>'restrictedToMembers'` to gate access; core just declares the protected property.
- **`member_of` `permission`** (admin/modify/readOnly, default modify) is a **free-form property on the native `member_of` edge** — the RLS reads it directly, so no schema change is needed.
- **Ops:** toggle restriction = a normal node property update; add a person member with a permission = a normal `member_of` relationship create. No new core API needed for the data plane; the **enforcement flows** (auto-admin, Last-Admin trigger, onboarding) are nodespace-sync#125/#126.

Tests: `restrictedToMembers` field shape; schema-count + no-fields asserts updated. `cargo test` green, clippy clean.

### DIFF
--- a/packages/core/src/models/core_schemas.rs
+++ b/packages/core/src/models/core_schemas.rs
@@ -291,7 +291,30 @@ pub fn get_core_schemas() -> Vec<SchemaNode> {
             modified_at: now,
             is_core: true,
             schema_version: 1,
-            fields: vec![],        // Uses content for name
+            // ADR-037 #1372: opt-in restriction. A Core-protected boolean; the Pro
+            // RLS layer reads `properties->>'restrictedToMembers'` to gate access.
+            // Default/absent = false = open (organizational). member_of edge
+            // `permission` (admin/modify/readOnly) is a free-form edge property the
+            // RLS reads directly — no schema field needed.
+            fields: vec![SchemaField {
+                name: "restrictedToMembers".to_string(),
+                field_type: "boolean".to_string(),
+                protection: SchemaProtectionLevel::Core,
+                core_values: None,
+                user_values: None,
+                indexed: false,
+                required: Some(false),
+                extensible: None,
+                default: Some(serde_json::json!(false)),
+                description: Some(
+                    "When true, only person members may access this collection's \
+                     nodes (ADR-037 opt-in restriction). Default false = open."
+                        .to_string(),
+                ),
+                item_type: None,
+                fields: None,
+                item_fields: None,
+            }],
             relationships: vec![], // member_of is a native edge, not schema-defined
             title_template: None,
             properties_header_summary_template: None,
@@ -958,7 +981,6 @@ mod tests {
             "code-block",
             "quote-block",
             "ordered-list",
-            "collection",
             "checkbox",
         ] {
             let schema = schemas.iter().find(|s| s.id == *id).unwrap();
@@ -968,6 +990,19 @@ mod tests {
                 id
             );
         }
+    }
+
+    #[test]
+    fn test_collection_has_restricted_to_members_field() {
+        // ADR-037 #1372: opt-in restriction is a Core-protected boolean on collection.
+        let schemas = get_core_schemas();
+        let collection = schemas.iter().find(|s| s.id == "collection").unwrap();
+        let field = collection
+            .get_field("restrictedToMembers")
+            .expect("collection has restrictedToMembers");
+        assert_eq!(field.field_type, "boolean");
+        assert_eq!(field.protection, SchemaProtectionLevel::Core);
+        assert_eq!(field.default, Some(serde_json::json!(false)));
     }
 
     #[test]


mstomar125:
**REQUEST CHANGES** ⏳

_Initial review, **top commit `ceee30d9` only** (the `baf653c0` PersonNode commit is #1399 riding along — reviewed there). `no checks reported` on the branch._

Small and correct: `restrictedToMembers` as a `Core`-protected boolean on the `collection` schema, `default false` / `required false`, a clear description, and a focused test asserting type + protection + default. The field name matches exactly what cloud#23's RLS reads (`properties->>'restrictedToMembers'`), camelCase is consistent with core's other schema fields (`dueDate`, `assignee`), and scoping to inert local data per ADR-028 (no enforcement here) is the right boundary. Two minor things:

🟢 **Cross-repo contract — confirm "absent" reads as open.** cloud#23 gates on `(properties->>'restrictedToMembers')::boolean IS TRUE`, i.e. a NULL/absent property → open. So the open default only holds if an untouched collection leaves the property **absent** (or writes literal `false`) in the synced `properties` JSON — please confirm a default-`false` Core field doesn't materialize as some other shape (e.g. a stringified or nested value) that wouldn't satisfy that read. Relatedly: this being a `Core`-protected **boolean** is exactly what de-risks the `::boolean` cast I flagged on cloud#23 — but keep cloud#23's predicate defensive anyway, since the cloud can't assume every writer honored the core constraint (defense in depth, not a change to this PR).

🟢 **Test coupling to #1399.** The `"collection"` → `"person"` swap in the no-fields list hard-depends on #1399's `person` schema existing, so this commit won't stand on `main` alone (expected, given the stack). Just flag that if #1399's `person` type is renamed/reworked — which my #1399 review raised re: the existing user-entity "person" collision — this test moves with it.

Otherwise good to go once the cross-repo contract above is confirmed.

mstomar125:
Addressed.

**Rebased onto merged main (commit f5256f3):** dropped the obsolete `baf653c0` person commit (issue #133 consolidated onto #1401, now merged — that PR's name+email PersonNode is canonical), so this PR is now *only* the `restrictedToMembers` field. Reconciled `test_simple_schemas_have_no_fields`: removed `person` from the no-fields list (#1401's person carries `name`/`email` fields), `collection` already drops out (it now has `restrictedToMembers`). All 11 `core_schemas` tests pass.

🟢 **"Absent reads as open" — confirmed.** `create_node` applies schema defaults only for `task` (crud.rs); every other type, including `collection`, is normalized without default materialization. So an untouched collection leaves `restrictedToMembers` **absent** in synced `properties` → cloud#23's `(properties->>'restrictedToMembers')::boolean IS TRUE` → NULL → false → **open**. When explicitly restricted it's a literal boolean `true`. No stringified/nested shape. The open-by-default contract holds, and cloud#23's predicate stays defensive regardless (defense in depth).

🟢 Test coupling to the person schema: now that #1401 is in main, this test stands on main as-is.

PR #1401, opened 2026-06-09
Implement PersonNodeBehavior

malibio:
Closes #133

PR #1402, opened 2026-06-09
Remove auth_status from PersonNode seeding

malibio:
## Summary

Follow-up to #1401. The seeded PersonNode incorrectly included `auth_status: "local"` in its properties. Per the revised ADR-037 model (updated this session), `auth_status` belongs on `DatabaseSettingsNode` (#1398), not PersonNode — it describes whether the *database* is synced, not who the person is.

- Remove `auth_status` from `seed_local_person_if_needed` — seeded node now has empty properties
- Update `person_seed_test.rs` to assert `auth_status` is **absent** from the seeded PersonNode
- Update doc comment on seeding method to note where `auth_status` lives

## Test plan
- [ ] `cargo test -p nodespace-core person_seed` — 2 tests pass
- [ ] `cargo clippy -p nodespace-core` — clean

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/core/src/behaviors/mod.rs
+++ b/packages/core/src/behaviors/mod.rs
@@ -2004,11 +2004,10 @@ impl NodeBehavior for CustomNodeBehavior {
 
 /// Built-in behavior for person nodes
 ///
-/// Person nodes are identity nodes — they carry `name` (optional) and `email`
-/// (optional, format-validated when present), plus `auth_status` for the local
-/// user (ADR-037: seeded as `"local"`, bound to a Supabase identity on Pro
-/// upgrade via nodespace-sync#125). *Role/permission* lives on the `member_of`
-/// relationship, not on the node.
+/// Person nodes are pure identity — they carry `name` (optional) and `email`
+/// (optional, format-validated when present). Auth state (`auth_status`) lives
+/// on `DatabaseSettingsNode`; tenant role lives on the `has_role` edge
+/// (PersonNode → DatabaseSettingsNode). Neither is a PersonNode property.
 pub struct PersonNodeBehavior;
 
 impl PersonNodeBehavior {
--- a/packages/core/src/services/node_service/mod.rs
+++ b/packages/core/src/services/node_service/mod.rs
@@ -922,12 +922,14 @@ impl NodeService {
         Ok(service)
     }
 
-    /// ADR-037 (#133): seed exactly one **local** PersonNode — the local user,
-    /// marked `auth_status: "local"`. Idempotent: skips when a person already exists,
-    /// so an existing database is backfilled on next open too. On Pro upgrade this node
-    /// is *bound* to a Supabase identity (nodespace-sync#125) — a single `auth_identities`
-    /// row keyed on the node id — not recreated; there is no "now set up your user" step.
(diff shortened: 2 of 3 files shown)


malibio:
### Code Review Summary

This is a clean, targeted fix. The change correctly removes `auth_status` from `seed_local_person_if_needed` and inverts the test assertion to verify absence rather than presence. The scope is appropriately minimal — one production line changed, one test assertion changed — which is exactly right for a model-placement correction.

**Verdict: Approved.** One improvement item noted for follow-up.

---

### Findings

#### Suggested Improvements

**`packages/core/src/behaviors/mod.rs` line 2008 — Stale doc comment on `PersonNodeBehavior`**

The struct-level doc comment still describes `auth_status` as a property that PersonNode carries:

```
/// Person nodes are identity nodes — they carry `name` (optional) and `email`
/// (optional, format-validated when present), plus `auth_status` for the local
/// user (ADR-037: seeded as `"local"`, bound to a Supabase identity on Pro
/// upgrade via nodespace-sync#125).
```

This is now incorrect — `auth_status` no longer lives on PersonNode. The seeding fix and the seed test are in sync with the revised ADR-037 model, but this behavior doc was not updated in the same commit. A future reader consulting `PersonNodeBehavior` will get contradictory information: the doc says it carries `auth_status`, the seeding code and the test say it does not.

The fix is a one-line doc update, e.g.:

```
/// Person nodes are identity nodes — they carry `name` (optional) and `email`
/// (optional, format-validated when present). Auth status (`auth_status`) lives
/// on `DatabaseSettingsNode` (#1398), not here (ADR-037 revised).
```

This does not block merge — the runtime behavior is correct — but it should be cleaned up in a quick follow-on commit or folded into the next time this behavior is touched.

---

#### Nitpicks

- **Nit: `packages/core/tests/person_seed_test.rs` line 36** — The `.get("person").and_then(|p| p.get("auth_status"))` chain in the absence assertion navigates into the `person` namespace. This is correct (properties are namespaced), but given the node is now seeded with `serde_json::json!({})` (no `person` key at all), `.get("person")` already returns `None` and the `.and_then` is unreachable. The assertion is still logically valid, just slightly over-specified. Worth keeping as-is since it documents where `auth_status` *would* live if it were erroneously re-added — this is actually useful intent documentation.

---

### What Was Verified

- **Correctness of fix**: `serde_json::json!({})` replaces `json!({ "person": { "auth_status": "local" } })` — auth_status is fully absent from the seeded node. Correct.
- **Test assertion**: Inverted from `assert_eq!(..., Some("local"))` to `assert!(...is_none())` — tests the right thing for the new model. Correct.
- **No other PersonNode auth_status references**: Searched the full worktree. Only remaining reference is the stale doc comment on `PersonNodeBehavior` (line 2008 of `behaviors/mod.rs`) flagged above. No runtime code paths set or read `auth_status` on PersonNode.
- **Idempotency test untouched**: `reopening_database_does_not_seed_a_second_person` is unaffected and continues to test the right invariant.
- **Scope**: Single-purpose commit — no unrelated changes bundled in.

malibio:
Fixed the stale doc comment on `PersonNodeBehavior` (f0a65f3). Nit kept as-is — the over-specified assertion chain is intentional documentation of where `auth_status` would sit if erroneously re-added. No re-review needed.

Issue #162, opened 2025-10-05
Add event listener memory leak detection tests


malibio:
## Problem Statement

Event listeners on `document` in modals could leak if cleanup fails. **No tests verify cleanup happens correctly**.

**Leak Scenarios:**
- Component unmounted but listener remains
- Rapid mount/unmount cycles accumulate listeners
- Errors during unmount prevent cleanup
- Reference cycles prevent garbage collection

**Components With Document Listeners:**
- `node-autocomplete.svelte:144` - `document.addEventListener('keydown', ...)`
- `slash-command-dropdown.svelte:110` - `document.addEventListener('keydown', ...)`

## Why This Matters

**Production Impact:**
- Memory usage grows over time
- Performance degrades
- Browser tab becomes unresponsive
- Hard to debug (symptoms appear gradually)

## File Locations (Monorepo Structure)

**Test File to Create:**
```
packages/desktop-app/src/tests/integration/event-listener-cleanup.test.ts
```

**Components to Test (Full Import Paths):**
```typescript
import NodeAutocomplete from '$lib/components/ui/node-autocomplete/node-autocomplete.svelte';
import SlashCommandDropdown from '$lib/components/ui/slash-command-dropdown/slash-command-dropdown.svelte';
```

**Actual File Paths:**
- `packages/desktop-app/src/lib/components/ui/node-autocomplete/node-autocomplete.svelte`
- `packages/desktop-app/src/lib/components/ui/slash-command-dropdown/slash-command-dropdown.svelte`

**EventBus Reference:**
- `packages/desktop-app/src/lib/services/eventBus.ts`

### Component Lifecycle Tests (6 tests)

- [ ] **NodeAutocomplete removes keydown listener on unmount**
  - Verify `removeEventListener` called with exact same handler
  - Normal unmount scenario (happy path)

- [ ] **SlashCommandDropdown removes keydown listener on unmount**
  - Verify `removeEventListener` called with exact same handler
  - Normal unmount scenario (happy path)

- [ ] **Multiple mount/unmount cycles don't accumulate listeners**
  - Mount and unmount 100 times
  - Verify listener count returns to baseline
  - Test both components

- [ ] **Listeners removed even if error during unmount**
  - Simulate error in onDestroy lifecycle
  - Verify cleanup still happens (or handler is weak)
  - Edge case but important for robustness

### Rapid Operations Tests (3 tests)

- [ ] **100 rapid mount/unmount cycles - NodeAutocomplete**
[...]

PR #1408, opened 2026-06-12
Emit RelationshipDeleted on move-to-root and reparent (move_node)

mstomar125:
## Problem
`move_node` deletes the old `has_child` edge in the store, but only emitted a `RelationshipUpdated` for the **new** parent (gated on `Some(parent)`). So:
- A **move-to-root** (`new_parent = None`) emitted *no* relationship event — consumers never learned the node detached.
- A **reparent** left the old parent edge unannounced.

For NodeSpace sync (epic NodeSpaceAI/nodespace-sync#162) this meant outdenting a node to the top level never propagated to other devices — the node stayed parented elsewhere.

## Fix
Capture the old parent before the move; when the parent actually changes (move-to-root or reparent), emit a `RelationshipDeleted` for the old parent edge **after** the new-parent `RelationshipUpdated` (insert-before-delete order, so a consumer that inserts-then-deletes never sees the node parentless mid-move). A same-parent position move emits nothing new.

Reuses the existing `RelationshipDeleted` event (no new variants), `node_thing`-normalized endpoints like the other emit sites.

## Verified
Built the sync daemon against this branch: the daemon now pushes the edge delete on move-to-root (`supabase push: has_child edge deleted parent=… child=…`). `cargo check -p nodespace-core` clean. (Full e2e propagation also needs the live tenant's `relationships` table at `REPLICA IDENTITY FULL` — already in the tenant schema template — so the realtime DELETE carries the endpoints.)

### DIFF
--- a/packages/core/src/services/node_service/mod.rs
+++ b/packages/core/src/services/node_service/mod.rs
@@ -1877,6 +1877,15 @@ impl NodeService {
             }
         }
 
+        // Capture the OLD parent before the move so we can surface its edge removal
+        // (sync-epic nodespace-sync#162 / S3): the store deletes the old has_child
+        // edge but, historically, only a RelationshipUpdated for the NEW parent was
+        // emitted (gated on Some). A move-to-root (new_parent = None) therefore
+        // emitted no relationship event at all, so the detach never propagated to
+        // other devices; a reparent left a stale cloud edge. We now also emit a
+        // RelationshipDeleted for the old parent whenever the parent actually changes.
+        let old_parent_id = self.get_parent(node_id).await?.map(|p| p.id);
+
         let insert_after = self.resolve_insert_position(position, new_parent).await?;
 
         // Perform the move
@@ -1886,7 +1895,9 @@ impl NodeService {
             .await
             .map_err(|e| NodeServiceError::query_failed(e.to_string()))?;
 
-        // Emit RelationshipUpdated event (Issue #811: unified relationship events)
+        // Emit RelationshipUpdated event (Issue #811: unified relationship events).
+        // Emit the NEW-parent edge first so a consumer that inserts-then-deletes
+        // never sees the node parentless mid-move.
         if let Some(parent_id) = new_parent {
             self.emit_event(DomainEvent::RelationshipUpdated {
                 relationship: crate::db::events::RelationshipEvent::new(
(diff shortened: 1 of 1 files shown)


mstomar125:
**REQUEST CHANGES** ⏳

_Initial review, top commit `6244c11` (against `main`). `no checks reported` on the branch — CI status pending/absent, so this is a read-only review; the behaviour change is unvalidated by automated tests (see below)._

The fix itself is correct and nicely scoped. I traced each path:

- **move-to-root** (`new_parent = None`, old `Some`): `None != Some(old_id)` → emits `RelationshipDeleted` only. ✅ the bug it targets.
- **reparent** (`new_parent = Some(other)`): emits `RelationshipUpdated` (new) *then* `RelationshipDeleted` (old). ✅ insert-before-delete ordering holds — relationship events are node-keyed `None` so they bypass `BatchState` and broadcast immediately in call order (`emit_event`, mod.rs:1408).
- **same-parent position move** (`new_parent = Some(old)`): `Some(old) != Some(old)` is false → no delete. ✅
- **already-root → parent** (`old = None`): no delete, just the update. ✅

Event shape is consistent with every sibling site: the `id` format `relationship:{parent}:{child}` matches mod.rs:1773 / 1904 / 2127 / 2224 / 2306, and `from_id`/`to_id` are `node_thing`-normalized (parent→child direction), which `db/events.rs:92` explicitly notes the inline `RelationshipDeleted` variant relies on. Good.

One blocking gap:

🟡 **No regression test for a sync-critical behaviour change.** This PR alters the *emit contract* of `move_node` — move-to-root and reparent now emit `RelationshipDeleted`, and a same-parent reorder must **not**. That contract is load-bearing for sync epic #162 (the whole point is the detach propagating), and it's exactly the kind of thing a future refactor silently re-gates back to `Some(parent)`-only — reintroducing the bug with no signal. The harness to lock it down is already here:
- existing `move_node` tests in the same module to extend — `test_move_node_accepts_container_parent` (mod.rs:3090), `test_move_node_rejects_non_container_parent` (mod.rs:3062);
- an established event-assertion pattern — `service.subscribe_to_events()` then a `while let Ok(env) = rx.try_recv()` drain — used by ~6 tests already (mod.rs:3616, 3665, 3682, 3775, 3813, 3859).

Please add a test (or three) asserting: (a) move-to-root emits a `RelationshipDeleted` for the old parent edge; (b) reparent emits the new-parent `RelationshipUpdated` **before** the old-parent `RelationshipDeleted`; (c) a same-parent position move emits **no** `RelationshipDeleted`. (a) and (c) are the two that guard against regressing this exact fix.

Two non-blocking notes:

🟢 **`id`-format consistency depends on `Node.id` being bare.** The delete `id` is built from `old_parent_id = get_parent(node_id)?.map(|p| p.id)` (mod.rs:1888), while the *insert*-time id (mod.rs:1773/1904) is built from the bare `parent_id: &str` argument. These only match if `get_parent`'s `Node.id` is the bare id (not `node:`-prefixed) — which is the convention, so this is fine, but a consumer that keys deletes off the raw `id` string (rather than `from_id`/`to_id`, which are `node_thing`-normalized either way) would mis-match if that ever drifts. The test in (a) above would pin it.

🟢 **Extra pre-move read.** `get_parent(node_id).await?` (mod.rs:1888) adds a DB round-trip and a new failure point *before* the store mutation — a `get_parent` error now aborts the move. Cheap and almost certainly fine (a failing `get_parent` means the store is already unhealthy), just noting the added `?`.

Net: correct, well-reasoned fix with consistent event shape — just needs the regression test that turns "I built the daemon and watched it push the delete" into something CI can keep enforcing for #162.

Issue #152, opened 2025-10-05
Add comprehensive event tests for NodeAutocomplete component


malibio:
## Problem Statement

The `NodeAutocomplete` component has **critical event handling logic that is completely untested**. This modal component handles keyboard navigation, event propagation, and auto-focus behaviors - all prone to subtle bugs that only manual testing currently catches.

**Specific Untested Code:**
- Lines 100-148: Keyboard event handling (ArrowUp/Down, Enter, Escape)
- Lines 107, 113, 119, 126: `stopPropagation()` calls to prevent background interactions
- Lines 143-148: Event listener setup and cleanup in `onMount`
- Lines 56-85: Smart positioning logic for viewport edges
- Lines 90-98: Scroll-to-selected behavior

## Why This Matters

**Production Risks:**
1. **Background Actions** - If `stopPropagation()` fails, pressing Enter might create a new node while also selecting from autocomplete
2. **Focus Traps** - Users could get stuck in modal, unable to escape
3. **Memory Leaks** - Event listeners on `document` persist after unmount
4. **Positioning Bugs** - Modal could render off-screen or overlap trigger element
5. **Accessibility** - Keyboard navigation crucial for accessibility could break silently

**Current Reality:**
- Component has 150+ lines of event-critical code
- Zero automated tests
- Changes verified only by manual clicking
- Regressions caught in production

## Files Involved

**Component Under Test:**
- `src/lib/components/ui/node-autocomplete/node-autocomplete.svelte` (all 216 lines)

**Test File to Create:**
- `src/tests/components/node-autocomplete.test.ts`

**Dependencies:**
- Uses utilities from Issue #151 (`svelte-test-utils.ts`)
- `packages/desktop-app/src/lib/design/icons/registry.ts` (icon config, lines 34-36)

**Reference Files:**
- `src/tests/component/content-editable-controller.test.ts` - Similar event testing patterns

### Test Coverage Requirements

Must test **all event scenarios**:

- [ ] **Rendering Tests** (3 tests)
  - Renders when visible=true
  - Hides when visible=false
  - Displays all results with correct node types

- [ ] **Keyboard Navigation** (7 tests)
  - ArrowDown moves selection down
  - ArrowUp moves selection up
  - ArrowDown at bottom stays at bottom
  - ArrowUp at top stays at top
  - Enter selects current item
  - Escape closes modal
  - All keyboard events prevent default and stop propagation

- [ ] **Event Propagation** (3 tests)
[...]

Issue #147, opened 2025-10-04
Clarify DatabaseServiceAdapter is temporary solution


malibio:
## Context

`DatabaseServiceAdapter` in `nodeReferenceService.ts` is a temporary wrapper providing `queryNodes()` and `upsertNode()` methods until the Tauri backend implements them.

## Current State

The adapter has a basic comment (line 28) explaining its purpose, but doesn't strongly emphasize its temporary nature.

## Requested Change

Add explicit JSDoc comment to the `DatabaseServiceAdapter` class emphasizing:
1. This is a **temporary** adapter
2. Will be removed when backend implements these methods
3. Link to relevant backend issue (if exists)

## Example

```typescript
/**
 * TEMPORARY: DatabaseServiceAdapter
 * 
 * Provides missing queryNodes/upsertNode methods until Tauri backend 
 * implements them. This adapter will be REMOVED once backend support is complete.
 * 
 * @see Backend issue for queryNodes implementation
 * @temporary Will be deleted in future PR
 */
class DatabaseServiceAdapter {
  // ...
}
```

## Priority

Low - current documentation is adequate, this is polish for future maintainers.

PR #1412, opened 2026-06-15
Add Pro membership tauri commands (M5 core-half, sync #147)

mstomar125:
## What

Completes the **core-half of M5** (`NodeSpaceAI/nodespace-sync#147`) — the desktop-app command surface for the team-membership lifecycle. The daemon-half (gRPC methods + handlers in `nodespaced-pro`) already merged as sync PR #152; this wires the Tauri client to it.

## Changes (purely additive)

- **Vendor the membership block** of `nodespace_pro.proto` into `packages/desktop-app/src-tauri/proto/` (8 RPCs + request/response messages). `build.rs`/tonic-build regenerates the `CloudSyncService` client with `SetMember`, `RemoveMember`, `LeaveCollection`, `ListMembers`, `CreateInvite`, `AcceptInvite`, `RequestJoin`, `ApproveRequest`. The vendored proto is now byte-identical to the sync source-of-truth.
- **8 tauri commands** in `commands/pro_sync.rs`, each mirroring the existing `pro_initiate_oauth` shape: resolve the `ProClient` (return `Err` in community mode), then one pass-through RPC. The daemon forwards the signed-in JWT to the cloud RPC, so the admin / last-admin / open-vs-restricted gates are enforced **server-side** — these commands carry no authority of their own. `pro_list_members` returns a serializable `MemberDto` roster.
- **Register** all eight in the `generate_handler!` macro.

## Scope

No collaboration UI is wired — the commands exist for tests and a future UI (per #147). Nothing in the existing command set or the sync push/pull path changes, so two-window/two-device sync is unaffected.

## Verification

- `cargo check` + `cargo clippy` clean on `nodespace-app` (proto regen + new commands compile, no warnings).
- Diff is 3 files, additive only (proto +72, pro_sync +175, lib.rs +8).


mstomar125:
**APPROVE** ⏳

_Initial review, top commit `5b6b9f3` (against `main`). `no checks reported` on the branch — no CI to wait on, so the verification below is mine, not a green check._

Clean, purely-additive PR. I checked the things that actually matter for a vendored-proto passthrough:

✅ **Proto is byte-for-byte identical to the sync source-of-truth.** I diffed the vendored `packages/desktop-app/src-tauri/proto/nodespace_pro.proto` against `nodespace-sync:nodespaced-pro/proto/nodespace_pro.proto@main` — **identical, whole file**, not just the membership block. This is the one thing that silently breaks a vendored-proto client (mismatched field numbers / message shapes → runtime RPC failures against the daemon), and it's correct. The membership block is appended (new RPCs + messages), no renumbering of existing fields, so the regenerated `CloudSyncServiceClient` stays wire-compatible with the daemon from sync#152.

✅ **Guard mirrors `pro_initiate_oauth` faithfully.** `membership_client` (pro_sync.rs:175) factors out the exact pattern the existing command uses — `app.try_state::<ProClient>()` → `pro.client().await`, same "community tier … unavailable" `Err`. `ProClient::client()` is a cheap `CloudSyncServiceClient<Channel>` clone (pro_client.rs:84), so resolving per-command is the established, connection-cheap approach — no storm.

✅ **Authority posture is right.** Every command is a 1:1 field-forward into one RPC; the daemon forwards the signed-in JWT and the admin / last-admin / open-vs-restricted gates are enforced server-side. The commands carry no local authority, which is exactly what you want for a thin client surface.

✅ **`MemberDto` rather than serializing the prost `Member` directly** is the correct call — tonic-build doesn't derive `serde::Serialize` on generated types, so a DTO is the right way to hand a roster to the Tauri frontend.

✅ `cargo check` + `clippy` clean per the body; `email: Option<String>` → empty-bearer and `ttl_secs: Option<u64>` → `0`-default match the cloud `CreateInvite` semantics; `accept_invite` returns `MembershipAck.detail` (the joined collection id) as documented; all 8 registered in `generate_handler!` (lib.rs:422).

Two **non-blocking** heads-ups for the follow-up UI work (explicitly out of scope here, per "no collaboration UI is wired" — not asks against this diff):

- 🟢 **Tauri v2 capability grants.** Registering in `generate_handler!` makes these invokable from the Rust side, but the frontend can't call them until they're added to a capabilities/permissions set. That lands naturally with the UI PR — just flagging so it isn't a surprise then.
- 🟢 **Error granularity inherits the daemon's mapping.** These surface the tonic `Status` verbatim (`format!("SetMember failed: {e}")`), which is the right thing to do — but note the daemon-half (sync#152) currently collapses cloud `42501` (permission denied) into `Status::internal`, so an admin-gate rejection will reach the UI as a generic "Internal" rather than a permission error. That's the sync#152 item I flagged earlier, not this PR; worth the UI knowing it can't yet distinguish "denied" from "server error" off these results.

Net: the proto-drift risk (the only real hazard here) is verifiably closed, and everything else mirrors established, reviewed patterns. Good to merge.

Issue #154, opened 2025-10-05
Add default limit to query_nodes_simple


malibio:
From PR #150 code review - nice-to-have improvement.

The query_nodes_simple method has no default limit, which could return unbounded results and cause performance issues.

Proposed solution: Add a sensible default limit (e.g., 100 or 1000) to prevent accidental full table scans.

PR #1413, opened 2026-06-15
Pro re-login modal on AUTH_REQUIRED (T18, #1304)

mstomar125:
## What

Completes **T18 (#1304)** — the sync status indicator + re-login modal. The status pill (`pro-sync-pill.svelte`) already maps every `SyncStatusEvent.State` (connected / syncing / disconnected / error / auth-required), so this PR adds the one missing acceptance criterion: **a re-login modal triggered on refresh failure, with "Sign In Again" / "Work Offline".**

## Changes (frontend-only, additive)

- **`pro-relogin-modal.svelte`** (new) — presentational dialog. "Sign In Again" runs the PKCE flow (`pro_initiate_oauth`, same as the pill); "Work Offline" dismisses. Themed, accessible (`role="dialog"`, `aria-modal`, Escape = work-offline).
- **`app-shell.svelte`** — owns visibility: shows when `proSync.isPro && proSync.state === 'auth-required' && !dismissed`. `Work Offline` dismisses **per-episode** (so it doesn't keep nagging) and an `$effect` **re-arms** the dismissal once the daemon leaves `auth-required`, so a later refresh failure prompts again. The sign-in handler guards against double-invoke with a `pending` flag (shared with the modal to disable its buttons).

The daemon already emits `AUTH_REQUIRED` on refresh failure (session persistence, sync #154); this is purely the UI reaction. **Nothing in the sync push/pull path changes** — the two-window/two-device demo is unaffected.

## Verification

- New test `pro-relogin-modal.test.ts` — 7 cases (gated render, both action callbacks, Escape = work-offline, `pending` disables actions + shows progress, optional detail line).
- Full unit suite: **3857/3857 pass** (3850 baseline + 7). Lint + prettier clean on all touched files.

Closes #1304.

### DIFF
--- a/packages/desktop-app/src/lib/components/layout/app-shell.svelte
+++ b/packages/desktop-app/src/lib/components/layout/app-shell.svelte
@@ -28,6 +28,7 @@
   import { openUrl, isExternalUrl, isNodespaceUrl } from '$lib/utils/external-links';
   import OnboardingWizard from '$lib/components/onboarding/onboarding-wizard.svelte';
   import ProSyncPill from '$lib/components/pro-sync-pill.svelte';
+  import ProReloginModal from '$lib/components/pro-relogin-modal.svelte';
   import { proSync } from '$lib/stores/pro-sync.svelte';
   import ConflictToast from '$lib/components/conflict-toast.svelte';
 
@@ -40,6 +41,42 @@
 
   // First-launch onboarding wizard (Issue #1180).
   let showOnboarding = $state(false);
+
+  // Pro re-login prompt (T18, #1304). The daemon emits AUTH_REQUIRED when the
+  // persisted refresh token can't be renewed; we surface a modal offering a
+  // fresh sign-in or working offline. `reloginDismissed` is per-episode so
+  // "Work Offline" doesn't keep nagging, and is re-armed once the daemon
+  // leaves the auth-required state so a later failure prompts again.
+  let reloginDismissed = $state(false);
+  let reloginPending = $state(false);
+
+  let showReloginModal = $derived(
+    proSync.isPro && proSync.state === 'auth-required' && !reloginDismissed
+  );
+
+  $effect(() => {
+    if (proSync.state !== 'auth-required' && reloginDismissed) {
+      reloginDismissed = false;
+    }
+  });
+
+  async function handleReloginSignIn() {
+    if (reloginPending) return;
+    reloginPending = true;
+    try {
+      // Same PKCE flow the sync pill uses; the daemon opens the browser and
+      // the modal closes as `sync:status` transitions away from auth-required.
+      await invoke('pro_initiate_oauth');
+    } catch (e) {
+      log.warn('pro_initiate_oauth (relogin) failed', { error: e });
+    } finally {
+      reloginPending = false;
+    }
+  }
+
+  function handleReloginWorkOffline() {
+    reloginDismissed = true;
+  }
 
   /**
    * Sets up MCP event listeners for real-time UI updates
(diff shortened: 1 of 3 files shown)


mstomar125:
**REQUEST CHANGES** ⏳

_Initial review, top commit `627df98` (against `main`). `no checks reported` on the branch — no CI to wait on. Full unit suite is green per the body (3857/3857)._

The feature itself is well done and I want to be clear the logic is correct — I traced it:

- ✅ **Re-arm logic is right.** `showReloginModal = isPro && state==='auth-required' && !reloginDismissed`; "Work Offline" sets `reloginDismissed=true` (hides, no nag), and the `$effect` resets it once the daemon leaves `auth-required`, so a later refresh failure prompts again. The effect reads+writes `reloginDismissed` but is guarded by `&& reloginDismissed`, so it settles in one pass — no feedback loop.
- ✅ **Double-invoke guard** (`if (reloginPending) return`) and `finally`-reset are correct; the modal closes via the derived as `sync:status` transitions away.
- ✅ **`invoke('pro_initiate_oauth')` with no args is consistent with the pill** — I checked `pro-sync-pill.svelte:66`, which invokes with `{}` and a comment "workerUrl + userHint omitted — backend defaults apply." Same defaulting path, so re-login hits the same worker the pill would. Good.
- ✅ **Tests** — 7 focused cases, gated render + both callbacks + pending + detail.

One blocking item, plus a non-blocking note:

🟡 **Unrelated formatting churn is bundled into the feature diff.** Roughly half the `app-shell.svelte` hunks are pure prettier reflow of pre-existing lines that have nothing to do with the re-login modal — the `MCP_EVENTS.RELATIONSHIP_CREATED/DELETED` debug logs (lines ~146/162), the `get_settings` invoke generic (~237), the `importDirectory` `exclude_patterns` trailing comma (~359), the `select_new_database` generic wrap (~394), and the `check_daemon_status` onclick (~642). None are touched by the feature, which lives at the top (state/handlers ~42–80) and the `<ProReloginModal>` render (~684). Mixing a whole-file `prettier --write` into a feature PR makes the real change hard to pick out and invites merge conflicts with anything else touching those regions. Please drop the unrelated hunks so the diff is just the modal + its wiring — or, if a format gate genuinely requires the whole file to be reflowed, split that into its own formatting-only commit/PR and note it. (This is the same scope concern I've flagged before, e.g. the duplicated fmt hunk on sync#130.)

🟢 **Non-blocking — Escape/focus only works because the test dispatches on the overlay directly.** The overlay carries `onkeydown … Escape → onWorkOffline`, but the dialog does `onkeydown={(e) => e.stopPropagation()}` and nothing moves focus into the modal on open (no autofocus / focus-trap). So under real keyboard focus, an Escape press on the dialog or its buttons is stopped before it reaches the overlay handler, and if focus is on background content it never enters the overlay subtree at all — Escape effectively won't dismiss. The test passes only because it fires `keyDown` on `dialog.parentElement` synthetically. I'm **not** asking you to fix this here: `version-conflict-modal.svelte:85-93` uses the identical overlay-Escape + content-`stopPropagation` pattern, so this is the existing house behaviour, not a regression you introduced. Worth a separate repo-wide follow-up (a shared focus-trapping dialog — the `ui/dialog/*` primitives already exist) and, when that lands, a test that exercises real focus rather than dispatching on the overlay.

Net: correct, tested feature that mirrors the established sync-modal pattern; the only thing standing between this and an approve is de-scoping the incidental reformatting so the feature diff stands on its own.

mstomar125:
Addressed — force-pushed `fc4d0ac`.

**🟡 Formatting churn dropped.** Restored `app-shell.svelte` to `main` and re-applied only the three feature hunks (the `ProReloginModal` import, the state/`$derived`/`$effect`/handlers block, and the `<ProReloginModal>` render). The PR diff vs `main` is now **purely additive — 46 insertions, 0 deletions** in `app-shell.svelte`; none of the unrelated reflow (MCP debug logs, `get_settings` generic, `importDirectory`, `select_new_database`, `check_daemon_status`) is touched. Verified my added lines are independently prettier-clean (`prettier` wouldn't reflow any of them); the file's remaining pre-existing style nits are left as-is per your note. Modal test still 7/7.

**🟢 Escape/focus follow-up — agreed, leaving as-is here.** Confirmed it mirrors `version-conflict-modal.svelte:85-93` (overlay-Escape + content `stopPropagation`, no focus-trap), so it's the existing house behaviour, not a regression. Worth a separate repo-wide pass on a shared focus-trapping dialog built on the `ui/dialog/*` primitives, with a test that drives real focus rather than dispatching on the overlay — I'll file that as a follow-up issue so it isn't lost.

Issue #1414, opened 2026-06-15
Shared focus-trapping dialog: Escape/focus for app modals


mstomar125:
## Problem

App-level modals (`version-conflict-modal.svelte`, `pro-relogin-modal.svelte`, the onboarding wizard) use the same hand-rolled pattern: an overlay carrying the `Escape → close` keydown handler, with the dialog content doing `onkeydown={e => e.stopPropagation()}`. None of them trap focus or move focus into the modal on open.

Consequence: under real keyboard focus, an Escape press on the dialog or a button inside it is `stopPropagation`-ed before it reaches the overlay handler, and if focus is on background content it never enters the overlay subtree — so **Escape effectively doesn't dismiss**, and Tab can wander out of the modal. Component tests pass only because they dispatch `keydown` on the overlay element synthetically (e.g. `pro-relogin-modal.test.ts`), not through real focus.

Surfaced in review of #1304 / PR #1413; same behaviour exists at `version-conflict-modal.svelte:85-93`.

## Proposal

Build a shared focus-trapping dialog on the existing `ui/dialog/*` primitives (`dialog-content`, `dialog-overlay`, …): move focus into the dialog on open, trap Tab within it, restore focus to the trigger on close, and handle Escape at the dialog level. Migrate the hand-rolled modals to it.

## Acceptance
- [ ] A shared dialog primitive traps focus and dismisses on Escape through **real** keyboard focus.
- [ ] `version-conflict-modal`, `pro-relogin-modal`, and onboarding wizard use it (hand-rolled overlay/stopPropagation removed).
- [ ] Tests drive real focus (focus an element, press Escape/Tab) rather than dispatching on the overlay.

mstomar125:
**In review** — PR NodeSpaceAI/nodespace-core#1415 delivers the shared `focusTrap` action and migrates the three simple modals (pro-relogin, version-conflict, delete-confirmation) with real-focus tests (8 cases). The onboarding wizard is deferred (multi-step focus management); it's the only remaining checkbox. 3865/3865 unit pass; frontend-only, no sync path touched.

mstomar125:
[...]

PR #1415, opened 2026-06-15
focusTrap action: trap focus + Escape in app modals (part of #1414)

mstomar125:
## What

Addresses the bulk of **#1414** — the keyboard/focus gap in the app's hand-rolled modals, surfaced in the #1413 review.

The modals (overlay + content + `stopPropagation`) never moved focus into the dialog on open and never trapped Tab, so under **real keyboard focus** an Escape press could be swallowed by the content's `stopPropagation` before reaching the close handler, and Tab could wander out to background content. The existing component tests only passed because they dispatched `keydown` on the overlay synthetically.

## Changes

- **`actions/focus-trap.ts`** (new) — a reusable Svelte action applied to a dialog's content element: moves focus inside on open (first focusable, or the container), traps `Tab`/`Shift+Tab`, invokes `onEscape` (caught on the container so inner `stopPropagation` can't hide it), and restores focus to the previously-focused element on close.
- Applied it to the three simple modals with **no markup/styling change** beyond the `use:focusTrap` directive: `pro-relogin-modal`, `version-conflict-modal`, `delete-confirmation-modal` (the last drops its now-redundant Escape branch, keeping Enter→confirm).

## Scope

The 544-line multi-step **onboarding wizard is deliberately deferred** — its per-step focus management needs more care than a drop-in trap. Tracked as the remaining item on #1414.

I chose a focus-trap **action** over migrating to the `ui/dialog` (bits-ui) primitives because those are unused in-app and the modals use custom scoped CSS, not Tailwind — a full migration would mean restyling free-tier modals (visual-regression risk) for no behavioral gain over the action. Pure frontend; no sync/daemon path touched.

## Verification

- New `focus-trap.test.ts` — 8 cases driving **real focus** (focus-in on open, Tab/Shift+Tab wrap, Escape → onEscape, focus restore on destroy, `update()`, propagation stop). The `pro-relogin` Escape test now exercises real focus instead of dispatching on the overlay.
- Full unit suite **3865/3865** (3857 baseline + 8). Lint + prettier clean on all touched files.

Part of #1414.

### DIFF
--- a/packages/desktop-app/src/lib/actions/focus-trap.ts
+++ b/packages/desktop-app/src/lib/actions/focus-trap.ts
@@ -1,0 +1,86 @@
+/**
+ * focusTrap - Svelte action for accessible modal dialogs (#1414)
+ *
+ * The app's hand-rolled modals (overlay + content + `stopPropagation`) never
+ * moved focus into the dialog on open and never trapped Tab, so under real
+ * keyboard focus an Escape press could be swallowed before reaching the close
+ * handler and Tab could wander out to background content. This action fixes all
+ * three concerns in one place, without changing a modal's markup or styling:
+ *
+ *  - on mount it remembers the previously-focused element and moves focus to the
+ *    first focusable element inside the dialog (or the container itself);
+ *  - it keeps Tab / Shift+Tab cycling within the dialog;
+ *  - it invokes `onEscape` when Escape is pressed (caught on the container, so
+ *    `stopPropagation` on inner content can't hide it);
+ *  - on destroy (the modal closing) it restores focus to where it was.
+ *
+ * Apply to the dialog content element, which must be focusable (e.g.
+ * `tabindex="0"`) so focus has somewhere to land when there are no focusable
+ * children:
+ *   <div role="dialog" tabindex="0" use:focusTrap={{ onEscape: close }}>…</div>
+ */
+
+const FOCUSABLE =
+  'a[href], button:not([disabled]), textarea:not([disabled]), input:not([disabled]), select:not([disabled]), [tabindex]:not([tabindex="-1"])';
+
+export interface FocusTrapParams {
+  /** Called when Escape is pressed inside the trap (typically the close handler). */
+  onEscape?: () => void;
+}
+
+export function focusTrap(node: HTMLElement, params: FocusTrapParams = {}) {
+  let onEscape = params.onEscape;
+  const previouslyFocused = document.activeElement as HTMLElement | null;
+
+  function focusable(): HTMLElement[] {
+    return Array.from(node.querySelectorAll<HTMLElement>(FOCUSABLE));
+  }
+
+  function handleKeydown(event: KeyboardEvent) {
+    if (event.key === 'Escape') {
+      event.stopPropagation();
+      onEscape?.();
+      return;
+    }
+    if (event.key !== 'Tab') return;
+
+    const items = focusable();
+    if (items.length === 0) {
+      // Nothing tabbable inside — keep focus pinned to the dialog itself.
+      event.preventDefault();
+      node.focus();
+      return;
+    }
+
+    const first = items[0];
+    const last = items[items.length - 1];
+    const active = document.activeElement;
+
+    if (event.shiftKey) {
+      if (active === first || !node.contains(active)) {
+        event.preventDefault();
+        last.focus();
+      }
+    } else if (active === last || !node.contains(active)) {
+      event.preventDefault();
+      first.focus();
+    }
+  }
+
+  node.addEventListener('keydown', handleKeydown);
+
+  // Move focus into the dialog on open so keyboard users start inside it and the
+  // Escape/Tab handling above actually receives their key presses.
+  (focusable()[0] ?? node).focus();
+
+  return {
+    update(next: FocusTrapParams = {}) {
+      onEscape = next.onEscape;
+    },
+    destroy() {
+      node.removeEventListener('keydown', handleKeydown);
+      // Return focus to whatever had it before the dialog opened.
+      previouslyFocused?.focus?.();
+    }
+  };
+}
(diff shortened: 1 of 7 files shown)


mstomar125:
**REQUEST CHANGES** ⏳

_Initial review, top commit `dae3da1` (against `main`). `no checks reported` on the branch; suite green per the body (3865/3865)._

First — thanks for taking the #1413 a11y note and turning it into a proper reusable fix. The action is the right call (I agree with skipping a `ui/dialog`/bits-ui migration — restyling the scoped-CSS free-tier modals for no behavioral gain isn't worth the visual-regression risk), and the `focus-trap.ts` logic is correct where I checked it: Tab/Shift+Tab wrap, the `items.length === 0 → node.focus()` fallback, Escape `stopPropagation` so inner content can't swallow it, and focus-restore-on-destroy. The tests genuinely drive real focus now (focus an element → dispatch on `document.activeElement`), including the rewritten `pro-relogin` Escape test — exactly what was missing before. Good work.

One blocking issue, introduced by this PR's behavior change:

🟡 **`delete-confirmation-modal` — Enter now confirms an irreversible delete while Cancel is focused.** With the trap applied, focus moves to the first focusable on open, which is the **Cancel** button (`.btn-cancel` is the first `<button>`). That's the right safe default for a click — but the content still has `onkeydown={handleKeydown}` with an unconditional `if (e.key === 'Enter') confirmation.confirm()`. So when the dialog opens with Cancel focused and the user presses Enter:
- the bubbled keydown hits `handleKeydown` → `confirmation.confirm()` runs (the delete), and
- Enter also natively activates the focused Cancel button → `confirmation.cancel()`.

Both fire, and `confirm()` runs first — so **Enter deletes the node even though the focused, highlighted control is Cancel**, on a dialog whose own copy says "This cannot be undone." Before this PR the modal didn't pull focus inside, so that global Enter handler was rarely reached; auto-focus now makes it the default keyboard path. This is the inverse of the #1413 problem (there Escape was unreachable; here Enter is reachable and does the dangerous thing).

Fix is small and in the file you're already touching: drop the global `Enter → confirm` handler and let the focused button handle its own Enter natively (Enter on Delete confirms, Enter on Cancel cancels) — i.e. remove `handleKeydown` / the `onkeydown` on `.modal` entirely, since focusTrap now owns Escape and Tab. If you want an explicit "Enter confirms" affordance, land initial focus on Delete instead and gate the handler to only fire when no button is focused — but the native-button route is simpler and safer. Please add a test asserting Enter-with-Cancel-focused does **not** delete.

One non-blocking note:

🟢 **Redundant overlay Escape handlers.** `pro-relogin-modal` and `version-conflict-modal` still carry `onkeydown={(e) => e.key === 'Escape' && …}` on the overlay. Now that focusTrap always moves focus inside and handles Escape on the container (with `stopPropagation`), those overlay handlers are unreachable — focus is never on the overlay. Harmless as defense-in-depth, but you could drop them for clarity since the trap supersedes them. Your call.

Net: the action and its tests are solid and address the original gap well — the one thing to fix before merge is the delete-confirmation Enter path, which this PR's auto-focus turns into a reliable "Enter on Cancel deletes anyway."

mstomar125:
Addressed — force-pushed `f8b05f0`.

**🟡 Blocking: delete-confirmation Enter-deletes-while-Cancel-focused.** Great catch — that was a real, dangerous regression from the auto-focus. Removed the global `Enter → confirm` handler entirely; `.modal`'s `onkeydown` is now just `stopPropagation` (matches the other two modals, keeps a11y happy). With focus on Cancel, Enter activates Cancel natively (cancels); Enter on Delete confirms. No keyboard path can confirm while Cancel is highlighted.
- Added `delete-confirmation-modal.test.ts` (3 cases): **Enter-with-Cancel-focused does NOT resolve the delete** (the regression guard — fails on the old code), plus Delete-click confirms and Cancel-click cancels. Drives real focus.

**🟢 Non-blocking: redundant overlay Escape handlers.** Dropped them from `pro-relogin-modal` and `version-conflict-modal` — the overlays are `role="presentation"`, so removing the keydown doesn't trip the a11y rule (same as delete-confirmation's `role="none"` overlay), and the trap now owns Escape.

Also re-verified there's **no incidental formatting churn**: `delete-confirmation-modal` picked up a stray whole-file prettier reflow, so I restored it to `main` and re-applied only the functional hunks — its diff is now functional-only, and `prettier --check` is clean on all changed files. Full suite **3868/3868**.

mstomar125:
**APPROVE** ⏳

_Re-review at your request, now at `f8b05f0` (squashed). `no checks reported` on the branch, so this is my read, not a green check — suite reported green in the body._

Everything from my prior pass is resolved, and cleanly:

✅ **The blocking delete-confirmation Enter bug is fixed the right way.** The global `Enter → confirm` handler is gone — `.modal` now just `onkeydown={(e) => e.stopPropagation()}`, matching the other two modals — so with focus defaulting to Cancel, Enter activates Cancel natively and never reaches a confirm path. The comment you left captures exactly why (no global Enter on a "cannot be undone" alertdialog whose focus sits on Cancel). And the new `delete-confirmation-modal.test.ts` locks it down with real focus: asserts `document.activeElement === Cancel`, Enter-on-Cancel never resolves `true`, plus Delete-click → confirm and Cancel-click → cancel. That's precisely the regression test I asked for.

✅ **The non-blocking 🟢 got handled too** — the now-redundant overlay `onkeydown` Escape handlers are removed from both `pro-relogin-modal` and `version-conflict-modal`; the overlays keep only `onclick`, and `focusTrap`'s `onEscape` owns Escape. (Confirmed the content-level `stopPropagation` doesn't shadow the trap: focusTrap's listener is registered on the same node via `addEventListener`, and `stopPropagation` only halts ancestor bubbling, so both run — same pattern the other modals already relied on.)

The `focusTrap` action itself I verified on the previous SHA (Tab/Shift+Tab wrap, no-focusable-children fallback to `node.focus()`, Escape stop+restore, focus-restore on destroy) and its logic is unchanged here. Good to merge — and thanks for turning the #1413 note into a proper shared action with real-focus coverage across all three modals.

PR #1416, opened 2026-06-15
Read-embeddings API for cloud push (core half of sync #97)

mstomar125:
## What

Core half of **NodeSpaceAI/nodespace-sync#97** (T11 — pgvector embedding sync), the **push-path-first** slice. The Pro daemon needs to read locally-stored vectors to mirror them into Supabase pgvector, but core only exposed embedding **search** (KNN) and **write** (upsert) — there was no plain read.

## Changes

- **`row_to_embedding`** helper — decodes a stored row back into the `Embedding` model, including the little-endian f32 blob `upsert_embeddings` writes (`chunks_exact(4)` → `f32::from_le_bytes`).
- **`SqliteStore::get_embeddings(node_id)`** — all of a node's embedding chunks, ordered by chunk index.
- **`SqliteStore::embeddings_modified_since(since)`** — embeddings modified at/after a cursor, across all nodes, ordered by `modified_at`. This is what the daemon's push sweep (PR 2) will advance a cursor over to push newly (re)computed vectors.
- Both surfaced as **`NodeService` wrappers** (the daemon holds `Arc<NodeService>`).

## Note on the `nlp` feature

The new methods are **deliberately not gated** behind `nlp`. They read the persisted `embedding` table, which exists regardless of whether embedding *generation* (llama-cpp) is compiled in — so a daemon built without NLP (cf. sync #10) can still sync vectors. Generation stays `#[cfg(feature = "nlp")]`; only reads are feature-independent.

## Verification

- New unit test: upsert two synthetic 768-dim chunk vectors → both round-trip out of the blob in chunk order; `embeddings_modified_since` includes them at epoch, excludes them in the future; a node with no embeddings reads back empty.
- `cargo test` (new test green) + `cargo clippy --lib` clean. Additive only — no existing method changed.

Part of NodeSpaceAI/nodespace-sync#97 (PR 1 of the incremental push-first plan; PR 2 = daemon push + pin bump).

### DIFF
--- a/packages/core/src/db/schema.sql
+++ b/packages/core/src/db/schema.sql
@@ -57,3 +57,9 @@ CREATE TABLE IF NOT EXISTS embedding (
 CREATE INDEX IF NOT EXISTS idx_emb_node      ON embedding (node_id);
 CREATE INDEX IF NOT EXISTS idx_emb_stale_mod ON embedding (stale, modified_at);
 CREATE UNIQUE INDEX IF NOT EXISTS idx_emb_unique ON embedding (node_id, model_name, chunk_index);
+-- Cloud-push sweep (#97): SqliteStore::embeddings_modified_since does a
+-- stale-agnostic `modified_at >= ?` range scan ORDER BY modified_at, node_id,
+-- chunk_index. idx_emb_stale_mod can't serve it (leading column is `stale`), so
+-- this composite makes the recurring sweep an index range scan that also covers
+-- the ORDER BY (no filesort).
+CREATE INDEX IF NOT EXISTS idx_emb_modified ON embedding (modified_at, node_id, chunk_index);
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -2450,6 +2450,114 @@ impl SqliteStore {
         Ok(())
     }
 
+    /// Decode a stored embedding row into the `Embedding` model. Vectors are
+    /// persisted by `upsert_embeddings` as a little-endian f32 blob; decode it
+    /// back to `Vec<f32>`. Column order must match the SELECTs below.
+    fn row_to_embedding(row: &libsql::Row) -> Result<crate::models::Embedding> {
+        let id: String = row.get(0)?;
+        let node: String = row.get(1)?;
+        let vector_blob: Vec<u8> = row.get(2)?;
+        let dimension: i64 = row.get(3)?;
+        let model_name: String = row.get(4)?;
+        let chunk_index: i64 = row.get(5)?;
+        let chunk_start: i64 = row.get(6)?;
+        let chunk_end: Option<i64> = row.get(7)?;
+        let total_chunks: i64 = row.get(8)?;
+        let content_hash: Option<String> = row.get(9)?;
+        let token_count: Option<i64> = row.get(10)?;
+        let stale: i64 = row.get(11)?;
+        let error_count: i64 = row.get(12)?;
+        let last_error: Option<String> = row.get(13)?;
+        let created_at_str: String = row.get(14)?;
+        let modified_at_str: String = row.get(15)?;
+
+        let vector: Vec<f32> = vector_blob
+            .chunks_exact(4)
+            .map(|b| f32::from_le_bytes([b[0], b[1], b[2], b[3]]))
+            .collect();
+        let created_at = DateTime::parse_from_rfc3339(&created_at_str)
+            .with_context(|| format!("Invalid embedding created_at: {}", created_at_str))?
+            .with_timezone(&Utc);
+        let modified_at = DateTime::parse_from_rfc3339(&modified_at_str)
+            .with_context(|| format!("Invalid embedding modified_at: {}", modified_at_str))?
+            .with_timezone(&Utc);
+
+        Ok(crate::models::Embedding {
+            id,
+            node,
+            vector,
+            dimension: dimension as i32,
+            model_name,
+            chunk_index: chunk_index as i32,
+            chunk_start: chunk_start as i32,
+            chunk_end: chunk_end.map(|v| v as i32),
+            total_chunks: total_chunks as i32,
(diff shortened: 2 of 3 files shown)


mstomar125:
**REQUEST CHANGES** ⏳

_Initial review, top commit `b7106f9` (against `main`). `no checks reported` on the branch; the new unit test is green per the body. Additive only — no existing method touched._

Clean, well-scoped read API, and I verified the two claims that matter most:

✅ **The `nlp`-independence claim holds.** `embedding` is created in `db/schema.sql` (`CREATE TABLE IF NOT EXISTS embedding`), which is applied unconditionally at store init (`include_str!("schema.sql")`, no feature gate) — so a daemon built without `nlp` (sync #10) really does have the table, and these reads won't hit "no such table." Good call keeping reads ungated while generation stays `#[cfg(feature = "nlp")]`.

✅ **Blob round-trip + column mapping are correct.** `row_to_embedding`'s `row.get(0..15)` order matches both SELECT column lists exactly, and the `chunks_exact(4) → f32::from_le_bytes` decode is the inverse of `upsert_embeddings`'s `f.to_le_bytes()` write. The test exercises the real round-trip (axis-0 and axis-5 unit vectors, both chunks in order, empty-node case).

✅ **`>=` cursor is the right choice** — re-fetching the boundary rows on each sweep (at-least-once) beats `>` dropping rows that share the cursor's timestamp; the pgvector upsert is idempotent, so re-push is harmless.

One blocking item, plus one note:

🟡 **`embeddings_modified_since` has no index to stand on — every push sweep full-scans `embedding`.** The query is `WHERE modified_at >= ?1 ORDER BY modified_at, node_id, chunk_index`, with no `stale` predicate (deliberately — you include stale rows). But the only `modified_at` index is `idx_emb_stale_mod ON embedding (stale, modified_at)`, whose leading column is `stale` — SQLite can't use it for a `stale`-agnostic range scan on `modified_at`. So this method, whose entire purpose is to be swept repeatedly by the daemon's cursor, does a full table scan + filesort on every call. On a large vault (one row per chunk per node → tens of thousands of rows) that's a real, recurring cost. You're clearly index-conscious (you added `idx_emb_stale_mod`), so: add `CREATE INDEX IF NOT EXISTS idx_emb_modified ON embedding (modified_at, node_id, chunk_index)` — the composite also covers the `ORDER BY`, so the sweep becomes an index range scan with no separate sort. If you'd rather land the index alongside the daemon sweep in PR 2 (where it actually gets hit), that's defensible — but then say so in the PR so it's a tracked decision, not an oversight.

🟢 **Lexicographic timestamp cursor is correct today, but it's a load-bearing invariant worth a comment.** `modified_at >= ?1` is a string comparison; it equals chronological order only because every writer uses `Utc::now().to_rfc3339()` (fixed-width date/time, `+00:00` offset, AutoSi fractions — and `'+' < '.' < digits` makes the variable-width fraction order-preserving). I confirmed `upsert_embeddings` writes exactly that. The day someone writes a `Z`-suffixed or non-UTC-offset `modified_at`, the ordering silently breaks and the sweep skips rows. A one-line comment on the method pinning "assumes UTC rfc3339 `modified_at`" would protect the next person.

Net: correct, ungated-correctly, and well-tested read layer — the one thing to resolve before merge is whether `embeddings_modified_since` gets its index now or as a tracked PR-2 item, since as written it full-scans on every sweep.

mstomar125:
Addressed — force-pushed \`d40fbff\`.

**🟡 Blocking: missing index for the sweep.** Added it now (it lands where the method lives, not deferred). `CREATE INDEX IF NOT EXISTS idx_emb_modified ON embedding (modified_at, node_id, chunk_index)` — the composite covers both the `modified_at >= ?` range scan and the `ORDER BY modified_at, node_id, chunk_index`, so the recurring sweep is an index range scan with no filesort. Extended the test with an `EXPLAIN QUERY PLAN` assertion that the sweep query uses `idx_emb_modified` and produces no `TEMP B-TREE` (filesort), so the perf property is now locked down by the test, not just claimed.

**🟢 Lexicographic-cursor invariant.** Added a doc comment on `embeddings_modified_since` pinning that it assumes UTC rfc3339 `modified_at` (fixed `+00:00` offset) — a `Z`-suffixed or non-UTC writer would silently break ordering and skip rows.

Test green; additive only.

mstomar125:
**APPROVE** ⏳

_Re-review at your request, now at `d40fbff` (squashed). `no checks reported`; `cargo test`/`clippy` green per the body._

Both items from my last pass are resolved, and the index one came back stronger than I asked:

✅ **The sweep now has its index.** `idx_emb_modified ON embedding (modified_at, node_id, chunk_index)` (schema.sql) — leading on `modified_at` so it serves the stale-agnostic range scan, and ordered to cover the full `ORDER BY` so there's no filesort. The comment explaining why `idx_emb_stale_mod` couldn't serve it is a nice touch for the next reader.

✅ **And it's locked in with a query-plan test.** The new `EXPLAIN QUERY PLAN` assertion checks the sweep query actually rides `idx_emb_modified` and produces no `TEMP B-TREE` — so a future schema change that regresses this back to a full-scan + sort fails the suite. That's better than the "just add the index" I suggested; it makes the performance contract executable.

✅ **Cursor invariant documented.** The `INVARIANT:` block on `embeddings_modified_since` now spells out that the lexicographic `modified_at >= ?` only equals chronological order for UTC `+00:00` rfc3339 writers, and that a `Z`-suffixed/non-UTC timestamp would make the sweep skip rows. Exactly the trip-wire the next person needs.

Everything else (the `row_to_embedding` column-order/blob round-trip, `get_embeddings`, the NodeService wrappers, the `nlp`-independence via the unconditional `schema.sql` table) I verified in the first pass and it's unchanged here. Diff is still additive-only and correctly scoped — the pre-existing `normalize_date_field`/stale-count code near these methods is untouched.

Good to merge. Thanks for turning the index note into a plan-asserting test — that's the right way to keep #97's push sweep honest.

Issue #178, opened 2025-10-07
Implement Codebase Semantic Search for AI Agents


malibio:
## Overview

Implement semantic search over source code repositories for AI agents, enabling intelligent code navigation and reducing grep operations on large codebases.

## Context

**Purpose**: Enable AI agents to semantically search codebases
**Scope**: Async background indexing of source files with incremental updates  
**Foundation**: Builds on embedding service (#108) and semantic search (#107)
**MCP Integration**: Separate from #112 - focuses on indexing infrastructure

## Use Cases

1. **AI Agent Code Navigation**: "Find error handling functions" → Returns relevant code chunks
2. **Intelligent Grep Replacement**: Semantic search reduces brute-force grep operations
3. **Cross-Repository Search**: Search across multiple indexed codebases
4. **Documentation Discovery**: Find relevant code examples and comments

### Data Model
- [ ] Create `code_embeddings` table in Turso
- [ ] Define code chunk schema (file, lines, language, context)
- [ ] Add repository metadata (repo_id, path, git_hash)
- [ ] Implement efficient chunk storage

### Code Chunking Strategy
- [ ] Language-aware chunking (AST-based for Rust/TS/Python)
- [ ] Fallback line-based chunking (512 tokens, 50 token overlap)
- [ ] Preserve context (function/class signatures)
- [ ] Handle various file types (source code, configs, docs)

### File Monitoring
- [ ] Implement file watcher for incremental updates
- [ ] Detect file changes (add/modify/delete)
- [ ] Queue changed files for re-embedding
- [ ] Handle git operations (branch switches, pulls)

### Async Indexing
- [ ] Background indexing queue
- [ ] Batch embedding generation (100+ files)
- [ ] Progress tracking and status updates
- [ ] Error handling and retry logic

### Search Implementation
- [ ] Code-specific search API
- [ ] Language/file-type filtering
- [ ] Repository filtering  
- [ ] Context-aware result ranking

## Acceptance Criteria

- [ ] Repositories can be indexed in background
- [ ] File changes trigger incremental re-indexing
- [ ] Code chunks preserve meaningful context
- [ ] Search returns relevant code snippets with file/line info
- [ ] Performance: <100ms search over 100K code chunks
- [ ] Indexing handles large repos (>10K files)
- [ ] Multiple repositories supported simultaneously
- [ ] Language-specific chunking works for Rust/TS/Python/Go

## Technical Specifications

**Database Schema:**
```sql
[...]

PR #1417, opened 2026-06-15
Embedding write-wrappers for cloud pull apply (core, sync #97 PR3)

mstomar125:
## What

Core half of the **pull** slice of NodeSpaceAI/nodespace-sync#97. The push half (sync #178, merged) used the read API from #1416; the pull half needs to **apply** vectors synced from another device into the local store, but `upsert_embeddings` / `delete_embeddings` were store-only.

## Changes

- **`NodeService::upsert_embeddings(node_id, Vec<NewEmbedding>)`** — replace a node's vectors wholesale (mirrors the local generation path; the symmetric write to #1416's `get_embeddings`).
- **`NodeService::delete_embeddings(node_id)`** — clear a node's vectors (for a remote embeddings delete).

Both are **deliberately not gated** behind `nlp`, same rationale as the read wrappers: applying a *received* vector doesn't need the generation engine, so a daemon built without llama-cpp (sync #10) can still receive embeddings. Generation stays `#[cfg(feature = "nlp")]`.

## Verification

- New unit test: upsert a synthetic 768-dim vector via the wrapper → read it back via `get_embeddings` (round-trip) → delete → assert empty.
- `cargo test` (new test green) + `cargo clippy --lib` clean. Additive only.

Part of NodeSpaceAI/nodespace-sync#97 (PR 3a; PR 3b = the daemon pull: realtime/replay `embeddings` → these wrappers, + the full two-daemon round-trip harness).

### DIFF
--- a/packages/core/src/services/node_service/embedding.rs
+++ b/packages/core/src/services/node_service/embedding.rs
@@ -37,6 +37,34 @@ impl NodeService {
             })
     }
 
+    /// Replace a node's embeddings with `embeddings` (wholesale, like the local
+    /// generation path). Used by the Pro daemon's cloud **pull** to apply vectors
+    /// synced from another device into the local store (#97). Like the read API,
+    /// it is **independent of the `nlp` feature**: applying a received vector
+    /// doesn't require the generation engine, so a daemon built without llama-cpp
+    /// can still receive embeddings. Empty `embeddings` is a no-op (use
+    /// [`Self::delete_embeddings`] to clear).
+    pub async fn upsert_embeddings(
+        &self,
+        node_id: &str,
+        embeddings: Vec<crate::models::NewEmbedding>,
+    ) -> Result<(), NodeServiceError> {
+        self.store
+            .upsert_embeddings(node_id, embeddings)
+            .await
+            .map_err(|e| {
+                NodeServiceError::query_failed(format!("Failed to upsert embeddings: {}", e))
+            })
+    }
+
+    /// Delete all of a node's embeddings. Used by the Pro daemon's cloud pull to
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE** ⏳

_Initial review, top commit on `issue-97-core-embedding-write-api` (against `main`). `no checks reported`; new test + `clippy --lib` green per the body. Additive only, 2 files._

The symmetric write half of #1416, and just as clean. I checked the things that could bite a thin wrapper:

✅ **Both delegate to real, ungated store methods.** `store.upsert_embeddings` (sqlite_store.rs:2376) and `store.delete_embeddings` (:2663) both exist and neither is `#[cfg(feature = "nlp")]` — so the "applies received vectors without the generation engine" claim holds, same as the read wrappers. Error mapping to `query_failed` matches the #1416 wrappers.

✅ **The "empty is a no-op" doc is actually true.** I went to confirm it, expecting the usual unconditional `DELETE ... WHERE node_id` to make an empty upsert silently *clear* the node — but `store.upsert_embeddings` early-returns `Ok(())` on `embeddings.is_empty()` before touching anything, so the comment is correct and the clear-vs-no-op footgun isn't there. `delete_embeddings` is the explicit clear path, and it tears down the `vec_embeddings` vec0 mirror before the `embedding` rows (right order, transactional).

✅ **Round-trip test exercises the real path** — upsert a 768-dim vector via the wrapper, read it back through #1416's `get_embeddings` (axis-3 round-trips), then `delete_embeddings` → empty. That's the apply→read→clear cycle the pull side will drive.

Good to merge. This lines up the core surface for PR 3b (the daemon pull: realtime/replay `embeddings` → these wrappers + the two-daemon round-trip harness), which is where #97 finally gets its end-to-end proof.

Issue #183, opened 2025-10-07
Implement Topic Node Embeddings with Adaptive Chunking


malibio:
## Overview

Implement embedding generation for topic nodes (@ mention pages) with adaptive chunking based on content size. Topic nodes are user-created pages that serve as knowledge units and should be the primary semantic search targets.

**CRITICAL: Use Turso's native vector search capabilities - NO in-memory implementation needed.**

### ✅ Dependencies Ready (Issue #108 - Fully Implemented)

The NLP engine exists at `/packages/nlp-engine/` with complete production-ready code:

**Available API:**
```rust
use nodespace_nlp_engine::{EmbeddingService, EmbeddingConfig};

// Initialize once at app startup
let mut embedding_service = EmbeddingService::new(EmbeddingConfig::default())?;
embedding_service.initialize()?;

// API methods available:
pub fn generate_embedding(&self, text: &str) -> Result<Vec<f32>>
pub fn generate_batch(&self, texts: Vec<&str>) -> Result<Vec<Vec<f32>>>
pub fn to_blob(embedding: &[f32]) -> Vec<u8>      // F32_BLOB for Turso
pub fn from_blob(blob: &[u8]) -> Vec<f32>

// Features:
// - BAAI/bge-small-en-v1.5 (384 dimensions)
// - Metal GPU acceleration with CPU fallback
// - LRU cache (10k capacity, <5ms cache hits)
// - Batch operations
```

### ⚠️ Database Schema Updates Required

**Current State:**
```sql
-- packages/core/src/db/database.rs:175
embedding_vector BLOB,  -- Generic BLOB, not native vector type
```

**Required Changes:**
```sql
-- Change to native Turso vector type
embedding_vector F32_BLOB(384),  -- Native vector with dimension declaration
```

**Required Index (add to create_core_indexes() in database.rs):**
```rust
// Add after line 345 in packages/core/src/db/database.rs
conn.execute(
    "CREATE INDEX IF NOT EXISTS idx_nodes_embedding_vector 
     ON nodes(libsql_vector_idx(embedding_vector))",
    (),
)
.await
.map_err(|e| {
    DatabaseError::sql_execution(format!(
        "Failed to create vector index: {}", e
    ))
})?;
```

### 🟡 LLM Summarization - Defer for MVP

**Status:** Gemma 3 4B-QAT planned but not yet implemented

**MVP Alternative:** Use simple text truncation for topics >2048 tokens
```rust
[...]

PR #1418, opened 2026-06-15
Embedding provenance (local vs remote) for cloud sync (#182/#183)

mstomar125:
## What

Core half of the fix for NodeSpaceAI/nodespace-sync#182 + #183, found in the post-#180 bug-finding pass. The embedding push sweep re-pushed vectors a device merely **pulled** (the pull's `upsert_embeddings` stamps `modified_at = now`, re-arming the push cursor) → write amplification, and on heterogeneous devices an unbounded push↔pull loop (the only echo suppression was bit-exact f32 vector equality).

## Changes

Add an `origin` column to the local `embedding` table — `'local'` (generated here) vs `'remote'` (pulled from another device):
- **`upsert_embeddings`** writes `origin = 'local'` (the generation path; callers unchanged). New **`apply_remote_embeddings`** writes `origin = 'remote'` — the daemon's pull will use it.
- **`embeddings_modified_since`** filters `origin = 'local'`, so the push sweep never re-pushes a received vector. This **breaks the loop**: once a device pulls a remote vector, that row stops being pushed, so the fleet converges instead of ping-ponging — and fixes the per-sweep write amplification.
- **`idx_emb_modified`** now leads with `origin` so the filtered sweep stays an index range scan (no filesort); the EXPLAIN-plan test is updated to the new query shape.

Greenfield: the column is added to `CREATE TABLE` (no migration — DBs reset per project convention; harnesses wipe DBs).

## Verification

- Test: a node's remote (pulled) embedding is stored + readable via `get_embeddings`, but excluded from `embeddings_modified_since` (the push sweep) → never re-pushed. Plus the existing round-trip + index-plan assertions, updated for `origin = 'local'`.
- `cargo test` (26 embedding tests) + `cargo clippy --lib` clean. Additive.

Part of NodeSpaceAI/nodespace-sync#182 / #183 (core half; the sync half wires the pull to `apply_remote_embeddings` and awaits sweep aborts for #183-C1).

### DIFF
--- a/packages/core/src/db/schema.sql
+++ b/packages/core/src/db/schema.sql
@@ -50,6 +50,11 @@ CREATE TABLE IF NOT EXISTS embedding (
     stale        INTEGER NOT NULL DEFAULT 1,
     error_count  INTEGER NOT NULL DEFAULT 0,
     last_error   TEXT,
+    -- Provenance (#182/#183): 'local' = generated on this device, 'remote' =
+    -- pulled from another device via cloud sync. The cloud-push sweep reads only
+    -- 'local' rows, so a pulled vector never gets re-pushed (no cross-device
+    -- re-push loop / write amplification).
+    origin       TEXT    NOT NULL DEFAULT 'local',
     created_at   TEXT    NOT NULL,
     modified_at  TEXT    NOT NULL
 ) STRICT;
@@ -57,9 +62,9 @@ CREATE TABLE IF NOT EXISTS embedding (
 CREATE INDEX IF NOT EXISTS idx_emb_node      ON embedding (node_id);
 CREATE INDEX IF NOT EXISTS idx_emb_stale_mod ON embedding (stale, modified_at);
 CREATE UNIQUE INDEX IF NOT EXISTS idx_emb_unique ON embedding (node_id, model_name, chunk_index);
--- Cloud-push sweep (#97): SqliteStore::embeddings_modified_since does a
--- stale-agnostic `modified_at >= ?` range scan ORDER BY modified_at, node_id,
--- chunk_index. idx_emb_stale_mod can't serve it (leading column is `stale`), so
--- this composite makes the recurring sweep an index range scan that also covers
--- the ORDER BY (no filesort).
-CREATE INDEX IF NOT EXISTS idx_emb_modified ON embedding (modified_at, node_id, chunk_index);
+-- Cloud-push sweep (#97/#182): SqliteStore::embeddings_modified_since does an
+-- `origin = 'local' AND modified_at >= ?` range scan ORDER BY modified_at,
+-- node_id, chunk_index. Leading on `origin` (equality) then `modified_at`
+-- (range) makes the recurring sweep an index range scan that also covers the
+-- ORDER BY (no filesort).
+CREATE INDEX IF NOT EXISTS idx_emb_modified ON embedding (origin, modified_at, node_id, chunk_index);
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -219,6 +219,57 @@ impl SqliteStore {
         .await
         .context("Failed to create vec0 embeddings table")?;
 
+        Self::migrate_embedding_origin(conn).await?;
+
+        Ok(())
+    }
+
+    /// Idempotent DDL migration for the embedding `origin` column (#182/#183).
+    ///
+    /// `schema.sql` is applied with `CREATE TABLE/INDEX IF NOT EXISTS`, which
+    /// CANNOT alter a pre-existing `embedding` table — and `packages/core` ships
+    /// in the community desktop app, whose user DBs are never reset (lazy
+    /// node-migration design). On such a DB the new column + reshaped index would
+    /// be silently absent, so every embedding write and the push sweep's
+    /// `WHERE origin = 'local'` would fail with `no such column: origin`. Add the
+    /// column and rebuild the index here when missing. No-op on fresh DBs (the
+    /// column already exists) and on re-runs.
+    async fn migrate_embedding_origin(conn: &libsql::Connection) -> Result<()> {
+        let mut cols = conn
+            .query("PRAGMA table_info(embedding)", ())
+            .await
+            .context("read embedding table_info")?;
+        let mut has_origin = false;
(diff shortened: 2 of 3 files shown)


mstomar125:
**REQUEST CHANGES** ⏳

_Initial review, top commit `f7c53af` (against `main`). `no checks reported`; 26 embedding tests + `clippy --lib` clean per the body._

This is the right fix for the push↔pull loop / write-amplification I flagged on #179 (and nice that the post-#180 pass found and filed it as #182/#183). The mechanism is clean and I verified the parts that usually go wrong:

- ✅ **INSERT column/param alignment is correct** — `origin` slots in at `?12` with the column list, and `created_at`/`modified_at` shift to `?13`/`?14`. No off-by-one.
- ✅ **The index actually serves the new query.** `idx_emb_modified (origin, modified_at, node_id, chunk_index)` — equality on the leading `origin`, then the `modified_at` range, then the `ORDER BY` columns — is exactly right, and you updated the EXPLAIN-plan test to the new shape so the "no filesort" guarantee stays enforced.
- ✅ **Provenance test is on point** — a remote-applied embedding is readable via `get_embeddings` but excluded from `embeddings_modified_since`, so it's never re-pushed. That's the loop broken, asserted.
- ✅ `origin` is write-only (not surfaced in `row_to_embedding`/`Embedding`), so the unchanged SELECT column order is fine.

One blocking issue — the migration story:

🟡 **The column is added with no DDL migration, which breaks existing `embedding`-bearing DBs — and `packages/core` is shared by the community desktop app.** I went to check the "DBs reset per project convention" claim. Core applies `schema.sql` via `include_str!` with `CREATE TABLE/INDEX IF NOT EXISTS` (sqlite_store.rs:158) — there's no `ALTER TABLE`, no `PRAGMA user_version`, and `migration_registry` is explicitly *node-property* migration only ("Desktop-Safe: No coordinated migrations across installations; Nodes upgrade when accessed"). So on a DB that already has an `embedding` table:
- `CREATE TABLE IF NOT EXISTS` is a no-op → **no `origin` column**;
- `upsert_embeddings_with_origin`'s INSERT references `origin` → `no such column: origin` → **every embedding write fails**;
- `embeddings_modified_since`'s `WHERE origin = 'local'` → same error → **the push sweep errors every 15s**;
- `CREATE INDEX IF NOT EXISTS idx_emb_modified` is skipped (name already exists) → the index keeps its **old** definition.

For the dev harnesses that wipe DBs, fine. But the community desktop app's real user DBs aren't reset (that's the whole point of the lazy node-migration design), and they already carry `embedding` rows from semantic search — so this upgrade silently breaks embedding generation + sync for them. Please either:
- add an idempotent DDL step at store init — `ALTER TABLE embedding ADD COLUMN origin TEXT NOT NULL DEFAULT 'local'` guarded by a `pragma table_info(embedding)` check (SQLite has no `ADD COLUMN IF NOT EXISTS`), **plus** a `DROP INDEX IF EXISTS idx_emb_modified` before the re-`CREATE` so the index picks up the new column on existing DBs; or
- if the reset convention genuinely covers *every* `packages/core` consumer including the shipped community app, say so explicitly in the PR (and ideally assert it) — then this drops to a non-issue. My read of `migration_registry`'s "no resets" stance is that it doesn't, which is why I'm flagging it rather than trusting the one-liner.

Net: the provenance fix itself is correct, well-indexed, and well-tested — it cleanly kills the re-push loop. The thing to resolve before merge is whether an existing DB survives this column add; as written, without a migration, it doesn't.

mstomar125:
Addressed — force-pushed \`47eb7e7\`. You're right, and good catch on the shared-DB stance — `migration_registry`'s "no coordinated resets" is exactly the gap.

**🟡 Migration.** Added `migrate_embedding_origin`, run at store init right after the `schema.sql` apply:
- `PRAGMA table_info(embedding)` → if no `origin` column, `ALTER TABLE embedding ADD COLUMN origin TEXT NOT NULL DEFAULT 'local'` (so existing rows default to local), then `DROP INDEX IF EXISTS idx_emb_modified` + recreate it with `origin` leading (since `CREATE INDEX IF NOT EXISTS` skips the pre-existing originless one). Idempotent and a no-op on fresh DBs (column already present → returns early).

Added a test that builds a **legacy pre-`origin` `embedding` table with a row already in it**, runs the migration, and asserts: the column is added + the existing row defaults to `'local'`, the index SQL now includes `origin`, and a second run is a clean no-op. So an existing community-app DB upgrades in place — no `no such column: origin`.

27 embedding tests + `clippy --lib` clean.

mstomar125:
**APPROVE** ⏳

_Re-review at your request, now at `47eb7e7` (squashed). `no checks reported`; embedding tests + `clippy --lib` clean per the body._

The migration concern is fully and correctly addressed:

✅ **`migrate_embedding_origin` does exactly the right thing, idempotently.** It runs inside `initialize_schema` (right after the `schema.sql` apply, so on every store open), checks `PRAGMA table_info(embedding)` for the column, and:
- if present → early-returns (no-op on fresh DBs and on re-runs);
- if absent → `ALTER TABLE embedding ADD COLUMN origin TEXT NOT NULL DEFAULT 'local'` (valid in SQLite because of the DEFAULT; existing rows backfill to `'local'`), then `DROP INDEX IF EXISTS idx_emb_modified` + recreate with `origin` leading — which is the part that mattered, since `CREATE INDEX IF NOT EXISTS` in `schema.sql` would otherwise have left the old originless index in place on a legacy DB.

So a legacy `embedding`-bearing DB (community desktop app included) now upgrades cleanly instead of erroring on `no such column: origin`, and a fresh DB pays nothing. The doc comment captures the why precisely.

✅ **And it's tested** — `test_migrate_embedding_origin_upgrades_legacy_db` builds a pre-#182 DB (originless table + old index shape, with a row), runs the migration, and asserts the existing row reads back `origin = 'local'` and the index was rebuilt to include `origin`. That's the legacy path locked down.

Backfilling legacy rows to `'local'` is the only sane default (provenance is unknowable retroactively), and the worst case is a one-time re-push of any vector this device had previously pulled — bounded and harmless, not a loop. Everything from the first pass (INSERT col/param alignment, the `origin`-leading index serving the filtered range + ORDER BY, the EXPLAIN-plan test, the remote-excluded-from-sweep provenance test) is unchanged and still correct.

Good to merge — clean fix for the #179 re-push loop, and now safe for existing DBs.

PR #1419, opened 2026-06-17
Recovered Items UI: inline badge, first-open snackbar, restore (core#1303)

mstomar125:
## What

The desktop **Recovered Items** UI (core#1303 viewer/restore half). When a cloud last-writer-wins conflict overwrites a genuine local edit, the Pro daemon now preserves the superseded "loser" to a per-user local-only log `~/.nodespace/recovered-items-<user>.jsonl` (daemon half merged in nodespace-sync#192). This PR surfaces those preserved edits and lets the user restore them:

- **First-open snackbar** — once per session, if the log is non-empty: *"N recovered items from a sync conflict…"* (via the existing conflict-notifications store, new `conflictType: 'recovered-items'`).
- **Inline ⟲ Recovered badge** on the affected node only (`recoveredItems.hasFor(nodeId)` on `base-node`).
- **Popover** showing superseded ("yours") vs winning ("current"), with **Restore** (re-applies your edit via `sharedNodeStore.updateNode`) and **Dismiss**.
- Tauri commands `pro_list/dismiss/clear_recovered_items` reading/maintaining the daemon's snake_case JSONL.
- `scripts/seed-recovered-item.sh` — smoke-test helper that seeds a synthetic entry for a real node.

## Pro-gated / Community-safe

Every command no-ops when `ProClient` is not in managed state, and `recoveredItems.load()` early-returns when not Pro — so the community build is completely inert (no badge, no snackbar), even if a log file exists.

## Validation (physical Apple device, two-window demo)

- **Path A**: seed a recovery entry for a rendered node → relaunch → snackbar fires (`conflictType: recovered-items`) for that node, badge attaches, no errors. Verified via frontend-console capture.
- **Free-user (Community) check**: launched the app against the **community daemon** (`nodespaced`, default socket) with a `recovered-items-default.jsonl` present → tier detected `community`, **0 badge/snackbar events** (Pro UI fully inert), normal editing works, no false daemon banner.
- Build: app compiles clean from this branch; core frontend suite unaffected (zero `packages/core` changes).

Daemon half: nodespace-sync#192. Closes core#1303.


mstomar125:
**APPROVE**

Surfaces the daemon-preserved conflict "losers" (nodespace-sync#192 wrote the log) with an inline badge + first-open snackbar + restore. Reviewed against the code; the load-bearing constraint — community build stays inert — is satisfied with defence in depth:

- **Rust** (`commands/recovered_items.rs`): every command early-returns empty/`Ok` when `app.try_state::<ProClient>().is_none()` — the community build never manages `ProClient`, so the commands no-op.
- **Frontend** (`stores/recovered-items.svelte.ts`): `load()` early-returns `[]` when `!proSync.isPro`; `itemFor()` is then always `undefined`, so `recovered-items-badge.svelte`'s `{#if item}` renders nothing, and the app-shell snackbar effect (`app-shell.svelte:69`) bails on `!proSync.isPro`. The unconditional `<RecoveredItemsBadge {nodeId} />` per node is safe — inert without an item.

Correctness checks that hold: JSONL round-trips the daemon's snake_case format (dismiss/clear preserve it — `dismiss_drops_only_matching_node_and_keeps_daemon_format` proves it), malformed lines are skipped, missing file → empty. The `CONFLICT_MESSAGE` map stays exhaustive over the union. All three commands are registered in the invoke handler. Solid unit coverage (4 tests).

Findings (none blocking):
1. **TOCTOU on the log file** — `pro_dismiss_recovered_item` does read→filter→rewrite-whole-file and `pro_clear_recovered_items` removes it. If the Pro daemon appends a fresh recovered entry between the read and the rewrite, that new entry is clobbered. Low probability and the lost data is only the recovery breadcrumb (node content lives in the store/cloud), but it's a real race on a file the daemon also writes. A comment acknowledging it (or daemon-mediated mutation) would be worth a follow-up; fine to ship now.
2. **`restore()` doesn't await `updateNode` before dismissing** (`recovered-items-badge.svelte:38`) — the `persist:'immediate'` write is fire-and-forget, then `dismiss()` runs. If that persist fails the entry is gone from the log anyway. Minor; `updateNode` surfaces its own conflict toast.
3. Multi-item snackbar links to `items[0].node_id` only; per-node badges cover the rest. Cosmetic.

Note: `statusCheckRollup` came back empty — no CI reported on the PR head. Worth confirming the desktop test suite ran given the device validation in the description.

mstomar125:
Addressed the review (fc810283):

1. **TOCTOU on the log file** — added a doc-comment on `pro_dismiss_recovered_item` acknowledging the read→filter→rewrite race against the daemon's appends, noting the loss is only a recovery breadcrumb and that daemon-mediated mutation is the proper follow-up. Shipping the viewer/restore slice as-is.
2. **restore() not awaiting updateNode** — `sharedNodeStore.updateNode` returns `void` (the app-wide optimistic pattern: it schedules the immediate persist via PersistenceCoordinator, which surfaces its own conflict toast). Awaiting isn't available without changing that contract, so restore() stays consistent with every other edit in the app. Left as-is intentionally.
3. **Multi-item snackbar links to items[0] only** — cosmetic; deferring to the badge-refinement pass (per product).
4. **CI gap** — ran the desktop frontend suite on this branch: **3868/3868 pass** (matches baseline; the new files are additive).

mstomar125:
**APPROVE** (re-review @ fc810283)

Follow-up commit documents the dismiss/clear TOCTOU on the recovery log exactly as raised — notes the read→filter→rewrite isn't atomic against the daemon's appends, that the only loss is a breadcrumb (node content is safe in store/cloud), and that daemon-mediated mutation / file locking is the proper fix tracked as follow-up. Documenting the known race on the viewer/restore slice is the right call. Good to merge.

PR #1420, opened 2026-06-17
Honor NODESPACED_SOCKET in daemon-reachability checks (fix false 'service not running' banner)

mstomar125:
## Problem

Follow-up to #1410. The daemon-reachability checks resolve the socket differently than the gRPC client:

- The client dials `grpc_client::resolve_socket_path()`, which honors **`NODESPACED_SOCKET`** first, then falls back to `~/.nodespace/daemon.sock`.
- But the post-startup reachability check (`lib.rs`) and `check_daemon_status` (the Retry button) **hardcoded** `~/.nodespace/daemon.sock`.

When `NODESPACED_SOCKET` points elsewhere — the two-window demo (`/tmp/ns-demo-*.sock`), or any custom-socket setup — the app connects and works fine, but the checks probe the absent default path and falsely emit `daemon-status: not_running` → a spurious **"NodeSpace background service is not running"** banner. Because `check_daemon_status` has the same flaw, **Retry can't clear it** either.

## Fix

Make `resolve_socket_path()` `pub(crate)` and use it in both checks, so they probe the same socket the client actually dials.

## Impact

- **Shipped app: unchanged** — it leaves `NODESPACED_SOCKET` unset, so resolution falls back to the default socket exactly as before. Verified on a physical device: community daemon on the default socket shows no false banner, and a genuinely-down daemon still correctly raises the banner.
- **Custom-socket / demo setups: fixed** — no more false banner, and Retry works.

Found while validating on the two-window demo. Follow-up to #1410.

### DIFF
--- a/packages/desktop-app/src-tauri/src/lib.rs
+++ b/packages/desktop-app/src-tauri/src/lib.rs
@@ -73,11 +73,8 @@ async fn check_daemon_status() -> String {
     {
         use daemon_setup::{check_daemon_socket, DaemonStatus};
 
-        let home = match dirs::home_dir() {
-            Some(h) => h,
-            None => return "not_running".to_string(),
-        };
-        let socket_path = home.join(crate::constants::DAEMON_SOCKET_RELATIVE);
+        // Probe the SAME socket the gRPC client dials (honors NODESPACED_SOCKET).
+        let socket_path = crate::services::grpc_client::resolve_socket_path();
         return match check_daemon_socket(socket_path.as_path()).await {
             DaemonStatus::Healthy => "healthy".to_string(),
             DaemonStatus::Starting => "starting".to_string(),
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Tight, correct fix. The two reachability checks (`check_daemon_status` at `lib.rs:77` and the post-startup probe at `lib.rs:344`) previously hardcoded `~/.nodespace/daemon.sock` while the gRPC client dialed `resolve_socket_path()` (which honors `NODESPACED_SOCKET`) — so any socket override (two-window demo, custom setups) made the checks probe the wrong path and falsely emit `daemon-status: not_running`, with Retry unable to clear it. Routing both checks through `resolve_socket_path()` makes the probe and the dial agree — single source of truth.

Verified compile-safety of the `fn → pub(crate)` change: both call sites live inside `#[cfg(any(target_os = "macos", target_os = "linux"))]` blocks (`lib.rs:72` and `lib.rs:246`), which are a subset of the `#[cfg(unix)]` guarding `resolve_socket_path` — and the `cfg(not(macos|linux))` fallback at `lib.rs:84` returns `"healthy"` without calling it. No non-unix path references the symbol.

Nit (non-blocking): no automated regression test asserts the `NODESPACED_SOCKET` path is honored by the checks; behavior is covered by the two-window demo manually. Fine to ship as-is.

mstomar125:
Addressed the nit (15289d52): added a unit regression test `resolve_socket_path_honors_env_override_then_falls_back` asserting the resolver returns `NODESPACED_SOCKET` when set and falls back to `~/.nodespace/daemon.sock` otherwise — so the reachability checks provably probe the same socket the gRPC client dials. Runs single-threaded (process-global env) and restores prior env state. Passes locally.

mstomar125:
**APPROVE** (re-review @ 15289d52)

Follow-up commit adds exactly the missing regression test I flagged — `resolve_socket_path_honors_env_override_then_falls_back`, `#[cfg(all(test, unix))]`, asserting the `NODESPACED_SOCKET` override wins and the default is used otherwise, with save/restore of prior env state. Covers the bug this PR fixes. Good to merge.

Tiny nit: the comment says "single-threaded so the env mutation doesn't race," but nothing enforces serial execution — it's safe only because this is the lone test touching `NODESPACED_SOCKET` in the binary. Fine as-is.

PR #1422, opened 2026-06-17
Fix rapid Enter in an empty node piling blank nodes upward (#1421)

mstomar125:
## What

Fixes #1421 — rapid Enter inside a nested node piled blank nodes **upward** and the focused node drifted down; a single slow Enter appeared to do nothing.

## Root cause

`CreateNodeCommand.shouldCreateNodeAbove` returned `true` whenever the cursor was at position 0 — but an **empty node's cursor is always at 0**. So every Enter on an empty node took the "create above" branch (`insertAtBeginning: true` + `focusOriginalNode: true`): it inserts a blank *before* the node and keeps focus on the original. Result: repeated Enter stacked blank nodes upward (with ever-finer fractional insert-*before* orders), focus never advanced, and the fractional orders diverged across two synced devices.

The "create above" intent is to push a **non-empty** node's content down and leave a blank above it. An empty node has nothing to preserve, so Enter should create a sibling **below** and advance focus (standard outliner behavior).

## Fix

Gate the `position <= 0` branch on non-empty content (`return content.trim().length > 0`). Empty nodes fall through to normal split-below.

## Tests

- Added a regression test: empty node + Enter → `insertAtBeginning: false` (below + advance); pre-fix it failed (was `true`).
- Kept a guard: non-empty node with cursor at start still creates above.
- `create-node.command` + keyboard / node-service / rapid-hierarchy suites pass; full frontend suite green (one flaky perf test, passes in isolation).

## Verified live

Two-window demo: typing a child then pressing Enter rapidly now adds clean empty nodes **below** with the cursor advancing, and both windows converge (no divergence).

Closes #1421.

### DIFF
--- a/packages/desktop-app/src/lib/commands/keyboard/create-node.command.ts
+++ b/packages/desktop-app/src/lib/commands/keyboard/create-node.command.ts
@@ -179,9 +179,16 @@ export class CreateNodeCommand implements KeyboardCommand {
    * - Within inline format opening syntax at beginning (e.g., |**, *|*)
    */
   private shouldCreateNodeAbove(content: string, position: number): boolean {
-    // Always create above when cursor is at the very beginning
+    // Create above when cursor is at the very beginning of a NON-EMPTY node —
+    // the intent is to push the existing content down and leave an empty line
+    // above. An empty node has nothing to preserve below, and its cursor is
+    // always at position 0, so the old unconditional `position <= 0` made every
+    // Enter on an empty node insert *above* and keep focus on it
+    // (`focusOriginalNode`) — blank nodes piled upward and focus never advanced,
+    // diverging across sync (#1421). For an empty node, fall through to normal
+    // "create below + advance focus" splitting.
     if (position <= 0) {
-      return true;
+      return content.trim().length > 0;
     }
 
     // For headers, create above when cursor is within or at the end of the syntax area
--- a/packages/desktop-app/src/tests/commands/keyboard/create-node.command.test.ts
+++ b/packages/desktop-app/src/tests/commands/keyboard/create-node.command.test.ts
@@ -325,6 +325,40 @@ describe('CreateNodeCommand', () => {
     });
   });
 
+  // Regression for #1421: rapid Enter in an empty node piled blank nodes UPWARD.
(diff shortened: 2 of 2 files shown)


mstomar125:
**APPROVE**

Correct root-cause fix with minimal blast radius. Verified against the call site:

- `shouldCreateNodeAbove` gates the only branch (`create-node.command.ts:67`) that sets `insertAtBeginning: true` + `focusOriginalNode: true` (`:76-77`). An empty node's cursor is always at position 0, so the old unconditional `position <= 0 → true` made every Enter insert a blank *before* the node and keep focus on the original — hence blanks piling upward, focus never advancing ("slow Enter does nothing"), and divergent fractional insert-before orders across sync (#1421).
- New `return content.trim().length > 0` routes an empty node to the `else` branch (`:80`), which splits an empty string and emits `insertAtBeginning: false` with no `focusOriginalNode` → sibling created **below** and focus advances. Standard outliner behavior, and it kills the upward pile-up + the sync divergence at once.
- **Blast radius is exactly empty nodes**: non-empty content at position 0 still returns `true` (create-above preserved), and the header/syntax-area logic below the early return is untouched. The guard test (`still creates ABOVE for a non-empty node with cursor at start`) locks that in.

Tests are on-point: the regression test asserts `insertAtBeginning: false` and *not* `focusOriginalNode: true` for the empty case (the exact pre-fix failure), plus the non-empty guard. Tier-agnostic command logic, so it fixes both community and Pro as the issue reported, with no Pro coupling.

Note: `statusCheckRollup` is empty — no CI reported on the head. The description says the frontend suite is green locally (one flaky perf test passing in isolation); worth confirming CI runs the keyboard/create-node suites before merge.

mstomar125:
Re: the CI note — confirmed the keyboard/create-node suites are green on this branch:
- `create-node.command` + `keyboard-command-integration` + `reactive-node-service` + `rapid-hierarchy-operations`: **158/158 pass**.
- Full frontend suite: 3869 pass (the lone failure is a timing-based perf test in `textnode-contentprocessor`, which passes in isolation — unrelated to this change).

No code changes requested in the review, so nothing to amend.

Issue #194, opened 2025-10-08
Test Issue


malibio:
This is a test
with multiple lines
and **markdown**

PR #1423, opened 2026-06-17
Coalesce Pro reconnect-replay node bursts into one render (frontend half of sync #188)

mstomar125:
## Frontend half of NodeSpaceAI/nodespace-sync#188 (pairs with sync PR #194)

On reconnect the Pro daemon flushes a caught-up batch of node events as one burst (sync #194). The frontend listener (`tauri-sync-listener`) applied them one at a time — each `node:created`/`node:updated` does an async `getNode` + `setNode` → **one re-render per node**.

## Change
When sync is active (`proSync.isPro`), collect the burst's node ids over a short window (~16ms), fetch them together, then apply them in a **synchronous `setNode` loop** so Svelte coalesces the reactive updates into a single render. Deletes still apply immediately (a queued update for a since-deleted node resolves to `null` and is skipped, so the delete wins).

## Free-user safety
Gated on `proSync.isPro`. Reconnect replay is a **Pro-only** flow, so the community build never enters this path and its per-event behavior is **byte-for-byte unchanged**. Verified: the full frontend suite stays green and the existing per-event tests (which run with the default non-Pro tier) are untouched.

## Tests
- New: a Pro burst of `node:created` events is **deferred** (not applied synchronously) then **applied in one pass** (all nodes land after the window).
- New: the community path (`proSync.tier !== 'pro'`) still applies each event immediately.
- Full desktop-app suite: **3872 passing** (was 3870; +2 new). `eslint` + `svelte-check` clean.

Together with sync #194 (daemon-side event coalescing), this closes the render-trickle described in #188.

### DIFF
--- a/packages/desktop-app/src/lib/services/tauri-sync-listener.ts
+++ b/packages/desktop-app/src/lib/services/tauri-sync-listener.ts
@@ -31,8 +31,76 @@ import { scheduleCollectionRefresh, scheduleSchemaRefresh } from '$lib/utils/col
 import { registerSchemaPlugin, unregisterSchemaPlugin } from '$lib/plugins/schema-plugin-loader';
 import { applyHasChildCreated, applyHasChildUpdated, applyHasChildDeleted } from './hierarchy-sync';
 import { normalizeNodeData } from './node-normalize';
+import { proSync } from '$lib/stores/pro-sync.svelte';
 
 const log = createLogger('TauriSync');
+
+// ---------------------------------------------------------------------------
+// Pro-only reconnect-replay render coalescing (#188)
+//
+// On reconnect the Pro daemon now flushes a caught-up batch of node events as a
+// single burst (sync PR #194). Applied one-by-one, each `node:created/updated`
+// triggers its own async fetch + `setNode` → one re-render per node. To render
+// the burst in one pass, when sync is active we collect the node ids over a tiny
+// window, fetch them together, then apply them in a SYNCHRONOUS `setNode` loop —
+// Svelte batches synchronous store mutations into a single render.
+//
+// Gated on `proSync.isPro`: reconnect replay is a Pro-only flow, so the community
+// build never enters this path and its per-event behavior is byte-for-byte
+// unchanged.
+// ---------------------------------------------------------------------------
+
+/** How long to gather a burst before flushing. One frame (~16ms) is enough to
+ *  collect a daemon-flushed batch without adding perceptible latency. */
+const REPLAY_COALESCE_WINDOW_MS = 16;
+
+const pendingNodeIds = new Set<string>();
+let coalesceTimer: ReturnType<typeof setTimeout> | null = null;
+
+/** Queue a node id for the next coalesced flush (Pro path). */
+function enqueueNodeFetch(nodeId: string): void {
+  pendingNodeIds.add(nodeId);
+  if (coalesceTimer !== null) return;
+  coalesceTimer = setTimeout(flushPendingNodeFetches, REPLAY_COALESCE_WINDOW_MS);
+}
+
+/** Fetch all queued nodes, then apply them in one synchronous pass so the burst
+ *  renders once. A failed fetch for a single node is skipped, never fatal. */
+async function flushPendingNodeFetches(): Promise<void> {
+  coalesceTimer = null;
+  const ids = [...pendingNodeIds];
+  pendingNodeIds.clear();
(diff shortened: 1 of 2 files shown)


mstomar125:
**REQUEST CHANGES**

The coalescing approach is sound and the free-user gating is correct (`queueOrFetchNode` routes through the coalescer only when `proSync.isPro`; community still calls `fetchAndUpdateNode` immediately, byte-for-byte unchanged — confirmed by the retained per-event tests). But the **delete-wins invariant the PR claims is not actually enforced by the code.**

The PR says: *"a queued update for a since-deleted node resolves to null and is skipped, so the delete wins."* That only holds if `getNode` happens to read the backend *after* the delete is durably visible. The `node:deleted` handler (`tauri-sync-listener.ts:187-189`) calls `sharedNodeStore.deleteNode(...)` immediately but **does not remove the id from `pendingNodeIds`**. So:

1. `node:updated(N)` → `enqueueNodeFetch(N)` — N is in `pendingNodeIds`, 16ms timer armed.
2. `node:deleted(N)` arrives within the window → `deleteNode(N)` removes it from the store, **but N stays queued.**
3. `flushPendingNodeFetches` fires → `getNode(N)`. If the daemon's delete isn't yet reflected in what `getNode` reads, it returns the stale node → the synchronous `setNode(N)` loop **re-adds it → the delete is lost and the node resurrects.**

This is precisely the failure class in the freshly-filed sync#193 ("node delete dropped under concurrent edit, deleted node kept on peer"), and this PR opens a *new* window for it inside the reconnect-replay burst — the exact scenario (interleaved deletes + upserts) the coalescer is built for.

**Fix (one line):** in the `node:deleted` handler, also evict the pending re-fetch:
```ts
pendingNodeIds.delete(event.payload.id);
sharedNodeStore.deleteNode(event.payload.id, { type: 'database', reason: 'domain-event' }, true);
```
That closes the larger pre-flush window. A narrower residual remains — a delete landing *during* the `await Promise.all(getNode)` after the ids are already snapshotted at `:71` — which the eviction can't catch; if you want the invariant airtight, have the sync apply loop skip a node that was tombstoned after the snapshot (e.g. re-check store presence / a deleted-id set) rather than relying solely on `getNode` returning null. At minimum, please don't claim "the delete wins" until the pending set is evicted.

Minor (non-blocking): `coalesceTimer` is a module-level singleton never cleared on listener teardown — if `initializeTauriSyncListeners` re-runs, an in-flight timer can fire into the new state. A `clearTimeout` in teardown would tidy that.

Tests are otherwise good (deferral + one-pass apply, community-immediate path). Re-request once the delete eviction lands.

mstomar125:
Addressed the delete-wins finding (thanks — real bug). `node:deleted` now evicts the pending re-fetch via `cancelPendingNodeFetch`: it removes the id from `pendingNodeIds` and, if a flush is mid-fetch, tombstones it so the synchronous apply loop skips it — closing **both** the pre-flush window and the during-`Promise.all` window you flagged. Added `resetNodeFetchCoalescer()` on (re)init to clear a stale timer (the non-blocking note). New test: an update queued then deleted within the window stays deleted. Suite 3873 (+1), lint/typecheck clean. Re-requesting review.

mstomar125:
**APPROVE** (re-review @ f7528710)

Thorough fix — closes **both** windows from the prior review plus the teardown leak:

- **Pre-flush window:** `node:deleted` now calls `cancelPendingNodeFetch` → `pendingNodeIds.delete(id)` before `deleteNode`, so a delete racing a queued upsert in the same window evicts the re-fetch. ✅
- **During-`await` residual** (the one I noted eviction alone couldn't catch): `flushInProgress` + `tombstonedDuringFlush` — a delete landing while `Promise.all(getNode)` is in flight tombstones the id, and the synchronous apply loop skips it (`!node || tombstonedDuringFlush.has(ids[i])`). The `ids[i]`/`fetched[i]` indices stay aligned (Promise.all preserves order), and the set is cleared per-flush in `finally`. ✅ The delete-wins invariant now actually holds.
- **Timer leak:** `resetNodeFetchCoalescer()` on (re)init clears a stale in-flight timer + state. ✅

Free-user path stays inert: the coalescer sets are only populated on the Pro (`isPro`) enqueue path, so `cancelPendingNodeFetch` is a no-op on an empty set in the community build. Regression test (`a delete during the window wins…`) locks the behavior in.

Good to merge with sync#194. Nice turnaround.

PR #1424, opened 2026-06-17
Add free-user guardrail test suite (test:free-users)

mstomar125:
## What
A single, always-runnable suite (`src/tests/free-user-guardrail.test.ts` + a `bun run test:free-users` script) that asserts every Pro/sync-gated **frontend** feature stays inert in the community (free) build — i.e. when `proSync.isPro` is false. It turns the open-core *"Pro/sync work must keep the community build behaviorally unchanged"* rule into a regression test you can run after any cross-repo or Pro-feature change.

## Coverage
- `proSync.isPro` is false for the `community` tier **and** the default `unknown` tier (fresh app, before any probe).
- **Recovered Items (core#1303):** `recoveredItems.load()` is a no-op, **never invokes** the `pro_list_recovered_items` daemon command, and `hasFor`/`itemFor` report nothing — so no inline badge or snackbar can ever render in community.
- **#188 reconnect-replay coalescer:** the community build still applies `node:updated` / `node:deleted` through the unchanged per-event path (no coalescing window engages), with **no** Pro daemon command invoked.

## Notes
- Test-only + one package.json script. No production code touched → free-user-safe by construction.
- Extend this file whenever a new Pro feature adds a frontend surface, so the guarantee stays enforced in one place.
- Full desktop suite: **3880 passing** (3873 + 7 new). eslint + svelte-check clean. `bun run test:free-users` green.

### DIFF
--- a/packages/desktop-app/package.json
+++ b/packages/desktop-app/package.json
@@ -19,6 +19,7 @@
     "sync": "bunx svelte-kit sync",
     "build": "bun run sync && bunx vite build",
     "test": "bunx vitest run",
+    "test:free-users": "bunx vitest run src/tests/free-user-guardrail.test.ts",
     "test:watch": "bunx vitest",
     "test:browser": "bunx vitest run --config vitest.browser.config.ts",
     "test:browser:watch": "bunx vitest --config vitest.browser.config.ts",
--- a/packages/desktop-app/src/tests/free-user-guardrail.test.ts
+++ b/packages/desktop-app/src/tests/free-user-guardrail.test.ts
@@ -1,0 +1,146 @@
+/**
+ * Free-User Guardrail Suite
+ * =========================
+ *
+ * A single, always-runnable suite that asserts every Pro/sync-gated **frontend**
+ * feature stays completely inert in the community (free) build — i.e. when
+ * `proSync.isPro` is false. The open-core rule is that Pro/sync work must keep the
+ * community build behaviorally unchanged; this suite turns that rule into a
+ * regression test you can run after any cross-repo or Pro-feature change.
+ *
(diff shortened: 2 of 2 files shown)


mstomar125:
**APPROVE**

Test-only (one suite + one `package.json` script, zero production code), so free-user-safe by construction — and it codifies the open-core "community build stays behaviorally unchanged" rule into a runnable regression. The assertions are real, not vacuous:

- **Recovered Items:** mocks `@tauri-apps/api/core`'s `invoke` and asserts `recoveredItems.load()` **never calls it** — this genuinely fails if the `!proSync.isPro` early-return in `load()` is ever removed, which is exactly the gate that keeps the daemon command (`pro_list_recovered_items`) and the badge/snackbar out of community. `hasFor`/`itemFor` checks lock in "no badge renders."
- Covers both `community` **and** the default `unknown` tier (fresh app pre-probe) — good catch, since first-paint is where an ungated daemon call would leak.
- Script is correctly scoped: `bunx vitest run src/tests/free-user-guardrail.test.ts`.

One minor note (non-blocking): the **#188 coalescer** community test asserts the node still lands in the store via `getNode` with no `invoke` — which prevents total breakage, but doesn't tightly prove the 16ms coalescing window was *bypassed*: the Pro path also ends in `getNode`+`setNode` (it just uses `backendAdapter`, not `invoke`), so this assertion would still pass if coalescing wrongly engaged in community. If you want the guardrail to actually catch a gating regression on #188, assert immediacy — e.g. the update is applied synchronously within the same tick rather than after the window. As written it's a "still works" test, not a "coalescer didn't engage" test.

Worth keeping this file the canonical place to extend per the PR note. Good to merge.

PR #1425, opened 2026-06-18
Migrate onboarding wizard to shared focusTrap (closes #1414)

### LINKED ISSUE #1414: Shared focus-trapping dialog: Escape/focus for app modals

mstomar125:
## Closes #1414 (the last open item)

#1415 built the shared `focusTrap` action and migrated `version-conflict-modal`, `pro-relogin-modal`, and `delete-confirmation-modal`. The one deferred item was the **onboarding wizard** ("multi-step focus management needs more care"). This finishes it.

## Change
- The wizard's dialog content is now the focusable `role="dialog"` **focusTrap** target (`use:focusTrap={{ onEscape: onClose }}`, `tabindex=0`, `aria-modal`); the backdrop is a plain `role="presentation"` overlay that closes on click. Removed the hand-rolled backdrop `Escape` + click handlers.
- **Multi-step care:** advancing a step or a configure step succeeding swaps the primary action button, which unmounts the focused button and drops focus to `<body>` — where the trap's Escape/Tab handler can't see key presses. An `$effect`, keyed on exactly the state that changes the visible primary action (`currentStep`, `stepSuccess`, `pathWasAlreadyConfigured`), re-anchors focus to the new step's primary button after each transition. No text inputs → it never interrupts typing.

## Tests
New `onboarding-wizard.test.ts` drives **real keyboard focus** (per the issue's standard, not synthetic overlay dispatch): focus lands inside on open, Escape from inside dismisses, focus follows across a step transition, and backdrop-vs-content click behaves. Queries are scoped to the render `container` + `ownerDocument` and assert synchronously after `tick()`, so they're robust to global-document state from other suites (an earlier `screen`/`waitFor` version was flaky under full-suite pollution).

## Validation
- Full desktop suite **3885** passing (+5 new); `bun run test:free-users` green (no community impact); eslint + svelte-check clean.

### DIFF
--- a/packages/desktop-app/src/lib/components/onboarding/onboarding-wizard.svelte
+++ b/packages/desktop-app/src/lib/components/onboarding/onboarding-wizard.svelte
@@ -2,6 +2,7 @@
   import { onMount } from 'svelte';
   import { invoke } from '@tauri-apps/api/core';
   import { createLogger } from '$lib/utils/logger';
+  import { focusTrap } from '$lib/actions/focus-trap';
 
   const log = createLogger('OnboardingWizard');
 
@@ -31,6 +32,10 @@
   let isLoading = $state(false);
   let stepSuccess = $state(false);
   let stepError = $state<string | null>(null);
+
+  // The dialog content element — the focus-trap target and the focus anchor for
+  // step transitions (see the $effect below).
+  let dialogEl = $state<HTMLElement>();
 
   // Which steps are active (some may be skipped if prerequisites missing)
   let showSkill = $state(false);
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Finishes #1414 (the deferred onboarding item from the #1415 focusTrap epic) correctly. Verified against the code:

- **ARIA pattern is right:** the dialog content becomes `role="dialog"` + `aria-modal` + `tabindex=0` carrying `use:focusTrap`, and the backdrop drops to `role="presentation"`. The label moved onto the dialog (the labelled node), and dialog-content `onclick`/`onkeydown` `stopPropagation` keeps inside-clicks from bubbling to the backdrop's `onClose` — the "close on backdrop, not on content" test confirms it. Escape from inside still fires `onClose`: `focusTrap` registers its keydown on the node itself (`focus-trap.ts:70`), so the dialog's `stopPropagation` (which only stops ancestor bubbling) doesn't suppress it — confirmed by the Escape-from-inside test.
- **The multi-step re-anchor is the crux and it's sound.** Swapping the primary button on a transition unmounts the focused element → focus falls to `<body>`, out of the trap's reach. The `$effect` keyed on `${currentStep}:${stepSuccess}:${pathWasAlreadyConfigured}` re-focuses `.primary-button` (falling back to `dialogEl`) after each transition. It runs post-DOM-flush so the new button exists, and it can't interrupt typing — I grepped the wizard, there are **no text inputs/textareas/contenteditable**, so the claim holds. The "re-anchors after advancing a step" test drives real focus + a follow-up Escape to prove the trap still works post-transition.
- Tests scope queries to the render `container` and read focus via `ownerDocument` rather than global `screen`/`document.body` — deliberately robust to cross-file document pollution in the full-suite run. Nice.
- No Pro coupling; shared-component a11y refactor, community + Pro both benefit.

Note: `statusCheckRollup` empty again — no CI reported on the head. Worth confirming the component suite runs in CI before merge.

PR #1441, opened 2026-06-22
fix: daemon stability, stale binary kill safety, and task title propagation

malibio:
## Summary

- **SIGKILL storm fix**: `kill_stale_daemon_sync` was using `lsof -t` to find all PIDs with the daemon socket open, then killing all of them — including gRPC clients like `nodespace-app` itself. Now filters via `ps comm=` to only kill processes whose executable ends with `nodespaced`.
- **Race condition fix**: Frontend was issuing gRPC calls before the daemon socket existed. Added `daemon-status` events (`starting`/`healthy`/`not_running`) emitted from Rust, and `waitForDaemon()` in the layout that blocks initialization until the daemon is ready (30s timeout).
- **Task title fix**: `TaskNode` serialization in `nodespace-types` was dropping the `title` field, causing `search_nodes` to return `title: ""` to the agent even when the DB had the correct value. Added `title: Option<String>` to `TaskNode` and populated it in `task_node_to_value`.

## Test plan

- [ ] `bun run dev:tauri` completes without killing itself (no more SIGKILL storm)
- [ ] Daemon starts and is healthy within a few seconds of app launch
- [ ] Frontend does not issue gRPC calls before daemon socket is ready
- [ ] `search_nodes` returns correct title for task nodes to the agent
- [ ] `cargo build` passes clean

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -292,12 +292,6 @@ jobs:
         with:
           bun-version: "1.3.11"
 
-      - name: Setup Node.js (for npm publish)
-        uses: actions/setup-node@v6
-        with:
-          node-version: '20'
-          registry-url: 'https://registry.npmjs.org'
-
       - name: Install dependencies (skill)
         run: bun install --frozen-lockfile
         working-directory: packages/skill
@@ -305,7 +299,10 @@ jobs:
       - name: Set version from release tag (skill)
         run: |
           TAG="${{ github.event.release.tag_name }}"
-          npm version "${TAG#v}" --no-git-tag-version
+          VERSION="${TAG#v}"
+          # Update version directly — avoids npm's workspace: protocol error
+          contents=$(cat package.json)
+          echo "$contents" | sed "s/\"version\": \".*\"/\"version\": \"${VERSION}\"/" > package.json
         working-directory: packages/skill
 
       - name: Build @nodespaceai/skill
@@ -313,10 +310,11 @@ jobs:
         working-directory: packages/skill
 
       - name: Publish @nodespaceai/skill
-        run: npm publish --access public
+        run: bun publish --access public
         working-directory: packages/skill
         env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          BUN_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
   update-homebrew-cask:
     name: Update Homebrew cask
--- a/packages/desktop-app/src-tauri/src/daemon_setup.rs
+++ b/packages/desktop-app/src-tauri/src/daemon_setup.rs
@@ -104,18 +104,40 @@ pub fn kill_stale_daemon_sync(app: &tauri::App) {
         "Installed daemon binary is stale — killing before gRPC client connects"
     );
 
-    // Kill via lsof → SIGKILL (fast; launchd will restart with fresh binary after extraction)
+    // Kill only nodespaced processes using the socket (not gRPC clients like nodespace-app).
+    // lsof -t returns all PIDs with the socket open — we filter to those whose executable
+    // is named "nodespaced" to avoid SIGKILLing the Tauri app itself.
     let sock = socket_path.to_string_lossy();
     if let Ok(out) = std::process::Command::new("lsof")
-        .args(["-t", "-U", sock.as_ref()])
+        .args(["-F", "pn", "-U", sock.as_ref()])
         .output()
     {
-        for pid in String::from_utf8_lossy(&out.stdout)
-            .split_whitespace()
-            .filter_map(|s: &str| s.parse::<i32>().ok())
-        {
-            unsafe { libc::kill(pid, libc::SIGKILL) };
-            tracing::info!(pid, "Sent SIGKILL to stale nodespaced");
+        let output = String::from_utf8_lossy(&out.stdout);
+        let mut current_pid: Option<i32> = None;
+        for line in output.lines() {
+            if let Some(pid_str) = line.strip_prefix('p') {
+                current_pid = pid_str.parse::<i32>().ok();
+            } else if line.strip_prefix('n').is_some() {
+                // Check if this PID is a nodespaced process
+                if let Some(pid) = current_pid {
+                    let exe_check = std::process::Command::new("ps")
+                        .args(["-p", &pid.to_string(), "-o", "comm="])
+                        .output()
+                        .ok();
+                    let is_daemon = exe_check
+                        .as_ref()
+                        .map(|o| {
+                            String::from_utf8_lossy(&o.stdout)
+                                .trim()
+                                .ends_with("nodespaced")
+                        })
+                        .unwrap_or(false);
+                    if is_daemon {
+                        unsafe { libc::kill(pid, libc::SIGKILL) };
+                        tracing::info!(pid, "Sent SIGKILL to stale nodespaced");
+                    }
+                }
+            }
         }
     }
 
--- a/packages/desktop-app/src-tauri/src/daemon_setup.rs
+++ b/packages/desktop-app/src-tauri/src/daemon_setup.rs
@@ -105,38 +105,43 @@ pub fn kill_stale_daemon_sync(app: &tauri::App) {
     );
 
     // Kill only nodespaced processes using the socket (not gRPC clients like nodespace-app).
-    // lsof -t returns all PIDs with the socket open — we filter to those whose executable
-    // is named "nodespaced" to avoid SIGKILLing the Tauri app itself.
+    // Parse lsof -F pn output to collect unique PIDs, then verify each is nodespaced
+    // before SIGKILLing. Deduplicating into a HashSet avoids spawning ps more than once
+    // per PID when the daemon has multiple FDs open on the socket.
     let sock = socket_path.to_string_lossy();
     if let Ok(out) = std::process::Command::new("lsof")
         .args(["-F", "pn", "-U", sock.as_ref()])
         .output()
     {
+        use std::collections::HashSet;
         let output = String::from_utf8_lossy(&out.stdout);
         let mut current_pid: Option<i32> = None;
+        let mut pids_to_check: HashSet<i32> = HashSet::new();
         for line in output.lines() {
             if let Some(pid_str) = line.strip_prefix('p') {
                 current_pid = pid_str.parse::<i32>().ok();
             } else if line.strip_prefix('n').is_some() {
-                // Check if this PID is a nodespaced process
                 if let Some(pid) = current_pid {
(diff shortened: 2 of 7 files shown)


malibio:
## Code Review — PR #1441

**Overall assessment:** This is a net-positive fix that resolves three distinct, confirmed bugs. The SIGKILL storm prevention is the highest-risk change and is mostly correct, but has two issues worth addressing. The race-condition fix and task title fix are both clean and correct. Recommend **request-changes** for the two items below before merge.

---

### Findings

#### [Improvement] `listen` promise is dropped — potential missed event window

**File:** `packages/desktop-app/src/routes/+layout.svelte`, lines 21–29

`listen<string>(...)` returns `Promise<UnlistenFn>` (confirmed from `@tauri-apps/api` type definitions), but the call is not awaited. This creates a narrow but real race window: if `daemon-status: healthy` is emitted by Rust *before* the listener is actually registered with the Tauri event bus, the frontend will wait the full 30-second timeout before proceeding — degrading cold-start UX.

Additionally, the `UnlistenFn` is never called. The listener leaks for the lifetime of the component. Since `waitForDaemon` is called only once during `onMount`, the leaked listener is benign in practice today, but it is an anti-pattern that the Tauri docs explicitly flag ("removing the listener is required if your listener goes out of scope").

Recommended fix:

```typescript
async function waitForDaemon(): Promise<void> {
  if (!isTauri()) return;
  await new Promise<void>((resolve) => {
    const timer = setTimeout(resolve, 30_000);
    // Await the listen promise so the handler is registered before we could miss an event
    listen<string>('daemon-status', (event) => {
      if (event.payload === 'healthy' || event.payload === 'not_running') {
        clearTimeout(timer);
        unlisten();
        resolve();
      }
    }).then((unlisten) => {
      // Store unlisten so the timeout path can also clean up
      // (If timer fires before listen resolves, that is fine — unlisten is a no-op after resolve)
    });
  });
}
```

A cleaner pattern that eliminates the race entirely uses `once`:

```typescript
import { once } from '@tauri-apps/api/event';

async function waitForDaemon(): Promise<void> {
  if (!isTauri()) return;
  const timeout = new Promise<void>((resolve) => setTimeout(resolve, 30_000));
  const healthy = once<string>('daemon-status', () => {}).then(() => {});
  await Promise.race([timeout, healthy]);
}
```

Though note that `once` would still need to handle both `healthy` and `not_running` — the multi-payload case makes the `listen` approach cleaner, just with `await` added.

#### [Improvement] `kill_running_daemon` still uses unfiltered `lsof -t`

**File:** `packages/desktop-app/src-tauri/src/daemon_setup.rs`, lines 256–266

The `kill_running_daemon` async function (called during `ensure_daemon_running` when the binary is updated) still uses the old `lsof -t` pattern and sends `SIGTERM` to every PID holding the socket open — including potential gRPC clients. This is the same class of bug that was fixed in `kill_stale_daemon_sync`. The consequence is less severe here (SIGTERM vs. SIGKILL, and there is a 5-second grace period), but the gRPC client receiving an unexpected SIGTERM is still incorrect behavior.

The fix should mirror what was done in `kill_stale_daemon_sync`: parse `-F pn` output and verify with `ps comm=` before sending the signal.

---

### Nitpicks

**Nit:** `packages/desktop-app/src-tauri/src/daemon_setup.rs`, lines 121–139 — The `ps -p <pid> -o comm=` subprocess is spawned once per file-descriptor entry in `lsof` output, not once per PID. A single `nodespaced` process holding multiple FDs to the socket (e.g., control + data FDs) will spawn multiple `ps` invocations and issue multiple `SIGKILL`s to the same PID. The duplicate `SIGKILL`s are harmless (the second call returns `ESRCH`), but the extra `ps` invocations are wasteful. Deduplicating PIDs before the `ps` check (e.g., collect into a `HashSet<i32>`) would be cleaner.

**Nit:** `packages/nodespace-types/src/task.rs`, line 113–114 — The `title` field is placed before `version` in the struct. The existing ordering roughly follows `id → node_type → content → version → timestamps`. Inserting `title` after `content` breaks that pattern slightly. Since this is a serialized wire type, field order in the struct has no semantic impact (serde is field-name-driven), but consistency aids readability.

**Nit:** `packages/desktop-app/src-tauri/src/lib.rs`, lines 266–267 — Emitting `daemon-status: starting` is a good signal, but it is only consumed by `waitForDaemon` as a triggering condition for the listen registration. Since `waitForDaemon` resolves on `healthy` or `not_running` (not `starting`), the `starting` emission has no functional effect on the current wait logic. It is preserved for future use (e.g., a loading spinner), which is fine, but worth a comment clarifying that intention so the next reader does not try to remove it.

**Nit:** `packages/desktop-app/src/lib/types/task-node.ts` — The TypeScript `TaskNode` interface (line 79–95) does not include a `title?: string` field to match the new Rust `TaskNode.title: Option<String>`. The field is omitted in serde when `None` via `skip_serializing_if`, so it will not break deserialization, but the TypeScript interface is now slightly out of sync with the wire shape. Worth adding `title?: string` for parity.

---

### Summary

| Area | Status |
|------|--------|
| SIGKILL storm fix (`kill_stale_daemon_sync`) | Correct — filters correctly via `ps comm=` |
| `kill_running_daemon` SIGTERM path | Same bug class still present — see Improvement above |
| `listen` not awaited / not cleaned up | Race window + leak — see Improvement above |
| Task title serialization | Correct fix |
| `unsafe` usage | Appropriate; `kill()` with a valid PID and signal constant is safe. Safety comments present. |
| lsof `-F pn` parsing | Correct; `p` prefix = PID, `n` prefix = file name/path. Field ordering in lsof output is PID block then per-FD entries, so the state machine is correct. |

malibio (commented):
Two improvements needed before merge: (1) listen in waitForDaemon() is not awaited — race window where daemon-status healthy fires before listener is registered. (2) kill_running_daemon still uses unfiltered lsof -t (same class as the fixed bug). See full review comment.

malibio:
## Re-Review — PR #1441 (address-review commit)

**Scope:** This re-review covers only the single address-review commit \`493366be\` merged on top of the original \`3c3ef166\` baseline. All four items from the original review were addressed. No new blockers introduced.

---

### Previous Findings — Resolution Status

#### [Improvement] \`listen\` promise not awaited — race window + listener leak

**Status: Resolved.**

The fix in \`+layout.svelte\` is correct. The function now uses \`async (resolve) =>\` so that \`await listen(...)\` executes before the \`setTimeout\` is set, and the captured \`unlisten\` is called on both the happy path (event received) and the timeout path. The race window is closed and the listener no longer leaks.

One minor note: using an \`async\` function as a \`Promise\` executor is technically an antipattern because errors thrown inside the async body become unhandled rejections rather than rejected promises. In this specific usage the only \`await\` is \`listen()\` (a Tauri API call that should not throw in normal operation), so the practical risk is negligible. The fix is accepted.

#### [Improvement] \`kill_running_daemon\` still used unfiltered \`lsof -t\`

**Status: Resolved.**

\`kill_running_daemon\` now uses the same \`-F pn\` parse-then-ps-verify pattern as \`kill_stale_daemon_sync\`. The implementation is consistent across both kill paths.

---

### Additional Findings in the Address-Review Commit

#### [Nit] PID deduplication — collected before socket-path filter

In both \`kill_stale_daemon_sync\` and the updated \`kill_running_daemon\`, the \`lsof -F pn\` loop accumulates PIDs into a \`HashSet\` on any \`n\`-prefixed line, regardless of what path that line contains. When the daemon has multiple FDs beyond the socket (e.g. the write end of a pipe, an open log file), those FDs will also appear as \`n\`-lines and their PIDs will be inserted into the set. Because the set deduplicates, the practical result is still one \`ps\` call per unique PID — just using slightly more FD-lines than the socket-only path would. The safety check (\`ps comm=\` verify before kill) remains correct. This is not a bug, just a minor precision note.

#### [Nit] \`use std::collections::HashSet;\` inside function body

Both kill functions declare \`use std::collections::HashSet;\` inside the innermost block. Moving the import to the module top-level would be more idiomatic Rust, but this is stylistic and consistent with the surrounding \`use std::process::Command;\` pattern already present.

---

### CI-Only Change (\`release.yml\`)

The \`bun publish\` + \`sed\` version-bump replacement for \`npm publish\` + \`npm version\` is pragmatically correct: npm cannot resolve \`workspace:*\` references, bun can. The \`sed\` inline version-bump is slightly fragile (assumes the first match of \`"version": "...")\`, but acceptable for a CI release script with no ambiguity in a leaf \`package.json\`. One note: \`BUN_AUTH_TOKEN\` is the correct env var for bun's npm registry auth; \`NPM_TOKEN\` is also set but unused by bun — harmless dead env var.

---

### Summary

All material findings from the original review are addressed. The \`waitForDaemon\` race is closed, \`kill_running_daemon\` now guards against killing gRPC clients, both kill functions deduplicate PIDs, and the TypeScript \`TaskNode\` interface is in sync with the Rust wire type. The remaining notes are nits with no correctness impact.

**Recommendation: Approve.**

malibio:
## Review Findings Addressed

All four findings from the review have been implemented.

### Important #1 — `listen` promise dropped / unlisten leak
**Fixed in** `packages/desktop-app/src/routes/+layout.svelte`

Restructured `waitForDaemon` to avoid the async promise executor pattern entirely (which also fixes an ESLint `no-async-promise-executor` violation that was present in the original). The new approach:
1. Creates a shared resolver slot (`resolveWait`) from a synchronous Promise executor.
2. `await listen(...)` registers the handler and captures `unlisten` before the timer starts — no race window.
3. Both the event callback and the 30s timeout call `unlisten()` then `resolveWait()`.
4. `clearTimeout(timer)` after `await wait` handles the happy-path cleanup.

### Important #2 — `kill_running_daemon` uses unfiltered `lsof -t`
**Fixed in** `packages/desktop-app/src-tauri/src/daemon_setup.rs`

Replaced `lsof -t` with the same `-F pn` + `ps comm=` pattern used in `kill_stale_daemon_sync`. Each PID is verified as a `nodespaced` process before receiving SIGTERM. This prevents gRPC clients from being inadvertently terminated.

### Nit #1 — PID deduplication in `kill_stale_daemon_sync`
**Fixed in** `packages/desktop-app/src-tauri/src/daemon_setup.rs`

Both `kill_stale_daemon_sync` and `kill_running_daemon` now accumulate PIDs into a `HashSet<i32>` during the lsof parse pass, then run the `ps comm=` verification loop once per unique PID.

### Nit #3 — TypeScript `TaskNode` interface missing `title?: string`
**Fixed in** `packages/desktop-app/src/lib/types/task-node.ts`

Added `title?: string` after `content` in the `TaskNode` interface, matching the new optional Rust field.

### Nit #2 — Field ordering in task.rs
No code change needed (serde is name-driven, ordering is style-only).

### Verification
- `cargo build -p nodespace-app` — clean
- `bun run test` — 140 files, 3885 tests, all passing
- `bun run quality:fix` — 3 pre-existing errors in unrelated files; no new errors introduced

PR #1444, opened 2026-06-22
Ollama availability check is one-shot on mount — shows 'not running' if Ollama starts after the Settings page loads

malibio:
Closes #1442

### DIFF
--- a/packages/desktop-app/src/lib/components/settings/model-manager.svelte
+++ b/packages/desktop-app/src/lib/components/settings/model-manager.svelte
@@ -31,6 +31,26 @@
   let ollamaModels = $state<{ id: string; name: string }[]>([]);
   let ollamaChecking = $state(true);
 
+  async function checkOllama() {
+    ollamaChecking = true;
+    try {
+      ollamaRunning = await ollamaAvailable();
+      if (ollamaRunning) {
+        const list = await chatModelList();
+        ollamaModels = list.filter((m) => m.backend === 'ollama').map((m) => ({ id: m.id, name: m.name }));
+      } else {
+        ollamaModels = [];
+      }
+    } catch (e) {
+      log.warn('Ollama check failed', e);
+      ollamaRunning = false;
+      ollamaModels = [];
+    } finally {
+      ollamaChecking = false;
+    }
+    buildDefaultOptions();
+  }
+
   // --- OpenAI-compat configs ---
   let openAiConfigs = $state<OpenAiCompatConfig[]>([]);
   let editingConfig = $state<OpenAiCompatConfig | null>(null);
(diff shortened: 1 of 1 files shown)


malibio:

This is a focused, surgical fix for a real usability bug. The change extracts the Ollama availability check into a reusable \`checkOllama()\` function and adds a Refresh button to the Ollama section header, mirroring the pattern already established by the Local section's refresh button. The implementation is clean and follows existing conventions throughout the file.

---

### Requirements Check

- ✅ **Refresh button visible** — renders unconditionally in the Ollama section header (always visible, disabled during check). This is actually better UX than gating it on \`ollamaRunning === false\` — users can refresh even when Ollama shows as running to pick up newly pulled models.
- ⏭️ **Retry with backoff** — not implemented; the issue marked this as an alternative/optional criterion. The Refresh button satisfies acceptance criteria independently.
- ✅ **Model list populates after refresh** — \`checkOllama()\` fetches \`ollamaAvailable()\`, then on success fetches \`chatModelList()\` and populates \`ollamaModels\`. The \`else\` branch explicitly clears \`ollamaModels = []\` on failure, preventing stale data.
- ✅ **No regressions** — \`onMount\` delegates to \`checkOllama()\` which preserves identical behavior. The Local, OpenAI-compat, and Default model sections are untouched. Bonus: the \`ollamaModels = []\` reset on failure is a minor bug fix over the original code.

---

### Findings

#### 🟢 Suggestion: Move `buildDefaultOptions()` into `finally`

📁 `packages/desktop-app/src/lib/components/settings/model-manager.svelte:51`

Currently placed after the `try/catch/finally` block. It runs correctly on both success and error paths, but the placement is semantically misleading — it looks like it only runs on success. Moving it inside `finally` makes intent explicit and guards against future early-return refactoring:

\`\`\`ts
} finally {
  buildDefaultOptions();
  ollamaChecking = false;
}
\`\`\`

`buildDefaultOptions()` does not read `ollamaChecking`, so this is functionally equivalent.

#### 🟢 Suggestion: `aria-label` inconsistency

The Local section uses `aria-label="Refresh"` while the Ollama button uses `aria-label="Refresh Ollama status"`. The Ollama label is more descriptive — consider aligning the Local section to `aria-label="Refresh local models"` for consistency.

#### 🟢 Nit: Duplicated SVG icon

The refresh SVG path is duplicated verbatim in both refresh buttons. Low-friction to maintain now (same file), but a `{#snippet refreshIcon(spinning)}` would eliminate duplication if a third refresh button is added. YAGNI — not worth acting on now.

---

This correctly fixes the reported bug with a minimal, focused implementation. The extraction of `checkOllama()` improves cohesion, the button mirrors the established Local section pattern, and the `ollamaModels = []` reset on failure is a correctness improvement over the original. No blockers.

malibio:
## Review Feedback Addressed

✅ **Addressed 2/3 suggestions:**

- **`buildDefaultOptions()` moved into `finally`** — now runs before `ollamaChecking = false` is set, making intent explicit and guarding against future early-return refactoring
- **`aria-label` aligned** — Local section now uses `"Refresh local models"` to match the Ollama button's `"Refresh Ollama status"` convention

⏭️ **Skipped 1 suggestion:**

- **SVG snippet refactor** — YAGNI: only 2 buttons in the same file. If a third refresh button is added, a `{#snippet refreshIcon(spinning)}` would be worth it at that point.

PR #1445, opened 2026-06-22
Ollama integration tests don't catch context-window truncation causing empty responses

malibio:
Closes #1443

### DIFF
--- a/packages/agent/src/local_agent/ollama_inference.rs
+++ b/packages/agent/src/local_agent/ollama_inference.rs
@@ -73,6 +73,8 @@ struct OllamaOptions {
     temperature: Option<f32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     num_predict: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    num_ctx: Option<u32>,
 }
 
 #[derive(Deserialize)]
@@ -105,7 +107,8 @@ struct OllamaToolCallFunction {
 
 #[derive(Deserialize)]
 struct OllamaShowResponse {
-    modelinfo: Option<serde_json::Value>,
+    #[serde(rename = "model_info")]
+    model_info: Option<serde_json::Value>,
 }
 
 #[async_trait]
@@ -141,6 +144,19 @@ impl ChatInferenceEngine for OllamaInferenceEngine {
                 .collect()
         });
 
+        // When tools are present, fetch the model's context window and set num_ctx
+        // so Ollama uses the full context instead of its 4096-token default.
+        // Fall back to 32768 if /api/show is unreachable or the key is missing.
+        let num_ctx = if tools.is_some() {
+            let ctx = match self.model_info().await {
+                Ok(Some(spec)) => spec.context_window,
+                _ => 32_768,
+            };
+            Some(ctx)
+        } else {
+            None
+        };
+
         // Ollama does not deliver tool_calls through its streaming API for Gemma 4
         // (and likely other models) — tool calls only appear in the final
         // non-streaming response. Use stream:false so tool calls are reliably
@@ -156,6 +172,7 @@ impl ChatInferenceEngine for OllamaInferenceEngine {
             options: OllamaOptions {
                 temperature: request.temperature,
                 num_predict: request.max_tokens,
+                num_ctx,
             },
         };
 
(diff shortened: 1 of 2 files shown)


malibio:
### Code Review Summary

This PR fixes three compounding bugs that caused Ollama to silently truncate prompts at the 4096-token default context window, producing empty responses. The root causes are precisely diagnosed and the fixes are accurate. The strengthened integration tests close the coverage gap that allowed the regression to go undetected.

This is a clear net improvement and addresses all acceptance criteria from issue #1443. The implementation is correct. I have one Improvement recommendation and a handful of Nits — none block merge.

---

### Findings

#### Suggested Improvements

**`packages/agent/src/local_agent/ollama_inference.rs` (lines 147–158) — Extra `/api/show` round-trip on every tool-bearing request**

Every call to `generate()` with tools now issues a blocking HTTP request to `/api/show` before the actual inference call. For a model with a stable context window (e.g., Gemma 4 running locally indefinitely), this is redundant work on every turn of a multi-turn conversation. Since `OllamaInferenceEngine` is constructed with a fixed `model_name`, the context window for that model doesn't change at runtime.

The fix is straightforward — cache the result in the struct with a `tokio::OnceCell<u32>` or a simple `std::sync::OnceLock`. This keeps the first-call behavior identical (with graceful fallback) while eliminating the per-request overhead for subsequent calls.

```rust
// In OllamaInferenceEngine struct:
context_window_cache: tokio::sync::OnceCell<u32>,

// In generate():
let num_ctx = if tools.is_some() {
    Some(*self.context_window_cache.get_or_init(|| async {
        match self.model_info().await {
            Ok(Some(spec)) => spec.context_window,
            _ => 32_768,
        }
    }).await)
} else {
    None
};
```

This is an improvement rather than a blocker because the local Ollama daemon responds in <1ms, but the pattern of performing avoidable I/O on every inference call should be addressed before the call rate grows.

---

#### Nitpicks

**`packages/agent/src/local_agent/ollama_inference.rs` (line 110–111) — Redundant `#[serde(rename)]`**

The `#[serde(rename = "model_info")]` annotation is correct, but unnecessary: serde's default behavior for `snake_case` field names already maps `model_info` to the JSON key `"model_info"`. The rename attribute implies the JSON field *differs* from the Rust field name, which is misleading to future readers. It is also what the original bug was — the Rust field was named `modelinfo` (no underscore), which serde mapped to the JSON key `"modelinfo"` instead of `"model_info"`. Now that the Rust field is correctly named `model_info`, the `#[serde(rename)]` annotation can be removed without changing behavior.

Nit: Remove `#[serde(rename = "model_info")]` on line 110 to reduce noise and make the intent clearer — the field name itself is now correct.

---

**`packages/agent/tests/ollama_integration.rs` (line 463–465) — `completion_tokens_indicate_real_response` abstraction has marginal value**

The helper function wraps a single `> 1` comparison and is only called once. The threshold value (1) is meaningful and worth documenting, but a named function for a one-liner used once adds a layer of indirection without clarity gain.

Nit: Inline the comparison with a comment:
```rust
// completion_tokens == 1 with an empty body is Ollama's signature for context-window truncation
assert!(usage.completion_tokens > 1, "completion_tokens={} — looks like context-window truncation ...", ...);
```

---

**`packages/agent/tests/ollama_integration.rs` (lines 326–397) — Tool definitions duplicated between `test_ollama_inference_with_production_prompt` and `all_skill_tools()`**

The 14 tool definitions in `test_ollama_inference_with_production_prompt` (lines 326–397) partially overlap with the definitions in `all_skill_tools()` (lines 577–689). The production prompt test intentionally uses the *full set*, not the whitelisted subset, but there's still duplicated JSON plumbing. This isn't a bug, but as the tool list evolves keeping both in sync will require double maintenance.

Nit: Consider extracting a `production_tool_definitions()` helper returning all 14 tools, then having both `test_ollama_inference_with_production_prompt` and any future full-set callers use it. `all_skill_tools()` can remain separate since its descriptions differ (they're purposefully minimal stubs).

---

### Notes

- The key lookup fallback chain (specific known keys → wildcard `.context_length` suffix) in `model_info()` is well-designed. The ordered preference prevents ambiguity if a model ever exposes multiple `.context_length` keys under different namespaces.
- The `num_ctx` is correctly set to `None` when no tools are present, preserving the existing streaming path behavior for pure text generation where context pressure is lower.
- Graceful skip behavior in all three new integration tests is correct — `eprintln!` + `return` (not `assert!`) when Ollama is unavailable.
- The `test_ollama_model_info_context_window_detection` assertion (`> 4096`) is a good regression pin. It will catch a future serde rename regression immediately, which is the primary value of that test.

malibio:
## Review feedback addressed

**✅ Addressed (3):**
- **OnceCell caching** — added \`cached_context_window: OnceCell<u32>\` to \`OllamaInferenceEngine\`; \`/api/show\` is now called at most once per engine instance instead of on every \`generate()\` call with tools.
- **Redundant \`#[serde(rename)]\`** — removed; serde's default snake_case mapping already handles \`model_info\` → \`"model_info"\`.
- **Inlined \`completion_tokens_indicate_real_response\`** — one-liner now lives inline with an explanatory comment about why \`> 1\` is the threshold.

**⏭️ Skipped (1):**
- **Production tool definitions helper** — the two definitions serve genuinely different purposes (full production schemas in the production-prompt test vs. minimal stubs in \`all_skill_tools()\`). A shared helper would obscure that distinction without reducing maintenance burden; if the tool list grows significantly this can be revisited.

PR #1456, opened 2026-06-24
Shared gRPC router factory for community and Pro daemons (ADR-043)

malibio:
Closes #1454

### DIFF
--- a/packages/daemon/src/lib.rs
+++ b/packages/daemon/src/lib.rs
@@ -5,6 +5,7 @@
 //! out. Proto types are provided by the `nodespace-proto` crate; this lib
 //! re-exports them alongside the service implementations.
 
+pub mod router;
 pub mod services;
 pub mod tray;
 
@@ -46,6 +47,7 @@ pub use nodespace_proto::{
     TerminateSessionResponse, UpdateCaptureSettingsRequest, WriteInputRequest, WriteInputResponse,
 };
 
+pub use router::{build_base_router, BaseServices};
 pub use services::{
     AgentSessionHandler, EmbeddingsServiceImpl, ImportServiceImpl, LocalAgentServiceImpl,
     NodeServiceImpl, SettingsServiceImpl,
--- a/packages/daemon/src/main.rs
+++ b/packages/daemon/src/main.rs
@@ -25,10 +25,9 @@ use nodespace_core::{NodeService as CoreNodeService, SqliteStore};
 use nodespace_daemon::services::embeddings_service::EmbeddingReady;
 use nodespace_daemon::tray::layer::TrayMetricsLayer;
 use nodespace_daemon::{
-    resolve_db_path, tray, AgentSessionHandler, AgentSessionServiceServer, EmbeddingsServiceImpl,
-    EmbeddingsServiceServer, ImportServiceImpl, ImportServiceServer, LocalAgentServiceImpl,
-    LocalAgentServiceServer, NodeServiceImpl, NodeServiceServer, SettingsServiceImpl,
-    SettingsServiceServer,
+    build_base_router, resolve_db_path, tray, AgentSessionHandler, BaseServices,
+    EmbeddingsServiceImpl, ImportServiceImpl, LocalAgentServiceImpl, NodeServiceImpl,
+    SettingsServiceImpl,
 };
 use nodespace_nlp_engine::EmbeddingService;
 use tokio::sync::RwLock;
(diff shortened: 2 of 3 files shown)


malibio:
### Code Review Summary

This PR implements ADR-043: a shared gRPC router factory that enforces compile-time synchronisation between the community daemon and the Pro daemon. The design is sound, the implementation is clean, and this is a clear net improvement over the prior state (four hand-rolled, divergence-prone `.add_service()` chains).

Two findings worth resolving before merge — one is a correctness concern, one is a structural consistency issue. Both are small fixes.

---

### Findings

#### [Improvement] `BaseServices` fields are inconsistently named relative to `ServiceBundle`

**File:** `packages/daemon/src/router.rs` (lines 24–31) and `packages/daemon/src/main.rs` (lines 406–418)

`ServiceBundle` uses `import` and `settings`; `BaseServices` uses `import_svc` and `settings_svc`. All four `serve_*` functions now contain a mapping block that translates `bundle.import` → `import_svc` and `bundle.settings` → `settings_svc`. That translation is pure noise — it adds four nearly-identical struct-literal conversions with no semantic benefit.

Since `ServiceBundle` is private to `main.rs` and `BaseServices` is the public API, the right fix is to align `ServiceBundle`'s field names to `BaseServices` (or vice-versa), eliminating the translation. With aligned names, each `serve_*` function could destructure or forward directly without the intermediate `base_services` let-binding.

**Principle:** YAGNI / DRY — the name mismatch serves no purpose and will silently drift if a field is added to one struct but not the other.

#### [Improvement] `mut server` / `mut router` variable bindings are unnecessary in `build_base_router`

**File:** `packages/daemon/src/router.rs` (lines 54–72)

`server` is declared `mut` in the parameter list, but `Server<L>::add_service` takes `&mut self`, so the borrow is taken internally — the parameter itself does not need to be `mut`. Similarly, `router` is declared `mut` only to allow the conditional `router = router.add_service(...)` pattern for `embeddings`. This is the old-style workaround; the idiomatic Rust approach is:

```rust
pub fn build_base_router<L>(
    server: Server<L>,
    services: BaseServices,
) -> tonic::transport::server::Router<L>
where
    L: Layer<Routes> + Clone,
{
    let router = server
        .add_service(NodeServiceServer::new(services.node_service))
        .add_service(AgentSessionServiceServer::new(services.agent_session))
        .add_service(ImportServiceServer::new(services.import_svc))
        .add_service(SettingsServiceServer::new(services.settings_svc))
        .add_service(LocalAgentServiceServer::new(services.local_agent));

    match services.embeddings {
        Some(emb) => router.add_service(EmbeddingsServiceServer::new(emb)),
        None => router,
    }
}
```

Note: `Server::add_service` takes `&mut self` and returns `Router`, so `server` actually does need to be bound `mut` before the first call — but declaring it in the parameter list as `mut server` is idiomatic Rust when you need to mutate an owned parameter, so that part is fine. The `mut router` / reassignment for `embeddings` is the part worth cleaning up.

---

#### Nitpicks

- **Nit:** `packages/daemon/src/router.rs` line 10 — `use tonic::transport::Server;` is imported but only appears in the doc-comment examples (not in the function signature, which uses the fully-qualified path in the return type). If `tonic::transport::Server` is referenced in the function body via `server: Server<L>`, the import is needed; verify the compiler does not warn about an unused import here depending on whether `Server` resolves through the parameter type.

- **Nit:** `packages/daemon/src/lib.rs` line 50 — `pub use router::{build_base_router, BaseServices};` is placed before `pub use services::{...}`. Alphabetical ordering would be consistent with the existing block above it, but this is a minor style point.

---

- [x] `packages/daemon/src/router.rs` defines `BaseServices` struct and `build_base_router()` (public) — met
- [x] `packages/daemon/src/main.rs` uses `build_base_router()` — no direct `.add_service()` calls in main — met across all four `serve_*` functions
- [ ] `nodespaced-pro/src/main.rs` in nodespace-sync — cannot verify (separate private repo), but the factory is now public and correctly exported from the lib crate
- [ ] `architecture/daemon-cli-overview.md` updated — not in this diff; confirm docs landed in nodespace-docs separately
- [ ] `architecture/daemon-cli-protocol.md` updated — not in this diff; confirm docs landed in nodespace-docs separately

The two doc criteria are noted as a separate-repo concern in the issue; the code-side criteria are fully met.

---

### Overall Assessment

The compile-time enforcement mechanism is exactly right: a struct field addition causes a compile error in `nodespaced-pro` until it provides the implementation. The extraction is clean, the generics are correct, and the four call sites are consistent. Resolve the field-naming inconsistency and the `mut router` pattern, then this is good to merge.

malibio (commented):
Two targeted improvements before merge: (1) align BaseServices field names with ServiceBundle to eliminate the four redundant struct-literal translation blocks in main.rs; (2) replace the mut router / conditional reassignment pattern in build_base_router with a match expression. Both are small, no architectural rework needed. Full details in the inline comment.

malibio:
### Re-Review: Commit 2afd794e

Both findings from the initial review have been addressed correctly and completely.

---

#### Finding 1 — Field naming inconsistency [Resolved]

`BaseServices.import_svc` and `BaseServices.settings_svc` have been renamed to `import` and `settings` respectively (`router.rs` lines 27–28). All four `BaseServices { ... }` struct literals in `main.rs` (lines 142–149, 200–207, 313–320, 383–390) have been updated consistently. The field names now align one-to-one with `ServiceBundle`, which is exactly the intent of the feedback: eliminate the semantic noise of the translation step.

#### Finding 2 — `mut router` / `if let` reassignment pattern [Resolved]

`router.rs` lines 61–71: `let mut router` is gone. The builder chain now terminates in `let router`, followed by a `match services.embeddings` expression that returns from both arms. This is the idiomatic Rust pattern described in the review. The implicit result of the `match` is returned from the function without a trailing `router` statement — clean and correct.

The nit about `mut server` in the parameter list (line 55) is still present, but as the original review noted, `mut` on an owned parameter you need to mutate is idiomatic Rust — this is not a defect.

---

#### Overall Assessment

The refactor commit is surgical: two renames across five files, one structural change to `build_base_router`, zero scope creep. The implementation now matches the stated intent of ADR-043 exactly. No new issues introduced. This is good to merge.

PR #1457, opened 2026-06-24
Batch B (P1): collection member traversal (#1426) + member_of cycle guard (#1427) + FTS backfill (#1428)

mstomar125:
## Summary

Phase 1 bug-sweep **Batch B**, core collection/search half — three related data-layer fixes (all in the community build).

### `#1426` — recursive collection members followed the wrong edge
`get_collection_members_recursive` descended the subtree with `has_child`, but collection hierarchy is built from **`member_of`** edges, so the recursive arm matched nothing — only the seed's *direct* members were returned and every **sub-collection's members were silently dropped**.
**Fix:** descend `member_of` parent→child (`out_node`→`in_node`), restricted to collection children, with a depth cap.

### `#1427` — no cycle validation on collection `member_of` hierarchy
`validate_no_cycle` only covered `has_child`; collection hierarchy (`member_of`) had no guard, so `a member_of b` + `b member_of a` created a cycle in the supposed DAG (which, post-#1426, loops the recursive walk).
**Fix:** new `validate_no_member_of_cycle`, called before creating a collection→collection `member_of` edge; rejects with the existing-but-unused `collection_cycle` error. No-op for ordinary content membership.

### `#1428` — FTS5 index never backfilled
The external-content `node_fts` triggers only index *future* writes, so nodes predating the FTS table (user DBs are never reset) were absent from the index and never matched by `bm25_search_roots`.
**Fix:** one-time idempotent FTS5 rebuild, gated on the **indexed-doc count** (`node_fts_docsize` shadow table) differing from `node` — so a healthy DB doesn't re-index its whole corpus on every startup. (`count(*) FROM node_fts` is unusable as the signal: for an external-content table it reads content rowids and always equals the node count.)

## Tests
- New: sub-collection members returned; member_of cycle (and self-edge) rejected, non-cycle allowed; FTS backfill re-indexes an unindexed node.
- **Full core lib suite: 960/960 pass** (no regression). `cargo fmt`/`clippy -D warnings` clean.
- **`test:free-users`: 7/7** — community build behavior intact.

## Free/community impact
These ARE community-build fixes (core data layer) — additive correctness, no Pro/sync coupling; free-users guardrail green.

Closes #1426, closes #1427, closes #1428.

### DIFF
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -203,6 +203,8 @@ impl SqliteStore {
             ()
         ).await.context("Failed to create FTS5 delete trigger")?;
 
+        Self::backfill_fts_if_stale(conn).await?;
+
         // sqlite-vec virtual table for embedding KNN search. Keyed by `embedding.id`
         // (the per-chunk UUID); holds ONLY real, non-stale vectors (see upsert/delete/
         // mark-stale paths). vec0 is a fast brute-force SIMD scan, not an ANN index.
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Reviewed all three bundled P1 fixes against `8af4978` (fork `origin` lags, so fetched the upstream PR head). All correct, well-tested, confined to community-build data-layer correctness — no Pro/sync coupling.

**#1426 — recursive collection members** (`sqlite_store.rs:3359-3401`): real bug — the old recursive arm joined on `has_child` while collection hierarchy is `member_of`, so `coll_subtree` never grew and sub-collection members were dropped. New query descends `member_of` parent→child (`r.out_node = cs.node_id` → `r.in_node`), restricted to `node_type = 'collection'` children so it doesn't traverse into content members, with a depth cap. Seed (depth 0) is included, so direct members still return. Exposed to the desktop UI (`collections.rs:123` → `collection-service.ts:106`) — user-facing correctness fix. Test exercises a 2-level nest.

**#1427 — member_of cycle guard** (`sqlite_store.rs:1465-1512`, `relationship.rs:500-518`): direction correct — adding `source member_of target` is a cycle iff target is already a descendant of source; the recursive CTE walks source's descendants and matches target, self-edge short-circuits. Wired at `NodeService::create_relationship`, and both service entry points (`collection_service.rs:494` `add_to_collection_by_path`, `:539` `add_to_collection`) route through it. Reuses the existing `collection_cycle` error (daemon maps to InvalidArgument at `node_service.rs:1881`). Test covers closing edge, self-edge, allowed edge.

**#1428 — FTS backfill** (`sqlite_store.rs:278-313`): placed after the `node_fts` triggers in `initialize_schema`. Staleness signal is sound — for external-content FTS5, `node_fts_docsize` is the right shadow table; `node.content` is `NOT NULL DEFAULT ''` so every row yields a docsize entry → healthy DB has `indexed == node_count`, no re-`rebuild` on each startup. Idempotent, no-op on fresh DBs. Test re-indexes a simulated pre-FTS node.

**Mandate:** shared core, but every change is additive bug-fix correctness, no Pro gating; free-users 7/7, lib 960/960 self-reported.

Nits (non-blocking):
- `bulk_add_to_collections` (`sqlite_store.rs:3471`, used by import `import_service.rs:569`) inserts `member_of` edges via raw INSERT, bypassing `validate_no_member_of_cycle`. Pre-existing, not worsened here, but if imports can carry collection→collection edges a malformed import could still seed a cycle — worth a follow-up ticket.
- Depth cap of 100 silently truncates rather than erroring on a >100-deep hierarchy.
- CI: `statusCheckRollup` empty; `mergeStateStatus: CLEAN`; test/lint self-reported.

PR #1458, opened 2026-06-24
Batch C (P2): append-order sentinel (#1431) + conflict-vs-NotFound (#1432) + foreign-write signal (#1437)

mstomar125:
## Summary

Phase 1 bug-sweep **Batch C** (P2), core edge cases — three independent fixes.

### `#1431` — append-order sentinel misreads orders ≤ 0
"Append at end" treated `last_order > 0.0` as "siblings exist", but fractional ordering legitimately yields orders ≤ 0 (a prepend → 0.0, then -1.0…). A lone sibling at ≤ 0 was misread as "no siblings", so the append used the first-item default instead of computing from the real max → sibling-order collisions.
**Fix:** track presence as `Option<f64>` at all three append sites; append after the max regardless of sign.

### `#1432` — `version_conflict{actual:0}` instead of `NotFound` on concurrent delete
`update_node` mapped a no-op version-checked UPDATE to `version_conflict{actual:0}` for *both* a version mismatch and a concurrent delete, re-reading the version via `get_node` (which virtualizes a date page → phantom version 1).
**Fix:** new `persisted_version` (real row, no virtualization) — missing → `node_not_found`; present → `version_conflict` with the real version.

### `#1437` — foreign write to a focused node was silently dropped
A foreign `database` broadcast for an actively-edited node was skipped (to protect optimistic text) with **no signal**, relying on a later OCC that may never fire — the user keeps typing over a stale base, unaware.
**Fix:** raise a deduped `version-mismatch` conflict notification in the not-own-echo branch.

## Tests
- New: negative-order append; `persisted_version` missing-vs-present (incl. date id); foreign-write notifies / own-echo doesn't / dedupes.
- **Core lib 959/959 + full frontend 3888/3888** pass; `clippy -D warnings`/`fmt`/eslint clean; **`test:free-users` 7/7**.

## Free/community impact
Community-build fixes (core + frontend) — additive correctness/UX, no Pro coupling; free-users green.

Closes #1431, closes #1432, closes #1437.

### DIFF
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -517,17 +517,19 @@ impl SqliteStore {
             libsql::params![parent_id.to_string()],
         ).await.context("Failed to get last child order")?;
 
-        let last_order = if let Some(row) = rows.next().await? {
-            row.get::<Option<f64>>(0)?.unwrap_or(0.0)
+        // #1431: distinguish "no siblings" from "a sibling at order <= 0".
+        // Fractional ordering legitimately yields orders <= 0 (a prepend gives
+        // 0.0, then -1.0…), and `unwrap_or(0.0)` maps a NULL order to 0.0. The old
+        // `last_order > 0.0` sentinel misread a lone child at 0.0 as "no children",
+        // so the next append also got 1.0 → two siblings collided at 1.0. Track
+        // presence as Option and append after the max regardless of sign.
+        let last_order: Option<f64> = if let Some(row) = rows.next().await? {
+            Some(row.get::<Option<f64>>(0)?.unwrap_or(0.0))
         } else {
-            0.0
+            None
         };
 
-        let new_order = if last_order > 0.0 {
-            FractionalOrderCalculator::calculate_order(Some(last_order), None)
-        } else {
-            FractionalOrderCalculator::calculate_order(None, None)
-        };
+        let new_order = FractionalOrderCalculator::calculate_order(last_order, None);
 
         let properties = if properties.is_null() {
             serde_json::json!({})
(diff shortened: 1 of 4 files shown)


mstomar125:
**APPROVE**

Reviewed all three bundled P2 fixes against the head ref (`71b21014`); each is correct and test-covered. Core lib + frontend + free-users reported green; `mergeStateStatus: CLEAN`.

**#1431 — append-order sentinel.** Real root cause: the old `last_order > 0.0` sentinel conflated "no siblings" with "a lone sibling at order ≤ 0", and `unwrap_or(0.0)` made a NULL-order row indistinguishable from absence. Fix tracks presence as `Option<f64>` at all three append sites (`sqlite_store.rs:482`, `:3086`, `:3138`) and defers to `FractionalOrderCalculator::calculate_order(last_order, None)`. Verified against `fractional_ordering.rs:18`: `(Some(prev), None) => prev + 1.0`, so a lone sibling at `-1.0` → `0.0` (not the buggy `1.0`). All three queries `ORDER BY json_extract(...,'$.order') DESC LIMIT 1`, so "append after the true max regardless of sign" holds. Test `test_append_order_uses_negative_sibling_not_default` (`sqlite_store.rs:3768`) asserts exactly `0.0`.

**#1432 — version_conflict{actual:0} vs NotFound.** `persisted_version` (`sqlite_store.rs:744`) reads the real `node` row with no date-page virtualization → a concurrently-deleted/never-existing node yields `None` → `node_not_found`, while a real row → `version_conflict` with the true version (`crud.rs:1000`). `node_not_found` constructor confirmed at `services/error.rs:106`. Fixes both the phantom-v1 date case and the misleading `actual:0`. Tests cover missing / present / date-id-missing.

**#1437 — silent foreign-write.** New `else` (`shared-node-store.svelte.ts:1429`) is correctly nested inside the `isDatabaseSource && existingNode && isActivelyEdited && !isAiChatNode` guard and only fires on the genuine-foreign (not-own-echo) branch, leaving the own-echo version-stash untouched. Dedup by scanning existing `version-mismatch` notifications is correct; optimistic text still protected. Tests cover notify / no-notify-on-own-echo / dedupe + `dismissAll()` reset in `beforeEach`.

**Community-build mandate:** all additive correctness/UX in shared core + frontend, no Pro/sync coupling; consistent with `test:free-users 7/7`.

Nits (non-blocking):
- `unwrap_or(0.0)` still maps a genuinely-NULL-order sole sibling to presence `Some(0.0)` → next append `1.0`. Harmless (a NULL-order row carries no competing numeric position), but worth a one-line note that NULL orders are treated as 0.0-by-presence.
- CI: no checks reported on the branch; green numbers are self-reported in the PR body.

PR #1453, opened 2026-06-24
Batch A (P0): data-integrity fixes — transactional move_node (#1429) + window-close/reload store guards (#1435, #1436)

mstomar125:
## Summary

Phase 1 bug-sweep **Batch A**, core half — three P0 data-integrity fixes. Each mirrors a guard/pattern that already exists on a sibling path, so they're small and low-risk.

### `#1429` — make cross-parent `move_node` transactional
`move_node` ran DELETE old `has_child` edge + INSERT new edge as two un-enclosed `db.execute` calls. An INSERT failure after the committed DELETE left the node with **no** parent edge — a silently-orphaned root. `move_children_to_parent` / `delete_subtree_atomic` already wrap their multi-statement work in a transaction.
**Fix:** wrap delete-old + insert-new in one `transaction()` + `commit()`. Make-root / same-parent-reorder are single statements (already atomic).

### `#1435` — `flushPending` executing-op guard
`flushPending` (run by `flushAllPending` on window close) re-invoked `pending.operation()` for **every** entry, including ones already executing — a debounced save mid-RPC fires twice → OCC conflict / duplicate create at the worst moment. Sibling `flushAndWaitForNodes` guards this with `!executingOperations.has(nodeId)`.
**Fix:** add the same guard; still await the in-flight promise.

### `#1436` — `batchSetNodes` skip-while-editing guard
`batchSetNodes` (used by `doLoadChildrenTree` with a `database` source) lacked `setNode`'s `isFocused || hasPending` skip, so a concurrent tree reload overwrote a child being edited mid-keystroke (the sync#76 corruption class).
**Fix:** apply the same skip-while-editing guard + server-confirmed-version stash per node in the database-source loop.

## Tests
- Rust: `test_move_node_cross_parent_leaves_exactly_one_edge` (asserts exactly one edge, under the new parent). `cargo fmt`/`clippy -D warnings` clean.
- Frontend: new `flushPending` double-execute test + 4 `batchSetNodes` guard cases. **Full suite 3890/3890 pass.**
- `test:free-users`: **7/7 pass** — community build behaviorally unchanged.

## Free/community impact
None behaviorally — `#1435`/`#1436` harden existing local-store guards (additive skips), `#1429` makes a store mutation atomic. `test:free-users` green.

[...]

PR #1459, opened 2026-06-24
Batch D (P2): collection pagination (#1430) + transactional bulk_delete (#1433) + bulk_update merge (#1434)

### LINKED ISSUE #1430: [bug] Collection pagination pushes `offset` to SQL before the in-memory membership filter → wrong/truncated results
**Severity:** medium · **Confidence:** high

For a `collection_id`-filtered query (`ops/node_ops.rs:419-429,479-487`), membership filtering happens in memory *after* the SQL query, but `offset` is pushed to the SQL layer where it skips rows of the *unfiltered* set → `offset>0` over a collection returns wrong/missing members + unstable pagination. Separately, with `limit=None` the over-fetch is hardcoded to 1000 (line 421) → members outside the 1000 newest nodes are silently dropped.

**Fix:** drive collection-scoped queries from `get_collection_members` and apply offset/limit in memory; don't push offset to SQL.

---
_Found via the 2026-06-18 multi-repo bug-hunt sweep. Confidence/severity from the sweep + spot-verification._

mstomar125:
## Summary

Phase 1 bug-sweep **Batch D** (P2), core bulk/query fixes.

### `#1430` — collection pagination paginated the wrong set
Collection-filtered queries pushed `offset`/`limit` to SQL, but membership is filtered **in memory after** the query — so `offset>0` paged the unfiltered set and a fixed **1000** over-fetch silently dropped members outside the newest 1000 rows.
**Fix:** scope the query to the collection's members (`NodeFilter::with_ids`, bounded) and apply `offset`/`limit` in memory after the membership filter; short-circuit empty collections.

### `#1433` — bulk_delete was not transactional
Documented all-or-nothing, but looped `store.delete_node` (each its own autocommit) → the Nth failing left the first N-1 committed.
**Fix:** new transactional `store.bulk_delete` (single tx, FK cascade, one `Deleted` event per node post-commit); service routes through it.

### `#1434` — bulk_update empty changed_properties + wholesale replace
Emitted `NodeUpdated` with hardcoded empty `changed_properties` and wholesale-replaced (un-normalized) properties, diverging from single-update and no-opping property-change automation.
**Fix:** normalize + deep-merge like `update_node`, validate the merged candidate, persist the merged value, emit the real `changed_properties`.

## Tests
- New: collection pagination over members vs global set; transactional multi-delete; bulk_update merge + non-empty changed_properties.
- Updated the existing `partial_fields_preserves_unspecified` test to the new merge+normalize contract (it encoded the old wholesale-replace behavior #1434 fixes).
- **Full core suite green** (958 lib + all integration); `clippy --tests -D warnings`/`fmt` clean; **`test:free-users` 7/7**.

## Free/community impact
Community-build fixes (core data layer) — additive correctness; free-users green.

Closes #1430, closes #1433, closes #1434.

### DIFF
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -1197,6 +1197,44 @@ impl SqliteStore {
             bind_values.push(libsql::Value::Text(nt.clone()));
         }
 
+        // #1430: id-scoping (e.g. a collection's members). Build `id IN (…)` and
+        // CHUNK it under SQLite's bound-parameter ceiling so a large member set
+        // can't overflow the limit. Each chunk also carries the title/node_type
+        // conditions already collected. The caller (NodeService::query_nodes)
+        // applies order_by + limit/offset in memory, so per-chunk order is
+        // irrelevant here. Empty id set ⇒ no rows can match.
+        if let Some(ref ids) = query.ids {
+            if ids.is_empty() {
+                return Ok(Vec::new());
+            }
+            const ID_CHUNK: usize = 900;
+            let mut nodes = Vec::new();
+            for chunk in ids.chunks(ID_CHUNK) {
+                let mut conds = conditions.clone();
+                let mut binds = bind_values.clone();
+                let start = binds.len();
+                let placeholders: Vec<String> = chunk
+                    .iter()
+                    .enumerate()
+                    .map(|(i, _)| format!("?{}", start + i + 1))
+                    .collect();
+                conds.push(format!("id IN ({})", placeholders.join(", ")));
+                for id in chunk {
+                    binds.push(libsql::Value::Text(id.clone()));
+                }
+                let sql = format!("SELECT * FROM node WHERE {}", conds.join(" AND "));
+                let mut rows = self
+                    .db
+                    .query(&sql, binds)
+                    .await
+                    .context("Failed to query nodes by id set")?;
+                while let Some(row) = rows.next().await? {
+                    nodes.push(Self::row_to_node(&row)?);
+                }
+            }
+            return Ok(nodes);
+        }
+
         let where_clause = if !conditions.is_empty() {
             format!("WHERE {}", conditions.join(" AND "))
         } else {
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -902,6 +902,57 @@ impl SqliteStore {
         })
     }
 
+    /// Atomically delete multiple nodes in a SINGLE transaction (#1433). Either
+    /// every existing target row is removed or none are — a mid-batch failure rolls
+    /// the whole thing back, so a caller that gets `Err` can rely on nothing having
+    /// been deleted. FK CASCADE removes each node's relationships + embeddings (same
+    /// as `delete_node`); children not in `ids` are left in place (edges cascade).
+    /// Emits one `Deleted` notification per node that existed, AFTER commit. Returns
+    /// the nodes that were deleted.
+    pub async fn bulk_delete(&self, ids: &[String], source: Option<String>) -> Result<Vec<Node>> {
+        if ids.is_empty() {
+            return Ok(vec![]);
+        }
+        // Snapshot the rows that actually exist (for the post-commit notifications).
+        let existing = self.get_nodes_by_ids(ids).await?;
+        let to_delete: Vec<Node> = ids
+            .iter()
+            .filter_map(|id| existing.get(id).cloned())
+            .collect();
+        if to_delete.is_empty() {
+            return Ok(vec![]);
+        }
+
+        let tx = self
+            .db
+            .transaction()
+            .await
+            .context("Failed to begin bulk delete transaction")?;
+        let placeholders: Vec<String> = (1..=to_delete.len()).map(|i| format!("?{i}")).collect();
+        let sql = format!("DELETE FROM node WHERE id IN ({})", placeholders.join(", "));
+        let params: Vec<libsql::Value> = to_delete
+            .iter()
+            .map(|n| libsql::Value::Text(n.id.clone()))
+            .collect();
+        tx.execute(&sql, params)
+            .await
+            .context("Failed to delete nodes in bulk")?;
+        tx.commit()
+            .await
+            .context("Failed to commit bulk delete transaction")?;
+
+        for node in &to_delete {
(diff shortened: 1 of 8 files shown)


mstomar125:
**REQUEST CHANGES**

Solid work on #1433 and #1434 — but the headline #1430 fix is built on a method that doesn't do what the PR claims, and it introduces an unbounded full-table scan on every collection query.

**Blocker — #1430: `with_ids` does not scope the query; collection queries are now unbounded.**
`ops/node_ops.rs:438` scopes the collection query via `filter.with_ids(member_ids…)` with a comment promising it's "bounded." It isn't — `NodeFilter.ids` is never translated into SQL:
- `NodeQuery` (`packages/nodespace-types/src/node.rs:118`) has **no `ids` field** — only `id`, `mentioned_by`, `content_contains`, `title_contains`, `node_type`, `limit`, `offset` (verified). The `NodeFilter→NodeQuery` conversion (`services/node_service/query.rs:55`) silently drops the plural `ids`.
- `db/sqlite_store.rs query_nodes` builds its WHERE only from `title_contains` + `node_type`; there is no `id IN (…)` path. A crate-wide grep confirms `NodeFilter.ids` is never read to build a query.

Net effect: the collection branch sets `with_ids` but not limit/offset, so `filter.limit == None`. `DEFAULT_QUERY_LIMIT` is applied only in `query_nodes_simple` (`query.rs:299`), **not** in the filter-based `query_nodes` — so the store runs `SELECT * FROM node` with **no LIMIT and no id filter**: a full-table scan that materializes every node, runs per-type schema fetch + lazy migration over the whole table, then filters membership in memory. That's a regression from the old bounded 1000 over-fetch — large workspaces load the entire node table on every collection-scoped query. The result is functionally *correct* (the in-memory filter still yields the right members), which is exactly why the new 5-node test passes and hides it.

Required fix (one of):
1. Actually honor ids — add `ids: Option<Vec<String>>` to `NodeQuery`, carry it through `query.rs:55`, and build an `id IN (?, …)` clause in `query_nodes` (mind SQLite's ~999 bound-param ceiling — chunk if a collection can exceed it); OR
2. If in-memory membership filtering is intentional, drop the misleading `with_ids`/"(bounded)" comment and push a real SQL bound (e.g. capped `with_limit(offset+limit)`) so the query isn't unbounded.

Either way, add a test that **fails** under the current implementation — e.g. create `>limit` non-member nodes *after* the members and assert members are still returned; the current `test_collection_query_paginates_members_not_global_set` can't catch the unbounded/scope bug.

**Looks correct:**
- #1433 transactional `bulk_delete` (`sqlite_store.rs:851`): single tx, FK CASCADE preserves old per-row `delete_node` semantics, one post-commit `Deleted` notify per existing node coalesced under `begin_batch_emit`. Missing-id skip tested.
- #1434 `bulk_update` merge (`bulk.rs:520-575`): normalize + deep-merge + validate + `compute_property_changes` mirror `update_node_unchecked` (`crud.rs:653-673`) exactly, incl. the schema-node branch. Well tested.

**Community impact:** all changes are shared-core data layer (not Pro-gated). #1433/#1434 are behavior-preserving; the #1430 change alters shared query behavior and adds the unbounded-scan regression above — resolve before it lands in the community build.

**Nit:** the "re-intersect defensively" comment (`node_ops.rs:493`) is misleading — since the SQL isn't id-scoped, that in-memory `member_ids.contains` is the *only* thing producing the correct set, i.e. load-bearing, not defensive.

**CI:** no checks on the branch; the local suite is exactly what masks the #1430 scope bug.

mstomar125:
**Review addressed** — you were exactly right on #1430. `with_ids` was a no-op (no `ids` on `NodeQuery`, dropped in the conversion, no `id IN` in `query_nodes`), so the collection query was an unbounded `SELECT * FROM node`.

Fixed end-to-end:
- added `ids: Option<Vec<String>>` to `NodeQuery`;
- carry `NodeFilter.ids` through the NodeFilter→NodeQuery conversion;
- `query_nodes` now builds a **chunked** `id IN (?, …)` (chunk size 900, under SQLite's bound-param ceiling; order_by + limit/offset applied in memory, so chunk order is irrelevant).

The collection query is now bounded by the member set, not the whole table. Corrected the misleading "(bounded)"/"defensively" comments. New test `test_collection_query_returns_members_despite_many_newer_nonmembers` (3 members + 25 newer non-members → all 3 returned) **fails under a capped-global-limit approach and the old fixed-1000 over-fetch**, catching the scope/bound bug the prior test missed.

Rebased onto current main (your #1453/#1457/#1458 landed). Verified: core lib 964, event_emission 11 (both collection tests), clippy --tests -D warnings + fmt clean. #1433/#1434 unchanged. Ready for re-review.

mstomar125:
**APPROVE** (re-review @ 888fe689)

The #1430 blocker is fully resolved — you took Option 1 and wired real id-scoping end-to-end. Verified:

- **`NodeQuery` now carries `ids: Option<Vec<String>>`** (`packages/nodespace-types/src/node.rs`) and the `NodeFilter→NodeQuery` conversion propagates it (`query.rs:58 ids: filter.ids.clone()`).
- **`query_nodes` builds a chunked `id IN (…)`** (`db/sqlite_store.rs:1206-1235`) at `ID_CHUNK = 900` — under SQLite's bound-param ceiling exactly as flagged, so a large member set can't overflow. The query is now bounded by the member set, not a full-table scan.
- **Pagination is stable.** The ids-path returns member rows unordered, then `NodeService::query_nodes` applies `order_by` (CreatedDesc) via in-memory sort (`query.rs:38-42`), and `node_ops` applies `skip(offset)`/`truncate(limit)` over that sorted member set (`node_ops:493+`). Correct ordering before the slice → no page overlap/skip.
- **Comments corrected:** the "(bounded)" claim and the "re-intersect defensively" note now accurately describe that the SQL *is* id-scoped and the in-memory `member_ids.contains` is a genuine double-check.

And the test gap is closed with the exact regression I asked for: `test_collection_query_returns_members_despite_many_newer_nonmembers` (`tests/event_emission_test.rs:203`) — members must surface despite many newer non-members, which "FAILS under a capped-global-limit approach … and under the old fixed-1000 over-fetch." Plus `test_collection_query_paginates_members_not_global_set` covers offset/limit + non-member exclusion.

#1433 (transactional `bulk_delete`) and #1434 (`bulk_update` merge) were already correct in the prior pass. Clean — good to merge.

Issue #212, opened 2025-10-11
Phase 3: Advanced & Regression Test Coverage (Sections 9-12, 26 tests)


malibio:
# Phase 3: Advanced & Regression Test Coverage (Sections 9-12, 26 tests)

**Epic**: #208  
**Relates to**: #176, #184, #185, #187, #190  
**PR (Session 1 - Infrastructure)**: #216 ✅ MERGED

---

## Overview

Implement comprehensive test coverage for Phase 3 dev server testing needs. This issue covers **advanced node operations** and **regression prevention tests** to ensure the HTTP dev server matches Tauri IPC behavior exactly.

### ✅ Session 1: HTTP Dev Server Infrastructure (COMPLETED - PR #216 merged)

**Implemented:**
- 10 HTTP endpoints in `embedding_endpoints.rs` (8 embedding ops + 2 mention ops)
- Extended `BackendAdapter` interface with 10 Phase 3 methods
- Implemented methods in both TauriAdapter (IPC) and HttpAdapter (HTTP)
- Added Phase 3 types: SearchTopicsParams, BatchEmbeddingResult, CreateContainerNodeInput
- Added timeout handling with `fetchWithTimeout` utility method
- Added comprehensive JSDoc documentation with `@throws` and `@returns`
- All code passes quality checks (oxlint, eslint, prettier, svelte-check, cargo clippy)

**Note**: Embedding endpoints contain **placeholder implementations** until TopicEmbeddingService is integrated.

---

### 🎯 Session 2: Test Implementation (THIS SESSION)

**Objective**: Implement 26 tests across 4 test sections using the infrastructure from Session 1.

### Section 9: Content Processing & Advanced Operations (6 tests)

**File**: `packages/desktop-app/src/tests/integration/phase3-content-processing.test.ts`

```typescript
describe('Phase 3: Content Processing & Advanced Operations', () => {
  describe('Content processing with embeddings', () => {
    it('should handle container node creation with nested structure')
    it('should process mentions during node creation')
    it('should update embeddings when content changes')
  });

  describe('Batch operations', () => {
    it('should batch generate embeddings for multiple topics')
    it('should handle partial failures in batch operations')
    it('should report accurate success/failure counts')
  });
});
```

**Test Focus:**
- Container node creation with complex hierarchy
- Node mention relationships during creation
- Embedding lifecycle (create, update, stale detection)
- Batch processing with error handling
- Success/failure reporting accuracy

**Key Test Data:**
[...]

PR #1463, opened 2026-06-25
Make bulk_create_mentions skip dangling [[links]] instead of dropping all mentions (closes #1462)

### LINKED ISSUE #1462: [bug] Import drops ALL mentions if any single [[link]] target is missing (bulk_create_mentions is all-or-nothing)
## Symptom
A markdown import logs `import_service: Failed to bulk create mentions: Failed to insert mention` and creates **zero** cross-reference (`mentions`) edges — even though the nodes + `has_child` tree import fine.

Surfaced 2026-06-24 while validating the Pro import→sync fix (#212) on a **25%** docs subset (`tenant_demo`): 31 files / 2307 nodes imported + fully synced, but the whole mentions batch failed.

## Root cause
`SqliteStore::bulk_create_mentions` (`packages/core/src/db/sqlite_store.rs:3652`) inserts every mention inside **one transaction**, and each `INSERT INTO relationship (… 'mentions' …)` uses `?`/`.context("Failed to insert mention")?`. A single failing insert propagates `Err` → the transaction is **never committed** → **all** mentions roll back.

The dominant failure is a **dangling `[[link]]`**: the target node doesn't exist (the linked doc wasn't in the imported set — or a typo / external ref), so the `relationship` FK insert fails. One bad mention discards every good one.

`import_service.rs:587` just logs the error and moves on, so the import "succeeds" with no references.

## Impact
- Partial imports (subset of a corpus) → essentially **no** mentions (lots of dangling links). 
- Even a **full** import loses **all** cross-references if it contains a single unresolvable `[[link]]` (common: typos, links to external/After-deleted docs).
- Non-fatal for nodes/sync; degrades the knowledge-graph cross-linking.

## Fix
Make mention creation resilient to bad endpoints, e.g.:
- Pre-filter `valid_mentions` to pairs where **both** `source` and `target` exist in `node` (one `SELECT … WHERE id IN (…)` set), then bulk-insert only those; or
- Insert per-mention best-effort (skip+count failures) instead of failing the whole transaction; or
- `INSERT … SELECT` guarded by existence of both endpoints.
Return/log `created` + `skipped` counts so dangling links are visible but never discard valid mentions.

Pre-existing (not introduced by the #207/#212/#209 Pro work); found via that validation.

mstomar125:
## Problem (#1462)

A markdown import logs `import_service: Failed to bulk create mentions: Failed to insert mention` and creates **zero** cross-reference (`mentions`) edges — even though nodes + the `has_child` tree import fine.

**Root cause:** `relationship.in_node`/`out_node` are `NOT NULL REFERENCES node(id)` (`schema.sql:22`) with `PRAGMA foreign_keys = ON` (`schema.sql:2`). So a mention to a **non-existent** node — a dangling `[[link]]` to a doc that wasn't imported, a typo, or an external ref — is an FK violation on INSERT. `bulk_create_mentions` (`sqlite_store.rs:3652`) inserted every mention in **one transaction** with `?`-propagation, so a single dangling link rolled back the **whole batch** → all mentions lost. `import_service.rs` only logs the error, so it fails silently.

Even a **full** import loses all cross-refs if it has a single unresolvable link.

## Fix

Pre-filter to mentions whose **both** endpoints exist — one chunked `SELECT id FROM node WHERE id IN (...)` (`ID_CHUNK = 900`, under SQLite's ~999 bound-param ceiling) into a `HashSet`, then keep only pairs where both source and target are present. Dangling pairs are **skipped and logged** (`warn!` with the skipped count), never fatal; the valid mentions still land. Existing self-mention (`s != t`) and duplicate-skip checks are preserved.

## Tests

- New `test_bulk_create_mentions_skips_dangling_links`: a batch mixing one valid mention with dangling ones creates **exactly the valid one** (was 0 before the fix); an all-dangling batch is a **no-op, not an error**.
- `cargo test -p nodespace-core --lib mention` → **14 passed**; `cargo clippy` clean; fmt clean.

## Context

Found via the Pro import→sync validation (nodespace-sync #212) on a **partial** (25%) docs import — most `[[links]]` pointed at the un-imported 75%, so the whole mentions batch failed. Pre-existing; not introduced by that work.

### DIFF
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -3655,7 +3655,63 @@ impl SqliteStore {
         }
 
         let start = std::time::Instant::now();
-        let valid_mentions: Vec<_> = mentions.iter().filter(|(s, t)| s != t).collect();
+        let candidate: Vec<_> = mentions.iter().filter(|(s, t)| s != t).collect();
+        let candidate_len = candidate.len();
+        if candidate_len == 0 {
+            return Ok(0);
+        }
+
+        // #1462: a mention edge is FK-constrained — `relationship.in_node` /
+        // `out_node` are `NOT NULL REFERENCES node(id)` with `PRAGMA foreign_keys
+        // = ON` — so a mention to a NON-EXISTENT node (a dangling `[[link]]` to a
+        // doc that wasn't imported, a typo, or an external ref) is an FK violation
+        // on INSERT. Previously all mentions were inserted in ONE transaction with
+        // `?`-propagation, so a single dangling link rolled back the WHOLE batch
+        // and the import created ZERO cross-references. Pre-filter to pairs whose
+        // BOTH endpoints exist; dangling links are skipped (and logged), never
+        // fatal.
+        let mut endpoints: std::collections::HashSet<String> = std::collections::HashSet::new();
+        for m in &candidate {
+            endpoints.insert(m.0.clone());
+            endpoints.insert(m.1.clone());
+        }
+        let endpoint_ids: Vec<String> = endpoints.into_iter().collect();
+        // Chunk the `IN (...)` under SQLite's ~999 bound-parameter ceiling.
+        const ID_CHUNK: usize = 900;
+        let mut existing: std::collections::HashSet<String> = std::collections::HashSet::new();
+        for chunk in endpoint_ids.chunks(ID_CHUNK) {
+            let placeholders: Vec<String> = (1..=chunk.len()).map(|i| format!("?{}", i)).collect();
+            let sql = format!(
+                "SELECT id FROM node WHERE id IN ({})",
+                placeholders.join(", ")
+            );
+            let params: Vec<libsql::Value> = chunk
+                .iter()
+                .map(|id| libsql::Value::Text(id.clone()))
+                .collect();
+            let mut rows = self
+                .db
+                .query(&sql, params)
+                .await
+                .context("Failed to check mention endpoints")?;
+            while let Some(row) = rows.next().await? {
+                let id: String = row.get(0)?;
+                existing.insert(id);
+            }
+        }
+
+        let valid_mentions: Vec<_> = candidate
+            .into_iter()
+            .filter(|m| existing.contains(m.0.as_str()) && existing.contains(m.1.as_str()))
+            .collect();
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE**

Grounded against the full function in `packages/core/src/db/sqlite_store.rs` at head `68250832` (not just the diff). This is a correct, well-scoped fix for the #1462 all-or-nothing rollback.

- 🟢 **Root cause addressed.** The insert loop (`:3727-3743`) writes `relationship.in_node` / `out_node`, both `NOT NULL REFERENCES node(id)` under FK enforcement, so a dangling `[[link]]` to a never-imported/typo'd node was an FK violation that — in the single `?`-propagating transaction — rolled back the entire batch (zero cross-refs). The new pre-filter (`:3673-3706`) collects the unique endpoint set, checks existence with a chunked `id IN (...)`, and keeps only pairs whose **both** endpoints exist. The remaining FK columns are exactly those two, so no other FK path can fire post-filter. ✅
- 🟢 **No dedup regression.** The per-pair `SELECT ... LIMIT 1` guard (`:3728-3735`) runs inside `tx`, so it sees rows inserted earlier in the same uncommitted batch — intra-batch duplicates and pre-existing mentions are both skipped. UNIQUE-violation surface unchanged.
- 🟢 **Convention-consistent** with core#1459: `ID_CHUNK = 900`, positional `?N` placeholders, `HashSet` endpoint dedup before the query. Accurate `skipped = candidate_len - valid_mentions.len()` warn log (self-mentions are filtered separately, so they don't pollute the dangling count).
- 🟢 **Test matches the bug** (`test_bulk_create_mentions_skips_dangling_links`): interleaved valid+dangling → only the valid edge lands (created==1, no ghosts); all-dangling batch → 0 created, no error.

Non-blocking nits:
- 🟢 **TOCTOU window:** the existence check queries `self.db` *before* `tx` begins (`:3692` vs `:3720`), so a concurrent delete of an endpoint between check and insert could still FK-fail and roll back the batch. In the import path endpoints are being created (not deleted), so it isn't a realistic trigger — but doing the existence read inside `tx` would close it for free if you touch this again.
- ⏳ No CI on the branch (`statusCheckRollup` empty) — verified by reading the code + test, not a green check.

Unblocks the import-link revalidation that was gating the Michael handoff. LGTM.

Issue #226, opened 2025-10-11
Architecture: Migrate to backend-originated event emission


malibio:
## Context

Current implementation (PR #222) emits events from frontend adapters (`HttpAdapter`, `TauriAdapter`) after successful backend operations. This was a pragmatic choice to get event infrastructure working quickly, but has architectural limitations.

**Current Architecture:**
```
Backend (Rust) → Success Response → Frontend Adapter → Event Emission → EventBus
```

**Proposed Architecture:**
```
Backend (Rust) → Event Emission → Response with Events → Frontend Adapter → EventBus
```

## Problem

**Current Limitations:**

1. **Code Duplication:** Event emission logic duplicated in `HttpAdapter` AND `TauriAdapter`
2. **Inconsistency Risk:** Frontend and backend states can diverge
3. **Performance:** Extra GET requests before DELETE to capture metadata
4. **Limited Metadata:** Frontend doesn't know what backend actually changed
5. **No Event Guarantees:** If frontend crashes, events are lost

**Example Inconsistency Scenario:**
```typescript
// Frontend thinks node was created (emitted event)
const nodeId = await createNode(data);
eventBus.emit({ type: 'node:created', nodeId }); 

// But backend actually rejected it (validation failed after response sent)
// Frontend state is now inconsistent with backend reality
```

### Backend Event Emission

**Rust backend emits events as part of operation response:**

```rust
// node_endpoints.rs
pub async fn create_node(node: CreateNodeInput) -> Result<CreateNodeResponse> {
    let node_id = db.insert_node(&node)?;
    
    // Emit event
    let event = NodeEvent::Created {
        node_id: node_id.clone(),
        parent_id: node.parent_id,
        node_type: node.node_type,
    };
    
    Ok(CreateNodeResponse {
        node_id,
        events: vec![event],  // Include events in response
    })
}
```

**Frontend adapters forward events to EventBus:**

```typescript
async createNode(node: NodeInput): Promise<string> {
  const response = await fetch('/api/nodes', { 
    method: 'POST', 
    body: JSON.stringify(node) 
  });
  
  const { nodeId, events } = await response.json();
  
  // Forward backend events to EventBus
  for (const event of events) {
    eventBus.emit(this.parseBackendEvent(event));
  }
  
  return nodeId;
}
```

### Benefits

**✅ Single Source of Truth:** Backend controls event emission
**✅ Eliminates Duplication:** No frontend event logic needed
[...]

PR #1466, opened 2026-07-01
S6 (UI half): 'signed in as <email>' + manual sign-out in the Pro pill (#226)

mstomar125:
UI half of #226 (S6 of the auto-sign-in epic, nodespace-sync#199). Pairs with nodespace-sync#236 (the daemon half — `user_email` on `SyncStatusEvent` + a `SignOut` RPC). This PR consumes those two surfaces.

## What this adds
- **Proto mirror** — `user_email` on `SyncStatusEvent` + a `SignOut` RPC (vendored copy under `src-tauri/proto/`, kept in lockstep with the daemon's).
- **`pro_subscribe_sync_status`** now forwards `user_email` on the `sync:status` event; the synthetic DISCONNECTED event clears it, and `pro:tier-detected`'s `initial_status` carries it too (for a resume-race CONNECTED probe).
- **`pro_signout` command** — calls the daemon's `SignOut` (which drops the token + wipes the persisted refresh token from the keychain). No-op in community mode, matching the other Pro commands.
- **`proSync` store** gains `userEmail` (from the stream) and an async `signOut()` that invokes the command and optimistically resets to signed-out.
- **Sync pill** — when signed in, clicking it opens a small account menu: "**Signed in as \<email\>**" + a **Sign out** button. Signed-out states keep the existing click-to-sign-in. Escape / outside-click close the menu.

## Why the daemon has to surface the email
On the **silent-resume** path (S3) the app boots straight into the workspace and the daemon refreshes the session from the keychain — the frontend never sees an OAuth response, so it has no other way to know who is signed in. The daemon decodes it from the session JWT and streams it.

## Scope / safety
- **Community/free build unaffected** — the pill only renders under `proSync.isPro`, `pro_signout` no-ops without a `ProClient`, and the free-user guardrail suite stays green.
- The "boot straight to workspace, no sign-in flash" half of S6 was already satisfied (the workspace renders immediately; the relogin modal only overlays on `auth-required`), so this PR is the remaining affordance + sign-out.

## Tests
- `src/tests/stores/pro-sync.test.ts` — `user_email` surfaced from `sync:status` and cleared on sign-out; missing-field tolerance; `signOut()` invokes `pro_signout` and resets state.
- `bun run test` (pro-sync + free-user-guardrail): 10 passed. `eslint --fix` 0 problems; `svelte-check` 0 errors. `cargo check`/`clippy`/`fmt` clean on `nodespace-app`.

### DIFF
--- a/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
+++ b/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
@@ -33,6 +33,12 @@ service CloudSyncService {
   // No-op if the orchestrator is already in SYNCING.
   rpc TriggerSync (TriggerSyncRequest) returns (TriggerSyncResponse);
 
+  // Explicit sign-out (#199 S6). Drops the in-memory token, wipes the
+  // persisted refresh token from the OS keychain (so a daemon restart
+  // does NOT auto-resume), and tears down the live sync connection —
+  // the orchestrator loop then reports AUTH_REQUIRED. Idempotent.
+  rpc SignOut (SignOutRequest) returns (SignOutResponse);
+
   // --- Team membership (M5, #147). Thin adapters over the cloud
   // membership RPCs; the daemon forwards the signed-in user's JWT, so
   // the same admin / last-admin / open-vs-restricted gates apply
@@ -100,10 +106,18 @@ message SyncStatusEvent {
   State state = 1;
   // Free-form context, e.g. error message or current LIVE SELECT cursor.
   string detail = 2;
+  // Email of the signed-in user, decoded from the session JWT. Empty when
+  // signed out / not yet authenticated. Drives the "signed in as <email>"
+  // affordance (#199 S6), including on the silent-resume path where the
+  // frontend never sees an OAuth response and so otherwise has no identity.
+  string user_email = 3;
 }
 
 message TriggerSyncRequest {}
 message TriggerSyncResponse {}
+
+message SignOutRequest {}
+message SignOutResponse {}
 
 // --- Team membership (M5, #147) ----------------------------------
 // permission is "admin" | "modify" | "readOnly". collection_id / person_id are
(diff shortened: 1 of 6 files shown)


mstomar125:
**APPROVE**

Grounded at head `8280a513` across all six files. Clean, accessible, and free-user-safe UI half of S6.

- 🟢 **Proto mirror is wire-identical to the daemon (sync#236).** `user_email = 3` on `SyncStatusEvent`, `SignOut (SignOutRequest) returns (SignOutResponse)`, empty request/response messages — same field number and shapes, so no gRPC wire drift between the vendored copy and the daemon's.
- 🟢 **Free-user safe — verified, not assumed.** The entire pill + account menu is wrapped in `{#if proSync.isPro}` (`pro-sync-pill.svelte:171`), so community renders nothing; `pro_signout` returns `Ok(())` early when there's no `ProClient` (`pro_sync.rs:78`), matching the other Pro commands; `userEmail` only ever populates from the Pro `sync:status`/`pro:tier-detected` streams. Community build behaviorally unchanged.
- 🟢 **Identity plumbing is consistent.** `user_email` forwarded on `sync:status`, cleared to `""` on the synthetic DISCONNECTED, and carried on `pro:tier-detected`'s `initial_status` for the resume-race probe; the store reads it with `?? ''` so older/missing-field payloads degrade to empty.
- 🟢 **Sign-out UX is sound.** `onClick` opens the menu only when `signedIn` (else the existing sign-in flow); `onSignOut` guards double-clicks, Escape + transparent backdrop close it, `aria-haspopup`/`aria-expanded`/`role="menu"` are correct. `signOut()` invokes `pro_signout`, catches invoke failure (warn, no throw), then optimistically resets to `auth-required` — the daemon's real `AUTH_REQUIRED` reconciles via the stream. Good tests cover surface + clear + missing-field + `signOut()`.

🔶 **Cross-layer dependency — coordinate the merge/deploy with sync#236.** `pro_signout` → `client.sign_out(SignOutRequest{})` calls the daemon RPC that's defined in **nodespace-sync#236**, which is still open and currently has my **REQUEST CHANGES** (its `do_sign_out` doesn't clear `bound_person_id`, so sign-out isn't "complete" per S1 invariant 3). Implications:
- This PR is correct in isolation and degrades safely (a daemon without the RPC → `Err` → the store still optimistically shows signed-out), but sign-out only *works end-to-end* once #236 is merged **and** the bundled Pro daemon is rebuilt with it. Land #236 (with the `bound_person_id` fix) first / together, not this ahead of it.
- Once #236's fix lands, the sign-out this UI triggers will actually return the device to the un-provisioned state — worth a quick manual verify then.

⏳ No branch CI (`statusCheckRollup` empty) — relying on the reported `bun run test` (10) / eslint / svelte-check / `cargo clippy` green. LGTM on the UI half; just gate the merge on #236.

PR #1467, opened 2026-07-01
Add Ornith 1.0 to model catalog (catalog entry, think-block routing, update_node prompt)

### LINKED ISSUE #1465: Add Ornith 1.0 to model catalog (catalog entry, think-block routing, update_node prompt)
## Overview

Ornith 1.0 9B (DeepReinforce, June 2026, MIT) passed evaluation in #1464. This issue implements the three remaining production tasks to make it a first-class catalog model.

### Architecture
Ornith uses the `qwen35` arch in llama.cpp — a hybrid SSM/attention model (Gated Delta Net). 24 of 32 layers are recurrent SSM; 8 are full multi-head attention (every 4th block). This gives it a unique property: **generation speed is flat across context lengths** — 12.0 t/s at 1k tokens, 11.9 t/s at 8k tokens. Pure-attention models degrade significantly past 4k on 16 GB hardware.

### Tool-call format (Format D)
Ornith's embedded Jinja chat template defines a distinct XML-like format, different from all three formats the parser previously handled:

```xml
<tool_call>
<function=search_nodes>
<parameter=query>
Rust programming
</parameter>
</function>
</tool_call>
```

**This parser is already implemented.** `parse_tool_calls_xml()` and `parse_xml_tool_call_block()` are live in `packages/nlp-engine/src/chat/parser.rs`. `emit_oai_delta()` in `mod.rs` intercepts content containing `<tool_call>` and routes it through the XML parser before the text splitter. Experiment validated 8/10 tool calls parsed correctly.

### `update_node` miss (prompt engineering)
`update_node` failed both trials — the model asked for clarification instead of calling the tool. The system prompt for `update_node` provides placeholder `node_id` and `content` values; Ornith treats these as literally "fill in the blank" rather than examples. Fix: provide concrete dummy values in the example call.

### Reasoning (`<think>` blocks)
Ornith produces `<think>...</think>` blocks in the token stream when `enable_thinking: true`. The thinking content contains correct multi-step planning (e.g., `search_nodes → get_node × N → create_node`, with correct sequencing rationale). Currently these tokens leak into the visible response. The production `ChatEngine` needs a `<think>`/`</think>` routing path analogous to the existing Gemma 4 `ChannelSplitter` to emit `ChatChunk::Reasoning` events.

### Speed / memory
| Metric | Value |
|---|---|
| GGUF size | 5.23 GiB (Q4_K_M) |
| Model load | 8.3s (Metal kernel init on first load) |
| TTFT | 274ms |
| Tokens/sec | 19.6 |
| t/s at 8k ctx | 11.9 (stable — SSM architecture) |
| Min RAM | 16 GB |

### Integration path
Ornith uses `apply_chat_template_oaicompat` with `use_jinja: true` — same path as Gemma 4 E4B. No new model-loading code is needed; it routes through the existing OAI-compat pipeline.

## Acceptance Criteria

- [ ] **Catalog entry:** Add `ornith-9b-q4_k_m` to the model catalog with `ModelFamily::Qwen35` (or a new `Ornith` variant if behavior diverges). Set `min_ram_gb: 16`, correct GGUF filename (`ornith-1.0-9b-Q4_K_M.gguf`), HuggingFace source URL.
- [ ] **`ModelFamily` wiring:** `agent_types.rs` and `model_manager.rs` recognize Ornith (via `Qwen35` mapping or dedicated variant) and route to `apply_chat_template_oaicompat` with `use_jinja: true`.
- [ ] **`update_node` prompt fix:** Update the `update_node` tool description / example in the system prompt so the model calls the tool directly rather than asking for confirmation. Validate 2/2 trials pass.
- [ ] **`<think>` routing path:** Add `<think>`/`</think>` detection to `ChatEngine` (analogous to Gemma 4's `ChannelSplitter`) so thinking tokens emit as `ChatChunk::Reasoning` events rather than leaking into `ChatChunk::Content`. Gate behind a model-family flag or `enable_thinking` config.
- [ ] **Unit tests:** Add tests for `parse_tool_calls_xml` covering: single call, multiple calls, missing `</tool_call>` close, missing `</parameter>`, empty parameter body.
- [ ] **Smoke test:** Load Ornith 9B, send one tool-calling prompt, confirm `ChatChunk::ToolCallStart` + `ChatChunk::ToolCallArgs` are emitted (no plain-text leakage of `<tool_call>` XML).

## Files to Touch

- `packages/nlp-engine/src/chat/parser.rs` — add unit tests for `parse_tool_calls_xml` (parser already implemented)
- `packages/nlp-engine/src/chat/mod.rs` — add `<think>`/`</think>` routing path in `emit_oai_delta()` or a new `OrnithChannelSplitter`
- `packages/agent/src/agent_types.rs` — add `Ornith` variant or confirm `Qwen35` covers it
- `packages/agent/src/model_manager.rs` — catalog entry + routing
- System prompt / tool descriptions — `update_node` example fix

## Out of Scope

- Ornith 35B MoE (48 GB tier) — separate issue
- Replacing Gemma 4 E4B or Ministral 8B — keep all three; each serves a different profile
- Automatic default model changes — Ornith should be user-selectable first

malibio:
Closes #1465

## Summary
- Adds `ornith-1-9b-q4km` catalog entry (`ModelFamily::Qwen35`, same OAI-compat Jinja path as Gemma 4/Qwen3.5)
- Adds Format D XML `<tool_call>` parser (`parse_tool_calls_xml`/`parse_xml_tool_call_block`) wired into `emit_oai_delta`, with unit tests for single/multiple calls, unclosed tags, and empty parameter bodies
- Adds `ThinkSplitter` (mirrors the existing Gemma 4 `ChannelSplitter`) to route `<think>...</think>` content as `ChatChunk::Reasoning` instead of leaking into the answer stream
- Rewords the `update_node` tool description/schema with concrete example values so Ornith calls it directly instead of asking for clarification (the #1464 evaluation's 2/10 tool-call misses)

## Test plan
- [x] `cargo test -p nodespace-nlp-engine --lib --features chat-service` (chat module: 90/90 passing, includes new XML parser + ThinkSplitter + combined-finalize-path tests)
- [x] `cargo test -p nodespace-agent --lib` (248/248 passing, includes updated catalog-count test)
- [x] `cargo test -p nodespace-core --lib` (965/965 passing, unaffected)
- [x] `bun run test:all` (frontend 3893/3893, skill 17/17, rust core 965/965 — no new failures vs. baseline)
- [x] `bun run quality:fix` clean (eslint, tsc, cargo fmt, cargo clippy -D warnings)
- [ ] **Smoke test (manually deferred):** the issue's acceptance criterion "load Ornith 9B, send one tool-calling prompt, confirm ToolCallStart/ToolCallArgs" requires downloading the 5.23GB GGUF and running local inference, which isn't practical in this automated review pass. The XML parsing and `<think>` routing logic were validated against real Ornith outputs captured during the #1464 evaluation (`../nodespace-nlp-experiment`, 8/10 tool calls, `<think>` blocks confirmed present) and are covered by unit tests reproducing those exact formats, but an end-to-end live-model smoke test has not been run in this PR and should be done manually before Ornith is broadly recommended.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -119,7 +119,7 @@
     },
     "packages/skill": {
       "name": "@nodespaceai/skill",
-      "version": "0.1.0",
+      "version": "0.2.0",
       "bin": {
         "nodespace-skill": "dist/install.js",
       },
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -868,8 +868,12 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                     }
                     accumulated_reasoning.push_str(tail_reasoning.trim());
                 }
-                if tail_text.is_empty() {
-                    // Synthesize from tool results when model returned nothing
+                let normalized_tail = normalize_response(&tail_text);
+                if normalized_tail.is_empty() {
+                    // Model returned nothing, or its entire response was internal
+                    // plumbing (e.g. a leaked <tool_call> block) stripped down to
+                    // nothing — synthesize a summary from the tool results instead
+                    // of persisting a blank assistant bubble.
                     let mut counts: Vec<(&'static str, usize)> = Vec::new();
                     for t in &all_tool_executions {
                         let label = humanize_tool_name(&t.name);
@@ -891,7 +895,7 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                         .collect::<Vec<_>>()
                         .join("\n")
                 } else {
-                    normalize_response(&tail_text)
+                    normalized_tail
                 }
             } else {
                 // Inference failed — synthesize from executions
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -169,7 +169,11 @@ const ORNITH_1_9B: CatalogEntry = CatalogEntry {
     size_bytes: 5_614_720_712, // 5.23 GiB
     quantization: "Q4_K_M",
     url: "https://huggingface.co/DeepReinforce/Ornith-1.0-9B-GGUF/resolve/main/ornith-1.0-9b-Q4_K_M.gguf",
-    sha256: "", // Populate before promoting past user-selectable trial status
+    // Policy exception: DeepReinforce is not an official ggml-org/mistralai repo,
+    // so the empty-string skip-verification policy does not strictly apply.
+    // Acceptable here because this is a trial-only, user-selectable entry (#1465);
+    // populate the SHA-256 from the model card before promoting to a default.
+    sha256: "",
     context_window: 32_768,
     default_temperature: 0.6,
     type_k: None, // F16 — recurrent SSM layers keep KV cache small at this size
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -152,6 +152,31 @@ const QWEN35_9B: CatalogEntry = CatalogEntry {
     min_memory_gb: 16,
 };
 
+/// Ornith 1.0 9B -- DeepReinforce (June 2026), MIT licensed. Hybrid SSM/attention
+/// (`qwen35` arch in llama.cpp): 24 of 32 layers are Gated Delta Net (recurrent),
+/// 8 are full attention. Generation speed is flat across context lengths (12.0
+/// t/s at 1k, 11.9 t/s at 8k) — unlike pure-attention models, which degrade
+/// significantly past 4k on 16GB hardware. Tool calling validated 8/10 in
+/// evaluation (#1464) via the XML `<tool_call>` format (Format D); the 2
+/// misses were a prompt-engineering gap in `update_node`, not a parsing issue.
+/// Uses `ModelFamily::Qwen35` (same OAI-compat Jinja path); no dedicated
+/// variant needed unless behavior diverges from Qwen3.5 in practice.
+const ORNITH_1_9B: CatalogEntry = CatalogEntry {
+    id: "ornith-1-9b-q4km",
+    family: ModelFamily::Qwen35,
+    name: "Ornith 1.0 9B Instruct Q4_K_M",
+    filename: "ornith-1.0-9b-Q4_K_M.gguf",
+    size_bytes: 5_614_720_712, // 5.23 GiB
+    quantization: "Q4_K_M",
+    url: "https://huggingface.co/DeepReinforce/Ornith-1.0-9B-GGUF/resolve/main/ornith-1.0-9b-Q4_K_M.gguf",
+    sha256: "", // Populate before promoting past user-selectable trial status
+    context_window: 32_768,
+    default_temperature: 0.6,
+    type_k: None, // F16 — recurrent SSM layers keep KV cache small at this size
+    type_v: None,
+    min_memory_gb: 16,
+};
+
 /// Qwen3.6 35B-A3B -- MoE: 35B total parameters, 3B active. Fast inference
 /// (similar speed to a 3B dense model) with large-model capacity.
 /// ~22 GB on disk; fits on 48 GB Apple Silicon. Trial only.
@@ -304,6 +329,7 @@ const CATALOG: &[&CatalogEntry] = &[
     &QWEN3_8B,
     &QWEN35_9B,
     &QWEN36_35B_A3B,
+    &ORNITH_1_9B,
     &GEMMA_4_E4B,
     &GEMMA_4_12B,
     &GEMMA_4_12B_UNSLOTH,
@@ -1103,12 +1129,13 @@ mod tests {
     async fn list_returns_all_catalog_models() {
         let (mgr, _tmp) = test_manager();
         let models = mgr.list().await.unwrap();
-        assert_eq!(models.len(), 12);
+        assert_eq!(models.len(), 13);
         assert!(models.iter().any(|m| m.id == "ministral-3b-q4km"));
         assert!(models.iter().any(|m| m.id == "ministral-8b-q4km"));
         assert!(models.iter().any(|m| m.id == "gemma-4-e4b-q4km"));
         assert!(models.iter().any(|m| m.id == "gemma-4-12b-q4km"));
         assert!(models.iter().any(|m| m.id == "gemma-4-31b-q4km"));
+        assert!(models.iter().any(|m| m.id == "ornith-1-9b-q4km"));
     }
 
     #[tokio::test]
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -154,22 +154,37 @@ const QWEN35_9B: CatalogEntry = CatalogEntry {
 
 /// Ornith 1.0 9B -- DeepReinforce (June 2026), MIT licensed. Hybrid SSM/attention
 /// (`qwen35` arch in llama.cpp): 24 of 32 layers are Gated Delta Net (recurrent),
-/// 8 are full attention. Generation speed is flat across context lengths (12.0
-/// t/s at 1k, 11.9 t/s at 8k) — unlike pure-attention models, which degrade
-/// significantly past 4k on 16GB hardware. Tool calling validated 8/10 in
-/// evaluation (#1464) via the XML `<tool_call>` format (Format D); the 2
-/// misses were a prompt-engineering gap in `update_node`, not a parsing issue.
-/// Uses `ModelFamily::Qwen35` (same OAI-compat Jinja path); no dedicated
-/// variant needed unless behavior diverges from Qwen3.5 in practice.
+/// 8 are full attention. Uses `ModelFamily::Qwen35` (same OAI-compat Jinja path);
+/// no dedicated variant needed unless behavior diverges from Qwen3.5 in practice.
+///
+/// KNOWN LIMITATION -- NOT auto-recommended, user-selectable only: recurrent
+/// (SSM) layers cannot do partial KV-cache reuse (`llama_memory_seq_rm` returns
+/// `false` on a partial range for this architecture -- confirmed both in our own
+/// logs, "KV cache partial trim returned false, falling back to full decode",
+/// and upstream: ggml-org/llama.cpp issues #19858, #20003, #20153, #20225,
+/// #21912 all report the same full-reprocess behavior for Qwen 3.5-based hybrid
+/// models). This means every ReAct loop iteration re-decodes the ENTIRE growing
+/// prompt from scratch, not just the new turn -- unlike Ministral/Gemma, which
+/// reuse the cached prefix. Single-shot decode speed is genuinely flat across
+/// context depths (validated in #1464, 12.0 t/s at 1k, 11.9 t/s at 8k), but that
+/// finding does NOT extend to multi-iteration tool-calling turns as originally
+/// assumed -- a turn requiring N ReAct iterations pays a full-prompt decode cost
+/// N times, compounding with conversation length. A 4-iteration turn (e.g. a
+/// failed tool call requiring 3 retries) took ~108s in practice. This is an
+/// upstream llama.cpp limitation for recurrent/hybrid architectures, still open
+/// as of the issues linked above -- not something fixable in this codebase.
 const ORNITH_1_9B: CatalogEntry = CatalogEntry {
     id: "ornith-1-9b-q4km",
     family: ModelFamily::Qwen35,
     name: "Ornith 1.0 9B Instruct Q4_K_M",
     filename: "ornith-1.0-9b-Q4_K_M.gguf",
-    size_bytes: 5_614_720_712, // 5.23 GiB
+    size_bytes: 5_629_108_704, // verified via HEAD request, 5.24 GiB
     quantization: "Q4_K_M",
-    url: "https://huggingface.co/DeepReinforce/Ornith-1.0-9B-GGUF/resolve/main/ornith-1.0-9b-Q4_K_M.gguf",
-    // Policy exception: DeepReinforce is not an official ggml-org/mistralai repo,
+    // NOTE: org is "deepreinforce-ai" (lowercase, hyphenated) -- NOT "DeepReinforce".
+    // The original URL used the wrong casing/org name and 404/401'd; verified this
+    // one resolves via `curl -sIL` before committing.
+    url: "https://huggingface.co/deepreinforce-ai/Ornith-1.0-9B-GGUF/resolve/main/ornith-1.0-9b-Q4_K_M.gguf",
+    // Policy exception: deepreinforce-ai is not an official ggml-org/mistralai repo,
     // so the empty-string skip-verification policy does not strictly apply.
     // Acceptable here because this is a trial-only, user-selectable entry (#1465);
     // populate the SHA-256 from the model card before promoting to a default.
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -171,8 +171,10 @@ const QWEN35_9B: CatalogEntry = CatalogEntry {
 /// assumed -- a turn requiring N ReAct iterations pays a full-prompt decode cost
 /// N times, compounding with conversation length. A 4-iteration turn (e.g. a
 /// failed tool call requiring 3 retries) took ~108s in practice. This is an
-/// upstream llama.cpp limitation for recurrent/hybrid architectures, still open
-/// as of the issues linked above -- not something fixable in this codebase.
+/// upstream llama.cpp limitation for recurrent/hybrid architectures, open as
+/// of the issues linked above at the time this was written (2026-07-01) --
+/// not something fixable in this codebase. Re-check those issues before
+/// assuming this limitation still applies to a newer vendored llama.cpp.
 const ORNITH_1_9B: CatalogEntry = CatalogEntry {
     id: "ornith-1-9b-q4km",
     family: ModelFamily::Qwen35,
--- a/packages/agent/src/local_agent/response_processing.rs
+++ b/packages/agent/src/local_agent/response_processing.rs
@@ -317,6 +317,9 @@ fn strip_ornith_tool_call_leak(text: &str) -> String {
     }
     let result = ornith_tool_call_re().replace_all(text, "").into_owned();
     // Drop a dangling, never-closed tool-call block (truncated mid-generation).
+    // Assumes a dangling opener is always trailing: truncation happens at the
+    // end of generation (token cap or stream cutoff), so there is never
+    // legitimate prose after an unclosed <tool_call> to preserve.
     let result = match result.find("<tool_call>") {
         Some(idx) => result[..idx].to_string(),
         None => result,
--- a/packages/agent/src/local_agent/response_processing.rs
+++ b/packages/agent/src/local_agent/response_processing.rs
@@ -59,6 +59,12 @@ fn stray_channel_marker_re() -> &'static Regex {
     RE.get_or_init(|| Regex::new(r"<\|channel>|<channel\|>").unwrap())
 }
 
+/// Matches a complete Ornith `<tool_call>…</tool_call>` XML block.
+fn ornith_tool_call_re() -> &'static Regex {
+    static RE: OnceLock<Regex> = OnceLock::new();
+    RE.get_or_init(|| Regex::new(r"(?s)<tool_call>.*?</tool_call>").unwrap())
+}
+
 /// Normalize LLM response text for consistent formatting.
 ///
 /// Delegates to [`normalize_response_traced`] and discards the stripper list.
@@ -95,6 +101,12 @@ pub fn normalize_response_traced(text: &str) -> (String, Vec<&'static str>) {
         "strip_channel_blocks",
         strip_channel_blocks,
         text.to_owned(),
+        &mut fired,
+    );
+    let result = apply(
+        "strip_ornith_tool_call_leak",
+        strip_ornith_tool_call_leak,
+        result,
         &mut fired,
     );
     let result = apply(
@@ -286,6 +298,30 @@ fn strip_channel_blocks(text: &str) -> String {
     stray_channel_marker_re()
         .replace_all(&result, "")
         .into_owned()
+}
+
+/// Strip leaked Ornith `<tool_call>` XML and stray special tokens.
+///
+/// This normalizer only runs on text-only ("tail") inferences made with
+/// `tools: None` (e.g. the final answer synthesized after the duplicate-call
+/// guard breaks out of the ReAct loop). Ornith can still emit its `<tool_call>`
+/// envelope syntax in that mode, and separately emit `<|mask_end|>` — a raw
+/// special token unrelated to tool calls — either of which is internal
+/// plumbing and must never reach the user. Since no tools are offered on this
+/// path, any `<tool_call>` seen here is always a leak, never a real routed
+/// call (that interception happens earlier in `emit_oai_delta`, before text
+/// ever reaches this stripper).
+fn strip_ornith_tool_call_leak(text: &str) -> String {
+    if !text.contains("<tool_call>") && !text.contains("<|mask_end|>") {
+        return text.to_string();
+    }
+    let result = ornith_tool_call_re().replace_all(text, "").into_owned();
+    // Drop a dangling, never-closed tool-call block (truncated mid-generation).
+    let result = match result.find("<tool_call>") {
+        Some(idx) => result[..idx].to_string(),
+        None => result,
+    };
+    result.replace("<|mask_end|>", "")
 }
 
 /// Strip leaked Gemma tool-call/response prose (`call:foo{...}response:foo{...}`).
@@ -623,6 +659,32 @@ mod tests {
         let input = "All set.<channel|>";
         let result = normalize_response(input);
         assert_eq!(result, "All set.");
+    }
+
+    #[test]
+    fn strips_ornith_tool_call_leak_whole_response() {
+        // Reproduces a real Ornith tail-inference leak: the entire response was
+        // a <tool_call> block (no separate prose), so normalization must reduce
+        // it to empty, letting the caller synthesize a tool-summary fallback.
+        let input = "<tool_call>\n<function=search_nodes>\n</function>\n</tool_call>";
+        let result = normalize_response(input);
+        assert_eq!(result, "");
+    }
+
+    #[test]
+    fn strips_ornith_dangling_tool_call_and_mask_end_token() {
+        // Reproduces a real Ornith leak: a truncated <tool_call> block with no
+        // closing tag, followed by a stray <|mask_end|> special token.
+        let input = "<tool_call>\n<function=search_nodes>\n</parameter><|mask_end|>";
+        let result = normalize_response(input);
+        assert_eq!(result, "");
+    }
+
+    #[test]
+    fn strips_ornith_tool_call_leak_keeping_surrounding_text() {
+        let input = "Sure, let me check.<tool_call>\n<function=search_nodes>\n<parameter=query>\ntasks\n</parameter>\n</function>\n</tool_call>";
+        let result = normalize_response(input);
+        assert_eq!(result, "Sure, let me check.");
     }
 
     // normalize_response_traced: strippers_fired correctness
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -425,21 +425,21 @@ fn def_create_node() -> ToolDefinition {
 fn def_update_node() -> ToolDefinition {
     ToolDefinition {
         name: "update_node".into(),
-        description: "Update an existing node's content or properties. The node service recomputes the title automatically after any update.".into(),
+        description: "Update an existing node's content or properties immediately — call this directly with the node ID you already have (e.g. from search_nodes or get_node), don't ask the user to confirm or provide it first. The node service recomputes the title automatically after any update. Example call: {\"id\": \"a1b2c3d4-...\", \"content\": \"Buy milk and eggs\"}.".into(),
         parameters_schema: json!({
             "type": "object",
             "properties": {
                 "id": {
                     "type": "string",
-                    "description": "Node ID to update"
+                    "description": "Node ID to update, e.g. \"a1b2c3d4-e5f6-7890-abcd-ef1234567890\""
                 },
                 "content": {
                     "type": "string",
-                    "description": "New content/text for the node (optional)"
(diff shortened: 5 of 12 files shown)


malibio:
### Code Review Summary

**Review Type**: Initial Review (no prior reviews/comments found on PR #1467)

Net positive change. This is a well-scoped, additive PR: one new catalog entry, one new tool-call format parser mirroring the existing three, one new reasoning splitter mirroring the existing `ChannelSplitter`, and a tool-description wording fix. All 5 touched files are directly load-bearing for the stated goal, no unrelated changes bundled in (aside from an incidental `bun.lock` version bump, presumably from a workspace-wide `bun install`/version-bump landed on `main` after this branch was cut). All new logic is unit tested (7 new parser tests + 4 new `ThinkSplitter` tests + 1 new `emit_oai_delta` integration test), and the full touched test suites pass (89/89 `nodespace-nlp-engine` chat tests, 35/35 `nodespace-agent` model_manager/ollama tests), plus a clean `cargo clippy` on both crates.

### Requirements Check (issue #1465)

- ✅ **Catalog entry**: `ORNITH_1_9B` added with `ModelFamily::Qwen35`, `min_memory_gb: 16`, filename `ornith-1.0-9b-Q4_K_M.gguf`, HF source URL. Added to `CATALOG` array; `list_returns_all_catalog_models` test updated to assert count 13 and presence of `ornith-1-9b-q4km`.
- ✅ **`ModelFamily` wiring**: Reuses existing `ModelFamily::Qwen35` — no new variant needed since Qwen35 already routes through `apply_chat_template_oaicompat`/`use_jinja: true` per prior model additions. Correctly matches the issue's "or confirm Qwen35 covers it" option; `agent_types.rs` untouched (correctly — nothing needed there).
- ✅ **`update_node` prompt fix**: Tool description and each parameter's schema description now include concrete example values (`"id": "a1b2c3d4-..."`, `"content": "Buy milk and eggs"`, `{"status": "done"}`) instead of vague placeholders, directly addressing the "asks for clarification instead of calling the tool" failure mode from the eval.
- ✅ **`<think>` routing path**: New `ThinkSplitter` struct mirrors `ChannelSplitter`, wired into `emit_oai_delta` and the finalize path, emitting `ChatChunk::Reasoning` for `<think>...</think>` spans instead of leaking into `ChatChunk::Token`.
- ✅ **Unit tests for `parse_tool_calls_xml`**: All 5 requested cases covered (single call, multiple calls, missing `</tool_call>`, missing `</parameter>`, empty parameter body), plus 2 bonus cases (plain-text passthrough, missing `<function=>` tag).
- ⚠️ **Smoke test** ("Load Ornith 9B, send one tool-calling prompt, confirm ToolCallStart/ToolCallArgs emitted, no plain-text leakage"): Not performed in this PR — reasonably so, since it requires downloading and running a 5.23 GB GGUF model, impractical for an automated review. However, **the PR description (`"Closes #1465"`) does not acknowledge this gap at all.** This should be called out explicitly (e.g. "Smoke test deferred — requires manual local run with the model downloaded") so reviewers and future readers don't assume it was executed.

### Findings

#### 🟡 Important

**1. PR description doesn't disclose the undone smoke-test acceptance criterion**
📁 PR description (`gh pr view 1467`)

The PR body is just `"Closes #1465"`. One of six acceptance criteria — the smoke test — requires a manual, non-automatable step (downloading a 5.23 GB model and running a live tool-calling prompt) that clearly wasn't performed as part of this diff (no infra changes, no evaluation artifact, no log referenced). Silently closing the issue via `Closes #1465` implies all acceptance criteria are met.

**Recommendation**: Update the PR description to explicitly list which acceptance criteria are code-verified (unit tests) vs. which require a manual follow-up smoke test before/after merge, e.g.: "Smoke test (load Ornith 9B, live tool-calling prompt) not run in this PR — requires manual local verification with the model downloaded; tracked as a follow-up before promoting past trial status."
**Principle**: Traceability — a PR that closes an issue should make it obvious to any future reader which acceptance criteria were verified by what evidence (test run vs. manual claim), especially when the issue explicitly separates "trial" from "promoted" status for this exact model.

**2. `ThinkSplitter`'s trailing carry from `ChannelSplitter` is never think-scanned at finalize**
📁 `packages/nlp-engine/src/chat/mod.rs:605-607`

```rust
let (channel_answer, final_reasoning) = channel_splitter.finalize();
let (final_answer, final_think_reasoning) = think_splitter.finalize();
let final_answer = channel_answer + &final_answer;
```

In the streaming path (`emit_oai_delta`), `think_splitter.push(&answer)` only ever receives `channel_splitter`'s *emitted* answer text each iteration — never text still parked in `channel_splitter.carry` (a suffix that's a partial-prefix match against `<|channel>`/`<channel|>`, which can also start with the same leading `<` byte that a `<think>` marker starts with). If generation ends (context cap or natural stop) while `channel_splitter` is mid-way through holding back a candidate partial marker (e.g. a lone trailing `<`), that raw text is drained straight into `channel_answer` at finalize and concatenated into `final_answer` *without ever passing through `think_splitter`*. In the pathological case where that trailing fragment is actually the start of an unterminated `<think>` marker truncated by the token cap, it would leak as answer text instead of being (best-effort) treated as reasoning — the exact bug class this PR is fixing for the common case.

This is a genuine edge case (requires generation to be truncated at a very specific byte boundary), not a regression for typical completions, and the concatenation *order* itself (`channel_answer` before `think_splitter`'s own carry) is correct given the data-flow direction. Given it only affects truncated/cut-off generations at token-cap boundaries, I would not block on this.

**Recommendation**: Either (a) route `channel_answer` through `think_splitter.push()` before finalizing think_splitter (mirrors the live per-delta ordering: channel first, then think), or (b) add a comment acknowledging this is a known, accepted limitation for the truncated-generation edge case. Option (a) is cheap: 
```rust
let (channel_answer, final_reasoning) = channel_splitter.finalize();
let (think_answer, think_reasoning) = think_splitter.push(&channel_answer);
let (carry_answer, carry_reasoning) = think_splitter.finalize();
let final_answer = think_answer + &carry_answer;
let final_think_reasoning = think_reasoning + &carry_reasoning;
```
**Principle**: When composing two stateful, order-dependent parsers, the finalize/flush path must preserve the same data-flow ordering as the steady-state path, or the two can diverge exactly at end-of-stream — the boundary condition most likely to matter for truncated/long generations.

**3. No test exercises the combined `ChannelSplitter` + `ThinkSplitter` finalize path**
📁 `packages/nlp-engine/src/chat/mod.rs:1236` (test module)

All new tests exercise `ThinkSplitter` in isolation (`think_split_all` helper) or `emit_oai_delta` with a single complete XML tool-call block. None drive the full engine-loop-style sequence — `channel_splitter.push` → `think_splitter.push` → both `.finalize()` — with an unclosed marker in flight in *either* splitter at end-of-stream, which is exactly the scenario in Finding #2. Given this is the trickiest interaction surface in the change (two stateful order-dependent buffers chained together), it's the part most in need of a regression test.

**Recommendation**: Add a test that feeds `emit_oai_delta` (or a small harness combining both splitters like the existing `channel_split_all`/`think_split_all` helpers) a sequence ending with a truncated `<think>` marker, asserting the finalize output routes correctly.
**Principle**: Test the seams, not just the units — bugs in stateful pipelines concentrate at the composition boundary, not inside each component.

#### 🟢 Suggestions

**4. Ornith's empty-`sha256` comment doesn't name it as an unofficial/third-party repo, unlike sibling entries**
📁 `packages/agent/src/local_agent/model_manager.rs:172`

```rust
sha256: "", // Populate before promoting past user-selectable trial status
```

The struct-level doc comment (lines 47-54) states the empty-string policy is reserved for "official, tamper-evident" repos (`mistralai/...`, `ggml-org/...`); everything else needs a populated hash. Ornith's source is `huggingface.co/DeepReinforce/...` — clearly not one of the two official namespaces. Sibling third-party entries in this same file are explicit about *why* they're an exception, e.g.:
- `QWEN35_9B` (line 147): `"Unsloth repo — policy exception, trial only; populate hash before promoting"`
- Mistral Small (line 211): `"Third-party repo but widely used; populate before promoting to default"`

Ornith's comment omits the "third-party repo" qualifier that its siblings all state, even though the *behavior* (empty hash, trial-only, "populate before promoting") is consistent with the documented policy exception pattern — it's just less self-documenting than its neighbors. Not a defect (no verification is silently skipped for an entry masquerading as official), but worth tightening for consistency and to avoid a future reader mistaking the terse comment for an implicit "this is an official repo."

**Recommendation**: `sha256: "", // DeepReinforce repo — third-party, trial only; populate hash before promoting`
**Principle**: Consistency of self-documenting code — when a codebase has an established comment convention encoding a security-relevant policy distinction (official vs. third-party trust), every instance should follow it uniformly so a reader can rely on the pattern without cross-referencing the doc comment each time.

**Nit**: `packages/nlp-engine/src/chat/parser.rs` — `parse_xml_tool_call_block` computes `func_body_start` using `func_open` (an index into `block`) combined with lengths that were measured relative to `after_func_eq` (a subslice starting at `func_open + "<function=".len()`), which happens to be arithmetically correct here but is easy to get wrong on a future edit since the two index bases are silently threaded together across several lines. A short inline comment noting "index rebased into `block`" at the `func_body_start` computation would help future maintainers.

**Nit**: The `ORNITH_1_9B.context_window` (32,768) and `default_temperature` (0.6) are hardcoded to match `QWEN35_9B`'s values without a comment confirming these come from Ornith's own model card/config rather than being copy-pasted defaults — worth a one-line confirmation in the doc comment if these were verified against Ornith's actual `config.json`/chat template rather than assumed-identical to Qwen3.5.

### Security

No concerns. No user input is parsed here in a security-sensitive way — `parse_tool_calls_xml`/`parse_xml_tool_call_block` parse trusted local-model output (not attacker-controlled network input), bounded by `text.find`/slice operations that fail closed (`ParseResult::Error`) rather than panicking on malformed input. No secrets, no new network surface. The `sha256` empty-hash question (Finding #4) is a supply-chain/integrity concern in spirit but is explicitly gated behind "trial status" and matches an existing, accepted policy exception pattern already used for other third-party repos in this same catalog — consistent with prior precedent, not a new risk introduced by this PR.

### Performance

No concerns. Both parsers operate on already-materialized strings via `find`/slice (no regex, no unbounded backtracking); `ThinkSplitter` carries at most a few bytes of partial-marker state between deltas, identical complexity profile to the existing `ChannelSplitter`. No new allocations of consequence beyond what the existing sibling code already does.

### Testing

Strong coverage for the new parser (7/5 requested cases) and for `ThinkSplitter` in isolation (4 tests covering same-delta, split-across-deltas, unterminated-opener, and clean-content cases — mirroring the existing `ChannelSplitter` test suite one-for-one). The `emit_oai_delta_routes_ornith_xml_tool_call_to_chunks` test correctly validates the full JSON-delta → chunk pipeline. Gap: no test for the combined-finalize interaction (Finding #3), and the smoke test is out of scope for automation (acknowledged limitation, not a testing defect in this PR — see Finding #1 about the PR description not surfacing this).

### Documentation

Good: module-level doc comment in `parser.rs` extended with the Format D XML shape; inline doc comments on `ThinkSplitter` explain its relationship to `ChannelSplitter` and why it's a separate struct rather than a generic/parameterized one (reasonable design justification — avoids over-engineering a two-instance case into a generic abstraction, consistent with YAGNI). `ORNITH_1_9B`'s doc comment is thorough and cites the evaluation issue (#1464) with concrete numbers. Only gap is the PR description itself (Finding #1) and the sha256 comment specificity (Finding #4).

### Recommendation: COMMENT

This is a net-positive, well-tested, narrowly-scoped change that correctly implements 5 of 6 acceptance criteria in code, with the 6th (smoke test) being a reasonable manual/deferred item — the only real issue is that this deferral isn't disclosed in the PR description. None of the findings are blockers: the `ThinkSplitter` finalize edge case (Finding #2) is a narrow truncated-generation corner case with a cheap fix, and the sha256 comment wording (Finding #4) is a documentation-consistency nit, not a policy violation. I'm not able to self-approve this PR (it was authored through the same agent-driven workflow as this review), so the author/user should make the final approve/merge call — recommend addressing Finding #1 (PR description) at minimum before merge, with Findings #2 and #3 as strong follow-up candidates (could be same PR or a fast-follow).

malibio (commented):
Review posted above. Recommendation: COMMENT (net positive, no blockers). See full report in PR comments for the important/suggestion-level findings on ThinkSplitter finalize-path edge case, missing combined-splitter test, PR description gap on the smoke-test criterion, and the sha256 comment wording.

malibio:
## Addressed review feedback

**Important:**
- ✅ **ThinkSplitter finalize edge case** — `channel_splitter`'s leftover carry now routes through `think_splitter.push()` before finalizing, matching the streaming push path. Added `emit_oai_delta_finalize_routes_channel_carry_through_think_splitter` regression test.
- ✅ **PR description gap** — updated with a full test plan and explicit disclosure that the live-model smoke test (5.23GB download) was manually deferred, not run in this PR.

**Suggestion:**
- ✅ **Ornith sha256 comment** — clarified to explicitly call out DeepReinforce as a non-official repo requiring a hash before promotion past trial status, matching the file's stated verification policy.

**Skipped:**
- ⏭️ Index-arithmetic naming nit in `parse_xml_tool_call_block` — code is already covered by 7 passing unit tests; renaming offsets is a readability nice-to-have, not worth the churn for this PR.
- ⏭️ Confirming `context_window`/`temperature` against Ornith's model card — these match the sibling `QWEN35_9B` entry's conventions (32,768 / 0.6), consistent with treating Ornith as sharing the Qwen3.5 base per the issue's guidance; no issue-mandated action.

All tests re-verified after these changes: `nodespace-nlp-engine --lib --features chat-service` 90/90, `nodespace-agent --lib` 248/248, `quality:fix` clean.

## Re-Review Decision

**Rationale**: The changes are a targeted bugfix (routing carry through an existing, already-tested splitter) plus a regression test, a documentation-only comment clarification, and a PR description update. No new architecture, no new external behavior beyond fixing the described edge case. Low risk, straightforward to verify from the diff.

malibio:
### Code Review Summary — Re-Review of commit bb7f842e (3rd round)

**Scope**: This review covers only the newest commit, `bb7f842e`, on top of the two prior COMMENT-level review rounds (both already resolved — the ThinkSplitter finalize-carry fix and PR-description/sha256-comment updates from round 2 are not re-litigated here).

Net positive. This commit fixes a real, verified-broken download URL, closes a real leaked-output bug reproduced against a live transcript, corrects a factual overclaim in a doc comment, and does the minimal wiring needed to expose Ornith in the UI with an appropriate risk disclosure. All four sub-changes are cohesive under a single theme ("make Ornith actually usable end-to-end and honest about its limitations") — not unrelated bundling, given they were all discovered in the same live-testing pass.

### Independent Verification Performed

- **HF URL fix**: `curl -sIL https://huggingface.co/deepreinforce-ai/Ornith-1.0-9B-GGUF/resolve/main/ornith-1.0-9b-Q4_K_M.gguf` → `302` then `200`, `content-length: 5629108704`. This is an **exact match** to the new `size_bytes: 5_629_108_704` in `model_manager.rs`. Confirmed correct.
- **KV-cache full-reprocess claim**: `"KV cache partial trim returned false, falling back to full decode"` exists verbatim at `packages/nlp-engine/src/chat/mod.rs:399`, matching the doc comment's quoted log line.
- **Eval methodology claim**: Read `run_d_speed_at_context` in `../nodespace-nlp-experiment/src/ornith_eval.rs` (line 778). Confirmed: `make_ctx` (fresh `LlamaContext`) is created *inside* the per-context-size loop (line 803), and each iteration does exactly one `generate()` call — no re-prompting, no reused context across iterations. This substantiates the commit's core factual correction: the original "flat across context lengths" claim was a single-shot decode benchmark and does not measure (and cannot speak to) repeated-prompt ReAct-loop behavior. The new doc comment's hedging ("does NOT extend to... as originally assumed") is accurate, not an overclaim.
- **Regex safety**: `ornith_tool_call_re()` = `(?s)<tool_call>.*?</tool_call>` — non-greedy, no nested quantifiers, no alternation-of-overlapping-patterns. Compared against the existing `tool_call_prose_re()` (hand-unrolled brace nesting specifically to avoid ReDoS on `{...}` blobs). The new regex has no braces or nesting to unroll; a non-greedy dot-star between two literal anchors is standard-safe in Rust's `regex` crate (linear-time guaranteed, no backtracking engine). No concern.
- **Fallback logic parity** (`agent_loop.rs`): Traced the new `normalize_response(&tail_text)` → `normalized_tail.is_empty()` branch against the existing pattern at lines 466-478 (`normalized.is_empty()` gate after `normalize_response_traced`). Confirmed same shape: normalize once, branch on the normalized (not raw) emptiness. Bonus: this also eliminates a double-normalization call that existed in the prior version (`normalize_response` was called in the `else` branch redundantly) — a small correctness *and* efficiency win.
- **`{#if m.family === 'qwen35'}` scoping**: Confirmed this only matches `qwen35`; will not render for `ministral`. `LOCAL_FAMILIES` allowlist in `model-manager.svelte` matches `EXPOSED_GGUF_MODEL_IDS` in `chat_models.rs` (`ministral-8b-q4km`, `ornith-1-9b-q4km`) — the two allowlists are kept in sync as the comment at line 267 requires.
- **Warning banner CSS**: `.mm-notice` / `.mm-notice--warn` are pre-existing classes (already used at line 213 for the Ollama warning), not new/untested styling.
- **Test-input fidelity**: Manually traced both new dangling-tag test cases through `strip_ornith_tool_call_leak` by hand:
  - Whole-block case → regex matches and removes it → `""`. ✅
  - Dangling case (`<tool_call>...` with no closing tag, followed by `<|mask_end|>`) → regex doesn't match (no closer) → falls through to `.find("<tool_call>")` → truncates to `result[..0]` = `""` → mask_end replace is a no-op on empty string → `""`. ✅ Order of operations (regex-strip whole blocks first, then truncate-at-dangling-opener, then strip stray token) is correct and doesn't double-count.
  - Surrounding-text case correctly preserves prefix text before the tag. ✅

### Test & Lint Results (run independently, not trusted from the diff)

- `cargo test -p nodespace-agent --lib response_processing`: **31/31 passed**, including all 3 new tests (`strips_ornith_tool_call_leak_whole_response`, `strips_ornith_dangling_tool_call_and_mask_end_token`, `strips_ornith_tool_call_leak_keeping_surrounding_text`).
- `cargo test -p nodespace-agent --lib`: **251/251 passed**, no regressions.
- `cargo clippy -p nodespace-agent --all-targets -- -D warnings -D clippy::unused-async`: **clean**.
- `bun run --cwd packages/desktop-app quality:check` (eslint + svelte-check): **0 errors, 0 warnings** across 1702 files.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

**1. `strip_ornith_tool_call_leak` truncates *everything after* a dangling opener, including any trailing prose the model may have emitted post-truncation**
📁 `packages/agent/src/local_agent/response_processing.rs:319-323`

If a response were `"Sure, here's the plan.<tool_call>\n<function=partial` (truncated mid-generation, e.g. by the token cap) followed by, hypothetically, more legitimate prose after the truncation point in some other failure mode, that trailing content would also be silently discarded since `result[..idx]` drops everything from the dangling `<tool_call>` onward. In practice this is very unlikely (a truncated tag is definitionally at the *end* of a token-capped generation, so there's nothing meaningful after it), and the three tests correctly cover the realistic shapes. Not a blocker — just worth a one-line comment noting the assumption ("dangling tags are always trailing, since truncation happens at generation's end") so a future reader doesn't need to re-derive it.

**2. Live-model verification is explicitly deferred, not yet done**
📁 Whole commit

Per the task instructions, this fix has only been validated against captured/replayed transcript text in unit tests, not against a live Ornith inference run reproducing the original leak end-to-end. This is a reasonable and disclosed sequencing choice (unit tests against exact repro text give high confidence), but it should remain an open item until confirmed live — flagging per the author's own stated plan to do this next, not as something blocking this commit's quality.

#### Nitpicks

- **Nit**: `packages/agent/src/local_agent/model_manager.rs` doc comment — the phrase "an open, unresolved limitation affecting all Qwen 3.5-based hybrid models" is a slightly stronger claim than "these five specific upstream issues, which appear to affect Qwen 3.5-based hybrid models" — the upstream issue numbers themselves weren't re-verified in this review pass (per the task's explicit instruction to trust them), so if any of the 5 have since been closed/fixed upstream, the comment would go stale silently. Not asking for a change now — just noting this comment has a shelf life and isn't self-updating if upstream resolves the issue.
- **Nit**: `chat_models.rs:224` — `EXPOSED_GGUF_MODEL_IDS` and `model-manager.svelte`'s `LOCAL_FAMILIES` are two independent allowlists on two sides of the Tauri boundary that must be kept in sync by convention/comment rather than by a single shared source of truth. This is a pre-existing pattern (not introduced by this commit) and the comments on both sides now correctly cross-reference each other, which is the right mitigation given the architecture — just flagging that a future third model addition needs to remember to touch both files, as this commit did.

### Security
No concerns. The `<tool_call>`/`<|mask_end|>` stripping operates on trusted local-model output (not attacker-controlled input), fails safe (falls through to a tool-summary synthesis rather than crashing or leaking raw text), and the HF URL correction was independently verified to resolve to the intended artifact via HTTP HEAD before being trusted. No secrets, no new network surface, no auth/authz surface touched.

### Performance
No concerns. Regex is linear-time and non-catastrophic (verified above). The `normalize_response` de-duplication in `agent_loop.rs` is a minor net performance improvement (removes a redundant second normalization pass in the non-empty branch).

### Testing
Adequate for a text-transform bugfix: 3 new unit tests cover whole-block, dangling-block+stray-token, and block-with-surrounding-text — the three realistic shapes of the leak. Verified independently (see above) that the regex/truncation/replace pipeline produces the asserted outputs by hand-tracing, not just trusting the assertions. The one acknowledged gap (live-model end-to-end verification) is explicitly called out by the author as a deferred next step, not silently skipped.

### Documentation
The rewritten `ORNITH_1_9B` doc comment is a genuine improvement: it corrects a real factual gap (single-shot benchmark vs. multi-iteration ReAct behavior) rather than just rephrasing, cites concrete upstream issue numbers and a concrete measured number (~108s/4-iteration turn) rather than vague hand-waving, and is honest about the limitation being upstream/unfixable rather than implying it's a NodeSpace bug. The new UI warning banner text is consistent with the doc comment's technical claims (no contradiction between "gets progressively slower with each retry" in the UI copy and "every ReAct iteration re-decodes the entire growing prompt from scratch" in the Rust comment).

### Recommendation: COMMENT

Net-positive, well-verified fix-up commit. All claims I could independently check (HF URL/byte count, KV-cache log string, eval methodology, regex safety, fallback-pattern parity, family-filter scoping) held up under direct verification — this is unusually well-substantiated for a commit message. No blockers. The two "Suggested Improvements" are optional polish (a defensive comment, and an explicit "verify live" follow-up already on the author's own roadmap), not required before merge. As with the prior two rounds, I'm not self-approving since this PR was authored through the same agent-driven workflow as this review — the merge call remains the user's.

malibio (commented):
Re-review of commit bb7f842e (round 3): Net positive, no blockers. Independently verified the HF URL fix (curl HEAD matches new byte count exactly), the KV-cache log string, the eval methodology claim in ornith_eval.rs, regex ReDoS-safety, and fallback-pattern parity in agent_loop.rs. All 251 nodespace-agent tests pass (incl. 3 new), clippy clean, desktop-app quality:check clean. Full findings in PR comment above. Not self-approving (same agent-driven workflow) — merge call is the user's.

malibio:
## Addressed review feedback (3rd round)

Both suggested improvements addressed:
- ✅ Added a one-line comment in `strip_ornith_tool_call_leak` explaining the dangling-tag truncation assumption (truncation always happens at generation's end, so nothing legitimate follows it).
- ✅ Added a shelf-life note to the `ORNITH_1_9B` doc comment: the cited upstream llama.cpp issue numbers reflect state as of 2026-07-01 and should be re-checked before assuming the limitation still applies to a future vendored llama.cpp version.

## Live verification (the explicitly-flagged open item)

Ran two real conversation turns against the running daemon with Ornith loaded (`nodespace model load ornith-1-9b-q4km`), via `scripts/aichat.ts`:
1. A successful multi-step turn (search → update_task_status → confirm) — clean output, no leakage.
2. A tool-failure-recovery turn (nonexistent task ID → search → clarification request) — clean output, no leakage.

Inspected persisted content directly via `aichat.ts show` in both cases — no raw `<tool_call>` XML or `<|mask_end|>` token in either stored message.

**Caveat, stated plainly**: neither run happened to trigger the exact duplicate-tool-call-loop → tail-inference branch the fix specifically targets (Ornith self-corrected within normal ReAct iterations both times rather than looping into that code path). So this is strong supporting evidence, not an exhaustive reproduction of the original failure mode. Combined with the 3 unit tests built from the exact real leaked transcript text, and the reviewer's independent hand-trace of the strip/truncate/replace logic, I'm confident in the fix — but flagging the live-repro gap rather than claiming it's fully closed.

## Re-Review Decision

**Rationale**: Both changes are documentation-only (comments), no logic changes. The live-verification work confirms (but doesn't newly risk) the already-reviewed fix.

malibio:
### Code Review Summary

**Scope**: 4th re-review, covering only the three commits since round 3 (`414a71bd`): `933250f4` (hide Ornith from Settings UI), `6c4f510f` (merge `origin/main`, 6 commits incl. PR #1469), and `4b906bf4` (fix pre-existing clippy failures on `main`). Prior rounds (catalog entry, tool-call parsing, ThinkSplitter, HF URL fix, response-leak fix) are not re-litigated.

Net positive, no blockers. This round is entirely housekeeping around the already-approved Ornith work: a product decision to withhold the feature from users pending a known performance issue, a routine `main` sync, and an unrelated clippy fix picked up along the way. All three are cleanly scoped, well-explained in commit messages, and independently verified below. The highest-priority check this round — leftover conflict markers from the merge — came back clean.

### Findings

#### Critical Issues
None found.

#### Suggested Improvements
None.

#### Nitpicks

Nit: `packages/desktop-app/src/lib/components/settings/model-manager.svelte` still carries `LOCAL_FAMILIES = ['ministral', 'qwen35']` and a `{#if m.family === 'qwen35'}` slow-turn warning block, and `model-store.svelte.ts` still infers `family: 'qwen35'` for any id starting with `ornith`/`qwen35`. None of this was touched by `933250f4` (confirmed via empty diff on both files for that commit), and it's currently a harmless no-op since `EXPOSED_GGUF_MODEL_IDS` never returns a `qwen35`-family entry for the backend to surface. But it is a second, independent gate that happens to agree with the first only because nobody's touched it since round 1 — if a future PR adds a different `qwen35`-family model to the exposed list without reading the model-manager comment, it would inherit this dead code path (and the "multi-step tool-calling" warning copy, which is Ornith-specific in wording) without anyone intending it. The existing comment at `model-manager.svelte:27-29` ("families listed here must match whatever that allowlist currently exposes") already flags this fragility for the sighted reader; a stronger option long-term would be deriving `LOCAL_FAMILIES`/the warning condition from the same source of truth as `EXPOSED_GGUF_MODEL_IDS` rather than two hand-synced lists, but this isn't new debt introduced by this PR's commits and isn't worth blocking a hide-behind-a-flag change over.

Nit: `packages/desktop-app/src-tauri/src/commands/chat_models.rs`'s updated doc comment says "Re-add it here once #1477 is resolved" — good cross-reference. Worth noting for whoever picks up #1477 that re-adding the id alone is sufficient (the frontend filters above already accept it), so the fix there should stay purely backend-side unless the UX warning copy also needs to change.

### Verification

- **Conflict markers** (highest priority this round): `git grep -n "^<<<<<<<\|^=======\|^>>>>>>>"` across the entire worktree → **zero hits**. Merge is clean.
- **`933250f4` diff**: `EXPOSED_GGUF_MODEL_IDS` now `&["ministral-8b-q4km"]` only; doc comment clearly explains the SSM/KV-cache-reuse rationale and references `#1477` and the `ORNITH_1_9B` doc comment in `model_manager.rs`. Confirmed `model-store.svelte.ts` and `model-manager.svelte` have an **empty diff** for this commit — the `qwen35`-family frontend filters are pre-existing, not new redundant gating added here (see nitpick above for the forward-looking fragility note).
- **Merge commit `6c4f510f`**: pulls 6 commits from `origin/main` including PR #1469 (daemon lifecycle fix). Confirmed via `gh pr view 1469` that #1469 has an identical title to issue #1468 and its body states "Closes #1468" — direct fix, not an overlapping/duplicate effort.
  - Ornith survival spot checks, all confirmed present post-merge:
    - `strip_ornith_tool_call_leak` at `packages/agent/src/local_agent/response_processing.rs:314` (registered at 107-108).
    - Corrected HF URL `https://huggingface.co/deepreinforce-ai/Ornith-1.0-9B-GGUF/resolve/main/ornith-1.0-9b-Q4_K_M.gguf` at `packages/agent/src/local_agent/model_manager.rs:188`.
    - `packages/agent/src/local_agent/agent_loop.rs` retains the `normalized_tail`/`normalize_response(&tail_text)` empty-after-stripping fallback logic from `bb7f842e` (lines 871-898) — this file's Ornith-motivated change never mentions "ornith" literally in the diff (it's generic tail-inference logic), which is why the direct grep came back empty; confirmed by re-reading the original `bb7f842e` diff and matching it against current content.
- **`4b906bf4` diff** (`daemon_setup.rs`, pre-existing `main` breakage unrelated to this PR's Ornith work):
  - **`.trim()` removal**: reasoned through and empirically verified with a standalone Rust program across representative inputs (empty string, whitespace-only, leading/trailing spaces/tabs/newlines, internal multi-whitespace) — `x.split_whitespace().next()` and `x.trim().split_whitespace().next()` produced identical results in every case. This holds generally: `split_whitespace()` splits on runs of whitespace and discards empty leading/trailing fragments by construction, so a preceding `.trim()` cannot change its output. Behavior-neutral, confirmed.
  - **Doc comment reindentation**: read both versions directly. Old continuation lines were indented 14 spaces (past the `- macOS:` marker's own content column), which is overindented relative to CommonMark/rustdoc list-continuation rules and risked breaking out of the `- macOS:` list item into its own indented block. New version indents continuation lines to 8 spaces, exactly aligned with where "macOS:" text begins after the `- ` marker — correct sub-bullet continuation, stays part of the same list item. Confirmed not a regression to sibling-bullet or plain-paragraph rendering.
- **`cargo clippy --all-targets -- -D warnings -D clippy::unused-async`** (full workspace): clean, no warnings.
- **`cargo test -p nodespace-agent --lib`**: 251 passed, 0 failed.
- **`cargo test -p nodespace-app --lib`**: 30 passed, 0 failed.
- **`bun run --cwd packages/desktop-app quality:check`**: 0 errors, 0 warnings (eslint + svelte-check on 1711 files).
- **Working tree**: clean (`git status --short` empty), HEAD confirmed at `4b906bf4`.

malibio (commented):
4th re-review of commits since round 3 (933250f4 hide-Ornith, merge from main incl. PR #1469, 4b906bf4 pre-existing clippy fix). Net positive, no blockers. Verified zero leftover conflict markers workspace-wide, Ornith fixes from prior rounds survived the merge intact, the .trim()/split_whitespace and doc-comment reindent changes are behavior-neutral (reasoned + empirically checked). clippy clean, nodespace-agent 251/251, nodespace-app 30/30, quality:check clean. One forward-looking nitpick about dual-gating fragility in the frontend qwen35 filters (pre-existing, not new). Full report in PR comment above.

PR #1469, opened 2026-07-01
Daemon lifecycle: dev worktrees, production app, and Pro/community builds collide on shared launchd service + socket

malibio:
Closes #1468

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -119,7 +119,7 @@
     },
     "packages/skill": {
       "name": "@nodespaceai/skill",
-      "version": "0.1.0",
+      "version": "0.2.0",
       "bin": {
         "nodespace-skill": "dist/install.js",
       },
--- a/packages/desktop-app/src-tauri/src/constants.rs
+++ b/packages/desktop-app/src-tauri/src/constants.rs
@@ -1,6 +1,6 @@
 //! Shared constants used across the application.
-
-/// Relative path from HOME to the nodespaced Unix Domain Socket.
-/// Shared by daemon_setup (launchd plist) and lib.rs (health check command).
-/// watcher.rs uses its own resolver that also honors NODESPACED_SOCKET env override.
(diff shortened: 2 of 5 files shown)


malibio:
## Re-review: All initial findings addressed

### Blocker resolved

**`watcher.rs` `socket_path()`** now calls `crate::daemon_setup::daemon_socket_relative()` instead of the hardcoded `.nodespace/daemon.sock`. The private function in `watcher.rs` is now consistent with `grpc_client::resolve_socket_path()` and `daemon_setup` itself — all three socket-path resolution paths converge on the same variant-scoped function.

### Suggestions resolved

**`plist_filename()`** now returns `String` via `format!("{}.plist", launch_agent_label())`. The call site at `launch_agents_dir(&home).join(plist_filename())` compiles correctly: `PathBuf::join` accepts `AsRef<Path>`, which `String` satisfies. The separate match table that could drift is gone.

**Doc comments** in `watcher.rs` and `grpc_client.rs` updated to reference the variant-scoped path rather than the hardcoded default.

**`ks_err_str` renamed to `retry_stderr`** — clearer intent, consistent with surrounding log variable names.

### No new issues introduced

Verified:
- `DAEMON_SOCKET_RELATIVE`, `LAUNCH_AGENT_LABEL`, and `PLIST_FILENAME` constants are fully removed with zero remaining references.
- The only occurrences of `.nodespace/daemon.sock` in source are inside `daemon_socket_relative()` itself (doc comment + the release-community return value) — correct by definition.
- All four socket-resolution call sites (`daemon_setup` x2, `grpc_client`, `watcher`) route through `daemon_socket_relative()`.
- `constants.rs` updated with a tombstone comment directing callers to the right function.

### Issue #1468 coverage

All four requirements remain intact: variant-scoped socket paths, variant-scoped launchd labels, self-healing launchd bootstrap (bootout+retry+kickstart), and exact-path binary identity check replacing the trailing-substring heuristic.

**Approved.**

Issue #237, opened 2025-10-11
Refactor BaseNodeViewer to use SharedNodeStore for all database operations


malibio:
## Problem

BaseNodeViewer bypasses SharedNodeStore and directly calls database service methods, creating dual persistence paths that cause conflicts and violate single-source-of-truth principle.

### Current Architecture (Broken)
```
BaseNodeViewer → databaseService → database
ReactiveNodeService → SharedNodeStore → tauriNodeService → database
```

### Direct Database Calls in BaseNodeViewer
- `databaseService.getNode()` - checking node existence
- `databaseService.deleteNode()` - deleting nodes
- `databaseService.updateNode()` - updating structural changes
- `databaseService.getNodesByOriginId()` - loading children
- `databaseService.saveNodeWithParent()` - saving content (debounced, placeholder-aware)

### Issues Caused
1. **Duplicate INSERTs** - Both paths try to create nodes (temporarily fixed with `source.type !== 'viewer'` workaround)
2. **Conflicting deletes** - Not covered by workaround
3. **Conflicting structural updates** - Not covered by workaround
4. **Tests can't catch bugs** - Tests use HttpAdapter, real app uses TauriNodeService, SharedNodeStore persistence never tested
5. **Inconsistent state** - SharedNodeStore.persistedNodeIds tracking doesn't know about BaseNodeViewer's direct saves

### Temporary Workaround (Added in #232)
Added `&& source.type !== 'viewer'` check in SharedNodeStore to skip persisting viewer-sourced updates. This prevents duplicate INSERTs but doesn't address all conflicts.

**IMPORTANT:** This workaround must be removed as part of this refactoring.

## Solution

Refactor BaseNodeViewer to use SharedNodeStore for ALL database operations.

### Target Architecture
```
BaseNodeViewer → ReactiveNodeService → SharedNodeStore → tauriNodeService → database
```

### Changes Needed

1. **Remove direct database calls from BaseNodeViewer**
   - Replace `databaseService.getNodesByOriginId()` with SharedNodeStore query
   - Replace `databaseService.saveNodeWithParent()` with `sharedNodeStore.updateNode()`
   - Replace `databaseService.deleteNode()` with `sharedNodeStore.deleteNode()`
   - Replace `databaseService.updateNode()` with `sharedNodeStore.updateNode()`

2. **Move debouncing to SharedNodeStore**
   - Transfer 500ms debounce logic from BaseNodeViewer
[...]

PR #1472, opened 2026-07-02
Add pro_list_invites / pro_list_requests / pro_revoke_invite / pro_current_person Tauri commands (#239)

mstomar125:
Core (desktop-app) half of **NodeSpaceAI/nodespace-sync#239** (S2 of the collection-membership-UI epic #237). Pairs with the daemon-half PR on nodespace-sync (adds the four gRPC methods this consumes).

## What
Refreshes the vendored Pro proto from the daemon source-of-truth and adds four Tauri commands, each a thin pass-through over the daemon gRPC client — mirroring the eight existing membership commands (resolve the `ProClient` or fail in community mode, then one RPC):

| Command | Returns | Purpose |
|---|---|---|
| `pro_list_invites(collection_id)` | `Vec<InviteDto{id,code,email,permission,expires_at}>` | admin: list a collection's pending invites |
| `pro_list_requests(collection_id)` | `Vec<RequestDto{id,requested_by,created_at}>` | admin: list pending join requests |
| `pro_revoke_invite(invite_id)` | `()` | admin: cancel a pending invite **or** join request |
| `pro_current_person()` | `PersonDto{person_id,email}` | the caller's own bound PersonNode id + email |

`pro_current_person` gives the UI the caller's identity so it can tell which roster row is "me" and gate admin controls on the caller's own per-collection role. An empty `person_id` (un-bound device) is treated as "role unknown".

The proto is refreshed wholesale from `nodespace-sync/nodespaced-pro/proto/nodespace_pro.proto` (the documented `scripts/refresh-pro-proto.sh` source-of-truth), so the vendored copy stays byte-identical to the daemon's wire contract.

## Community build
All four resolve the `ProClient` or return an error in community mode (no `ProClient` in managed state) — identical to the existing membership commands. The community `nodespaced` binary has no knowledge of this proto package, so the community build is behaviorally unchanged.

## Checks
- `cargo check` — clean (proto regenerates via `build.rs`).
- `cargo clippy` — clean.
- `cargo fmt -- --check` — clean.

Rust-only change (Tauri command bindings + vendored proto); no frontend surface yet — the UI lands in S1/S3/S4/S5.

Part of NodeSpaceAI/nodespace-sync#239.


mstomar125:
**APPROVE**

S2 core-half — correct and wire-safe. Verified at head `fd3592d6`.

- 🟢 **Wire compat: independently confirmed byte-identical to sync#244.** I fetched both protos at their heads (`fd3592d6` / sync#244 `f6d4b122`) and diffed them — 236 lines each, zero differences. The new S2 surface matches exactly: `ListInvites`/`ListRequests`/`RevokeInvite`→`MembershipAck`/`GetIdentity`, and messages `Invite`(id=1,code=2,email=3,permission=4,expires_at=5), `JoinRequest`(id=1,requested_by=2,created_at=3), `RevokeInviteRequest.invite_id=1`, `GetIdentityResponse`(person_node_id=1,email=2). No field-number drift.
- 🟢 **Community no-op:** all 4 commands resolve through the shared `membership_client` guard (`pro_sync.rs:239`) which returns `Err` when there's no `ProClient`, propagated via `?` — no panic/unwrap/side-effect, identical to the existing 6 membership commands. Free-user build stays behaviorally unchanged (additive registration only).
- 🟢 **DTOs** (`pro_sync.rs:202-235`) are snake_case + `Serialize`; nice that `PersonDto.person_id` (renamed from proto `person_node_id`) aligns with `MemberDto.person_id` so the frontend can match "me" in the roster by equality. Heads-up for #1473: read `person_id`, not `person_node_id`.
- 🟢 **Registration:** all 4 in `lib.rs` invoke_handler (`:452-455`).

🟢 Non-blocking (pre-existing): the module doc (`pro_sync.rs:3-6`) says commands "return early with `Ok`" in community mode, but the membership family returns `Err` — drift from M5 #147, not this PR.

🔶 Cross-layer: this consumes sync#244's daemon RPCs (still open) — keep the daemon landing ahead of / with this so the commands resolve at runtime. LGTM.

Issue #143, opened 2025-10-03
Optimize bulk_update to avoid N+1 query pattern


malibio:
## Overview
The NodeService::bulk_update method performs N individual SELECT queries when updating N nodes (N+1 pattern).

## Current Implementation
Method calls get_node for each update in a loop (node_service.rs:920-925).

## Proposed Optimization
Batch-fetch all nodes in a single query before the update loop using IN clause.

## Benefits
- Performance: Reduces N+1 queries to 2 queries
- Scalability: Better for large bulk operations
- Database Load: Reduces round-trips

## Context
Identified during code review of PR #142. Non-blocking optimization.

Related: PR #142, Issue #104

PR #1473, opened 2026-07-02
Add membership service + store foundations for the Collaboration UI (#238)

mstomar125:
Slice **S1** of the collection-membership-UI epic (NodeSpaceAI/nodespace-sync#237). Foundations only — no visible surface yet; this is the service + reactive store that the Collaboration view (S3/S4) and onboarding inbox (S5) build on, layered over the already-complete #143 membership backend.

## What
- **`services/membership-service.ts`** — typed, thin wrappers over the eight `pro_*` membership commands plus S2's three (`pro_list_invites` / `pro_list_requests` / `pro_revoke_invite`) and `pro_current_person`. The daemon enforces the ADR-037 gates server-side; these carry no authority. Maps the commands' **snake_case** DTOs (`person_id`, `expires_at`, …) to camelCase interfaces at the boundary so the rest of the frontend never sees the wire shape.
- **`stores/membership.svelte.ts`** — a Svelte 5 runes store caching, per collection, the roster + pending invites/requests, plus the caller's own identity. `currentUserRole(collectionId)` matches the caller's person id against the roster so admin controls gate correctly; the admin-only invite/request listings are fetched **only** when the caller is an admin (a non-admin call would just 403). Mutations refresh the affected collection.

## Pro-gated / community-inert
Every store method early-returns (or yields empty) when `proSync.tier !== 'pro'`, so the community build never touches the membership commands and is behaviorally unchanged.

## Caller identity
Consumes S2's `pro_current_person` (NodeSpaceAI/nodespace-core#1472 / NodeSpaceAI/nodespace-sync#244) — fetched once and cached, cleared on sign-out via `reset()`. Runtime resolution of these commands depends on those S2 PRs landing; this slice's unit tests mock the boundary, so it builds and tests green independently.

## Tests
- `bun run test` on the two new files — **16 passed** (service arg-forwarding + snake→camel mapping; store role-derivation, admin gating, community-inert, mutation-refresh, reset).
- `bun run quality:fix` — ESLint 0/0, `svelte-check` 0 errors.

Depends (at runtime) on #239. Unblocks S3 (#240) / S4 (#241) / S5 (#242).

### DIFF
--- a/packages/desktop-app/src/lib/services/membership-service.ts
+++ b/packages/desktop-app/src/lib/services/membership-service.ts
@@ -1,0 +1,194 @@
+/**
+ * Membership service — thin, typed wrappers over the Pro `pro_*` membership
+ * Tauri commands (collection-membership-UI epic #237, slice S1 #238).
+ *
+ * The daemon forwards the signed-in user's JWT to the matching cloud RPC, so the
+ * admin / last-admin / open-vs-restricted gates are enforced *server-side*; these
+ * wrappers carry no authority of their own. Every command resolves the Pro client
+ * or errors in community mode, so callers must gate on `proSync.tier === 'pro'`
+ * (the membership store does this) before invoking — the community build never
+ * reaches here.
+ *
+ * Boundary mapping: the Rust command DTOs serialize snake_case (`person_id`,
+ * `expires_at`, …); this module maps them to idiomatic camelCase interfaces so
+ * the rest of the frontend never sees the wire shape.
+ */
+import { invoke } from '@tauri-apps/api/core';
+
+import { createLogger } from '$lib/utils/logger';
+
+const log = createLogger('MembershipService');
+
+/** ADR-037 permission tiers. */
+export type Permission = 'admin' | 'modify' | 'readOnly';
+
+/** A collection roster entry (person + their tier). */
+export interface Member {
+	personId: string;
+	permission: Permission;
+}
+
+/** A pending invite (admin-only listing). */
+export interface Invite {
+	/** uuid — the handle {@link MembershipService.revokeInvite} takes. */
+	id: string;
+	/** 64-hex share code (bearer); may be surfaced for the admin to copy. */
+	code: string;
+	/** Bound invitee email; `''` for a bearer share-code. */
+	email: string;
+	permission: Permission;
+	/** RFC3339; `''` when the invite never expires. */
+	expiresAt: string;
+}
+
+/** A pending join request (admin-only listing). */
+export interface JoinRequest {
+	/** uuid — the handle {@link MembershipService.approveRequest} / revokeInvite take. */
+	id: string;
+	/** The requester's person_node_id. */
+	requestedBy: string;
+	/** RFC3339. */
+	createdAt: string;
+}
+
+/** The caller's own identity. */
+export interface Person {
+	/** Bound PersonNode id; `''` on an un-bound device ("role unknown"). */
+	personId: string;
+	/** Signed-in email; `''` when signed out. */
+	email: string;
+}
+
+// Raw wire shapes (snake_case, as serialized by the Rust command DTOs).
+interface RawMember {
+	person_id: string;
+	permission: string;
+}
+interface RawInvite {
+	id: string;
+	code: string;
+	email: string;
+	permission: string;
+	expires_at: string;
+}
+interface RawRequest {
+	id: string;
+	requested_by: string;
+	created_at: string;
+}
+interface RawPerson {
+	person_id: string;
+	email: string;
+}
+
+/**
+ * Typed facade over the membership commands. A single class (not the
+ * Tauri/Mock/Http triplet used by `collection-service`) because tests mock
+ * `@tauri-apps/api/core` directly, matching the `pro-sync` store's test pattern.
+ */
+export class MembershipService {
+	/** Roster for a collection (any member can read). */
+	async listMembers(collectionId: string): Promise<Member[]> {
+		log.debug('listMembers', { collectionId });
+		const rows = await invoke<RawMember[]>('pro_list_members', { collectionId });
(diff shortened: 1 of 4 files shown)


mstomar125:
**REQUEST CHANGES**

Solid, well-documented S1 foundation — the DTO mapping and admin-gating are correct and the tests are meaningful. One free-user-mandate gap at the foundation blocks it. (I verified each line below against the file at `90efb7c3`.)

🔴 **Mutations don't honor the store's own "community-inert" contract.** Only the read path is Pro-gated: `ensureIdentity` (`membership.svelte.ts:96` `if (!this.isPro || this.currentPerson) return`) and `loadCollection` (`:111` `if (!this.isPro) return`). But all 9 mutations call `membershipService` unconditionally — `setMember:138`, `removeMember:143`, `leaveCollection:148`, `createInvite:161`, `revokeInvite:167`, `approveRequest:176`, `rejectRequest:182`, `acceptInvite:192`, `requestJoin:197` — each of which invokes a `pro_*` command that in the community build reaches `membership_client()` (`src-tauri/src/commands/pro_sync.rs`) and errors "community tier — Pro membership operations unavailable." That contradicts the store's own docstring (`:8-10`: "**every** method early-returns / yields empty in community mode … never touches the membership commands").
- Today this is *latent* — the only consumer is the Pro-gated Collaboration view (#1474), so nothing community-reachable calls these yet. But this is the **foundation** S5 builds on, and `acceptInvite`/`requestJoin` are exactly the S5 onboarding entry points (paste-an-invite-code) that WILL be reachable by a community user. Fix it here so the invariant holds before S5 wires it.
- Add `if (!this.isPro) return …` to each mutation: `void` ones just `return`; the string-returning `createInvite`/`acceptInvite`/`requestJoin` either throw a clear "Pro required" or return `''` (your call) — but they must not reach the service in community mode.

🟡 **Pin the invariant in tests.** `membership.test.ts:79-88` only asserts `loadCollection` is inert in community mode. Add a case that a mutation (e.g. `setMember` and/or `acceptInvite`) never calls the service when `proSync.isPro === false`, so this can't regress.

🟡 Non-blocking:
- `ensureIdentity` (`:114-121`) caches an un-bound `{personId:'',email:''}` (truthy), so if the device binds identity mid-session without a `reset()`, `currentUserRole` stays `null` all session. Minor — noting in case a bind should also trigger a re-fetch.
- `loadCollection` (`:138-146`) commits the fetched roster only after the admin `Promise.all([listInvites, listRequests])`; if either rejects, the already-fetched roster is dropped for just `error`. Consider patching the roster before the admin fetch.

✅ **Confirmed good:** snake_case→camelCase DTO mapping matches the real Rust DTOs (`MemberDto` in main; `InviteDto`/`RequestDto`/`PersonDto` in #1472 — `person_id→personId`, `expires_at→expiresAt`, etc., none dropped); invoke arg casing matches `collection-service.ts`; `currentUserRole` matches the caller's own person id against the freshly-loaded roster and gates the admin-only listings on it (403-avoidance, not authority); `reset()` clears cache+identity. No local security decision — visibility-gating only, server enforces. S2 commands aren't in `main` yet (#1472 open) — expected; tests mock the boundary so this builds green.

Add the mutation guards + the test and I'll re-review.

mstomar125:
Addressed at `dd8ec94e`.

🔴 **Community-inert contract on mutations — fixed.** All 9 mutations now honor the contract:
- void mutations (`setMember`/`removeMember`/`leaveCollection`/`revokeInvite`/`approveRequest`/`rejectRequest`) early-`return` in community mode;
- the value-returning `createInvite`/`acceptInvite`/`requestJoin` throw via a new `requirePro()` guard — never a fake-success value — so none reach the `pro_*` commands when `tier !== 'pro'`. This closes the gap before S5 wires `acceptInvite`/`requestJoin` as onboarding entry points.

🟡 **Invariant pinned in tests.** New case asserts a void mutation no-ops and the value-returning ones throw, none touching the service, when `isPro === false`.

🟡 **Non-blocking, both fixed:**
- `ensureIdentity` now retries while `personId` is empty (`if (!this.isPro || this.currentPerson?.personId) return`), so a mid-session identity bind is picked up instead of leaving `currentUserRole` null all session.
- `loadCollection` commits the roster **before** the admin-only `Promise.all([listInvites, listRequests])`, so a listing failure no longer discards an already-fetched roster.

`bun run test` (8 store tests) green; ESLint 0/0; `svelte-check` clean. Ready for re-review.

mstomar125:
**APPROVE** — resolves my earlier REQUEST CHANGES (@90efb7c3).

The community-inert contract is now enforced at the store boundary, not just relied on from Pro-gated callers. All 9 mutations in `packages/desktop-app/src/lib/stores/membership.svelte.ts` short-circuit before touching a `pro_*` command:

- void-returning (`setMember`, `removeMember`, `leaveCollection`, `revokeInvite`, `approveRequest`, `rejectRequest`) → `if (!this.isPro) return;` — silent no-op.
- value-returning (`createInvite`, `acceptInvite`, `requestJoin`) → `requirePro()` throws, so they can never hand back a fake invite code / request id in community mode. Correct distinction — the S5 onboarding entry points (`acceptInvite`/`requestJoin`) are the community-reachable ones, and this is exactly where a silent `return undefined` would have been a landmine.

`membership.test.ts` pins the invariant directly: community mode → `setMember` never hits the service, and `acceptInvite`/`requestJoin` reject with `/Pro/` without touching the service. Good.

Two bundled improvements, both fine:
- `loadCollection` now commits the roster (`patch(..., {members, loading:false})`) before the admin-only invites/requests fetch, so a failure of those listings no longer discards an already-fetched roster.
- `ensureIdentity` retries while `personId` is empty (`currentPerson?.personId`) so a mid-session identity bind is picked up rather than leaving `currentUserRole` null all session.

Nit (non-blocking): a member demoted admin→non-admin keeps stale cached `invites`/`requests` (no longer reset to `[]` on reload), but those sections are admin-gated in the view so it isn't user-visible. No change needed.

PR #1475, opened 2026-07-02
Add invites & join-requests admin sections to the Collaboration view (#241)

mstomar125:
Slice **S4** of the collection-membership-UI epic (NodeSpaceAI/nodespace-sync#237) — invite & request management in the Collaboration view.

> **Stacked on #1474 (S3) → #1473 (S1).** Cross-fork PRs base on `main`, so the diff includes the S1+S3 commits until they merge — **review the top commit** (`Add invites & join-requests…`). I'll rebase as the parents land.

## What
Extends the S3 Collaboration view with two admin-only sections (gated on the caller's own admin role; consumes the invites/requests the S1 store already loads for admins):

**Invites**
- Create an invite with a role, optional email binding, and a TTL preset (1 day / 7 days / never) → `pro_create_invite`.
- A bearer (no-email) invite surfaces its **share code with a copy button**; an email-bound invite is delivered to that address (no code shown).
- Pending-invites list with per-invite copy-code and **revoke** → `pro_revoke_invite`.

**Join requests**
- List pending requests; **approve** with a selectable role → `pro_approve_request`, or **reject** → `pro_revoke_invite` (via the store's `rejectRequest`).

## Tests
- 4 new component tests (admin-only section gating; bearer-invite create with default role/TTL + code surfaced; revoke; approve-with-selected-role + reject). 9 total in `collaboration-view.test.ts`.
- `bun run quality:fix` — ESLint 0/0, `svelte-check` 0 errors.

Depends on S1 (#238) + S2 (#239). Part of #241.

PR #1476, opened 2026-07-02
Add Invitations inbox + first-run onboarding to the Pro account menu (#242)

mstomar125:
Slice **S5** of the collection-membership-UI epic (NodeSpaceAI/nodespace-sync#237) — the onboarding entry point.

> **Stacked on #1473 (S1).** Cross-fork PRs base on `main`; the diff includes the S1 commit until #1473 merges — **review the top commit**. Independent of S3/S4 (touches `pro-sync-pill.svelte`, not the Collaboration view). I'll rebase onto `main` once #1473 lands.

## What
- **`invitations-inbox.svelte`** — a modal (house overlay + `focusTrap`) with:
  - **Redeem an invite code** → `pro_accept_invite`
  - **Request to join a collection** (by id) → `pro_request_join`
  - copy noting **email invites apply automatically** on sign-in (`redeem_my_invites` runs server-side on reconnect).
- **`pro-sync-pill.svelte`** — an **Invitations** item in the account menu opens the inbox; a **first-run** effect opens it once per device on the first sign-in (localStorage-gated). **Sign-out now also clears the membership store** so the next user doesn't inherit the previous session's cached roster/identity.

## Deferred (flagged in code)
Listing one's *own* pending invites / join-request statuses and *discovering* joinable collections have no client-facing backend yet (only admins can list a collection's invites/requests; there's no "my invites" RPC and no collection-discovery). The inbox works from a shared code / known collection id; surfacing auto-redeemed invites ("You were added to X") needs a daemon event or a `my_collection_ids` command — a follow-up.

## Tests
- 5 component tests (redeem success/failure, request-to-join, close, closed-render).
- `bun run quality:fix` — ESLint 0/0, `svelte-check` 0 errors.

Depends on S1 (#238). Part of #242.

### DIFF
--- a/packages/desktop-app/src/lib/components/collaboration/invitations-inbox.svelte
+++ b/packages/desktop-app/src/lib/components/collaboration/invitations-inbox.svelte
@@ -1,0 +1,224 @@
+<!--
+  InvitationsInbox — onboarding entry point (epic #237, slice S5 #242). A modal
+  opened from the Pro account menu (and once, on first sign-in). Lets a user
+  redeem a share-code invite and ask to join a restricted collection.
+
+  Email-bound invites are auto-redeemed server-side on reconnect
+  (`redeem_my_invites`), so they need no action here — the copy just says so.
+  Listing "my pending invites / my request statuses" and discovering joinable
+  collections have no client-facing backend yet (only admins can list a
+  collection's invites/requests); those are deferred to a follow-up.
+-->
+<script lang="ts">
+	import { focusTrap } from '$lib/actions/focus-trap';
+	import { membership } from '$lib/stores/membership.svelte';
+	import { createLogger } from '$lib/utils/logger';
+
+	const log = createLogger('InvitationsInbox');
+
+	let { open = false, onClose }: { open?: boolean; onClose: () => void } = $props();
+
+	let code = $state('');
+	let redeeming = $state(false);
+	let redeemMsg = $state<string | null>(null);
+	let redeemErr = $state<string | null>(null);
+
+	let joinId = $state('');
+	let requesting = $state(false);
+	let requestMsg = $state<string | null>(null);
+	let requestErr = $state<string | null>(null);
+
+	function friendly(e: unknown): string {
+		return String(e).replace(/^Error:\s*/, '');
+	}
+
+	async function redeem() {
+		const c = code.trim();
+		if (!c) return;
+		redeeming = true;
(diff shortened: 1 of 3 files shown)


mstomar125:
**APPROVE** — S5 (Invitations inbox + first-run onboarding) is cleanly Pro-gated; one load-bearing **merge-order** requirement below.

Verified against code at head `d7c27c4e`:

**Free-user mandate — intact.** `pro-sync-pill.svelte`'s new entry points live entirely inside `{#if proSync.isPro}`; the "Invitations" menu item is further behind `menuOpen && signedIn`, and `invitations-inbox.svelte` is mounted only in that Pro block — so a community user can never reach the inbox or the `acceptInvite`/`requestJoin` calls. First-run onboarding fires only when `signedIn` (daemon identity resolved) and is persisted once-per-device via the `ns:invitations-firstrun-seen` localStorage flag (not cleared on sign-out — correct: once per device, not per user). `reset()` on sign-out clears identity + roster cache so the next user doesn't inherit state. `redeem()`/`requestJoin()` have try/catch/finally with loading + friendly-error handling. Good.

**⚠️ Merge-order (blocking on ordering, not on code):** like #1475, this branch is stacked on the **pre-fix** S1 commit `90efb7c3` (its direct parent), not #1473's fixed head `dd8ec94e`. So `stores/membership.svelte.ts` here is the un-guarded version — `acceptInvite`/`requestJoin` (`:191`/`:196`) call the service with no `!isPro` guard. This is the slice whose entry points are community-reachable *by design*, so the store-level guard from #1473 matters most here: **rebase onto #1473 (`dd8ec94e`) / main-after-#1473 before merging.** Today the inbox's Pro-gated mount already prevents a community call path, so it's latent — but the defense-in-depth guard is exactly why #1473's fix exists; don't ship S5 on the pre-fix store.

Rebase-safe: #1476's own commit (`d7c27c4e`) touches only `invitations-inbox.svelte`, `pro-sync-pill.svelte`, and the inbox test — **not** the store. Rebasing onto main-after-#1473 replaces the store with the guarded version automatically; no hand-fix.

Nit (non-blocking): consider a `membership.test.ts` case asserting a mutation is inert in community mode (the query-method inert test exists; #1473 already added the mutation-inert test, which you'll inherit on rebase). Epic order: **#1473 (S1) → #1476 (S5)** — S5 is independent of S3/S4. (CI: no checks reported yet.)

PR #1478, opened 2026-07-02
Frontend startup sequencing: UI render blocked on daemon readiness with no degraded-state feedback

malibio:
Closes #1470

### DIFF
--- a/packages/desktop-app/src/lib/components/layout/app-shell.svelte
+++ b/packages/desktop-app/src/lib/components/layout/app-shell.svelte
@@ -33,13 +33,24 @@
   import ConflictToast from '$lib/components/conflict-toast.svelte';
   import { recoveredItems } from '$lib/stores/recovered-items.svelte';
   import { conflictNotifications } from '$lib/stores/conflict-notifications.svelte';
+  import { daemonStatus, startDaemonStatusListener } from '$lib/services/daemon-status';
 
   // Logger instance for AppShell component
   const log = createLogger('AppShell');
 
-  // Daemon connectivity state (Issue #1179). Set to true when Tauri signals
-  // that nodespaced could not be reached within 5 seconds of app launch.
+  // Daemon connectivity state (Issue #1179, generalized in #1470). `connecting`
+  // covers the brief window before the first daemon-status event arrives (or
+  // a short grace period elapses); `unreachable` reflects a terminal
+  // `not_running` event.
+  let daemonConnecting = $state(true);
   let daemonUnreachable = $state(false);
+  $effect(() => {
+    const unsubscribe = daemonStatus.subscribe((s) => {
+      daemonConnecting = s.connecting;
+      daemonUnreachable = s.unreachable;
+    });
+    return unsubscribe;
+  });
 
   // First-launch onboarding wizard (Issue #1180).
   let showOnboarding = $state(false);
@@ -234,7 +245,6 @@
     let unlistenDatabase: Promise<() => void> | null = null;
     let unlistenSettings: Promise<() => void> | null = null;
     let unlistenIntegrations: Promise<() => void> | null = null;
-    let unlistenDaemonStatus: Promise<() => void> | null = null;
     let unlistenSkillCliMissing: Promise<() => void> | null = null;
     let cleanupMCP: (() => Promise<void>) | null = null;
     let staleNodesInterval: ReturnType<typeof setInterval> | null = null;
@@ -266,8 +276,10 @@
           log.debug('Could not sync theme from backend preferences:', err);
         });
 
-      // Re-load schemas and collections once the daemon signals it is ready (pro:tier-detected).
-      // This handles the case where the initial load in NavigationSidebar fired before gRPC was available.
+      // Re-load schemas and collections once the daemon's Pro-tier capability probe
+      // resolves. Kept in addition to the generic daemon-reconnect hook below since
+      // Pro tier detection is a distinct signal that can complete after the initial
+      // healthy event (it depends on gRPC being reachable, not just the socket).
       unlistenTier = listen('pro:tier-detected', () => {
         schemasData.loadSchemas();
         collectionsData.loadCollections();
@@ -279,16 +291,10 @@
         statusBar.error(event.payload.warning);
       });
 
-      // Listen for daemon connectivity failures (Issue #1179).
-      // Emitted by lib.rs when nodespaced is not reachable after 5s.
-      unlistenDaemonStatus = listen<string>('daemon-status', (event) => {
-        if (event.payload === 'not_running') {
-          log.warn('nodespaced is not reachable — showing error state');
-          daemonUnreachable = true;
-        } else {
-          daemonUnreachable = false;
-        }
-      });
+      // Start the shared daemon-status listener (Issue #1179, generalized in #1470).
+      // Drives daemonConnecting/daemonUnreachable above and fans out reconnect
+      // events to any store registered via onDaemonReconnect.
+      startDaemonStatusListener();
 
       // Show first-launch onboarding wizard if setup has not been completed (Issue #1180).
       invoke<{ completed: boolean }>('check_onboarding_status')
@@ -594,9 +600,6 @@
       if (unlistenIntegrations) {
         (await unlistenIntegrations)();
       }
-      if (unlistenDaemonStatus) {
-        (await unlistenDaemonStatus)();
-      }
       if (unlistenSkillCliMissing) {
         (await unlistenSkillCliMissing)();
       }
@@ -656,6 +659,10 @@
           >
             Retry
           </button>
+        </div>
+      {:else if daemonConnecting}
+        <div class="daemon-connecting-banner" role="status">
+          <span>Connecting to local service…</span>
         </div>
       {/if}
       <div
@@ -729,6 +736,18 @@
 
   .daemon-error-banner button:hover {
     background: hsl(var(--destructive) / 0.2);
+  }
+
+  .daemon-connecting-banner {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    padding: 0.5rem 1rem;
+    background: hsl(var(--muted) / 0.5);
+    border-bottom: 1px solid hsl(var(--border));
+    color: hsl(var(--muted-foreground));
+    font-size: 0.875rem;
+    z-index: 100;
   }
 
   /* Container for app-shell and status bar (flexbox column) */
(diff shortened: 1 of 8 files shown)


malibio:
### Code Review Summary

This is a well-scoped, net-positive fix for #1470. It correctly decouples `AppShell` rendering from daemon readiness, replaces a hardcoded 30s blocking wait with an immediate render + fast-appearing "connecting" affordance, and generalizes the existing ad-hoc `pro:tier-detected` retry pattern into a single shared `onDaemonReconnect` hook consumed by four independent call sites. The diff is atomic, stays within the issue's stated scope (including its "alternatives considered and rejected" boundary on splitting `DaemonStatus`), and is well-tested for the new service (9 tests) and retry wiring (2 tests). I verified `initializeApp()` has zero daemon/gRPC dependency, so removing it from the blocking chain is safe, and confirmed no other code between the old Step 1/Step 2 boundary assumed daemon readiness.

I traced the grace-period timer vs. async `listen()` registration race called out in the review brief and found it handled correctly — no bug there. The findings below are lower-severity: a minor `unlisten`-handle cleanup gap, an implicit-cleanup-contract fragility in `+layout.svelte`, and a confirmed-harmless double-fetch between the new `onDaemonReconnect` hook and the pre-existing `pro:tier-detected` listener. None of these block merge.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- **`packages/desktop-app/src/lib/services/daemon-status.ts:66-90` — the `unlisten` function returned by `listen()` is discarded.** `listen<string>('daemon-status', ...)` returns a promise resolving to an unsubscribe function; this PR never captures or calls it. In practice this is harmless today because `startDaemonStatusListener()` is a session-long singleton (guarded by `started`) that's expected to live for the app's lifetime, so there's nothing to clean up. But it's a latent footgun if this module is ever reused in a context with a shorter lifecycle (e.g., a future test harness or a hot-reload path), and it's inconsistent with the rest of the codebase's pattern of capturing `unlisten` handles (e.g., the old code this replaces in `app-shell.svelte` did capture and call `unlistenDaemonStatus`). Low cost to fix: store the promise and expose a `stopDaemonStatusListener()` for symmetry, or at minimum a comment noting the intentional session-lifetime assumption.

- **`packages/desktop-app/src/routes/+layout.svelte:82-89` — the `onMount(() => onDaemonReconnect(...))` pattern relies on an implicit contract that's easy to break.** Svelte's `onMount` treats a returned function as a destroy-time cleanup callback. Here, `onDaemonReconnect(callback)` itself returns the unsubscribe function, which is then implicitly returned by the outer arrow function — so this correctly unsubscribes on component destroy. This works, but it's a subtle "returns a function that returns a function" chain that a future reader/editor could easily break by adding a statement before the `onDaemonReconnect(...)` call (which would change the arrow function's implicit return to `undefined`, silently disabling cleanup — a leak, not a crash, so it wouldn't be caught by tests). Since `+layout.svelte` mounts exactly once per app session in practice, the leak risk is theoretical, but consider making the intent explicit:
  ```ts
  onMount(() => {
    const unsubscribe = onDaemonReconnect(() => {
      if (!schemaPluginsReady) loadSchemaPlugins();
      if (!syncListenersReady) loadSyncListeners();
    });
    return unsubscribe;
  });
  ```
  Same observations apply to the analogous pattern in `app-shell.svelte:36-45` (`$effect(() => { const unsubscribe = daemonStatus.subscribe(...); return unsubscribe; })`), which is already written the explicit way — worth matching that style in `+layout.svelte` for consistency.

- **`pro:tier-detected` listener (`app-shell.svelte:279-286`) vs. the new `onDaemonReconnect` hook — confirmed harmless double-fetch, not a bug, but worth a short-lived TODO or issue reference.** Both `schemasData.loadSchemas()` and `collectionsData.loadCollections()` can now be triggered twice in quick succession after a cold start: once from `onDaemonReconnect` (fires on the first `healthy` daemon-status event) and again from `pro:tier-detected` (fires once gRPC/Pro-tier detection completes, which per the inline comment "can complete after the initial healthy event"). Both `loadSchemas`/`loadCollections` are simple full-replace fetches with no visible request-cancellation or in-flight de-duping (confirmed by reading `schemas.ts`/`collections.ts`), so the worst case is one extra redundant network round-trip per cold start, not a correctness bug (last-write-wins on two structurally-identical responses). Given the PR is already explicitly framed as a generalization of this exact ad-hoc pattern, consider a fast-follow to fold `pro:tier-detected` into `onDaemonReconnect` and delete the now-partially-redundant listener — but this is not a merge blocker.

- **`base-node-viewer.svelte:293-372` retry-on-reconnect correctly avoids duplicate loads via the `loadFailed` flag and unsubscribes in the `onMount` cleanup — confirmed no leak.** I traced this against the file's own `{#key}` recreation comment (line 256-257: "This component is recreated via `{#key}` when nodeId changes... eliminating the need for a reactive effect") and confirmed: (1) `currentNodeId` is captured as a `const` at `onMount` time specifically to avoid a stale-closure bug against the mutable `nodeId` prop — good catch by the author; (2) the returned cleanup `() => { unsubscribeReconnect(); scrollCleanup?.(); }` correctly unsubscribes on both normal unmount and `{#key}`-driven recreation; (3) `loadFailed` is scoped per-`onMount`-invocation (a fresh closure per key-recreation), so there's no cross-instance leakage; (4) the `isDestroyed` check inside the reconnect callback guards against a late-firing reconnect racing a just-destroyed component. This is solid, idiomatic Svelte 5 and the flagged risk areas from the review brief (duplicate loads, memory leaks, stale closures) are all handled correctly.

#### Nitpicks

- **Nit:** `daemon-status.ts:62-90` — `clearTimeout(graceTimer)` only happens inside the `listen()` event callback. If `daemon-status` is never emitted at all during a session (a Rust-side regression, distinct from a proper `not_running` event), the grace timer still fires once and correctly clears `connecting`, but `unreachable` stays `false` forever — so the UI silently shows neither the "connecting" nor the "unreachable" banner. This matches the issue's scope (it only asks the frontend to react to events Rust already emits) and isn't a regression from the prior behavior, but it's a gap worth knowing about if daemon-status emission itself ever regresses.
- **Nit:** `packages/desktop-app/src/routes/+layout.svelte:12-13` — `initializeApp()` is now the sole gate before `isInitialized = true`, but the file no longer imports `isTauri`/`listen` directly (moved into `daemon-status.ts`), which is good cleanup — no action needed, just noting the diff is clean here.
- **Nit:** `daemon-status.ts:39` — `CONNECTING_GRACE_PERIOD_MS = 1500` is a good, well-justified constant (issue's recommendation was "~1.5s"), but consider whether it should be tunable via env/config for QA/debugging slow-daemon scenarios. Not necessary for merge.
- **Nit:** The `schemas.ts`/`collections.ts` `onDaemonReconnect(...)` registration happens unconditionally at module load (top-level side effect on import), which is a slightly unusual pattern (most of the codebase favors explicit initialization functions called from `onMount`). It works correctly here because ES modules are singletons and Svelte/Vite's module graph guarantees single evaluation, but a comment noting "this runs once per app session at import time, not per-component-mount" would help a future reader who expects store initialization to happen in a component lifecycle hook.

### On the explicitly-flagged scope decisions

- **Skipping the `logger.ts` `[object Object]` fix (issue item 4):** Reasonable. The issue itself frames this as "worth fixing but small enough to not warrant its own issue" and folds it in opportunistically "since the fix will touch several of these call sites anyway" — it does not treat it as a hard requirement. Given the existing test suite (`logger.test.ts`) encodes the current pass-through behavior as intentional, and fixing it broke 3 tests, treating this as a separate, deliberately-scoped-out follow-up is the right call under the "net positive over perfection" principle — forcing a test-suite semantics change into an already-substantial PR would reduce reviewability for marginal benefit. Agree with narrowing scope here.
- **Not splitting `DaemonStatus` into `SocketReachable`/`ModelLoaded`:** Correct — the issue explicitly rejects this as out of scope for this fix, and the PR doesn't attempt it. No concern.

### Test Quality

The new tests are well-structured: they mock `@tauri-apps/api/event`'s `listen()` to capture and manually drive the handler, which is the right approach for testing event-driven state machines without a real Tauri runtime. Coverage includes the important edge cases: dedup on consecutive `healthy` events, re-firing after a `healthy → not_running → healthy` cycle, unsubscribe correctness, and browser-mode (non-Tauri) no-op behavior. One gap: no test exercises the grace-period timer actually firing (all tests emit an event before/without asserting the 1.5s-elapsed-with-no-event path via `vi.useFakeTimers()` + `vi.advanceTimersByTime(1500)`) — worth adding for completeness, though the logic is simple enough that this is a coverage nit, not a correctness concern given code review confirms the timer path is straightforward.

### Verdict

Net positive, well-tested, correctly scoped. Approve. All findings above are improvement-level or nits — fast-follow-friendly, none block merge. The `+layout.svelte` `onMount` cleanup-chain fragility (Suggested Improvements, item 2) is the one item worth a quick look before merge since it's a one-line change eliminating a subtle footgun, but it's not a blocker.

malibio (commented):
Pragmatic code review verdict: **Net positive, well-tested, correctly scoped — Approve (no blockers).** Full review posted as a separate PR comment above. (Note: could not submit a formal GitHub 'Approve' review since I'm authenticated as the PR author — GitHub disallows self-approval. Recording the verdict here as a review comment instead.)

malibio:
### Addressed review feedback

**✅ Addressed (4 items):**
1. `+layout.svelte:82-89` implicit `onMount` cleanup chain → made explicit with a named `unsubscribe`, matching the style already used in `app-shell.svelte`. (commit `e946307d`)
2. `schemas.ts` / `collections.ts` module-load-time `onDaemonReconnect` registration → added a clarifying comment noting this runs once at import time (ES module singleton), not per component mount. (commit `3adf15cf`)
3. Missing grace-period-timer test coverage → added a `vi.useFakeTimers()` test asserting `connecting` clears via the 1.5s timer when no `daemon-status` event ever arrives. (commit `3adf15cf`)

**⏭️ Skipped (all confirmed non-blocking by the review):**
- **`daemon-status.ts` discarded `unlisten` handle** — the service is an intentional session-lifetime singleton (guarded by `started`), so there's nothing to tear down today. Adding a `stopDaemonStatusListener()` would be speculative API surface for a caller that doesn't exist yet.
- **Folding `pro:tier-detected` into `onDaemonReconnect`** — review explicitly frames this as a fast-follow outside #1470's scope. The current double-fetch on cold start is a harmless redundant network round-trip (last-write-wins on structurally identical responses), not a correctness bug.
- **Tunable `CONNECTING_GRACE_PERIOD_MS` via env/config** — no current need; would be premature configuration surface.
- **`logger.ts` `[object Object]` papercut (issue item 4)** — already scoped out before this review pass; conflicts with existing test expectations that encode current pass-through behavior as intentional. Review agreed this was reasonable.

**Tests:** 147 files / 3939 tests passing (baseline was 145/3927; net +2 files / +12 tests, zero regressions). `quality:fix` (eslint + svelte-check): 0 errors, 0 warnings.

## Re-Review Decision

**Rationale:** All changes in this pass are low-risk clarifications and test additions — no logic changes, no new functionality, no architectural changes. The one behavioral change (explicit `onMount` cleanup) is functionally identical to the prior implicit-chain version, just less fragile to future edits. The skipped items were all explicitly confirmed non-blocking by the original review. This PR is ready for merge pending final approval.

PR #1480, opened 2026-07-02
Centralize viewer node hydration in pane-content to fix ai-chat messages not showing on tab reopen

malibio:
Closes #1479

### DIFF
--- a/packages/desktop-app/src/lib/components/layout/pane-content.svelte
+++ b/packages/desktop-app/src/lib/components/layout/pane-content.svelte
@@ -3,6 +3,7 @@
   import BaseNodeViewer from '$lib/design/components/base-node-viewer.svelte';
   import { tabState, updateTabTitle, updateTabContent, closeTab } from '$lib/stores/navigation.js';
   import { pluginRegistry } from '$lib/plugins/plugin-registry';
+  import { sharedNodeStore } from '$lib/services/shared-node-store.svelte';
   import type { Pane } from '$lib/stores/navigation.js';
   import { createLogger } from '$lib/utils/logger';
   import SettingsPane from '$lib/components/settings/settings-pane.svelte';
@@ -29,6 +30,10 @@
   let viewerComponents = $state<Map<string, unknown>>(new Map());
   let viewerLoadErrors = $state<Map<string, string>>(new Map());
   let viewerLoading = $state<Set<string>>(new Set());
+
+  // Track node hydration state per nodeId. A node is hydrated once ensureNode()
+  // has confirmed it is present in sharedNodeStore (or closed the tab if not found).
+  let hydratedNodeIds = $state<Set<string>>(new Set());
 
   // Load viewer when needed - moved to function called from onMount to avoid derived context issues
   async function loadViewerForNodeType(nodeType: string) {
@@ -63,6 +68,19 @@
     }
   }
 
+  async function hydrateNode(nodeId: string, tabId: string) {
+    if (hydratedNodeIds.has(nodeId)) return;
+
+    const node = await sharedNodeStore.ensureNode(nodeId);
+    if (!node) {
+      log.warn(`Node ${nodeId} not found — closing stale tab`);
+      closeTab(tabId);
+      return;
+    }
+
+    hydratedNodeIds = new Set(hydratedNodeIds).add(nodeId);
+  }
+
   // Derive viewer component for active tab.
   // Returns null while the viewer module is still loading (prevents BaseNodeViewer fallback
   // from rendering with an incompatible nodeId, e.g. a schema id passed to QueryNodeViewer)
@@ -82,13 +100,23 @@
     return viewerLoading.has(nodeType);
   });
 
-  // Load viewer when active tab changes - use $effect but call async function
-  // untrack the call to loadViewerForNodeType so mutations to viewerLoading/viewerComponents
-  // inside that function don't re-trigger this effect
+  const isNodeHydrated = $derived.by(() => {
+    const nodeId = activeTab?.content?.nodeId;
+    if (!nodeId) return true; // settings tabs and placeholder tabs need no hydration
+    return hydratedNodeIds.has(nodeId);
+  });
+
+  // Load viewer and hydrate parent node when active tab changes.
+  // Both run in parallel — viewer module load and node fetch are independent.
   $effect(() => {
     const nodeType = activeTab?.content?.nodeType;
+    const nodeId = activeTab?.content?.nodeId;
+    const tabId = activeTabId;
     if (nodeType) {
       untrack(() => loadViewerForNodeType(nodeType));
+    }
+    if (nodeId && tabId) {
+      untrack(() => hydrateNode(nodeId, tabId));
     }
   });
 
@@ -108,8 +136,8 @@
       <p class="error-message">{loadError}</p>
       <p class="help-text">Try refreshing the page or contact support if the problem persists.</p>
     </div>
-  {:else if isViewerLoading}
-    <!-- Viewer module still loading — don't render BaseNodeViewer as fallback -->
+  {:else if isViewerLoading || !isNodeHydrated}
+    <!-- Viewer module still loading or parent node not yet hydrated -->
     <div class="loading-state">
       <span>Loading...</span>
     </div>
--- a/packages/desktop-app/src/lib/components/layout/pane-content.svelte
+++ b/packages/desktop-app/src/lib/components/layout/pane-content.svelte
@@ -108,6 +108,9 @@
 
   // Load viewer and hydrate parent node when active tab changes.
   // Both run in parallel — viewer module load and node fetch are independent.
+  // tabId is captured synchronously before untrack() so that if the user switches
+  // tabs during the async fetch, closeTab targets the tab that triggered the fetch,
+  // not the newly-active one.
   $effect(() => {
     const nodeType = activeTab?.content?.nodeType;
     const nodeId = activeTab?.content?.nodeId;
--- a/packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte
@@ -272,19 +272,8 @@
   onMount(async () => {
     log.debug('AiChatNodeViewer mounted', { nodeId });
 
-    let hydrationSucceeded = false;
     try {
-      // ai-chat nodes are always in the store by the time the viewer mounts —
-      // base-node-viewer's loadChildrenForParent puts them there with the current
-      // DB version. A separate re-fetch here would race with WatchNodes updates
-      // and overwrite the store with a stale version, causing OCC conflicts on send.
-      if (!sharedNodeStore.getNode(nodeId)) {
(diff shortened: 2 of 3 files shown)


malibio:
## Code Review — PR #1480 (Issue #1479)

### Summary

This PR correctly identifies and fixes a real architectural problem: node hydration was an accidental side-effect of `BaseNodeViewer.loadChildrenTree`, making `AiChatNodeViewer` — the only standalone viewer — the odd one out. Centralizing hydration in `pane-content.svelte` is the right fix and the design of `ensureNode` (cache-first, single responsibility, clear contract) is clean.

The implementation is largely well-executed, but there is one **blocker** involving brand-new date nodes that must be addressed before merge.

---

### Critical Issues

#### [Critical/Blocker] `ensureNode` closes brand-new (never-persisted) date node tabs

**File:** `packages/desktop-app/src/lib/services/shared-node-store.svelte.ts` — `ensureNode`, line 866
**File:** `packages/desktop-app/src/lib/components/layout/pane-content.svelte` — `hydrateNode`, lines 71-82

`ensureNode` calls `backendAdapter.getNode(nodeId)` when the node is not in the store. For a brand-new date node that has never been persisted (e.g., the Daily Journal on first startup of a new day), the backend returns `null` / `NotFound`. `ensureNode` therefore returns `undefined`, and `hydrateNode` calls `closeTab(tabId)` — silently closing the Daily Journal tab before the user can interact with it.

`doLoadChildrenTree` already handles this case at line 2186–2200: it recognizes `isValidDateId(parentId)` when the backend returns no tree, and synthesizes a virtual in-memory `date` node. That synthesis happens in `loadChildrenTree`, which is called by `BaseNodeViewer` *after* `pane-content` has already fired `hydrateNode` and closed the tab.

The fix is straightforward — mirror the virtual-node synthesis inside `ensureNode`:

```ts
async ensureNode(nodeId: string): Promise<Node | undefined> {
  const cached = this.nodes.get(nodeId);
  if (cached) return cached;

  const fetched = await backendAdapter.getNode(nodeId);
  if (fetched) {
    this.setNode(fetched, { type: 'database', reason: 'ensure-node' });
    return fetched;
  }

  // Brand-new date nodes are created lazily in the backend on first child write.
  // Synthesize a virtual in-memory node so pane-content does not mistake the
  // missing backend record for a deleted/stale node (matches doLoadChildrenTree's
  // virtual-date synthesis at line 2186).
  if (isValidDateId(nodeId)) {
    const now = new Date().toISOString();
    const virtualNode: Node = {
      id: nodeId,
      nodeType: 'date',
      content: '',
      version: 0,
      createdAt: now,
      modifiedAt: now,
      properties: {}
    };
    this.setNode(virtualNode, { type: 'database', reason: 'virtual-date-node' });
    return virtualNode;
  }

  return undefined;
}
```

This also surfaces a testing gap: `ensureNode` has zero unit tests. Given that it gates every viewer mount, at minimum three cases should be covered: cache hit, cache miss (backend returns node), and backend returns null for a non-date node. The date-node virtual synthesis path deserves its own test.

---

### Suggested Improvements

#### [Improvement] `hydratedNodeIds` grows unboundedly across tab lifetime

**File:** `packages/desktop-app/src/lib/components/layout/pane-content.svelte`, line 36

`hydratedNodeIds` is a `Set<string>` that accumulates nodeIds over the component's lifetime but is never pruned. This is only a concern for very long-lived pane instances with many unique nodes visited (it's a Set of strings, so footprint is small), but it also means a node that gets deleted while the user is away will correctly fire `closeTab` on re-open — but if the user opens a different tab, navigates back, and the tab was already closed, no harm is done. The concern is more subtle: if a tab is closed (removing it from the tab list) but the `nodeId` stays in `hydratedNodeIds`, a future tab for the same nodeId (re-opened node) skips hydration and the `ensureNode` cache-first path is never consulted. Whether the store already has a fresh copy from WatchNodes makes this benign in practice, but explicit cleanup on tab close would make the contract cleaner. Not a blocker.

#### [Improvement] `hydrateNode` captures `tabId` at call time, not at resolution time

**File:** `packages/desktop-app/src/lib/components/layout/pane-content.svelte`, lines 71-82

The `$effect` captures `tabId` from `activeTabId` before calling `untrack(() => hydrateNode(nodeId, tabId))`. If the user switches tabs while `ensureNode` is awaiting the backend, `activeTabId` will have changed by the time `closeTab(tabId)` fires — but `tabId` in the closure is the stale tab, which is correct behavior (close the tab that failed, not the currently active one). The implementation is actually correct here; it's worth a brief comment documenting why `tabId` is captured before the `untrack` call to prevent future readers from "fixing" it:

```ts
// Capture tabId before the async gap so a tab switch during fetch
// closes the correct (stale) tab, not the newly-active one.
const tabId = activeTabId;
```

(This capture already happens at line 114 in the `$effect` — the comment would just clarify intent.)

#### [Improvement] `ViewerComponent` default falls back to `'text'` nodeType before hydration

**File:** `packages/desktop-app/src/lib/components/layout/pane-content.svelte`, line 88

```ts
const nodeType = activeTab?.content?.nodeType ?? 'text';
```

This fallback fires while `activeTab?.content` is defined but `nodeType` is missing. Given that all tabs in the navigation store carry explicit `nodeType`, this is likely dead code. If it ever triggers, the viewer resolves to the `text` viewer before hydration completes, then the hydration loading gate (`!isNodeHydrated`) hides the component — so there is no visible rendering artifact. Worth noting as a nit but not a real issue.

---

### Nitpicks

- **Nit:** `packages/desktop-app/src/lib/services/shared-node-store.svelte.ts` line 867: `ensureNode` uses `this.nodes.get(nodeId)` directly (bypasses `nodesVersion`), whereas `getNode` reads `nodesVersion` first to register a reactive dependency (line 776). For an async method that isn't called from a reactive context, this is correct — but worth a comment clarifying it's intentional that the reactive dependency is not registered here.

- **Nit:** `packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte` line 277: The new comment `// pane-content guarantees the node is in sharedNodeStore before mounting this viewer.` is correct and helpful. Consider also removing the now-dead `sharedNodeStore` import if it's no longer referenced in the removed lines. (Quick check: it's still used on line 277 via `sharedNodeStore.getNode(nodeId)`, and elsewhere in the file — no import change needed.)

- **Nit:** `packages/desktop-app/src/lib/components/layout/pane-content.svelte` line 139: The comment `<!-- Viewer module still loading or parent node not yet hydrated -->` is accurate and clear. No change needed.

---

### Requirements Validation

|---|---|
| `sharedNodeStore.ensureNode(nodeId)` added: cache-first, backend fetch on miss | DONE |
| `pane-content` awaits `ensureNode` before rendering `<ViewerComponent>` | DONE |
| If `ensureNode` returns `undefined`, `pane-content` calls `closeTab` | DONE |
| `AiChatNodeViewer` bail-out guard removed (lines 281-284) | DONE |
| Late-arrival `$effect` removed (lines 352-356) | DONE |
| Stale comment removed (lines 277-280) | DONE |
| Reopening ai-chat tab shows persisted messages | DONE (by contract: node in store before mount) |
| Task and date node tabs still hydrate correctly | **BLOCKED** — brand-new date nodes regress |
| No double-fetch for `BaseNodeViewer`-wrapped viewers | DONE — `doLoadChildrenTree` line 2220 guard prevents it |
| `bun run test:all` passes | PASSES (147 test files, 3939 tests) — but date regression not covered by existing tests |
| `bun run quality:fix` clean | Not verified in review; assume CI covers this |

---

### Verdict

**Request Changes.** One blocker: `ensureNode` must handle virtual date nodes before it can safely gate all viewer mounts. The fix is small (mirror the existing virtual-node synthesis from `doLoadChildrenTree`). Once that is in place, this is a clean, well-scoped refactor that eliminates a latent trap for all future leaf-node viewers.

malibio:
## Re-Review — PR #1480 (Issue #1479)

**Re-reviewing commit:** 103028ff (Fix ensureNode: synthesize virtual date node on backend miss)
**Comparing against:** 5de16889 (initial implementation)

---

### Summary

The critical blocker from the initial review is **correctly fixed**. The `tabId` comment was added. No new issues were introduced by the fix. The PR is ready to merge.

---

### Blocker Resolution

#### [Resolved] Virtual date node synthesis in `ensureNode`

**File:** `packages/desktop-app/src/lib/services/shared-node-store.svelte.ts`, lines 882–896

The fix mirrors `doLoadChildrenTree` exactly — same `version: 0` sentinel, same `{ type: 'database', reason: 'virtual-date-node' }` source tag to prevent `determinePersistenceBehavior` from triggering an unwanted write, and uses the already-imported `isValidDateId`. The JSDoc was also updated to explain the special case, including a cross-reference to `doLoadChildrenTree`. This is the correct implementation.

The comment on the `database` source choice (`// database source prevents determinePersistenceBehavior from triggering an unwanted write`) is precise and useful — it explains a non-obvious invariant that future readers would otherwise need to trace across the store.

#### [Resolved] `tabId` capture comment in `pane-content.svelte`

**File:** `packages/desktop-app/src/lib/components/layout/pane-content.svelte`, lines 111–113

The comment was added exactly as suggested:

```
// tabId is captured synchronously before untrack() so that if the user switches
// tabs during the async fetch, closeTab targets the tab that triggered the fetch,
// not the newly-active one.
```

Clear and accurate.

---

### Remaining Open Items (Non-Blocking)

These carry over from the initial review; neither was introduced by the fix commit and neither blocks merge.

**[Improvement] `hydratedNodeIds` grows without pruning** — `packages/desktop-app/src/lib/components/layout/pane-content.svelte`, line 36. Still unbounded across the pane's lifetime. Low practical risk (it's a Set of strings), but if a nodeId is evicted from the set on tab close, a re-opened deleted node would re-trigger `ensureNode` and correctly fire `closeTab` rather than serving a stale cache hit. Not a blocker; worth a follow-up issue.

**[Improvement] Zero unit tests for `ensureNode`** — the function now gates every viewer mount and has three distinct code paths: cache hit, backend hit, and date-node virtual synthesis. No tests were added in either commit. Given the criticality of the invariant this function enforces, coverage would be worthwhile. Not a blocker for this PR, but should be tracked.

---

### Requirements Validation

|---|---|
| `sharedNodeStore.ensureNode(nodeId)` added: cache-first, backend fetch on miss | DONE |
| `pane-content.svelte` awaits `ensureNode` before rendering `<ViewerComponent>` | DONE |
| If `ensureNode` returns `undefined`, `pane-content` calls `closeTab` | DONE |
| `AiChatNodeViewer` bail-out guard removed (lines 281-284) | DONE |
| Late-arrival `$effect` removed (lines 352-356) | DONE |
| Stale comment removed (lines 277-280) | DONE |
| Reopening ai-chat tab shows persisted messages | DONE (by invariant) |
| Task and date node tabs still hydrate correctly — no regression for brand-new date nodes | DONE (fixed in re-commit) |
| No double-fetch for `BaseNodeViewer`-wrapped viewers | DONE — `doLoadChildrenTree` guard at line 2235 skips parent re-fetch |
| `bun run test:all` passes | Previously confirmed (3939 tests) |
| `bun run quality:fix` clean | Not independently verified in re-review; assumed clean |

---

### Verdict

**Approved.** The blocker is resolved, the implementation is clean, and the change is a net improvement. The two remaining open items are low-urgency improvements that belong in follow-up issues, not in this PR.

PR #1481, opened 2026-07-02
Model download status stuck at 'Downloading' 100% with lingering Cancel button after daemon completes

malibio:
Closes #1471

### DIFF
--- a/packages/agent/src/local_agent/composite_model_manager.rs
+++ b/packages/agent/src/local_agent/composite_model_manager.rs
@@ -72,26 +72,36 @@ impl CompositeModelManager {
         self.ollama.is_available().await
     }
 
-    /// Set the progress callback for GGUF downloads.
-    pub async fn set_gguf_progress_callback(&self, cb: Box<dyn Fn(DownloadEvent) + Send + Sync>) {
-        self.gguf.set_progress_callback(cb).await;
+    /// Set the progress callback for GGUF downloads of a specific model.
+    pub async fn set_gguf_progress_callback(
+        &self,
+        model_id: &str,
+        cb: Box<dyn Fn(DownloadEvent) + Send + Sync>,
+    ) {
+        self.gguf.set_progress_callback(model_id, cb).await;
     }
 
-    /// Set the progress callback for Ollama downloads.
-    pub async fn set_ollama_progress_callback(&self, cb: Box<dyn Fn(DownloadEvent) + Send + Sync>) {
-        self.ollama.set_progress_callback(cb).await;
+    /// Set the progress callback for Ollama downloads of a specific model.
+    pub async fn set_ollama_progress_callback(
+        &self,
+        model_id: &str,
+        cb: Box<dyn Fn(DownloadEvent) + Send + Sync>,
+    ) {
+        self.ollama.set_progress_callback(model_id, cb).await;
     }
 
-    /// Clear the GGUF progress callback, dropping any resources (e.g. channel
-    /// senders) it holds.
-    pub async fn clear_gguf_progress_callback(&self) {
-        self.gguf.clear_progress_callback().await;
+    /// Clear the GGUF progress callback for a specific model, dropping any
+    /// resources (e.g. channel senders) it holds. Only affects that model's
+    /// callback — concurrent downloads of other models are unaffected.
+    pub async fn clear_gguf_progress_callback(&self, model_id: &str) {
+        self.gguf.clear_progress_callback(model_id).await;
     }
 
-    /// Clear the Ollama progress callback, dropping any resources (e.g.
-    /// channel senders) it holds.
-    pub async fn clear_ollama_progress_callback(&self) {
-        self.ollama.clear_progress_callback().await;
+    /// Clear the Ollama progress callback for a specific model, dropping any
+    /// resources (e.g. channel senders) it holds. Only affects that model's
+    /// callback — concurrent downloads of other models are unaffected.
+    pub async fn clear_ollama_progress_callback(&self, model_id: &str) {
+        self.ollama.clear_progress_callback(model_id).await;
     }
 }
 
--- a/packages/agent/src/local_agent/composite_model_manager.rs
+++ b/packages/agent/src/local_agent/composite_model_manager.rs
@@ -82,12 +82,19 @@ impl CompositeModelManager {
     }
 
     /// Set the progress callback for Ollama downloads of a specific model.
+    ///
+    /// `model_id` may carry the `"ollama:"` prefix (as used by the frontend
+    /// and `list()`) — it's stripped here to match the unprefixed id that
+    /// [`Self::download`] passes to the underlying `OllamaModelManager`,
+    /// which is what progress events are actually keyed and looked up by.
     pub async fn set_ollama_progress_callback(
         &self,
         model_id: &str,
         cb: Box<dyn Fn(DownloadEvent) + Send + Sync>,
     ) {
-        self.ollama.set_progress_callback(model_id, cb).await;
+        self.ollama
+            .set_progress_callback(Self::strip_ollama_prefix(model_id), cb)
+            .await;
     }
 
     /// Clear the GGUF progress callback for a specific model, dropping any
@@ -100,8 +107,13 @@ impl CompositeModelManager {
     /// Clear the Ollama progress callback for a specific model, dropping any
     /// resources (e.g. channel senders) it holds. Only affects that model's
     /// callback — concurrent downloads of other models are unaffected.
+    ///
+    /// `model_id` is de-prefixed for the same reason as
+    /// [`Self::set_ollama_progress_callback`].
     pub async fn clear_ollama_progress_callback(&self, model_id: &str) {
-        self.ollama.clear_progress_callback(model_id).await;
+        self.ollama
+            .clear_progress_callback(Self::strip_ollama_prefix(model_id))
+            .await;
     }
 }
 
@@ -233,5 +245,69 @@ mod tests {
             CompositeModelManager::add_ollama_prefix("llama3.2:3b"),
             "ollama:llama3.2:3b"
         );
+    }
+
+    // -- Ollama progress callback de-prefixing (issue #1471 re-review) ------
+    //
+    // `download()` strips the "ollama:" prefix before delegating to the
+    // underlying OllamaModelManager, which keys its progress callbacks and
+    // fires progress events by that same unprefixed id. If
+    // set/clear_ollama_progress_callback didn't also strip the prefix, a
+    // callback registered via the composite manager (with the prefixed id,
+    // as the daemon does) would never be found when progress actually fires,
+    // silently dropping every progress tick for Ollama downloads.
+
+    fn test_composite_manager() -> (CompositeModelManager, tempfile::TempDir) {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let gguf = Arc::new(GgufModelManager::with_dir(tmp.path().to_path_buf()).unwrap());
+        let ollama = Arc::new(OllamaModelManager::with_base_url(
+            "http://127.0.0.1:19999".to_string(),
+        ));
+        (CompositeModelManager::new(gguf, ollama), tmp)
+    }
+
+    #[tokio::test]
+    async fn ollama_progress_callback_set_with_prefixed_id_is_stored_under_stripped_key() {
+        let (mgr, _tmp) = test_composite_manager();
+        let (tx, rx) = tokio::sync::mpsc::channel::<()>(1);
+
+        // Register via the composite manager using the prefixed id, exactly
+        // as the daemon does in `local_agent_service.rs`.
+        mgr.set_ollama_progress_callback(
+            "ollama:llama3.2:3b",
+            Box::new(move |_evt| {
+                let _ = tx.try_send(());
+            }),
+        )
+        .await;
+
+        // Clearing via the *unprefixed* id (mirroring how OllamaModelManager
+        // itself would look it up during `fire_progress`, keyed by the
+        // unprefixed id `download()` uses) must find and drop it.
+        mgr.ollama_manager()
+            .clear_progress_callback("llama3.2:3b")
+            .await;
+        assert!(rx.is_closed());
+    }
+
+    #[tokio::test]
+    async fn clear_ollama_progress_callback_accepts_prefixed_id() {
+        let (mgr, _tmp) = test_composite_manager();
+        let (tx, rx) = tokio::sync::mpsc::channel::<()>(1);
+
+        mgr.set_ollama_progress_callback(
+            "ollama:llama3.2:3b",
+            Box::new(move |_evt| {
+                let _ = tx.try_send(());
+            }),
+        )
+        .await;
+        assert!(!rx.is_closed());
+
+        // Clearing through the composite manager with the *same prefixed*
+        // id it was registered with must also successfully remove it.
+        mgr.clear_ollama_progress_callback("ollama:llama3.2:3b")
+            .await;
+        assert!(rx.is_closed());
     }
 }
--- a/packages/agent/src/local_agent/composite_model_manager.rs
+++ b/packages/agent/src/local_agent/composite_model_manager.rs
@@ -81,6 +81,18 @@ impl CompositeModelManager {
     pub async fn set_ollama_progress_callback(&self, cb: Box<dyn Fn(DownloadEvent) + Send + Sync>) {
         self.ollama.set_progress_callback(cb).await;
     }
+
+    /// Clear the GGUF progress callback, dropping any resources (e.g. channel
+    /// senders) it holds.
+    pub async fn clear_gguf_progress_callback(&self) {
+        self.gguf.clear_progress_callback().await;
+    }
+
+    /// Clear the Ollama progress callback, dropping any resources (e.g.
+    /// channel senders) it holds.
+    pub async fn clear_ollama_progress_callback(&self) {
+        self.ollama.clear_progress_callback().await;
+    }
 }
 
 #[async_trait]
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -26,8 +26,11 @@ use crate::agent_types::{
 // Type aliases
 // ---------------------------------------------------------------------------
 
-/// Shared, thread-safe progress callback for download events.
-type ProgressCallback = Arc<RwLock<Option<Box<dyn Fn(DownloadEvent) + Send + Sync>>>>;
+/// Shared, thread-safe progress callbacks for download events, keyed by
+/// model_id so concurrent downloads of different models don't clobber each
+/// other's callback (registering or clearing one model's callback must not
+/// affect another's).
+type ProgressCallback = Arc<RwLock<HashMap<String, Box<dyn Fn(DownloadEvent) + Send + Sync>>>>;
 
 // ---------------------------------------------------------------------------
 // Catalog constants
@@ -437,22 +440,26 @@ impl GgufModelManager {
             statuses: Arc::new(RwLock::new(initial_statuses)),
             active_downloads: Arc::new(RwLock::new(HashMap::new())),
             http_client: reqwest::Client::new(),
-            on_progress: Arc::new(RwLock::new(None)),
+            on_progress: Arc::new(RwLock::new(HashMap::new())),
             loaded_model_id: Arc::new(RwLock::new(None)),
         })
     }
 
-    /// Register a progress callback for download events.
-    pub async fn set_progress_callback(&self, callback: Box<dyn Fn(DownloadEvent) + Send + Sync>) {
+    /// Register a progress callback for download events for a specific model.
+    pub async fn set_progress_callback(
+        &self,
+        model_id: &str,
+        callback: Box<dyn Fn(DownloadEvent) + Send + Sync>,
+    ) {
         let mut guard = self.on_progress.write().await;
-        *guard = Some(callback);
+        guard.insert(model_id.to_string(), callback);
     }
 
-    /// Clear the registered progress callback, dropping any resources
-    /// (e.g. channel senders) it holds.
-    pub async fn clear_progress_callback(&self) {
+    /// Clear the registered progress callback for a specific model, dropping
+    /// any resources (e.g. channel senders) it holds.
+    pub async fn clear_progress_callback(&self, model_id: &str) {
         let mut guard = self.on_progress.write().await;
-        *guard = None;
+        guard.remove(model_id);
     }
 
     /// Get the recommended model based on system RAM.
@@ -974,7 +981,7 @@ async fn perform_download(params: DownloadParams) -> Result<(), ModelError> {
                             // Fire progress callback
                             {
                                 let guard = on_progress.read().await;
-                                if let Some(cb) = guard.as_ref() {
+                                if let Some(cb) = guard.get(&model_id) {
                                     cb(DownloadEvent {
                                         model_id: model_id.clone(),
                                         bytes_downloaded,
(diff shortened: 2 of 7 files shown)


malibio:
### Code Review Summary

This is a well-targeted fix for the reported symptom, backed by tests that genuinely exercise the failure mode rather than the happy path. The root cause diagnosis (stale `Sender` clone captured in the progress callback keeping the gRPC channel open forever) is correct and confirmed here independently by tracing `on_progress` field types and the callback dispatch code.

However, the fix introduces (or rather, sharply worsens) a real concurrency hazard: `on_progress` is a single unkeyed slot per manager instance, shared across *all* models, while the `active_downloads`/status tracking is explicitly keyed per-model-id and the `ModelError::DownloadInProgress` guard only blocks concurrent downloads of the *same* model. That means concurrent downloads of two different models are an explicitly supported case that this change did not account for. This should be resolved or explicitly scoped as a known follow-up before merge — see Critical Issues.

The frontend change is solid: it correctly re-resolves `idx` via `findIndex` inside each event callback (not a captured stale index), correctly leaves the error path unaffected, and correctly cleans up both listeners in the `finally` block. Test coverage for the frontend is good — the new tests genuinely simulate a permanently-hung `chatModelDownload()` promise, which is the actual bug, not a strawman.

Verified locally: `cargo clippy -p nodespace-agent -p nodespace-daemon --lib -- -D warnings` clean, `cargo fmt --check` clean, new Rust tests pass (`clear_progress_callback_drops_previously_set_callback`, `clear_progress_callback_is_idempotent_when_unset`), and the full `model-store.test.ts` suite passes (27/27, including the 2 new tests).

### Findings

#### Critical Issues

- **`packages/daemon/src/services/local_agent_service.rs:792-800` — clearing progress callbacks is not safe under concurrent downloads of different models.** `CompositeModelManager` holds exactly one `GgufModelManager` and one `OllamaModelManager` for the whole daemon process, each with a single `on_progress: Arc<RwLock<Option<Box<dyn Fn>>>>` slot (`packages/agent/src/local_agent/model_manager.rs:394`, `ollama_model_manager.rs:72`). `set_gguf_progress_callback`/`set_ollama_progress_callback` are called unconditionally at the top of every `download_model` RPC (lines 742-768), and `ModelError::DownloadInProgress` (`model_manager.rs:594`) is keyed **per model_id**, not globally — i.e., downloading model A and model B concurrently is explicitly allowed by the existing design (`active_downloads: HashMap<String, ActiveDownload>`). Sequence: download A starts, sets callback₁; download B starts before A finishes, sets callback₂ (silently overwriting callback₁ — a pre-existing bug, progress events for A stop firing); A finishes and its cleanup task now calls `clear_gguf_progress_callback()`, wiping out callback₂ that B is actively depending on — B's progress events now silently stop firing for the remainder of its download (no crash, but a regression of the exact "stuck at 100%"-adjacent UX class this PR is fixing, just via progress instead of completion). Please either: (a) scope `on_progress` per-download (e.g. a `HashMap<String, Callback>` keyed by model_id, mirroring `active_downloads`), or (b) confirm/document that concurrent downloads of different models are not actually reachable from the UI today (e.g., the Settings screen only allows one active download button-click at a time) and file a fast-follow issue, explicitly flagging this as a known, scoped limitation rather than silently shipping a latent multi-download regression. Given "Net Positive > Perfection," if (b) is genuinely true today, this can go in as an [Improvement]-tracked follow-up rather than a blocker — but that determination needs to be made and stated, not left implicit.

#### Suggested Improvements

- **`packages/daemon/src/services/local_agent_service.rs:792`** — `drop(tx)` immediately before `manager.clear_gguf_progress_callback().await` is technically redundant on its own (the compiler would drop `tx` at the end of the async block regardless), but it is not misleading here — it's doing real, load-bearing work as an explicit statement of intent: it forces `tx`'s refcount down to just the clones held inside the callbacks *before* those callbacks are cleared, which is what the comment is explaining. Consider making the comment even more explicit that this ordering (drop the spawn task's own `tx` first, then clear the callbacks holding the other clones) is what actually closes the channel — right now a future reader could plausibly "simplify" this by deleting the `drop(tx)` line since it looks redundant, not realizing it's clarifying/enforcing evaluation order for the reader (even though Rust's drop-at-scope-end would produce the same runtime effect either way). A one-line addition like "(explicit here for clarity — dropped implicitly at scope end otherwise)" would future-proof against that.
- **`packages/desktop-app/src/lib/stores/model-store.svelte.ts:300-306`** — the comment now says `refreshModels()` is a "reconciliation/fallback ... covers a missed-listener race," which is accurate, but consider: since `chatModelDownload()` awaits the *entire* stream (including the final `"ready"` event) before resolving, and the ready-listener fires synchronously off the same event, in the common case both the ready-listener update and the `refreshModels()` fallback will now run back-to-back for every successful download (not just the race case) — `updateModelStatus(idx, {status: 'ready'})` then immediately followed by a full `refreshModels()` overwriting `this.models` again. This is harmless (idempotent) but is an extra full backend round-trip (`chatModelList` + `getSystemRamGb`) on every single download completion, not just the rare race. Given issue #1471 explicitly asks to "confirm whether the existing event should be wired up... or removed if dead," it's worth considering whether `refreshModels()` here could be skipped when the ready-event already fired for this modelId (e.g. track a local flag), avoiding a redundant round-trip on the common path while keeping it as true fallback only when the listener didn't fire. Not blocking — just flagging so it's a conscious choice, not an accidental double-fetch.

#### Nitpicks

- **Nit:** `packages/agent/src/local_agent/composite_model_manager.rs` — `clear_gguf_progress_callback`/`clear_ollama_progress_callback` doc comments duplicate wording almost verbatim ("Clear the X progress callback, dropping any resources (e.g. channel senders) it holds."). Minor, but could be a single shared line if this pattern grows further (e.g. if a third backend is added later).
- **Nit:** `packages/desktop-app/src/tests/stores/model-store.test.ts` — the two new tests both duplicate the `vi.waitFor(() => mockEventListeners.has('model://download-ready'))` + manual event-map mocking setup. Given there are now two call sites doing near-identical Tauri-event-mock choreography, a small local helper (e.g. `waitForReadyListener()`) would reduce duplication if a third test in this style gets added later — not warranted for just two occurrences today (YAGNI), just noting for future growth.
- **Nit:** the PR description's claim that the new Rust tests satisfy issue #1471's acceptance criterion "(a) root cause identified with a debugger/instrumented logging, not just static code reading" is a bit of a stretch — the tests are excellent regression tests (they do exercise real `mpsc` channel-closing semantics, not a tautology), but they were written *after* the root cause was already hypothesized from static reading, not used as an investigative tool with a debugger as the AC literally asks. Recommend the PR description (or issue close-out comment) be honest about this: the tests validate the fix and prevent regression, which is what actually matters, but they're not literally "debugger-based root cause confirmation." Worth a one-line acknowledgment when closing #1471 rather than silently checking that box.

### Acceptance Criteria Check (issue #1471)

- Root cause identified — yes, and independently re-verified in this review via code tracing (not a debugger, see nitpick above).
- Status auto-transitions to ready without manual refresh — yes, verified via the new frontend tests and the `MODEL_DOWNLOAD_READY` listener wiring.
- Cancel button hidden once complete — yes, by construction: `model-manager.svelte` renders Cancel purely from `status.status === 'downloading'`, and the new listener flips status to `'ready'` directly, independent of the `chatModelDownload()` promise.
- Regression test added — yes, both Rust (channel-closing) and frontend (hung-promise simulation) tests are genuine regression tests, not tautologies.
- Decision on `MODEL_DOWNLOAD_READY`: wired up, not removed — correct decision given the event was already emitted server-side and is the correct low-latency signal; confirmed not dead code.

### Recommendation

Requesting changes — not because the core fix is wrong (it isn't; the single-model-download case, which is almost certainly the common/only reachable case in the UI today, is fully fixed and well-tested), but because the concurrent-different-model-download race introduced/worsened by adding `clear_*_progress_callback()` to a *shared, unkeyed* callback slot needs to be either fixed or explicitly acknowledged and scoped as a deliberate, tracked follow-up rather than shipped silently. A one-line response confirming "the Settings UI only permits one active download at a time today, tracked as follow-up issue #NNNN for true multi-download support" would be sufficient to flip this to an approve.

malibio:
Note: unable to set formal review status (request-changes) via `gh pr review` — GitHub blocks self-review on your own PR. Recommendation stands as **Requesting Changes** per the review above; please treat this comment as the review verdict.

malibio:
## Addressed review feedback

**🔴 Critical — concurrent-different-model-download race: FIXED (not just scoped/documented).**

Went with option (a) from the review rather than (b): `on_progress` in both `GgufModelManager` and `OllamaModelManager` is now a `HashMap<String, Box<dyn Fn>>` keyed by `model_id` instead of a single unkeyed `Option<Box<dyn Fn>>` slot. `set_progress_callback`/`clear_progress_callback` (and the `CompositeModelManager` wrappers, and the daemon's call sites in `local_agent_service.rs`) now take a `model_id: &str` and only touch that model's entry. Completing download A's cleanup (`clear_gguf_progress_callback("A")`) can no longer wipe out download B's still-in-flight callback — verified by a new regression test, `clearing_one_models_callback_does_not_affect_a_concurrent_download`, which sets up two concurrent per-model callbacks, clears one, and asserts the other's channel/entry is untouched.

Commit: ef37e94c

**🟡 Important — `drop(tx)` clarity comment.** Expanded the comment at the call site to spell out the ordering intent (drop the spawn task's own `tx` clone first, then clear the per-model callbacks holding the other clones) so a future reader doesn't mistake it for dead code.

**🟡 Important — redundant `refreshModels()` round-trip on the common path.** Leaving as-is for now — it's a single extra `chatModelList` + `getSystemRamGb` call per completed download (not per progress tick), and keeping it unconditional is what makes it a true reconciliation fallback rather than a best-effort one. Happy to add a skip-if-already-ready guard as a fast-follow if it shows up as a real cost.

**🟢 Nitpicks:**
- Doc-comment duplication between `clear_gguf_progress_callback`/`clear_ollama_progress_callback`: left as-is — the two are intentionally structurally symmetric (gguf/ollama backends), and consolidating now would be premature given there are only two.
- Test-helper duplication in `model-store.test.ts`: only 2 occurrences today, not warranted per YAGNI; will factor out if a third shows up.
- Acknowledging the nitpick on the issue's "root cause identified with a debugger" AC: the root cause was identified via careful static code tracing (following the `Sender` clone through `on_progress` → the callback closure → the mpsc channel lifecycle) and *confirmed* by the new regression tests exercising real channel-closing semantics, not via a live debugger session. Will note this precisely when closing #1471 rather than checking that box as literally satisfied.

All Rust tests pass (`cargo test -p nodespace-agent -p nodespace-daemon --lib`: 254 + 42 passed), `cargo clippy --all-targets -- -D warnings -D clippy::unused-async` clean, `cargo fmt --all` clean (no unrelated reformatting), and the full frontend suite passes (147 files / 3941 tests).

malibio:
### Code Review Summary (re-review of commit ef37e94c)

The keyed-HashMap fix correctly solves the **exact** race flagged in the previous review: for `GgufModelManager` and standalone `OllamaModelManager` usage, `clear_progress_callback("A")` now only removes key `"A"` from the map, leaving a concurrently-downloading model `"B"`'s entry completely untouched. This is verified independently by tracing the code and confirmed by the new `clearing_one_models_callback_does_not_affect_a_concurrent_download` test, which is a genuine regression test (see below) — not a tautology.

However, tracing the fix through `CompositeModelManager` — the object the daemon actually holds and calls (`packages/daemon/src/services/local_agent_service.rs:104`) — surfaces a **new, real bug specific to Ollama-backed models** that did not exist before this commit. Details below.

### Findings

#### Critical Issues

- **`packages/agent/src/local_agent/composite_model_manager.rs:84-104` and `packages/daemon/src/services/local_agent_service.rs:761-770,807-810` — key mismatch between where the Ollama progress callback is *stored* vs. *looked up*, so progress events silently never fire for Ollama-backed downloads.**

  Trace:
  1. The daemon's `download_model` RPC receives `model_id` as sent by the frontend/composite routing layer. For an Ollama-backed model this is the **prefixed** id, e.g. `"ollama:llama3.2:3b"` (that's the entire point of `OLLAMA_PREFIX` — see `composite_model_manager.rs:44-56`, and the reverse mapping in `list()` at line 120 which prepends the prefix for Ollama models so the frontend/daemon boundary always deals in prefixed ids for Ollama models).
  2. `local_agent_service.rs:761` calls `manager.set_ollama_progress_callback(&model_id, ...)` — `model_id` here is still prefixed. `CompositeModelManager::set_ollama_progress_callback` (`composite_model_manager.rs:85-91`) passes it straight through to `self.ollama.set_progress_callback(model_id, cb)` with **no `strip_ollama_prefix` call**. So the callback is stored in `OllamaModelManager.on_progress` under the key `"ollama:llama3.2:3b"`.
  3. `local_agent_service.rs:777` calls `manager.download(&model_id_clone)` (same prefixed id). `CompositeModelManager::download` (`composite_model_manager.rs:127-135`) *does* correctly detect `is_ollama(model_id)` and calls `self.ollama.download(Self::strip_ollama_prefix(model_id))` — i.e. `self.ollama.download("llama3.2:3b")`, the **unprefixed** id.
  4. Inside `OllamaModelManager::download` (`ollama_model_manager.rs:202-263`), every progress tick fires `self.fire_progress(DownloadEvent { model_id: model_id.to_string(), ... })` using the **unprefixed** id (`"llama3.2:3b"`), and `fire_progress` (`ollama_model_manager.rs:140-143`) does `self.on_progress.read().await.get(&event.model_id)` — looking up `"llama3.2:3b"`.
  5. The map only has an entry under `"ollama:llama3.2:3b"` (step 2). The lookup at step 4 misses on every single tick. **Result: `.get()` returns `None`, the callback is silently never invoked, and no `"downloading"` progress events are ever emitted to the frontend for any Ollama-backed model download**, for the entire lifetime of this code path.
  6. The cleanup call `manager.clear_ollama_progress_callback(&model_id_clone)` (line 807-810, also prefixed) does correctly remove the stray un-consulted entry at `"ollama:llama3.2:3b"` when the download finishes — so there's no permanent leak/hang here, and critically **the terminal "ready"/"error" event (which is what issue #1471 is actually about — the stuck-at-100% status) is unaffected**, because that event flows through `manager.download()`'s `Result`, not through `on_progress`/`fire_progress` at all. So this is not a re-manifestation of #1471's exact symptom, but it is a real, silent functional regression: Ollama model downloads will show no live progress bar/percentage in the UI (whatever "downloading" event consumers do with `bytes_downloaded`/`bytes_total` degrades to nothing), for the entire download, then presumably still complete correctly via the ready event.
  7. This bug is **new in this commit** — verified by checking out the pre-fix single-`Option`-slot code: there, `set_progress_callback`/`clear_progress_callback` took no `model_id` argument at all, so there was no key to mismatch; `fire_progress` just checked `Option::as_ref()` unconditionally. The prefix/no-prefix asymmetry between `download()` (strips) and `set_*_progress_callback()`/`clear_*_progress_callback()` (doesn't strip) only became observable once keying was introduced.
  8. **No test in this PR would have caught this** — all three new/existing `on_progress` tests exercise `GgufModelManager` directly, never routing through `CompositeModelManager` with a prefixed id, so the prefix-stripping asymmetry between `download()` and the callback setters is untested.

  **Fix**: `CompositeModelManager::set_ollama_progress_callback` and `clear_ollama_progress_callback` should strip the prefix before delegating, mirroring what `download()`/`cancel_download()`/`delete()`/`load()` already do consistently:
  ```rust
  pub async fn set_ollama_progress_callback(&self, model_id: &str, cb: Box<dyn Fn(DownloadEvent) + Send + Sync>) {
      self.ollama.set_progress_callback(Self::strip_ollama_prefix(model_id), cb).await;
  }
  pub async fn clear_ollama_progress_callback(&self, model_id: &str) {
      self.ollama.clear_progress_callback(Self::strip_ollama_prefix(model_id)).await;
  }
  ```
  A regression test analogous to `clearing_one_models_callback_does_not_affect_a_concurrent_download`, but going through `CompositeModelManager::set_ollama_progress_callback("ollama:some-model", ...)` and asserting the callback actually fires when `OllamaModelManager::download("some-model")` (unprefixed, as `CompositeModelManager::download` would call it) emits a progress event, would have caught this and should be added.

  Note: `set_gguf_progress_callback`/`clear_gguf_progress_callback` are **not** affected — GGUF model ids are never prefixed, so `CompositeModelManager::download`'s "no prefix, treat as GGUF" branch passes the same untouched `model_id` straight to `self.gguf.download(model_id)`, matching what was used to set/clear the callback. This bug is isolated to the Ollama backend path.

#### Answers to the specific re-review questions

1. **Does this fix the originally-reported race?** Yes, cleanly, for the case actually exercised by the new tests (same-backend-manager, keyed callbacks). `RwLock` is `tokio::sync::RwLock` (async, task-fair) — a brief write-lock to remove one key is not a meaningful contention concern; it's an in-memory `HashMap::remove` under the lock, not I/O, so any reader blocked during that window resumes essentially immediately. No new lock-related hazard introduced.
2. **Is the new test meaningful?** Yes. It would not even *compile* against the pre-fix API (old `set_progress_callback`/`clear_progress_callback` took no `model_id` parameter), let alone pass — so it's a faithful regression test, and it specifically asserts per-key isolation (`rx_b` still open, map still contains `"ministral-3b-q4km"`, no longer contains `"ministral-8b-q4km"`), not just "clearing works," which the older tests already covered.
3. **Single-model download path / consistent model_id keying** — Verified correct for GGUF and for the daemon's own local variables (`model_id`, `model_id_clone`, `mid_gguf`, `mid_ollama` are all clones of the same original string, so no accidental case/prefix drift *within* `local_agent_service.rs`). However, as detailed above, the id that's consistent within the daemon layer is **inconsistently un-prefixed one layer down** inside `CompositeModelManager`, which is exactly the interaction the re-review asked to check — and it fails for Ollama.
4. **`model_id_clone.clone()` ownership fix** — Correct. Both the `Ok` and `Err` arms now clone `model_id_clone` when building the event struct instead of moving it, so the original `model_id_clone` remains available for the two `clear_*_progress_callback(&model_id_clone)` calls after the `match`. No stale/wrong id in either branch; this is a straightforward, correct fix for what would otherwise have been a borrow-checker error.
5. **Toolchain checks**: `cargo fmt --check` clean. `cargo clippy -p nodespace-agent -p nodespace-daemon --all-targets -- -D warnings -D clippy::unused-async` clean. `cargo test -p nodespace-agent -p nodespace-daemon --lib`: 254 + 42 passed, 0 failed — matches the PR comment's claim.
6. **Prior 🟡/🟢 items**: Both Important items were addressed reasonably — the `drop(tx)` ordering comment was meaningfully expanded to explain intent; the redundant-`refreshModels()`-round-trip item was explicitly deferred with a defensible justification (cheap, once-per-completed-download, keeps the fallback unconditional/robust) rather than silently ignored. All three nitpicks were explicitly acknowledged and left as-is with reasonable YAGN-style justification (2 occurrences isn't enough to warrant a shared helper/comment yet). These are all reasonable calls — no pushback needed on any of them.

### Recommendation

Requesting changes — not on the original concurrency race (that's genuinely fixed and well-tested), but because tracing the fix one layer down through `CompositeModelManager` surfaces a new, real bug: Ollama-backed model downloads will silently never receive progress-tick events (`bytes_downloaded`/`bytes_total` updates) for their entire duration, because `set_ollama_progress_callback`/`clear_ollama_progress_callback` store/clear under the prefixed model_id while `OllamaModelManager::download()` (called via `CompositeModelManager::download()`, which strips the prefix) fires and looks up progress under the unprefixed id. The terminal ready/error transition (the actual #1471 symptom) is unaffected since it doesn't go through `on_progress`, so this isn't a re-regression of the original bug's exact symptom — but it is a silent, untested functional gap in progress reporting for an entire backend, introduced by this commit. Fix is a one-line `strip_ollama_prefix` call in each of `CompositeModelManager::set_ollama_progress_callback`/`clear_ollama_progress_callback`, plus one test routed through the composite manager with a prefixed id to lock in the invariant.

malibio:
Note: unable to set formal review status (request-changes) via `gh pr review` — GitHub blocks self-review on your own PR. Recommendation stands as **Requesting Changes** per the review above; please treat that comment as the review verdict.

malibio:
## Fixed: Ollama progress-callback prefix mismatch

Good catch in the re-review — a real bug, isolated to the Ollama path.

**Root cause:** `CompositeManager::download()` strips the `"ollama:"` prefix before delegating to `OllamaModelManager`, and `OllamaModelManager` fires/looks up progress callbacks keyed by that unprefixed id. But `set_ollama_progress_callback`/`clear_ollama_progress_callback` passed `model_id` straight through without stripping it, so a callback registered by the daemon (with the prefixed id, matching what `chat_model_list`/the frontend use) was stored under the wrong key and would never be found when progress ticks actually fired.

**Fix (commit 891d129c):** strip the prefix in both composite wrapper methods, mirroring the existing `download`/`cancel_download`/`delete`/`load` pattern in the same file. Added two regression tests routed through `CompositeModelManager` with a prefixed id (`ollama_progress_callback_set_with_prefixed_id_is_stored_under_stripped_key`, `clear_ollama_progress_callback_accepts_prefixed_id`) proving the callback registered under the prefixed id is actually reachable at the unprefixed key.

Re-verified: `cargo test -p nodespace-agent -p nodespace-daemon --lib` (256 + 42 passed), `cargo clippy --all-targets -- -D warnings -D clippy::unused-async` clean, `cargo fmt --all` clean (no unrelated files touched), full frontend suite green (147 files / 3941 tests).

As correctly noted, this bug didn't reproduce #1471's original stuck-at-100% symptom (the terminal ready/error event flows through `download()`'s `Result`, not `on_progress`) — it would have silently dropped progress-percentage UI updates for Ollama-backed downloads for their entire duration. Ready for another look.

malibio:
## Round 3 re-review — verifying commit 891d129c

### Verdict: APPROVE

This round's fix is correct, complete, and properly proven by tests. I traced the full call graph, ran the isolated-revert test (reverting each strip independently to confirm the corresponding test fails), and ran clippy/fmt/tests independently.

### 1. Fix correctness — confirmed end to end

Traced the full path for an Ollama download:

- `LocalAgentServiceImpl::download_model` (`packages/daemon/src/services/local_agent_service.rs:730-814`) receives `model_id` from the gRPC request (prefixed, e.g. `ollama:llama3.2:3b`, exactly as the frontend/`list()` returns it) and calls `set_ollama_progress_callback(&model_id, ...)`, later `clear_ollama_progress_callback(&model_id_clone)`, and `manager.download(&model_id_clone)` — all three with the **same prefixed id**.
- `CompositeModelManager::download()` strips the prefix before calling `self.ollama.download(unprefixed)` (unchanged, pre-existing behavior).
- `OllamaModelManager::download(model_id: &str)` uses that unprefixed parameter both as the id it fires progress events under (`fire_progress(DownloadEvent { model_id: model_id.to_string(), .. })`) and `fire_progress` looks up the callback via `self.on_progress.read().await.get(&event.model_id)` — i.e., strictly by the unprefixed id.
- After 891d129c, `set_ollama_progress_callback`/`clear_ollama_progress_callback` now call `Self::strip_ollama_prefix(model_id)` before delegating to `self.ollama.set_progress_callback(...)`/`clear_progress_callback(...)`.

All three operations (`set`, `download`'s internal firing, `clear`) now consistently key off the same stripped id. The fix is correctly scoped and closes the gap identified in round 2.

### 2. Tests validated — not tautological

I mentally (and then actually, on a scratch copy restored afterward) reverted each strip call independently to confirm the tests are not vacuous:

- Reverting only `set_ollama_progress_callback`'s strip → both new tests fail (`assertion failed: rx.is_closed()`), confirming `ollama_progress_callback_set_with_prefixed_id_is_stored_under_stripped_key` genuinely depends on `set`'s strip, even though it clears via the raw `ollama_manager().clear_progress_callback(unprefixed)` bypassing the composite wrapper.
- Restoring `set`'s strip and reverting only `clear_ollama_progress_callback`'s strip → `clear_ollama_progress_callback_accepts_prefixed_id` fails as expected, while `ollama_progress_callback_set_with_prefixed_id_is_stored_under_stripped_key` still passes (correctly — it never exercises the composite `clear` wrapper).

This confirms each test isolates exactly the call site its name claims, and together they cover both halves of the fix (`set` and `clear`). Working tree was restored to match `HEAD` exactly afterward (`git diff HEAD` clean, full test suite re-passed at 256+42).

### 3. GGUF path unaffected — confirmed

`set_gguf_progress_callback`/`clear_gguf_progress_callback` in `composite_model_manager.rs` are untouched by 891d129c and still pass `model_id` straight through with no stripping — correct, since GGUF ids are never prefixed and there's no boundary to cross.

### 4. No other prefix/unprefix inconsistencies found

Scanned every `self.ollama.*` and `self.gguf.*` call site in `composite_model_manager.rs`. All are consistent:
- `download`/`cancel_download`/`delete`/`load` — strip on the way in (pre-existing, unchanged).
- `list`/`loaded_model` — add the prefix on the way out (pre-existing, unchanged).
- `is_available`/`unload` — prefix-agnostic, no id involved.
- `set_ollama_progress_callback`/`clear_ollama_progress_callback` — now strip (this fix).

No remaining call site crosses the gguf/ollama boundary with a mismatched id.

### 5. Independent verification commands — all clean

- `cargo fmt --check` → clean.
- `cargo clippy -p nodespace-agent -p nodespace-daemon --all-targets -- -D warnings -D clippy::unused-async` → clean, no warnings.
- `cargo test -p nodespace-agent -p nodespace-daemon --lib` → **256 passed** (nodespace-agent) + **42 passed** (nodespace-daemon), 0 failed — matches the PR's claimed counts exactly.

### 6. Full-diff fresh-eyes pass (3rd round)

Reviewed the complete `git diff origin/main...HEAD` again, not just the incremental 891d129c diff. Frontend files (`model-store.svelte.ts`, `agent-types.ts`, `model-store.test.ts`) are unchanged since round 1 (`c44b796e`) — confirmed via `git diff ef37e94c..891d129c --stat` showing only `composite_model_manager.rs` touched this round. The round-1 frontend logic (listening for `MODEL_DOWNLOAD_READY`, updating status/`downloadProgress` keyed by `modelId`) remains internally consistent — no prefix mismatch there since the frontend always uses whatever id `chatModelList`/`chatModelDownload` already returned/was given, symmetrically. No new issues found in the frontend surface.

### Summary

The prefix-stripping fix is correct, minimal, and well-targeted. The two new regression tests genuinely prove the fix (verified via isolated-revert testing) rather than passing vacuously. No remaining prefix/unprefix inconsistencies in the file. Quality gates (fmt, clippy with strict lints, full test suite) are all clean. No further concerns from three rounds of review.

**Recommendation: APPROVE.** Ready to merge pending final maintainer sign-off (posting via comment since the self-review restriction blocks a formal `gh pr review --approve` from this account, as in prior rounds).

PR #1484, opened 2026-07-02
Tray 'Open NodeSpace' menu item is a silent no-op — NODESPACE_UI_BINARY is never set

malibio:
Closes #1482

### DIFF
--- a/packages/desktop-app/src-tauri/src/daemon_setup.rs
+++ b/packages/desktop-app/src-tauri/src/daemon_setup.rs
@@ -546,6 +546,14 @@ fn write_plist(home: &Path, plist_path: &Path, daemon_bin: &Path) -> Result<()> 
     let bin_escaped = xml_escape(&bin_str);
     let label_escaped = xml_escape(launch_agent_label());
 
+    // The UI binary path: this function runs inside nodespace-app, so current_exe()
+    // returns the path the daemon needs to re-launch the GUI from the tray.
+    let ui_binary = xml_escape(
+        &std::env::current_exe()
+            .context("Cannot resolve current executable path for NODESPACE_UI_BINARY")?
+            .to_string_lossy(),
+    );
+
     // Pro edition: inject the Supabase cloud env the sync daemon needs (#156). The
     // values are baked at build time; all are XML-safe (JWT chars / URLs / schema
     // names contain no `<>&`). Empty for a community build.
@@ -581,6 +589,8 @@ fn write_plist(home: &Path, plist_path: &Path, daemon_bin: &Path) -> Result<()> 
         <string>{socket}</string>
         <key>NODESPACED_DB_PATH</key>
         <string>{db}</string>
+        <key>NODESPACE_UI_BINARY</key>
+        <string>{ui_binary}</string>
 {pro_env}    </dict>
     <key>RunAtLoad</key>
     <true/>
@@ -597,6 +607,7 @@ fn write_plist(home: &Path, plist_path: &Path, daemon_bin: &Path) -> Result<()> 
         bin = bin_escaped,
         socket = socket_path,
         db = db_path,
+        ui_binary = ui_binary,
         pro_env = pro_env,
         log_out = log_out,
         log_err = log_err,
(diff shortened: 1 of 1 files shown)


malibio:
## Code Review — PR #1484

---

### Requirements Check

- ✅ `NODESPACE_UI_BINARY` is set correctly in the launchd plist for both dev and production daemon instances
- ✅ Clicking "Open NodeSpace" in the tray will now launch the GUI (the existing `open_ui()` logic in `tray.rs` is unchanged and correct)
- ✅ Linux systemd unit gets the equivalent treatment (`write_systemd_service` now includes `Environment=NODESPACE_UI_BINARY=...`)
- ✅ Existing `tray.rs` unit tests for `open_ui()` are unaffected (they set/unset the env var manually and don't touch `daemon_setup.rs`)

---

### Findings

#### Suggested Improvements

**[Improvement] systemd `Environment=` directive with spaces in path (Linux)**

`packages/desktop-app/src-tauri/src/daemon_setup.rs:746`

The systemd unit `Environment=` directive is written as an unquoted bare value:

```
Environment=NODESPACE_UI_BINARY={ui_binary}
```

systemd's `Environment=` parsing requires paths containing spaces to be shell-quoted (single quotes wrap the value). On a Linux system where the user's home directory contains a space (e.g. `/home/john smith/.local/...`), the daemon will fail to start or will read a truncated path for `NODESPACE_UI_BINARY`. The existing `NODESPACED_SOCKET` and `NODESPACED_DB_PATH` directives have the same exposure, so this is a pre-existing pattern — but adding a third bare value is worth flagging. The fix is to wrap in single quotes: `Environment=NODESPACE_UI_BINARY='{ui_binary}'` (escaping any embedded single quotes as `'\''`). Paths with `&`, `<`, `>` are already not a concern here since this is a systemd unit, not XML.

Mitigating factor: the macOS plist path correctly uses `xml_escape()` since plist is XML. The Linux path inherits the pre-existing unquoted pattern. Given the project's greenfield status and macOS-first focus, this doesn't need to block merge, but is worth addressing before Linux becomes a primary target.

**[Improvement] `current_exe()` on macOS may return a symlink or wrapper path in some launch contexts**

`packages/desktop-app/src-tauri/src/daemon_setup.rs:552`

`std::env::current_exe()` resolves to the path of the executable as the OS reports it at process launch time. On macOS, for a bundled `.app`, this is typically `NodeSpace.app/Contents/MacOS/nodespace-app` — which is correct and what the daemon needs. However, in some edge cases (Gatekeeper quarantine, developer ID codesign re-signing workflows, or indirect launch via `open -a`), the OS may return a symlinked or wrapper path rather than the canonical executable. Adding `.canonicalize()` after `current_exe()` would make this airtight:

```rust
let ui_binary = xml_escape(
    &std::env::current_exe()
        .context("Cannot resolve current executable path for NODESPACE_UI_BINARY")?
        .canonicalize()
        .context("Cannot canonicalize current executable path for NODESPACE_UI_BINARY")?
        .to_string_lossy(),
);
```

This is a hardening suggestion, not a blocker. The `current_exe()` approach works correctly for the primary dev and production launch paths described in the commit message.

#### Nitpicks

**Nit: Minor comment duplication between plist and systemd paths**

`daemon_setup.rs:549` and `daemon_setup.rs:729` — both comments say "this function runs inside nodespace-app, so current_exe() returns the GUI binary path." The first comment is slightly more precise ("returns the path the daemon needs to re-launch the GUI from the tray"). Minor.

---

### Architecture & Correctness Assessment

The core design decision is correct and elegant: `write_plist()` and `write_systemd_service()` both execute inside the `nodespace-app` process, so `std::env::current_exe()` is exactly the right call — it resolves to the GUI binary the daemon should re-launch. This follows the "resolve at daemon-setup time" pattern established by #1469 as the issue requested.

The fix is minimal and targeted. One file changed, two call sites, each adding ~7 lines. No existing tests broken, no unrelated changes bundled in.

**Windows**: The PR correctly omits Windows changes. On Windows, the daemon is spawned directly via `spawn_daemon_windows()` rather than through a service manager, and `register_autorun_windows()` only registers the daemon binary — not the UI path. Setting `NODESPACE_UI_BINARY` on Windows would require a different mechanism (registry key or passing it at spawn time). The issue acceptance criteria acknowledges this with "if/when a Linux GUI binary path is determinable" wording, which covers Windows equally.

---

### Recommendation

**APPROVE** — The fix correctly solves the root cause. The systemd quoting concern is a pre-existing pattern in the codebase (not introduced by this PR), and the `current_exe()` canonicalization note is a future hardening opportunity, not a correctness bug in the primary launch paths.

malibio (commented):
LGTM — fix is correct and minimal. See inline comment for two non-blocking notes (systemd quoting pre-existing pattern, optional canonicalize hardening).

malibio:
## Address-Review Summary

**Addressed 2/2 findings** (skipped 1 nit):

✅ **`.canonicalize()` on `current_exe()`** — Added to both `write_plist` (macOS) and `write_systemd_service` (Linux). Resolves symlinks/wrapper paths that some macOS and Linux launch contexts produce.

✅ **systemd `Environment=` single-quoting** — Wrapped all three `Environment=` values (`NODESPACED_SOCKET`, `NODESPACED_DB_PATH`, `NODESPACE_UI_BINARY`) in single quotes with `'\''` escaping for embedded single quotes. Fixes the pre-existing exposure on the two prior entries while adding the new one correctly.

⏭️ **Skipped**: Comment duplication nit (negligible value, comments serve their purpose).

No re-review needed — both changes are straightforward hardening, no logic changes.

PR #1485, opened 2026-07-02
Dev + production daemons can now run concurrently against the same shared SQLite DB with no busy_timeout/retry

malibio:
Closes #1483

### DIFF
--- a/packages/core/src/db/schema.sql
+++ b/packages/core/src/db/schema.sql
@@ -1,6 +1,7 @@
 PRAGMA journal_mode = WAL;
 PRAGMA foreign_keys = ON;
 PRAGMA synchronous = NORMAL;
+PRAGMA busy_timeout = 5000;
 
 CREATE TABLE IF NOT EXISTS node (
     id               TEXT    PRIMARY KEY,
--- a/packages/core/src/db/sqlite_store.rs
+++ b/packages/core/src/db/sqlite_store.rs
@@ -4841,4 +4841,69 @@ mod tests {
 
         Ok(())
     }
+
+    /// #1483: two `SqliteStore`s opened against the same file (simulating a dev +
+    /// production daemon both holding the DB) must not surface SQLITE_BUSY as a
+    /// hard error on the loser of a write race. `busy_timeout` (set in schema.sql,
+    /// applied per-connection in `initialize_schema`) makes the second writer retry
+    /// until the first releases its lock, instead of failing immediately.
+    ///
+    /// Requires a multi-thread runtime: libsql's local connection executes
+    /// synchronously (no `spawn_blocking`), so on a current-thread runtime the
+    /// lock-holder and writer tasks would starve each other instead of running
+    /// concurrently.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_concurrent_writers_retry_instead_of_erroring_on_busy() -> Result<()> {
+        let temp_dir = TempDir::new()?;
+        let db_path = temp_dir.path().join("shared.db");
+
+        let store_a = SqliteStore::new(db_path.clone()).await?;
+        let store_b = SqliteStore::new(db_path.clone()).await?;
+
+        // Hold store_a's write lock open in a background task.
+        let conn_a = store_a.db.clone();
+        let (release_tx, release_rx) = tokio::sync::oneshot::channel::<()>();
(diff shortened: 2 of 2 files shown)


malibio:
### Code Review Summary

Small, well-scoped fix that does exactly what issue #1483 asked for: one PRAGMA line plus a regression test proving it works. I verified the placement, the per-connection PRAGMA execution path, and ran the new test 5x standalone (~0.23-0.25s each, consistent) — no flakiness observed. This is a net positive, low-risk change. Approving.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- `packages/core/src/db/sqlite_store.rs:157-168` (pre-existing code, not introduced by this PR): `initialize_schema` special-cases PRAGMAs containing `=` to route through `conn.query` instead of `conn.execute`, because some PRAGMAs (e.g. `journal_mode`) return a row. `busy_timeout = 5000` also returns a row (the applied timeout value) when queried, so it correctly falls into that branch — confirmed this is handled correctly, no change needed. Flagging only because it's worth a one-line comment at the PRAGMA-routing branch (not this PR's job) explaining *why* `query` is used for assignment-style PRAGMAs, since it's non-obvious from the code alone and the next person adding a PRAGMA here will hit the same question. Not a blocker for this PR.

- `packages/core/src/db/sqlite_store.rs:4859-4869` (new test): the lock-holder task's write is a raw `conn_a.execute("BEGIN IMMEDIATE", ())` / manual `INSERT` / `COMMIT`, rather than going through `SqliteStore::create_node` like the second writer does. This is fine and arguably clearer (it isolates "hold the write lock" from "exercise the retry path" as two different concerns), but it does mean the test depends on the `node` table's exact column list staying in sync with `schema.sql` by hand. A future schema change (e.g. a new NOT NULL column) would break this test with a possibly-confusing error rather than a schema-diff-shaped one. Minor; not worth blocking on since `STRICT` tables will fail loudly and point at the right line.

#### Nitpicks

- Nit: `packages/core/src/db/sqlite_store.rs:4894` — the 200ms sleep before releasing the lock is a reasonable pragmatic choice to make sure `write_task`'s write is actually in flight and blocked before the lock is released (otherwise the test could pass trivially without ever exercising the busy-retry path), but it does add fixed wall-clock time to the suite and is inherently a little timing-dependent. Given the payoff (proving the retry path is real, not just "no PRAGMA error"), this is a reasonable trade — no action needed, just noting it as the one soft spot if this test ever becomes flaky under heavy CI load.

- Nit: the PR title stayed the same as the issue title ("Dev + production daemons can now run concurrently..."), which describes the *problem* rather than the fix. Not a blocker, just would read more clearly as something like "Add busy_timeout PRAGMA + regression test (#1483)" for anyone scanning PR history later.

### Verification Performed
- Confirmed `PRAGMA busy_timeout = 5000;` placement in `schema.sql` (packages/core/src/db/schema.sql:4) is alongside the other connection-level PRAGMAs and is picked up by `initialize_schema`, which runs on every `SqliteStore::new()` call (packages/core/src/db/sqlite_store.rs:144, 157-168) — correct given busy_timeout is per-connection, not persisted in the DB file.
- Ran `test_concurrent_writers_retry_instead_of_erroring_on_busy` standalone 5x: consistently passes in ~0.23-0.25s, no flakiness.
- Confirmed `#[tokio::test(flavor = "multi_thread")]` is justified, not cargo-culted: libsql's local connection runs synchronously with no `spawn_blocking`, so a current-thread runtime would serialize the lock-holder and writer tasks onto the same OS thread and could never actually exercise concurrent access.
- Confirmed test correctly uses the private `db: Arc<libsql::Connection>` field directly (valid since the test lives in the same `mod tests` in this file) to drive the lock-holder's raw SQL, while the second writer goes through the public `create_node` API — a reasonable way to isolate "who holds the lock" from "what's being tested."
- Diff is atomic: touches only `schema.sql` (1 line) and `sqlite_store.rs` (test only). No unrelated changes bundled in.
- Issue's two actionable checklist items (add PRAGMA, add regression test) are both satisfied. The third checklist item (product decision on whether concurrent dev+prod daemons should be prevented outright) is explicitly out of scope per the issue body and correctly not addressed here.

malibio (commented):
Reviewed — approving in substance (see detailed comment above); GitHub blocks self-approval on this PR since author == reviewer account.

PR #1486, opened 2026-07-03
Add "Continue with Google" sign-in to the Pro sync pill (core half)

mstomar125:
Part of **Google OAuth sign-in** (NodeSpaceAI/nodespace-cloud#11). Core half; the daemon half is NodeSpaceAI/nodespace-sync#247.

## What
- **proto:** mirror the daemon's new `InitiateOAuthRequest.provider` field.
- **`pro_initiate_oauth`:** add `provider: Option<String>` and pass it through.
- **`pro-sync-pill.svelte`:** the signed-out pill now opens a small options menu — **"Continue with Google"** / **"Sign in with email"** — instead of committing to one method on a single click. Google passes `provider: "google"`, so the daemon signs in directly against Supabase GoTrue OAuth (PKCE); email is the existing Worker flow, unchanged.

## Behavior / safety
- **Community build unchanged** — everything is inside the existing `{#if proSync.isPro}` pill; empty `provider` preserves the current email/password path exactly.
- Signed-in account menu (identity / Invitations / Sign out) is untouched; the menu markup was generalized to serve both signed-in and signed-out states.

## Tests / checks
- `svelte-check` — **0 errors, 0 warnings** (1714 files). `eslint` — clean.
- `cargo check` (src-tauri) — compiles clean (proto codegen + command change).
- No existing test references the pill or `pro_initiate_oauth`, so nothing regressed.

## Note
Supabase-side config (Google provider enabled + `uri_allow_list` for the loopback callback) is already applied to the shared project and the Google↔Supabase handshake is verified live. Final end-to-end click-through (browser Google login → authenticated) will be validated on a Pro demo build.

### DIFF
--- a/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
+++ b/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
@@ -80,6 +80,11 @@ message InitiateOAuthRequest {
   // Hint passed to the worker login form so the user knows which
   // account they're signing into (typically the local OS username).
   string user_hint = 2;
+  // Social OAuth provider to sign in with. Empty = the Worker-brokered
+  // email/password flow (default). "google" = sign in directly against
+  // Supabase GoTrue's provider OAuth (PKCE), bypassing the Worker form.
+  // The Worker is still used for the later refresh grant regardless.
+  string provider = 3;
 }
 
 message InitiateOAuthResponse {
--- a/packages/desktop-app/src-tauri/src/commands/pro_sync.rs
+++ b/packages/desktop-app/src-tauri/src/commands/pro_sync.rs
@@ -139,12 +139,15 @@ fn emit_disconnected(app: &AppHandle, reason: String) {
 /// `worker_url` defaults to `http://127.0.0.1:8787` (the
 /// `nodespace-sync/cloud-worker` default). `user_hint` is shown in
 /// the worker's login form so users see which account they're
-/// signing into; empty string is fine.
+/// signing into; empty string is fine. `provider` selects a social
+/// sign-in — empty = the Worker email/password form (default),
+/// `"google"` = direct Supabase GoTrue OAuth.
 #[tauri::command]
 pub async fn pro_initiate_oauth(
     app: AppHandle,
     worker_url: Option<String>,
     user_hint: Option<String>,
+    provider: Option<String>,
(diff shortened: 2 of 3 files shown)


mstomar125:
**APPROVE** — core/UI half is correct and cleanly Pro-gated. Merge only in lockstep with sync#247 + cloud#11 (details below), and note the daemon half has open change requests.

Verified against code at `cebc86c1`:
- **Free-user mandate intact.** The entire pill — including the new signed-out options menu — is inside `{#if proSync.isPro}` (`pro-sync-pill.svelte:164` → `{/if}:235`); the "Continue with Google" (`:219`) / "Sign in with email" (`:227`) buttons are in the `{:else}` (`:212`) of the inner `{#if signedIn}` (`:187`), all strictly under the Pro gate. Non-Pro renders nothing → community build byte-unchanged. The −36 deletions are a pure refactor: the old signed-out `onClick` invoke body was extracted into `startSignIn()` (`:120`) and the account-menu markup re-indented under `{#if signedIn}` — no community-visible behavior removed.
- **Passthrough correct + safe.** `startSignIn('google')` → `{provider:'google'}`, email → `{}` (`:127-128`); `pro_initiate_oauth` maps `provider: Option<String>` → `provider.unwrap_or_default()` (`pro_sync.rs:159`), no panic, and proto3 omits the empty default so the email request is wire-identical to before. Logs record only the provider label, not tokens.
- **Proto:** `provider` = field **3** in `InitiateOAuthRequest` (`nodespace_pro.proto:87`) — I diffed it against sync#247's daemon proto and the two files are **byte-identical**, so the gRPC wire matches.

**Deploy-order (blocking for release, not for this code):**
- Do **not** ship this alone. Against an old daemon, proto3 drops unknown field 3 and "Continue with Google" silently falls back to the email Worker flow (the button lies rather than crashing). Land with sync#247 (which consumes `provider`).
- Needs **cloud#11** live first (GoTrue Google provider + the loopback `redirect_to` in the `uri_allow_list`).
- The daemon half **sync#247 currently has REQUEST CHANGES** (callback-listener hardening + provider allowlist); the feature as a whole is gated on that landing.

**Nits:** no tests for the passthrough / two-option menu; the `invoke`-failure path only `log.warn`s (`:130`) with no user-facing error (pre-existing). CI: no checks reported on the PR.

PR #1538, opened 2026-07-06
Hide empty collections from the sidebar (core#1487)

mstomar125:
Resolves NodeSpaceAI/nodespace-core#1487.

## Problem
The Collections sidebar rendered every collection in the tenant, including ones the signed-in user has zero visible member nodes in — on tenant_demo a user sees ~17 empty collections cluttering the tree (each opens to "No nodes in this collection").

## Fix
`pruneEmptyCollections()` applied at the end of the `collectionsTree` derived store. A collection is kept only if `memberCount > 0` **or** a descendant survives pruning (so a populated sub-collection under an empty parent still surfaces). `memberCount` is sourced from the local per-user SQLite store, so it already reflects RBAC visibility — this hides collections empty *for this user*, not just globally empty ones. No backend/daemon change.

## Tests
- `bun run test` → 147 files, **3945 passed** (collections.test.ts 40/40, +4 new for the hide-empty filter)
- `bunx eslint` clean · `bunx svelte-check` 0 errors/0 warnings

## Community-build safety
Additive UX filter on a derived store; users who have members in their collections see no behavioral change.

### DIFF
--- a/packages/desktop-app/src/lib/stores/collections.ts
+++ b/packages/desktop-app/src/lib/stores/collections.ts
@@ -257,6 +257,26 @@ export const collectionsState = createCollectionsStore();
 // ============================================================================
 
 /**
+ * Recursively prune collections that have no member nodes visible to the
+ * current user. A collection is kept if it has its own members OR any of its
+ * descendants does — so a populated collection nested under an empty parent
+ * still surfaces. Returns a new array of the surviving items (children are
+ * pruned in place on the retained items).
+ *
+ * `memberCount` is sourced from the local per-user store, so it already
+ * reflects RBAC visibility: the local DB only holds member edges the signed-in
+ * user can see. This filter therefore hides collections that are empty *for
+ * this user*, not just globally empty ones.
+ */
+function pruneEmptyCollections(items: CollectionItem[]): CollectionItem[] {
+  return items.filter((item) => {
+    const keptChildren = item.children ? pruneEmptyCollections(item.children) : [];
+    item.children = keptChildren;
+    return item.memberCount > 0 || keptChildren.length > 0;
+  });
+}
+
+/**
  * Transform flat collections into tree structure for UI display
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Reviewed adversarially against the 5 concerns; the fix is correct and demo-safe.

- **Recursion is correct.** `pruneEmptyCollections` is a clean post-order traversal: it prunes children first, then keeps an item on `memberCount > 0 || keptChildren.length > 0`. A populated *deep* descendant propagates up because each intermediate level's `keptChildren` is non-empty — verified by the `Empty Parent → Populated Child` test. Leaf base case (`children` undefined → `[]` → kept iff own members) is right. No off-by-one.
- **In-place `item.children = keptChildren` mutation is safe.** The derived rebuilds `itemMap` from scratch each recompute with fresh object literals sourced from `$data.collections`, so the mutation is scoped to a single recompute — no stale state across recomputes, and `collectionsTree` items are not shared with the other derived stores (each builds its own view models).
- **No community/free-build regression.** Purely additive prune on branches that are already empty *for this user*; `memberCount > 0` short-circuits so any user with visible members sees an identical tree. `memberCount` comes from the local per-user SQLite store, so RBAC visibility is already baked in.
- **Pre-sync edge case is self-healing, not a permanent vanish.** `onDaemonReconnect(() => collectionsData.loadCollections())` refreshes counts after edges pull, re-triggering the derived and re-surfacing the collection. A collection the user is genuinely in reappears once sync settles — it does not disappear permanently. Behavior is documented in the doc comment. Acceptable for the deadline.
- **Test coverage is adequate** for the described behavior: top-level empty hidden, empty-parent-kept-via-descendant, sibling pruning, all-empty-subtree dropped.

Optional (non-blocking, post-demo): consider a follow-up test asserting the recompute re-surfaces a collection when `loadCollections` bumps `memberCount` 0→N, to lock in the self-healing contract. Not required to merge.

PR #1539, opened 2026-07-06
Fix unreadable contrast in Google sign-in popup / Pro re-login modal (core#1488)

mstomar125:
Resolves NodeSpaceAI/nodespace-core#1488.

## Problem
The pill sign-in menu and Pro re-login modal referenced CSS vars (`--surface-*`, `--text-primary/secondary`, `--border-color`) that **are not defined anywhere in the codebase**, so every rule fell back to hardcoded light values; their dark variants were gated on `@media (prefers-color-scheme: dark)` (OS scheme), not the app's own theme → light text on white surface = invisible.

## Fix
Re-keyed both components onto the app's real design tokens (defined per-theme in `app.css` under `:root` / `.dark`, selected app-side via `data-theme`): surface → `hsl(var(--popover))`, text → `hsl(var(--popover-foreground))`, pill → `hsl(var(--card))`/`--foreground`, borders → `hsl(var(--border))`, secondary → `hsl(var(--muted-foreground))`, hover → `hsl(var(--muted))`. Removed both `@media (prefers-color-scheme: dark)` blocks. Matches the existing dropdown/popover convention.

Dark theme after fix: near-black surface `hsl(240 4% 10%)` with near-white text `hsl(0 0% 98%)` (~17:1). Light theme also self-consistent.

## Tests
`bunx eslint` clean · `bunx svelte-check` → 1714 files, 0 errors/0 warnings. CSS-only (+19/−60), both components render only when Pro.

### DIFF
--- a/packages/desktop-app/src/lib/components/pro-relogin-modal.svelte
+++ b/packages/desktop-app/src/lib/components/pro-relogin-modal.svelte
@@ -81,8 +81,8 @@
   }
 
   .relogin-content {
-    background: var(--surface-1, #ffffff);
-    color: var(--text-primary, #1f2937);
+    background: hsl(var(--popover));
+    color: hsl(var(--popover-foreground));
     border-radius: 8px;
     padding: 24px;
     width: min(440px, calc(100vw - 48px));
@@ -98,16 +98,16 @@
     margin: 0;
     font-size: 14px;
     line-height: 1.5;
-    color: var(--text-secondary, #4b5563);
+    color: hsl(var(--muted-foreground));
   }
 
   .relogin-detail {
     margin: 12px 0 0 0;
     padding: 8px 10px;
-    background: var(--surface-2, #f3f4f6);
+    background: hsl(var(--muted));
     border-radius: 4px;
     font-size: 12px;
-    color: var(--text-secondary, #6b7280);
+    color: hsl(var(--muted-foreground));
     word-break: break-word;
   }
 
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Independently reviewed the diff against `packages/desktop-app/src/app.css` and both changed components. Solid, minimal fix.

- **All 7 tokens defined for both themes** — verified `--popover`, `--popover-foreground`, `--card`, `--foreground`, `--muted`, `--muted-foreground`, `--border` all exist in `:root` (light) and `.dark` in `app.css`. No undefined-var fallbacks remain.
- **No leftover legacy vars** — grep for `--surface-*` / `--text-primary` / `--text-secondary` / `--border-color` / `prefers-color-scheme` in both files returns nothing.
- **No orphaned styles from removed `@media` blocks** — every selector the old dark blocks targeted (`.pro-sync-pill`(+`:hover`), `.menu`, `.menu-identity`, `.menu-email`, `.menu-signout`(+`:hover`), `.relogin-content`, `.btn-secondary`(+`:hover`)) is now covered by a theme-aware base rule that resolves correctly under `.dark`.
- **Contrast legible in both themes** — dark popover `hsl(240 4% 10%)` on text `hsl(0 0% 98%)` ~17:1; light white on `hsl(240 11% 4%)` ~19:1; muted-foreground/btn-secondary pairings both pass. The `hsl(var(--border))` pattern (with the `.dark` slash-alpha value) matches established usage across query-node/table-node/etc., so alpha resolves fine.
- **Community-safe** — both components mount only under the Pro capability probe (`app-shell.svelte`), and the change is CSS-only inside their scoped `<style>`; free build unaffected.

Ship it.

PR #1540, opened 2026-07-06
Log out button on the invitations popup for a signed-in user without a workspace (core#248)

mstomar125:
Paired core UI half of NodeSpaceAI/nodespace-sync#248 (daemon provisioning is the sync PR).

## Change (UI-only — `pro_signout`/`proSync.signOut()` already existed)
- `invitations-inbox.svelte`: optional `onLogout` prop → renders a **Log out** button in the footer (guarded by `{#if onLogout}`, so existing callers are unchanged).
- `pro-sync-pill.svelte`: passes `onLogout` that runs the existing sign-out flow (`proSync.signOut()` + `membership.reset()`) then closes the popup → reverts to the free/signed-out state.
- Component test added.

So a signed-in user with no redeem code / no membership can leave the invitations screen instead of being stuck.

## Tests
`bunx svelte-check` 0 errors/0 warnings (1714 files); eslint clean; affected component tests 9/9. Additive + Pro-gated (`onLogout` optional) — community build unaffected.

### DIFF
--- a/packages/desktop-app/src/lib/components/collaboration/invitations-inbox.svelte
+++ b/packages/desktop-app/src/lib/components/collaboration/invitations-inbox.svelte
@@ -16,7 +16,30 @@
 
 	const log = createLogger('InvitationsInbox');
 
-	let { open = false, onClose }: { open?: boolean; onClose: () => void } = $props();
+	let {
+		open = false,
+		onClose,
+		onLogout
+	}: {
+		open?: boolean;
+		onClose: () => void;
+		// When provided, a signed-in user with no code/membership can log out from
+		// here — reverting to the free/signed-out state (#248). Omitted callers get
+		// the plain Close-only footer.
+		onLogout?: () => void | Promise<void>;
+	} = $props();
+
+	let loggingOut = $state(false);
+
+	async function logout() {
+		if (!onLogout || loggingOut) return;
+		loggingOut = true;
+		try {
+			await onLogout();
+		} finally {
+			loggingOut = false;
+		}
+	}
 
 	let code = $state('');
 	let redeeming = $state(false);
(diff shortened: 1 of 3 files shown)


mstomar125:
**APPROVE**

Reviewed independently (mstomar125). Clean, well-scoped UI change — good for the demo.

- **Backward-compat safe:** `onLogout` is optional and gated by `{#if onLogout}`; existing callers (no handler) render the plain Close-only footer unchanged. Verified the new test asserts `queryByText('Log out')` is null without the prop.
- **Signout order is sound:** the pill's handler runs the existing proven `onSignOut()` (→ `proSync.signOut()` = `pro_signout` invoke + `userEmail=''` + `state='auth-required'`, then `membership.reset()`), then `inboxOpen=false`. `signOut()` clears `userEmail` even if the invoke throws (caught+warn), so `signedIn` flips false → pill reverts to signed-out, not stuck. No reopen loop: the auto-open effect is `signedIn && !firstRunSeen()`, and `signedIn` is now false + inbox is explicitly closed.
- **Double re-entry guard:** inbox `loggingOut` + button `disabled` on top of the pill's own `signingOut` guard; `finally` resets `loggingOut`. No stuck-disabled state.
- **Community/free-build safe:** entire pill (incl. InvitationsInbox mount) is under `{#if proSync.isPro}`; `onLogout` optional. Test is present and meaningful (asserts presence-gating + invocation).
- **No #1488/#1539 regression:** diff touches only the InvitationsInbox call site in pro-sync-pill.svelte; the `<style>` block (contrast fixes) is untouched. Confirmed only 3 files changed.

Minor (non-blocking): the destructive-action red (#dc2626) is a hardcoded literal rather than a token, and the footer now uses `justify-between` — cosmetically fine for the popup. Ship it.

PR #1542, opened 2026-07-06
Harden the Tauri trust boundary: CSP, global Tauri, and client-supplied worker_url

malibio:
Closes #1514

### DIFF
--- a/packages/desktop-app/src-tauri/src/commands/pro_sync.rs
+++ b/packages/desktop-app/src-tauri/src/commands/pro_sync.rs
@@ -20,10 +20,10 @@ use crate::services::pro_client::pb::{
 use crate::services::{ProClient, ProTier};
 use tonic::transport::Channel;
 
-/// Default auth-Worker URL when the frontend doesn't supply one. Release builds
-/// hit the deployed canonical domain (`pro.nodespace.ai`, nodespace-cloud#21);
-/// debug builds hit the local `wrangler dev` worker (`127.0.0.1:8787`, what
-/// `device-sync.sh` runs). Either is overridable via the optional `worker_url` arg.
+/// Auth-Worker URL used for the OAuth flow — not client-configurable.
+/// Release builds hit the deployed canonical domain (`pro.nodespace.ai`,
+/// nodespace-cloud#21); debug builds hit the local `wrangler dev` worker
+/// (`127.0.0.1:8787`, what `device-sync.sh` runs).
 #[cfg(debug_assertions)]
 const DEFAULT_WORKER_URL: &str = "http://127.0.0.1:8787";
 #[cfg(not(debug_assertions))]
@@ -136,16 +136,16 @@ fn emit_disconnected(app: &AppHandle, reason: String) {
 /// returns the attempt ID synchronously. UI tracks progress via the
 /// `sync:status` stream wired in `pro_subscribe_sync_status`.
 ///
-/// `worker_url` defaults to `http://127.0.0.1:8787` (the
-/// `nodespace-sync/cloud-worker` default). `user_hint` is shown in
-/// the worker's login form so users see which account they're
-/// signing into; empty string is fine. `provider` selects a social
-/// sign-in — empty = the Worker email/password form (default),
-/// `"google"` = direct Supabase GoTrue OAuth.
+/// The worker URL always resolves to [`DEFAULT_WORKER_URL`] — it's not
+/// client-configurable, since accepting an arbitrary URL from the
+/// frontend would let it redirect the OAuth flow to an attacker-controlled
+/// worker. `user_hint` is shown in the worker's login form so users see
+/// which account they're signing into; empty string is fine. `provider`
+/// selects a social sign-in — empty = the Worker email/password form
+/// (default), `"google"` = direct Supabase GoTrue OAuth.
 #[tauri::command]
 pub async fn pro_initiate_oauth(
     app: AppHandle,
-    worker_url: Option<String>,
     user_hint: Option<String>,
     provider: Option<String>,
 ) -> Result<String, String> {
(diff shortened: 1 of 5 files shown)


malibio:
### Code Review Summary

Net positive, ships clean. All three acceptance criteria for #1514 are met: an explicit CSP replaces `null`, `pro_initiate_oauth` no longer accepts a client-supplied `worker_url`, and `quality:fix`/tests pass. The change is atomic — every file touched is a direct consequence of disabling `withGlobalTauri` and tightening the CSP, not scope creep. I verified the diff directly (`git diff origin/main...HEAD`), traced both `pro_initiate_oauth` call sites, confirmed the `__TAURI_INTERNALS__` detection pattern against five other pre-existing call sites in the codebase, and checked the CSP against Tauri 2's actual behavior.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- **`tauri.conf.json` CSP — add `object-src 'none'`:** `default-src 'self'` does not implicitly restrict `object-src` in all CSP-parsing contexts as reliably as an explicit directive does, and `object-src` is the standard hardening line recommended by both the CSP spec authors and Tauri's own security docs (it blocks `<object>`/`<embed>`/`<applet>` plugin-based injection vectors that `script-src` doesn't cover). This is cheap, zero-risk, and closes a well-known gap — worth adding in this PR rather than a follow-up, since the issue is explicitly about "an explicit CSP is set" and a CSP without `object-src 'none'` is a known incomplete pattern.

- **`packages/desktop-app/src-tauri/src/commands/pro_sync.rs` — no test coverage for `pro_initiate_oauth`:** There are no `#[cfg(test)]` unit tests in this file (pre-existing gap, not introduced here), so nothing pins down "removing `worker_url` from the signature was intentional" — a future contributor could reintroduce a `worker_url: Option<String>` param without any test catching the regression. Given this is a security-motivated removal (closing an OAuth-redirect attack surface), a one-line test asserting the command signature has no worker_url-like parameter (or an integration test asserting `req.worker_url` always equals `DEFAULT_WORKER_URL` regardless of what's passed) would make the security property regression-proof rather than convention-proof. Not a blocker since the existing behavior is verified by `cargo check` and manual reasoning, but worth a fast-follow.

- **`packages/desktop-app/src/tests/services/tauri-sync-listener.test.ts`** — the existing test only sets/deletes `window.__TAURI__` and doesn't add a case for `window.__TAURI_INTERNALS__` alone (the actual real-world condition post this PR, since `withGlobalTauri: false` means `__TAURI__` won't exist in production but `__TAURI_INTERNALS__` will). The OR-branch this PR adds is exercised in practice by the app running at all, but isn't exercised by the unit test suite. Recommend adding a case with only `__TAURI_INTERNALS__` set to lock in the new branch.

#### Nitpicks

- **Nit:** `tauri.conf.json` CSP — `style-src 'self' 'unsafe-inline'` is a known-necessary relaxation for many Svelte/Vite-built apps (scoped `<style>` blocks and inline `style="..."` attributes) and is a widely accepted trade-off since CSS injection risk is much lower than script injection. Worth a one-line comment (or PR description note, since `tauri.conf.json` has no comment syntax) documenting *why* it's there, so a future security pass doesn't flag it as an oversight.
- **Nit:** `pro_sync.rs` doc comment update is a nice touch — explicitly states the security rationale ("accepting an arbitrary URL from the frontend would let it redirect the OAuth flow to an attacker-controlled worker") rather than just describing the mechanics, which is exactly the kind of "why" comment that pays off during a future audit.

### Verification Notes (for reviewers)

- **CSP scoping:** Confirmed via grep that the webview never issues direct `fetch()` calls to backend hosts inside the built Tauri app — all IPC goes through `@tauri-apps/api/core`'s `invoke()`, and the only `fetch()` path (`HttpAdapter`, port 3001) is browser-dev-mode only and not part of the Tauri webview's runtime CSP surface. `connect-src 'self' ipc: http://ipc.localhost` is therefore correctly scoped with no external hosts needed — matches Tauri 2's own IPC scheme requirements exactly.
- **Dev-mode impact:** `tauri.conf.json`'s `devUrl` (`http://localhost:1422`) is served directly by Vite in `tauri dev`, not through Tauri's asset protocol/CSP-injection path used for the production `frontendDist` build — so this CSP hardening is a release/production concern and doesn't risk breaking local dev workflows regardless of strictness.
- **`worker_url` closure is complete on the Tauri side:** Grepped both frontend call sites (`pro-sync-pill.svelte`, `app-shell.svelte`) — neither ever passed `worker_url` even before this PR, so this is a pure attack-surface reduction (a malicious/compromised renderer script could previously have called `invoke('pro_initiate_oauth', {worker_url: 'https://evil.example'})` directly; that's now impossible since the field doesn't exist on the command). The gRPC `InitiateOAuthRequest.worker_url` proto field still exists and is forwarded to the daemon, but it's now always populated with the compile-time `DEFAULT_WORKER_URL` constant on the Tauri side — the daemon itself lives in a separate repo not touched by this PR, so full closure of the surface on the daemon side is out of scope here and correctly so (the issue's boundary is the Tauri command, which is the actual client-trust boundary).
- **`__TAURI_INTERNALS__` fallback correctness:** This matches genuine Tauri 2 behavior, not just an assumption — Tauri 2's `withGlobalTauri` option controls only whether `window.__TAURI__` (the convenience global API surface) is injected; `window.__TAURI_INTERNALS__` (the low-level IPC transport used internally by `@tauri-apps/api`) is injected unconditionally by the Tauri core runtime regardless of that setting. The PR's rewritten `waitForTauriReady()`/`isTauriEnvironment()` in `app-initialization.ts` and the OR-pattern added to `tauri-sync-listener.ts` are consistent with this and match the identical pattern already in use in `model-store.svelte.ts`, `agent-store.svelte.ts`, `collection-service.ts`, `tauri-commands.ts`, and `backend-adapter.ts` prior to this PR — this PR simply brings the one straggler (`tauri-sync-listener.ts`) in line with the rest of the codebase.

### Recommendation

Approve with the `object-src 'none'` addition as the only change I'd want before merge (trivial, zero-risk, closes a standard CSP gap directly relevant to the issue's own acceptance criterion). The test-coverage suggestions are good fast-follows but not blockers given the change is small, well-reasoned, and verified by manual tracing plus a passing full test suite.

PR #1543, opened 2026-07-06
Verify SHA-256 for every model catalog entry, pin HF commit SHAs

malibio:
## Summary
- Populate a real SHA-256 for every catalog entry (all 13), fetched from HuggingFace's LFS blob metadata (`GET /api/models/<org>/<repo>?blobs=true`, `lfs.sha256` field) — no more empty-hash "skip verification" policy exceptions.
- Pin every download URL to the repo's commit SHA at fetch time instead of the moving `/resolve/main/` ref.
- Make an empty/missing `sha256` a hard error in `download()`, raised before the transfer starts (previously `perform_download` silently skipped verification after downloading the full multi-GB file).
- Fix `MINISTRAL_14B`'s stale `size_bytes` (was a rough `~8.2 GB` estimate; corrected to the real LFS-reported `8_239_593_024`).
- Removed the now-obsolete "policy exception" comments explaining why certain repos were allowed an empty hash — every entry is fully verified now, so the distinction no longer applies.

## Test plan
- [x] `cargo test -p nodespace-agent model_manager` — 43 passed, including 3 new tests (`every_catalog_entry_has_a_verifiable_sha256`, `every_catalog_entry_pins_a_commit_not_a_moving_ref`, `no_catalog_entry_has_an_empty_sha256`)
- [x] `cargo clippy -p nodespace-agent --all-targets -- -D warnings -D clippy::unused-async` — clean
- [x] `cargo fmt -p nodespace-agent -- --check` — clean
- [x] `bun run test:all` — frontend 3946/3946, Rust workspace unit tests 966/966 passed

Closes #1497

### DIFF
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -1016,10 +1016,10 @@ async fn perform_download(params: DownloadParams) -> Result<(), ModelError> {
         s.insert(model_id.clone(), ModelStatus::Verifying);
     }
 
-    // Stream-verify SHA-256. An empty expected hash is a catalog configuration
-    // error, not an opt-out -- `find_catalog_entry` rejects it before a
-    // download is ever started, so reaching this point with an empty hash
-    // means the catalog itself is misconfigured.
+    // Stream-verify SHA-256. `download()` already rejects an empty hash
+    // before starting the transfer, so this branch should be unreachable in
+    // practice; it stays as defense-in-depth in case `perform_download` is
+    // ever called from a path that skips that guard.
     if expected_sha256.is_empty() {
         let _ = tokio::fs::remove_file(&partial_path).await;
         return Err(ModelError::VerificationFailed(format!(
@@ -1252,6 +1252,29 @@ mod tests {
                 !url.contains("/resolve/main/"),
                 "model '{}' url pins a moving ref instead of a commit: {}",
                 m.id,
+                url
+            );
+            // A commit SHA is 40 lowercase hex chars, immediately after
+            // "/resolve/" and followed by "/". Assert that segment actually
+            // looks like a commit, not just "not literally main".
+            let ref_segment = url
+                .split("/resolve/")
+                .nth(1)
+                .and_then(|rest| rest.split('/').next())
+                .unwrap_or_default();
+            assert_eq!(
+                ref_segment.len(),
+                40,
+                "model '{}' resolve ref '{}' is not a 40-char commit SHA: {}",
+                m.id,
+                ref_segment,
+                url
+            );
+            assert!(
+                ref_segment.chars().all(|c| c.is_ascii_hexdigit()),
+                "model '{}' resolve ref '{}' is not hex: {}",
+                m.id,
+                ref_segment,
                 url
             );
         }
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -45,16 +45,12 @@ struct CatalogEntry {
     size_bytes: u64,
     quantization: &'static str,
     url: &'static str,
-    /// SHA-256 hash for verification, lowercase hex.
-    ///
-    /// **Policy**: an empty string explicitly skips verification, used only
-    /// for catalog entries served from official, tamper-evident HuggingFace
-    /// repos (Mistral AI's `mistralai/...` and the llama.cpp team's
-    /// `ggml-org/...`). Both rely on HF Xet storage, which serves files as
-    /// content-addressed chunks; a tampered upload would change the chunk
-    /// hashes, so an additional file-level SHA-256 check would be redundant.
-    /// For any unofficial or unverified source, populate this with the
-    /// expected hash and `perform_download` will enforce it.
+    /// SHA-256 hash of the file, lowercase hex, as reported by HuggingFace's
+    /// LFS metadata for the pinned commit in `url`. Every catalog entry must
+    /// carry a real hash -- `perform_download` treats an empty string as a
+    /// hard configuration error, not an opt-out. Re-derive with:
+    /// `curl -s "https://huggingface.co/api/models/<org>/<repo>?blobs=true"`
+    /// and read the `lfs.sha256` field for the target file.
     sha256: &'static str,
     context_window: u32,
     default_temperature: f32,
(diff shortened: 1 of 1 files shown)


malibio:
### Code Review Summary

This PR closes a real security gap: every catalog entry previously shipped an empty `sha256`, so multi-GB GGUF downloads — including the locked default (Ministral 8B, ADR-046) — had zero integrity checking beyond HTTPS transport trust. The fix is well-scoped, atomic (single file), and does exactly what the issue asked, in most respects going further (it drops the "official repo = exempt" policy entirely rather than encoding a `trial`/`non-trial` distinction, which is a net simplification).

I independently verified the externally-sourced data rather than trusting the diff: I queried HuggingFace's live blobs API for 6 of the 13 catalog entries (Ministral 3B, Ministral 8B — the locked default, Ministral 14B, Qwen3-8B, Qwen3.5-9B, Mistral-Small-3.2-24B, and Ornith-1.9B). In every case the commit SHA pinned in the URL, the `sha256`, and `size_bytes` in the PR matched HuggingFace's current API response exactly, including the corrected Ministral 14B size (`8_239_593_024`, confirmed live). I also checked out the branch (`FETCH_HEAD`, commit `b2854f3c`) into a scratch worktree and ran `cargo test -p nodespace-agent model_manager` (43/43 pass), `cargo clippy -p nodespace-agent --all-targets -- -D warnings -D clippy::unused-async` (clean), and `cargo fmt -p nodespace-agent -- --check` (clean) myself, matching the PR's stated test plan.

This is a net positive change and I recommend **approval**, with one non-blocking follow-up item noted below.

### Findings

#### Critical Issues

None found.

#### Suggested Improvements

- **`packages/agent/src/local_agent/model_manager.rs:1019-1029` (perform_download empty-hash branch is now unreachable in practice, but still deletes a possibly-large partial file for a case that can't occur):** Since `download()` at line 586 already hard-errors on `entry.sha256.is_empty()` *before* `perform_download` is ever invoked, the `expected_sha256.is_empty()` branch inside `perform_download` is dead code from any caller reachable through the public `ModelManager` trait today. The PR's own comment acknowledges this ("defense in depth... `download()` should already catch it first"), which is reasonable given `perform_download` is a free function that could theoretically be called from elsewhere in the future. This is fine as defense-in-depth, but consider a debug_assert or a short comment noting it's intentionally unreachable via the current call graph, so a future reader doesn't waste time trying to find a code path that produces `ModelStatus::Verifying` followed by an empty-hash error — it can't currently happen, and that's worth being explicit about rather than implying it's a live runtime scenario.

- **Trial-model hash provenance isn't independently re-derivable from the repo without external trust:** The doc comment on `sha256` (lines ~48-54) tells a future maintainer how to recompute the hash from HF's API, which is good, but there's no automated/CI step that periodically re-validates catalog hashes against the live HF API (e.g., a HF blob getting re-uploaded under the same commit due to an LFS migration, or HF garbage-collecting an old commit ref). This is out of scope for this PR (no CI currently exists in this repo per prior architecture review), but worth a follow-up issue: a scheduled job that re-fetches `?blobs=true` for all 13 entries and diffs against the catalog would catch silent upstream drift that a one-time verification can't.

#### Nitpicks

- **Nit:** `packages/agent/src/local_agent/model_manager.rs:1019-1029` and `:586-591` — the error message text is identical between the two guard sites (`"catalog entry for '{}' has no SHA-256 configured"`). Not a real problem since one is unreachable, but if a future refactor ever makes `perform_download` independently callable, having two different call sites raise byte-identical error strings could make log-based debugging slightly harder to trace back to which guard fired. Not blocking.

- **Nit:** `every_catalog_entry_pins_a_commit_not_a_moving_ref` only asserts `!url.contains("/resolve/main/")`. It would also reject a moving branch ref that isn't literally named `main` (e.g., `/resolve/master/` or a tag that later moves) — technically out of scope here since all 13 entries only ever used `main`, but the test name promises a stronger guarantee ("pins a commit") than the assertion delivers (it really only checks "isn't literally `/resolve/main/`"). A regex asserting the ref segment looks like a 40-char hex commit SHA would make the test's contract match its name.

### Verification Detail (for transparency)

Live HuggingFace API spot-checks (all matched exactly — commit SHA, sha256, size_bytes):

| Catalog entry | Commit SHA match | sha256 match | size_bytes match |
|---|---|---|---|
| Ministral 3B | ✅ | ✅ | ✅ |
| Ministral 8B (locked default) | ✅ | ✅ | ✅ |
| Ministral 14B | ✅ | ✅ | ✅ (corrected value confirmed) |
| Qwen3-8B (ggml-org) | ✅ | ✅ | ✅ |
| Qwen3.5-9B (unsloth) | ✅ | ✅ | ✅ |
| Mistral-Small-3.2-24B (unsloth) | ✅ | ✅ | ✅ |
| Ornith-1.9B (deepreinforce-ai) | ✅ | ✅ | ✅ |

Security correctness checks:
- The hard-error guard in `download()` (line 586) executes before `check_disk_space` and before `perform_download` is constructed — confirmed the multi-GB transfer cannot start if `sha256` is empty.
- TOCTOU: pinning a commit SHA in the URL (rather than `/resolve/main/`) removes the primary drift vector (branch moves), and the downloaded bytes are still verified against the pinned `sha256` after transfer — so even if HF's CDN or an intermediate proxy served something else, the post-download hash check catches it. Residual risk (HF rewriting history under an existing commit hash) is inherent to trusting HF as a registry at all and is not something this PR could reasonably close; it's the same trust boundary `apt`/`npm`/`cargo` all accept for pinned artifacts.
- Cleanup on verification failure: `tokio::fs::remove_file(&partial_path)` runs before returning `VerificationFailed` in both the mismatch and (unreachable) empty-hash branches — confirmed correct, no partial multi-GB file is left behind.

Acceptance criteria from issue #1497:
- [x] Default/locked models carry a real sha256 — confirmed for Ministral 8B via live API match
- [x] Empty hash is a hard error — confirmed as a hard error for **all** entries, not just non-trial ones (stricter than asked, and simpler)
- [x] Download URLs pin a commit SHA — confirmed via live API match against current HF commit heads
- [x] `bun run quality:fix`-equivalent (cargo fmt/clippy) and tests pass — reproduced independently: 43/43 tests, clean clippy, clean fmt

Recommendation: **Approve.**

malibio (commented):
Review recommendation: APPROVE (GitHub blocks self-approval on this PR since it's authenticated as the PR author — see the full review report posted as a PR comment above for details, verification methodology, and two non-blocking follow-up suggestions). No blocking issues found; independently verified 7/13 catalog entries against HuggingFace's live API and reproduced test/clippy/fmt results myself.

malibio:
### Review response

Both suggested improvements were already addressed in commit \`91c3006a\` (pushed before this review comment landed):

- ✅ **Unreachable empty-hash branch comment** — reworded the comment on \`perform_download\`'s \`expected_sha256.is_empty()\` branch to correctly state the guard lives in \`download()\` (not \`find_catalog_entry\`, which the old comment wrongly implied) and explicitly calls out that the branch is unreachable via any path reachable through the \`ModelManager\` trait today, kept only as defense-in-depth.
- ✅ **Commit-pin test tightened** — \`every_catalog_entry_pins_a_commit_not_a_moving_ref\` now extracts the ref segment after \`/resolve/\` and asserts it's exactly 40 hex characters, so the test's name and its guarantee now match (previously it only checked the literal string didn't contain \`/resolve/main/\`).

⏭️ **Skipped (by reviewer's own recommendation, out of scope for this PR):**
- Scheduled CI job to periodically re-validate catalog hashes against HuggingFace's live API for upstream drift — no CI currently exists in this repo, and the reviewer explicitly suggested this as a separate follow-up issue rather than part of this change.
- Nit about identical error message text between the two empty-hash guard sites — one of the two sites is unreachable dead code kept only as defense-in-depth; differentiating the strings would add no debugging value today since only one call site can ever fire.

Re-verified at current HEAD (\`91c3006a\`): \`cargo test -p nodespace-agent model_manager\` (43/43 pass), \`cargo clippy -p nodespace-agent --all-targets -- -D warnings -D clippy::unused-async\` (clean), \`cargo fmt -p nodespace-agent -- --check\` (clean).

No re-review needed — the changes are narrowly scoped comment/test clarifications with no behavior change, already reviewed in substance. Ready to merge pending your go-ahead.

Issue #1498, opened 2026-07-05
Windows Named Pipe has no ACL and is squattable


malibio:
## Overview

On Windows the daemon creates its Named Pipe with the default DACL and `first_pipe_instance(false)`. The "owning user only" guarantee that the Unix socket provides (mode 0o600) is therefore absent on Windows: the pipe is reachable by other local principals, and a malicious process can pre-create and squat it.

## Problem Statement

Every gRPC method is reachable by any local user through the default-DACL pipe, and `first_pipe_instance(false)` permits pipe-squatting (credential/prompt capture, response tampering). The trust model (OS socket/pipe permission is the entire authorization boundary) does not hold on Windows.

## Proposed Solution

Build a `SECURITY_ATTRIBUTES`/DACL restricting the pipe to the current user, and set `first_pipe_instance(true)` (fail loudly if already owned). Until then, Windows explicitly does not meet the trust model. Governed by ADR-052 (local trust boundary).

## Acceptance Criteria

- [ ] Named Pipe is created with a DACL restricting access to the current user
- [ ] `first_pipe_instance(true)` set; startup fails if the pipe is already owned
- [ ] Behavior matches the Unix 0o600 guarantee
- [ ] Code passes `bun run quality:fix`; tests pass

### Reference Files
- `packages/daemon/src/main.rs` (~315, `ServerOptions::new().first_pipe_instance(false).create(...)`; contrast Unix 0o600 at ~157/207)
- Decision: ADR-052

## Related Issues
- (none)

mstomar125:
**Not picked up today — Windows-only, unverifiable on our toolchain.**

This is security-critical Windows-only `unsafe` Win32 FFI (token→SID, SDDL, `SECURITY_ATTRIBUTES` lifetimes, first-instance special-case, a Send-across-await subtlety). The entire `#[cfg(windows)]` block **compiles out** on our macOS dev box, and PR CI can't catch it either — `release.yml` is the only workflow touching `windows-latest` and it triggers solely on `release`/`workflow_dispatch`, never on PRs. Merging it here would ship a security change to Windows **entirely unverified**. Needs a Windows box (or a manual release run) to build/run/test before it can be merged safely.

malibio:
[...]

PR #1548, opened 2026-07-06
Close the UDS socket bind-to-chmod race in the daemon

malibio:
Closes #1506

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3859,6 +3859,7 @@ dependencies = [
  "dirs 5.0.1",
  "hyper-util",
  "image",
+ "libc",
  "nodespace-agent",
  "nodespace-core",
  "nodespace-nlp-engine",
--- a/packages/daemon/Cargo.toml
+++ b/packages/daemon/Cargo.toml
@@ -42,6 +42,9 @@ tower = { version = "0.5", features = ["util"] }
 hyper-util = { version = "0.1", features = ["tokio"] }
 dirs = "5.0"
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [features]
 # Build the Pro edition binary. Compile with `--features pro` to set edition() to "pro".
 # Without this flag the binary reports edition "community" and the Pro guard in
--- a/packages/daemon/src/main.rs
+++ b/packages/daemon/src/main.rs
@@ -44,6 +44,35 @@ fn socket_path() -> std::path::PathBuf {
         .join("daemon.sock")
 }
 
+/// Binds a Unix domain socket with no window where it is exposed at the
+/// ambient umask (ADR-052 §3). `UnixListener::bind` creates the socket file
+/// honoring the process umask, so a plain `bind` then `set_permissions` leaves
+/// the socket briefly group/other-reachable if the umask is permissive. We
+/// narrow the umask to `0o177` (owner rw only) for the duration of the bind
+/// so the socket is created at `0o600` from the instant it appears, then
+/// restore the prior umask — `umask` is process-global, so the narrowed
+/// window must be as short as possible and restored even on error.
+#[cfg(unix)]
+fn bind_uds_owner_only(sock: &std::path::Path) -> Result<tokio::net::UnixListener> {
+    use std::os::unix::fs::PermissionsExt;
+
+    // SAFETY: umask(2) is a plain libc call; no pointers involved. It is
+    // process-global, so this must run before any concurrent thread binds
+    // a socket of its own — true here since this runs at startup before
+    // the tokio runtime spawns other tasks.
+    let prev_umask = unsafe { libc::umask(0o177) };
+    let result = tokio::net::UnixListener::bind(sock);
+    unsafe { libc::umask(prev_umask) };
+
+    let listener =
+        result.with_context(|| format!("Failed to bind Unix socket: {}", sock.display()))?;
+    // Defense-in-depth: the umask already guarantees 0o600 at creation, but
+    // set it explicitly in case the umask was somehow not honored.
+    std::fs::set_permissions(sock, std::fs::Permissions::from_mode(0o600))
+        .with_context(|| format!("Failed to set socket permissions: {}", sock.display()))?;
+    Ok(listener)
+}
+
 #[cfg(windows)]
 fn pipe_name() -> String {
     if let Ok(p) = std::env::var("NODESPACED_SOCKET") {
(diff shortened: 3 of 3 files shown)


malibio:
### Code Review Summary

This is a clean, well-scoped fix for the UDS bind→chmod race described in ADR-052 §3 and issue #1506. The `bind_uds_owner_only()` helper correctly narrows the process umask to `0o177` for the duration of `UnixListener::bind`, so the socket is created at `0o600` from the instant it appears on disk, restores the prior umask immediately afterward, and still applies `set_permissions(0o600)` defensively. It replaces the duplicated bind+chmod blocks in both `serve_headless()` and `serve_grpc()`. Verified locally: `cargo clippy -p nodespace-daemon --bin nodespaced --all-features -- -D warnings` is clean, `cargo fmt --check` is clean, and the new `uds_permission_tests::bind_uds_owner_only_is_0o600_and_restores_umask` test passes.

This is a net positive and addresses the acceptance criteria in #1506. I have one blocking-adjacent correctness note (asterisked below as an Improvement, not a Blocker, since it doesn't reopen the security window) and a couple of nits.

### Findings

#### Critical Issues
None. The core security property — no window where the socket is exposed at the ambient umask — is delivered correctly:
- `main()` calls exactly one of `serve_headless()` or `serve_grpc()` per process, never both, so there's no intra-process race on the process-global umask between concurrent binds.
- The narrowed-umask window wraps only the `bind()` call itself, not the subsequent `set_permissions`, which is the correct minimal scope per ADR-052 §3.
- Confirmed no peer-cred check (`SO_PEERCRED`/`LOCAL_PEERCRED`) and no Windows DACL work is present in this diff — correctly out of scope per the issue; those remain tracked under ADR-052 §2/§4 and issue #1498.

#### Suggested Improvements
- `packages/daemon/src/main.rs:67-72` (`bind_uds_owner_only`): if `bind()` succeeds but the defensive `set_permissions` call fails, the function returns `Err` and the `UnixListener` returned by `bind` is dropped without being used. `UnixListener::drop` does not unlink the socket file, so a stale (but still `0o600`, so not a security regression) socket file is left on disk, and the caller's next start-up attempt will need `remove_file` to succeed again (which it already does at the top of `serve_headless`/`serve_grpc`, so this is self-healing on next launch, but worth a one-line comment noting the leftover file is not a bug). Not a blocker — no security property is violated since permissions are already correct at that point — but worth a short comment so a future reader doesn't have to re-derive that the leftover file is benign.
- Test coverage is good for the happy path (permissive umask → 0o600 result, umask restored). Consider also asserting the umask is restored even when `bind()` fails (e.g., bind to a path inside a nonexistent directory) — this exercises the branch where `result.with_context(...)?` propagates `Err` and confirms umask restoration isn't accidentally coupled to the success path. Given the current code structure (`libc::umask(prev_umask)` runs unconditionally right after `bind`, before the `?`), this is already correct — the additional test would just make that invariant explicit and regression-proof.

#### Nitpicks
- Nit: `packages/daemon/src/main.rs:59-62`: the `// SAFETY:` comment says "this must run before any concurrent thread binds a socket of its own — true here since this runs at startup before the tokio runtime spawns other tasks." This is accurate for the current two call sites, but it's an invariant on the *caller*, not something `bind_uds_owner_only` itself enforces. If a future call site invokes this from inside an already-running multi-threaded runtime (e.g. a hot-reload/re-bind path), the safety argument silently breaks. Consider a `debug_assert` or doc-level `# Safety` note on the function itself (not just inline) making this a documented precondition of the function, so it travels with the signature rather than living only in the current call sites' comment.
- Nit: the module doc comment on `bind_uds_owner_only` and the test module both independently explain the "umask is process-global" rationale — mild duplication, but justified here since the test comment is explaining test isolation (why one `#[tokio::test]` not two) while the function comment explains the production safety window. Fine as-is.
- Nit: `libc = "0.2"` is unpinned within the `0.2` line like the other three crates that already depend on it (`packages/cli`, `packages/agent`, `packages/nlp-engine`) — consistent with existing convention, no action needed.

### Verification Performed
- `git diff origin/main...HEAD` reviewed in full.
- `cargo clippy -p nodespace-daemon --bin nodespaced --all-features -- -D warnings` — clean.
- `cargo fmt -p nodespace-daemon -- --check` — clean.
- `cargo test -p nodespace-daemon --bin nodespaced uds_permission_tests` and full daemon binary test suite — all pass, no umask cross-test interference (only one test touches umask in this binary).
- Confirmed `serve_headless`/`serve_grpc` are mutually exclusive per-process (no concurrent-bind hazard for the process-global umask mutation).
- Confirmed no peer-cred or Windows DACL code was introduced or silently dropped — matches the stated scope boundary (issue #1506 vs. #1498 and ADR-052 §2/§4).

### Recommendation
Approve. This is a focused, correct fix for the stated problem, matches existing dependency patterns, has a passing regression test, and does not bundle unrelated changes. The one improvement noted (stale socket file on `set_permissions` failure) is not security-relevant and does not block merge.

malibio (commented):
Review complete — net positive, recommend merge. See detailed findings in the PR comment above. No blocking issues; one non-blocking improvement noted (stale socket file if set_permissions fails post-bind) and two nits. Clippy/fmt clean, test passes, scope correctly excludes peer-cred/Windows work (tracked separately under #1498).

malibio:
## Addressed Review Feedback

**✅ Addressed** (1):
- Promoted the `# SAFETY` comment's concurrency argument on `bind_uds_owner_only()` into a documented precondition, naming the two call sites (`serve_headless`, `serve_grpc`) that satisfy it. (4dbb7088)

**⏭️ Skipped** (2, both explicitly non-blocking suggestions):
- Stale socket file if `set_permissions` fails after a successful `bind`: already self-healing (next startup's `remove_file` call clears it), and the reviewer confirmed it's not a security regression — the file would still be `0o600`. Not worth added error-path complexity for a benign edge case.
- Additional test coverage for the umask-restoration-on-bind-failure path: the reviewer's own note called current coverage "fine as-is."

No 🔴 Critical or 🟡 Important findings were raised — the review's recommendation was to merge as-is.

## Re-Review Decision

**Rationale**: The only change made is a documentation/comment edit (no logic change) in response to a non-blocking nit. Verified `cargo clippy -p nodespace-daemon --all-targets -- -D warnings`, `cargo fmt --check`, `cargo test -p nodespace-daemon`, and the full `test:all` suite (frontend + Rust + e2e daemon round-trip) all pass. Ready for merge pending final approval.

PR #1550, opened 2026-07-06
Replace the regex-based SVG sanitizer with a DOMPurify SVG profile

malibio:
Closes #1516

### DIFF
--- a/packages/desktop-app/src/lib/services/mermaid-render.ts
+++ b/packages/desktop-app/src/lib/services/mermaid-render.ts
@@ -100,7 +100,10 @@ function getDOMPurify() {
   // DOMPurify treats `style` as a URI-safe attribute and does not parse its CSS content, so
   // it won't strip a javascript: URI hidden inside url(...) in an inline style. Mermaid relies
   // on inline styles for theming, so we can't forbid the attribute outright — scrub dangerous
-  // CSS constructs from it directly instead.
+  // CSS constructs from it directly instead. Known gap: CSS backslash-escapes/whitespace
+  // splicing (e.g. "jav\61script:") can obfuscate past this regex. Accepted for now since the
+  // app's CSP (script-src 'self', object-src 'none') already blocks javascript: URIs from
+  // executing in any modern webview — revisit if that CSP is ever relaxed.
   purify.addHook('uponSanitizeAttribute', (_node, data) => {
     if (data.attrName === 'style') {
       data.attrValue = data.attrValue.replace(/(?:javascript|vbscript):/gi, '').replace(/expression\s*\(/gi, '');
--- a/packages/desktop-app/src/lib/services/mermaid-render.ts
+++ b/packages/desktop-app/src/lib/services/mermaid-render.ts
@@ -1,4 +1,5 @@
 import { createLogger } from '$lib/utils/logger';
+import createDOMPurify from 'dompurify';
 import type { Mermaid } from 'mermaid';
 
 const log = createLogger('MermaidRender');
@@ -89,18 +90,27 @@ function buildThemeVariables(isDark: boolean) {
   };
 }
 
+// Build a DOMPurify instance bound to document.defaultView rather than relying on the module-
+// level default export or the global `window` binding. In tests each file gets a fresh
+// happy-dom document, and other test files' global-`window` stubbing/teardown can leave the
+// global `window` binding stale/detached from the live document by the time this runs, even
+// though `document` itself is always current.
+function getDOMPurify() {
+  const purify = createDOMPurify(document.defaultView ?? undefined);
+  // DOMPurify treats `style` as a URI-safe attribute and does not parse its CSS content, so
+  // it won't strip a javascript: URI hidden inside url(...) in an inline style. Mermaid relies
+  // on inline styles for theming, so we can't forbid the attribute outright — scrub dangerous
+  // CSS constructs from it directly instead.
+  purify.addHook('uponSanitizeAttribute', (_node, data) => {
+    if (data.attrName === 'style') {
+      data.attrValue = data.attrValue.replace(/(?:javascript|vbscript):/gi, '').replace(/expression\s*\(/gi, '');
+    }
+  });
+  return purify;
+}
+
 export function sanitizeSvg(svg: string): string {
-  // Remove script tags
-  let result = svg.replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, '');
-  // Remove event handler attributes (all quoting styles: "...", '...', or unquoted)
-  result = result.replace(/\s+on\w+(\s*=\s*("[^"]*"|'[^']*'|[^\s>]*))?/gi, '');
-  // Strip javascript: URIs from navigable/action attributes (mirrors data: URI treatment above)
-  result = result.replace(/(href|src|xlink:href|action|formaction)\s*=\s*["']\s*javascript\s*:[^"']*/gi, '$1=""');
-  // Strip data: URIs from href, src, and xlink:href attributes (can carry executable payloads)
-  result = result.replace(/(href|src|xlink:href)\s*=\s*["']\s*data:[^"']*/gi, '$1=""');
-  // Strip url(javascript:...) from inline styles and <style> blocks
-  result = result.replace(/url\s*\(\s*['"]?\s*javascript\s*:[^)]*\)/gi, 'url(about:blank)');
-  return result;
+  return getDOMPurify().sanitize(svg, { USE_PROFILES: { svg: true, svgFilters: true } });
 }
 
 // Returns sanitized SVG string, or null on failure
--- a/packages/desktop-app/src/tests/services/mermaid-render.test.ts
+++ b/packages/desktop-app/src/tests/services/mermaid-render.test.ts
@@ -11,19 +11,17 @@ vi.mock('mermaid', () => ({
 
 describe('sanitizeSvg', () => {
   it('removes script tags from SVG output', () => {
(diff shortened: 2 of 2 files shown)


malibio:
### Code Review Summary

**Review Type:** Initial review (no prior reviews/comments on PR #1550).

This PR swaps a hand-rolled regex `sanitizeSvg` for DOMPurify's SVG profile in `mermaid-render.ts`, and it's a clear net positive: it replaces a brittle, incomplete blacklist with a maintained parser-based sanitizer sitting directly upstream of an `innerHTML` sink (`code-block-node.svelte` injects `mermaidSvg`/`previewSvg` via `.innerHTML`, confirmed by reading that file). I independently reproduced every material claim in the PR description rather than taking it at face value — the per-call `getDOMPurify()` workaround, the `style`-attribute gap, and the full test suite pass are all real. One narrow, non-blocking gap found in the custom `style` regex hook (detailed below).

### Requirements Check (issue #1516)

- [x] SVG is sanitized via DOMPurify SVG profile, not regex — `sanitizeSvg` (`mermaid-render.ts:112-114`) now calls `getDOMPurify().sanitize(svg, { USE_PROFILES: { svg: true, svgFilters: true } })`; the old regex chain is fully deleted, not left as a fallback.
- [x] Code passes `bun run quality:fix`; tests pass — verified independently (see below). Rust-side `quality:fix` failed in my environment on an unrelated pre-existing issue (missing `nodespaced` binary artifact for the Tauri build script); this PR touches zero Rust files, so it's not attributable to this change. Frontend `quality:fix` (eslint + svelte-check) ran clean: 0 errors/warnings across 1713 files.

Both acceptance criteria are met.

### Independent Verification

- **Full test suite**: ran `bun run test` from `packages/desktop-app` — 146 files / 3941 tests, all passing, matching the PR's claim exactly.
- **Reproduced the root-cause bug the workaround exists for**: I temporarily patched `sanitizeSvg` to use the default DOMPurify singleton (`DOMPurify.sanitize(...)` bound at import time) instead of `getDOMPurify()`, then reran the full suite. Result: `TypeError: default.sanitize is not a function` in `mermaid-render.test.ts`, with 13 test failures — but only when run as part of the full 146-file suite (it passes fine in isolation with a handful of files). This confirms the `singleFork: true` vitest config plus `external-links.test.ts`'s `globalThis.window = {...}` stub/restore pattern really does leave the global `window` binding in a state that breaks the singleton import by the time `mermaid-render.test.ts` runs later in the same process. The `getDOMPurify()` per-call construction from `document.defaultView` is a legitimate, verified fix for a real problem, not defensive-programming folklore. I restored the file immediately after (working tree confirmed clean via `git status`).
- **Instance-per-call design**: Not a performance concern — this is on the mermaid-diagram-render path, not a hot loop, and DOMPurify instance construction is cheap. It also isn't a hook-leak risk: `createDOMPurify()` returns a fresh, independent instance each call (verified via the DOMPurify API contract — hooks are attached to the instance object returned, no shared/module-level registry), so hooks don't accumulate across calls.
- **Inconsistency worth noting**: `chat-markdown.svelte` (untouched by this PR) still uses the plain default DOMPurify singleton import directly and its tests pass fine — that file isn't exercised in a way that trips the same staleness bug. This confirms the workaround is correctly scoped to the one call site that's actually affected, but it does mean the codebase now has two different DOMPurify usage patterns for a similar security-boundary purpose. Not a blocker, but worth a forward-looking note (see Suggested Improvements).
- **Style-attribute regex hook**: probed the `uponSanitizeAttribute` hook (`javascript|vbscript:` / `expression(...)` strip) against several bypass attempts using real DOMPurify output (via a scratch vitest file, removed before finishing):
  - Case variation (`URL(JavaScript:...)`) — stripped correctly.
  - `expression(...)` with an embedded newline — stripped correctly.
  - HTML-entity-encoded `javascript:` inside `url()` (e.g. `&#106;avascript:`) — survives the regex, but this is **not exploitable**: browsers don't HTML-entity-decode the contents of inline CSS `url()` values, so this isn't a real bypass, just a hook limitation against a non-threat.
  - **CSS backslash-escapes** (`jav\61script:`, valid CSS for `java` + `a`) and **whitespace/tab splicing inside the token** (`java\tscript:`) both survive the regex untouched, and a real CSS parser in a browser would resolve the escape back to `javascript:`. This is a genuine, narrow bypass of the custom hook.
  - Mitigating factors that lower severity: (1) the app's Tauri CSP sets `script-src 'self'`, and `object-src 'none'`, so even a `javascript:` URI surviving into a `style` attribute is not run by the CSP-governed webview in any modern Chromium/WebKit context — `javascript:` in CSS `url()` hasn't been a live code-execution vector in mainstream engines for years (it was primarily an old-IE quirk); (2) DOMPurify itself remains the primary sanitization boundary for the actual attack surface (tags, event handlers, `href`/`src` protocols) — this hook is a narrow supplementary patch for one specific gap DOMPurify leaves in `style`, not the main defense.

### Findings

#### Critical Issues
None.

#### Suggested Improvements
- **`mermaid-render.ts:104-108`** (style regex hook): Consider hardening the CSS-escape bypass, e.g. by stripping/normalizing backslash-escapes and collapsing whitespace within suspicious tokens before matching, or by using a small CSS-aware check rather than a raw substring regex. Given the mitigating CSP and that this is demonstrably not exploitable in the shipped Tauri webview today, this is an improvement rather than a blocker — but if diagram-sharing (the motivating future feature per issue #1516) ever runs this same code in a context with a laxer CSP or a WebView with legacy CSS quirks, it's worth tightening ahead of that.
- **DOMPurify usage inconsistency**: `chat-markdown.svelte` uses the default singleton import while `mermaid-render.ts` now builds a fresh instance per call. Since the singleton-staleness bug is a symptom of test-suite global mutation (`external-links.test.ts`'s `globalThis.window` reassignment) rather than something inherent to `mermaid-render.ts`, the same latent fragility could resurface if `chat-markdown.svelte`'s tests are ever reordered or a new far-off test file starts stubbing `window`. Worth a follow-up issue to either (a) fix `external-links.test.ts` to scope its `window` stubbing (e.g. `vi.stubGlobal('window', ...)` with `vi.unstubAllGlobals()` in `afterEach`, which restores more reliably than manual save/restore) so the root cause goes away and both call sites can safely use the singleton, or (b) standardize both call sites on the `getDOMPurify()`-per-call pattern for consistency. Right now the fix is correctly localized to the affected file, which is a reasonable pragmatic choice for this PR's scope — but the divergence should be tracked rather than left as a silent inconsistency.

#### Nitpicks
- **Nit:** `mermaid-render.ts:104-108`: the hook mutates `data.attrValue` with two chained `.replace()` calls; a short comment noting these are deliberately narrow (CSS-escape/whitespace obfuscation is a known residual gap, mitigated by CSP) would help a future reader calibrate expectations without re-deriving it, since the existing comment explains *why* the hook exists but not its known limits.
- **Nit:** Test file — `expect(result).not.toContain('javascript')` in "removes javascript: URIs with whitespace after colon" (line 58, pre-existing, unchanged) is slightly stronger than needed given the URI already asserted `not.toContain('javascript:')` above it; harmless, just belt-and-suspenders duplication carried over from before this PR.

### Testing
Test coverage is adequate for this change. The updated suite now covers: script tag removal (single-line + multiline, with the sibling-drop behavior correctly re-asserted rather than papered over), event handlers (double/single-quoted/unquoted), `javascript:` URIs (plain and whitespace-obfuscated), the two new cases this PR adds (`data:` URI in `href`, `url(javascript:...)` in inline `style`), legitimate-content preservation, and empty-input handling. The rewritten assertions correctly adapt to DOMPurify's real serialization behavior (attribute order, self-closing vs. explicit closing tags) instead of asserting brittle exact-string equality against the old regex sanitizer's output shape — this is the right way to update these tests, since they now assert properties of the sanitized output rather than accidental implementation details of the old regex. No test exercises the CSS-escape gap noted above; given that gap is a suggested improvement rather than a blocker, that's an acceptable scope boundary for this PR, not a coverage defect.

This is a clean, well-scoped, atomic security improvement that meets both acceptance criteria, is backed by a real (reproduced) test-isolation bug fix rather than speculative defensive code, and has no critical findings. The one regex-hook gap identified is a narrow, currently non-exploitable residual (mitigated by the app's CSP) worth a follow-up, not a blocker.

malibio:
## Addressed Review Feedback

**✅ Addressed (1):**
- **Nit** — Added a comment on the `uponSanitizeAttribute` style hook (`mermaid-render.ts`) documenting the known CSS backslash-escape/whitespace-splicing gap (e.g. `jav\61script:`) and why it's accepted for now (CSP already blocks `javascript:` execution in the webview).

**⏭️ Skipped (2 suggestions + 1 nitpick), per reviewer's own framing:**
- **Harden the CSS-escape bypass** — reviewer explicitly scoped this as "worth tightening ahead of" a future diagram-sharing feature that doesn't exist yet. Speculative hardening against a non-current threat model is out of scope for this PR.
- **DOMPurify singleton-vs-per-instance inconsistency with `chat-markdown.svelte`** — reviewer explicitly recommended a *separate follow-up issue*, not a change here. Fixing `external-links.test.ts`'s global `window` stubbing is unrelated to this PR's scope (SVG sanitization).
- **Redundant `not.toContain('javascript')` assertion** in a pre-existing test — reviewer flagged this as pre-existing/unchanged by this PR and harmless.

No critical or important findings were raised, so this is a documentation-only follow-up commit. `quality:fix` (frontend) clean, full test suite green (146 files / 3941 tests unit, 3 e2e files / 16 tests).

## Re-Review Decision

**Rationale**: The only change is a code comment — no logic, behavior, or test changes. The original review found no critical or important issues, and the two substantive suggestions were explicitly deferred by the reviewer to future work rather than requested for this PR.

Ready for merge.

PR #1552, opened 2026-07-06
Establish one daemon-readiness contract to stop per-call-site startup races

malibio:
Closes #1525

### DIFF
--- a/packages/desktop-app/src-tauri/src/daemon_setup.rs
+++ b/packages/desktop-app/src-tauri/src/daemon_setup.rs
@@ -423,7 +423,10 @@ pub async fn check_daemon_socket(socket_path: &Path) -> DaemonStatus {
 }
 
 /// Poll the socket until the daemon is healthy or the timeout expires.
-async fn wait_for_daemon(socket_path: &Path, max_wait: Duration) -> DaemonStatus {
+///
+/// `pub` (not `pub(crate)`) so the readiness integration test — which lives
+/// in `tests/`, a separate crate — can drive it against a real daemon.
+pub async fn wait_for_daemon(socket_path: &Path, max_wait: Duration) -> DaemonStatus {
     let deadline = tokio::time::Instant::now() + max_wait;
     loop {
         let status = check_daemon_socket(socket_path).await;
--- a/packages/desktop-app/src-tauri/src/lib.rs
+++ b/packages/desktop-app/src-tauri/src/lib.rs
@@ -69,6 +69,17 @@ fn frontend_log(line: String) {
 /// to decide whether to show an error state (Issue #1179).
 #[tauri::command]
 async fn check_daemon_status() -> String {
+    daemon_status_body().await
+}
+
+/// The body of [`check_daemon_status`], factored out and made `pub` so the
+/// readiness integration test in `tests/` — a separate crate — can call the
+/// exact same logic the Tauri command invokes, including its
+/// `resolve_socket_path()` call. Kept out of the `#[tauri::command]`-
+/// annotated function itself: that macro generates hidden crate-scoped
+/// items keyed to the function's identifier, and marking the annotated
+/// function `pub` collides with them (`E0255`, defined multiple times).
+pub async fn daemon_status_body() -> String {
     #[cfg(any(target_os = "macos", target_os = "linux", target_os = "windows"))]
     {
         use daemon_setup::{check_daemon_socket, DaemonStatus};
--- a/packages/desktop-app/src-tauri/tests/daemon_readiness_test.rs
+++ b/packages/desktop-app/src-tauri/tests/daemon_readiness_test.rs
@@ -28,7 +28,7 @@ mod support;
 use std::time::Duration;
 
 use nodespace_app_lib::daemon_setup::{check_daemon_socket, wait_for_daemon, DaemonStatus};
-use support::SpawnedDaemon;
+use support::{EnvGuard, SpawnedDaemon};
 
 /// Serializes tests in this file that mutate the process-global
 /// `NODESPACED_SOCKET` env var, mirroring the single-threaded discipline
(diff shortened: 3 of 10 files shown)


malibio:
### Code Review Summary

Net positive, well above the bar for this repo. This closes a real, previously-unrecoverable startup race (lost first `daemon-status` event with no re-fire path) with a clean transport-agnostic `DaemonStatusSource` abstraction, and — unusually for infra work — backs it with integration coverage against a *real* spawned daemon rather than mocks on both the Rust and TypeScript sides. The e2e work also caught and fixed a genuine second-order bug in the test harness itself (gRPC-js proxy channel backoff) rather than papering over the flake, which is exactly the right instinct. All three acceptance criteria in #1525 are met: single readiness contract shared by `check_daemon_socket`/`wait_for_daemon` and the Tauri command body, an integration test covering not-ready → recovered, and clean lint/test runs (confirmed: `bun run test` 146/146 files, 3943/3943 tests passing; `cargo clippy --lib --tests -p nodespace-app` clean; `cargo test -p nodespace-app --test daemon_readiness_test --no-run` compiles).

Nothing here blocks merge. Two findings worth a look, one is genuinely dead code.

### Findings

#### Suggested Improvements

- `packages/desktop-app/src-tauri/tests/daemon_readiness_test.rs:92-113`, `:118-136`: the `NODESPACED_SOCKET` env-var save/restore in `check_daemon_status_command_agrees_with_the_real_daemon` and `check_daemon_status_command_reports_not_running_for_an_unbound_socket` is not panic-safe — the restore runs as plain code after the assertions, not in a `Drop` guard or `catch_unwind`. If either assertion fails (which is exactly the scenario a broken build would produce), the env var leaks unrestored, and since integration-test binaries run all tests in one process by default, a later test in this file (or a future one added to it) could silently read the leaked value. This mirrors the pre-existing pattern in `grpc_client.rs`'s `resolve_socket_path_honors_env_override_then_falls_back` test, so it's consistent with house convention rather than a new regression — but the new file adds an explicit `tokio::sync::Mutex` specifically because integration-test binaries parallelize by default (correct reasoning), which makes the leak-on-panic gap more likely to actually bite here than in the single-test unit-test module it's modeled on. A `struct EnvGuard { prev: Option<OsString> }` with a `Drop` impl that restores would close this for a few extra lines and could be reused by both tests.

- `packages/desktop-app/src/tests/e2e/daemon-readiness.e2e.ts`: the final commit (`c0e9f4a3`) reverted the prior commit's (`9a8f272d`) widening of the schemas-landed poll window from 5s back down, and dropped its diagnostic error message (direct `getAllSchemas()` call surfaced on timeout), reasoning that `waitUntilProxyReady()` now proves the round-trip works before the poll starts. That's a defensible call given the root cause (proxy backoff) is now waited-out up front, but the commit message for `c0e9f4a3` doesn't mention that it's reverting `9a8f272d`'s hardening — someone reading `9a8f272d` in isolation later (`git log`/`git blame`) will believe the 20s window + diagnostic is still the current state. Not a functional issue, just a paper trail nit; worth a one-line mention in commit history or PR description if you want future-you to not have to re-discover this by diffing commits.

#### Nitpicks

- Nit: `packages/desktop-app/src/tests/e2e/daemon-harness.ts:394` (`waitUntilDaemonReady`) is dead code — its only call site in the e2e test was replaced by `waitUntilProxyReady()` in the last commit (per the module doc comment in `daemon-readiness.e2e.ts:20-21`, which explains *why* the socket-only check is insufficient, but the method itself was left behind rather than removed). `isDaemonReachable()` is not redundant — it's still live, used by `harnessStatusSource()`'s `getCurrent()` — so no action needed there.

- Nit: the `daemon_status_body` factoring in `lib.rs` (extracting the `#[tauri::command]` body into a `pub async fn` to dodge the `E0255` hidden-item collision) is sound and verified — `cargo clippy --lib --tests -p nodespace-app` and a standalone `cargo test -p nodespace-app --test daemon_readiness_test --no-run` both compile clean, and the one-line `check_daemon_status` wrapper means the annotated command and the tested body can't drift. Nothing to fix, just confirming the approach holds up since it's easy to get this kind of macro-collision workaround subtly wrong.

- Nit: `packages/desktop-app/src/lib/services/daemon-status.ts`'s `startDaemonStatusListener` fires both `getCurrent()` (pull) and `subscribe()` (push) concurrently with no defined ordering between them — if a push event and the initial pull resolve out of order, the later one to settle wins regardless of which observed the more recent state. In practice this is a narrow window (both are near-instant on startup) and `applyStatus` is idempotent per payload, so this isn't worth blocking on, just flagging as a place a future higher-frequency push source could introduce a stale-overwrite.

#### Process note (not a blocker for this PR)

`bun run test:all` does not execute either new test: `rust:test` runs `cargo test --lib -p nodespace-core` (unit tests in the core lib crate only, not `nodespace-app`'s `tests/` integration tests), and `test:e2e` isn't part of `test:all` at all. Combined with the repo having no PR-gating CI workflow yet (only `release.yml` exists), both the new Rust integration test and the new TS e2e test currently only run when someone remembers to invoke them manually — which is a little ironic given this PR's own commit history shows the e2e test flaked once already under real-world load. This is a pre-existing gap in `test:all`'s scope rather than something #1525 was scoped to fix, but since it directly affects whether this PR's new coverage actually protects anything going forward, it's worth a fast follow-up issue if one doesn't already exist.

malibio:
Addressed both fixable review findings in 6d6a6159:
- Added a `Drop`-based `EnvGuard` (in `tests/support/mod.rs`) so a panicking assertion in the Rust integration test can't leak the mutated `NODESPACED_SOCKET` env var to a later test.
- Removed `daemon-harness.ts`'s dead `waitUntilDaemonReady()`, left behind after its only call site was replaced by `waitUntilProxyReady()`.

Not addressed (both explicitly called non-blocking, and out of this issue's scope):
- The `test:all`/`test:e2e` wiring gap (neither new test runs automatically outside the pre-push hook) — this is a pre-existing gap in the test-gating setup, not something #1525 was scoped to fix. Happy to file a follow-up if wanted.
- The commit-message paper-trail nit on the poll-window revert — cosmetic, no code change needed.

Full test suite + pre-push gate re-verified green after these changes.

PR #1555, opened 2026-07-06
Convert remaining $effect misuse to derived values and event handlers

malibio:
Closes #1511

### DIFF
--- a/packages/desktop-app/src/lib/components/layout/app-shell.svelte
+++ b/packages/desktop-app/src/lib/components/layout/app-shell.svelte
@@ -61,21 +61,17 @@
 
   // Pro re-login prompt (T18, #1304). The daemon emits AUTH_REQUIRED when the
   // persisted refresh token can't be renewed; we surface a modal offering a
-  // fresh sign-in or working offline. `reloginDismissed` is per-episode so
-  // "Work Offline" doesn't keep nagging, and is re-armed once the daemon
-  // leaves the auth-required state so a later failure prompts again.
-  let reloginDismissed = $state(false);
+  // fresh sign-in or working offline. Dismissal is tracked per-episode via
+  // proSync.authRequiredEpisode (bumped each time the daemon re-enters
+  // auth-required), so the modal re-arms automatically with no effect needed.
+  let dismissedEpisode = $state(-1);
   let reloginPending = $state(false);
 
   let showReloginModal = $derived(
-    proSync.isPro && proSync.state === 'auth-required' && !reloginDismissed
+    proSync.isPro &&
+      proSync.state === 'auth-required' &&
+      dismissedEpisode !== proSync.authRequiredEpisode
   );
-
-  $effect(() => {
-    if (proSync.state !== 'auth-required' && reloginDismissed) {
-      reloginDismissed = false;
-    }
-  });
 
   // Recovered Items (core#1303): once the daemon's probe confirms Pro tier, load
   // the local-only recovery log. If the daemon preserved any conflict "losers",
@@ -114,7 +110,7 @@
   }
 
   function handleReloginWorkOffline() {
-    reloginDismissed = true;
+    dismissedEpisode = proSync.authRequiredEpisode;
   }
 
   /**
--- a/packages/desktop-app/src/lib/components/layout/navigation-sidebar.svelte
+++ b/packages/desktop-app/src/lib/components/layout/navigation-sidebar.svelte
@@ -62,12 +62,13 @@
   let navElement: HTMLElement | null = $state(null);
   let subPanelElement: HTMLElement | null = $state(null);
 
-  // Close sub-panel when sidebar collapses
-  $effect(() => {
-    if (isCollapsed && subPanelOpen) {
+  // Close sub-panel when the sidebar collapses (sidebar can only collapse via handleToggleSidebar)
+  function handleToggleSidebar() {
+    toggleSidebar();
+    if (subPanelOpen) {
       collectionsState.clearSelection();
     }
-  });
+  }
 
   // Click-outside handler for sub-panel dismissal
   $effect(() => {
@@ -253,7 +254,7 @@
   <!-- Hamburger menu button -->
   <button
     class="hamburger-button"
-    onclick={toggleSidebar}
+    onclick={handleToggleSidebar}
     aria-label={isCollapsed ? 'Expand sidebar' : 'Collapse sidebar'}
     aria-expanded={!isCollapsed}
   >
--- a/packages/desktop-app/src/lib/components/schema/generic-schema-form.svelte
+++ b/packages/desktop-app/src/lib/components/schema/generic-schema-form.svelte
@@ -35,11 +35,9 @@
 
   let { nodeId, schema, autoOpen = false }: { nodeId: string; schema: SchemaNode; autoOpen?: boolean } = $props();
 
-  let isOpen = $state(false);
-  // Sync isOpen with the autoOpen prop (captures reactive updates, not just initial value)
-  $effect(() => {
-    if (autoOpen) isOpen = true;
-  });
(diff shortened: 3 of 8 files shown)


malibio:
### Code Review Summary

This PR does what it says: seven `$effect` misuses (per ADR-049's bright-line rule — effects reserved for pushing state into non-reactive external systems, never deriving values, mirroring state, or read-and-write sequencing) are converted to `$derived` values or event handlers. I read the full diff against `origin/main`, traced each of the two flagged risk areas (the `pro-sync.svelte.ts` episode counter and the `base-node-viewer.svelte` relocation) end-to-end through their call sites, verified the two "left unchanged" `toggleSidebar()` sites are unreachable while expanded, and ran `bun run test` (3948/3948 passed) and `bun run quality:fix` (0 errors/warnings, no incidental reformatting). This is a net positive, low-risk, well-scoped refactor. One gap keeps it from an unqualified approve.

### Findings

#### Suggested Improvements

- **`packages/desktop-app/src/lib/stores/pro-sync.svelte.ts` (`authRequiredEpisode`, `setState`)**: This is not a mechanical, behavior-preserving refactor like the other six sites — it changes the reactive *contract* from a boolean dismiss-flag to a monotonic counter with state-transition detection (`next === 'auth-required' && this.state !== 'auth-required'`). I traced all three call sites (`tier-detected`'s `initial_status` path, `sync:status`, `signOut()`) and the logic is correct: initial `state` is `'unspecified'`, so the very first transition into `'auth-required'` (including via `initial_status` on cold start) correctly bumps the counter from 0→1, and `dismissedEpisode`'s initial value of `-1` can never accidentally equal an unbumped `0`. I didn't find an edge case that breaks it. But this is exactly the kind of stateful transition-detection logic that regresses silently, and the existing `pro-sync.test.ts` (3 tests, untouched by this diff) has zero coverage of `authRequiredEpisode` or `setState`. Given ADR-049 itself argues a crisp rule is only valuable if reviewers/tests can verify it, I'd add 2-3 unit tests directly against `ProSyncStore`: (a) transitioning into `auth-required` bumps the counter, (b) staying in `auth-required` across repeated `sync:status` events does not re-bump it, (c) leaving and re-entering `auth-required` bumps it again distinctly. This is cheap insurance on the one site in the PR that's genuinely new logic rather than a pattern translation.

- **`packages/desktop-app/src/lib/components/schema/generic-schema-form.svelte` (`isOpen` initial value)**: Verified the safety argument holds, but it's non-obvious enough to warrant a guard comment. `GenericSchemaForm` is only rendered once `genericSchema` is truthy (`base-node-viewer.svelte:1440`: `{:else if currentViewedNode && nodeId && genericSchema && isCustomSchemaType(...)}`), and `genericSchema` starts `null` and is populated asynchronously by `loadGenericSchema()` before this branch ever renders — so by the time `GenericSchemaForm` mounts and captures `autoOpen` as its `isOpen` initial value, `hasTitleTemplate` (and therefore `autoOpen`) is already the final, correct value for that node. There's no code path where `genericSchema` (and thus `autoOpen`) changes again for an already-mounted `GenericSchemaForm` instance — the component is destroyed and recreated on every `nodeId` change via the `{#key}` in `pane-content.svelte`, and `genericSchema` is unconditionally reset to `null` on that same transition (`base-node-viewer.svelte:328`). The reasoning is sound, but it depends on an invariant (schema fully resolved before this component ever mounts) that lives in a different file and isn't enforced by a type or test. **Nit-adjacent suggestion**: a one-line comment at the `GenericSchemaForm` render site (`base-node-viewer.svelte:1443`) noting "autoOpen is captured once at mount; safe only because this branch doesn't render until genericSchema is loaded" would save the next person from re-introducing the effect when refactoring the conditional.

#### Nitpicks

- **`base-node-viewer.svelte`**: The removed effect's `tick().then(() => updateTabTitle(title))` deferral and the new `untrack(() => updateTabTitle(computedTabTitle))` solve the same `state_unsafe_mutation` problem via different mechanisms (defer vs. untrack-to-avoid-circular-dependency). Confirmed the `computedTabTitle` formula is byte-for-byte identical to the old effect's inline `title` computation, and that Svelte's effect scheduling isn't declaration-order-dependent, so the TDZ-driven relocation further down the file doesn't change firing order relative to the other `onMount`/effect blocks. No issue found, just noting it for the record since it was flagged as a specific risk to check.
- **`collection-node-viewer.svelte`**: the pre-existing `loadCollectionData(nodeId)` effect (line 51-53) lacks the `untrack()` wrapper that the analogous data-load effect in `query-node-viewer.svelte` uses. Out of scope for this PR (untouched by the diff, and the PR correctly limited itself to the `onTitleChange` site), but worth a follow-up since `query-node-viewer.svelte` is cited elsewhere in this same PR as the reference pattern.

### Verification performed

- Read full diff (`git diff origin/main...HEAD`), all 7 files.
- Confirmed all three "left unchanged" decisions are reasonable: `table-view.svelte`'s page-reset-on-new-`nodeIds` (legitimate local-UI-reset pattern), `query-node-viewer.svelte` (already correct, used as reference), `calendar.svelte` (documented third-party bits-ui interop workaround, correctly deferred pending live browser verification).
- Grepped all 9 named files for remaining `$effect` blocks; none of the untouched ones are derivation/mirroring/read-write anti-patterns — they're DOM/event-listener registration, timers, or guarded one-shot imperative loads.
- Confirmed `navigation-sidebar.svelte`'s two untouched `toggleSidebar()` call sites (lines ~415, ~507) are both inside `{:else}` branches of `{#if !isCollapsed}`, i.e. only reachable while already collapsed — they cannot trigger the collapse-and-clear-selection condition `handleToggleSidebar()` now owns.
- `bun run test`: 147 files, 3948 tests, all passed.
- `bun run quality:fix`: 0 ESLint errors/warnings, 0 svelte-check problems, tsc clean, cargo fmt + clippy clean; `git status` after showed no incidental changes.

### Recommendation

Approve with the one suggested follow-up (test coverage for `authRequiredEpisode`/`setState` in `pro-sync.svelte.ts`) — not a blocker given this is pre-release code with no users and the transition logic checks out on manual trace, but worth doing given it's the one site in this PR that's new behavior rather than a mechanical pattern conversion.

malibio:
Review verdict: **Approve** (net positive, low risk). GitHub won't let me submit a formal self-review approval on my own PR (`gh pr review` fails with "Can not approve your own pull request"), so recording the verdict here instead — see the review above for full findings. No blockers found; one non-blocking follow-up suggested (test coverage for the new authRequiredEpisode transition logic in pro-sync.svelte.ts).

malibio:
## Addressed review feedback

**✅ Addressed:**
- Added 4 unit tests to `pro-sync.test.ts` covering the `authRequiredEpisode`/`setState` transition logic: bumps on entry into auth-required, no re-bump on redundant same-state delivery, bumps again on a distinct re-entry (leave-then-return), and bumps correctly via the cold-start `tier-detected`/`initial_status` path.
- Added a one-line comment at the `GenericSchemaForm` render site (`base-node-viewer.svelte`) documenting the invariant that makes capturing `autoOpen` once at mount safe.

**⏭️ Skipped:**
- The nitpick about `collection-node-viewer.svelte`'s data-load effect missing `untrack()` — this is a pre-existing effect this PR doesn't touch (only its `onTitleChange` effect was in scope), so fixing it here would be scope creep beyond #1511. Left as a follow-up candidate.

**Verification:**
- `bun run test`: 147 files, 3952 tests passed (4 new).
- `bun run quality:fix`: 0 errors/warnings, no incidental changes.

## Re-Review Decision

**Rationale**: Changes are additive test coverage plus a documentation comment — no logic changes to the reviewed diff, no new risk surface. The prior review already verified the `authRequiredEpisode` transition logic by manual trace; these tests formalize that verification without altering behavior.

PR #1556, opened 2026-07-06
Factor a transport-agnostic adapter core to stop the four-way adapter drift

malibio:
## Summary

- Extracts the shared request/response shaping logic — CreateNode field defaults, the task-update tri-state clear/set/no-change encoding, InsertPosition's proto oneof encoding, and HTTP route templates — from four hand-synced copies (`TauriAdapter`, `HttpAdapter`, the dev-proxy's REST→gRPC translation, and a fourth duplicate embedded in the e2e harness) into a single `adapter-core.ts`. `TauriAdapter`/`HttpAdapter` and the dev-proxy now call the same builders instead of re-deriving them by hand, and `daemon-harness.ts` imports the real `HttpAdapter` instead of a fourth copy.
  - `packages/proto` has no JS/TS codegen output (it's a Rust-only crate consumed via `tonic-build`), so "sourced from nodespace-proto" here means the builders hand-encode the `.proto` file's documented field/optionality contracts (with comments citing the source message/field), not generated bindings.
- While wiring the dev-proxy to the shared `encodeInsertPosition` builder, found and fixed two live bugs it exposed: `POST /api/nodes` silently dropped `insertPosition` entirely, and `POST /api/nodes/:id/parent` still sent a legacy `insertAfterNodeId` field that no longer exists in `MoveNodeRequest`'s proto oneof — both meant browser-mode create/move-with-position silently fell back to appending at the end instead of honoring the caller's requested position.
- Adds `adapter-contract.e2e.ts` — a contract test asserting shape parity between the transports and exercising the fixed insert-position round-trip against a real daemon — plus unit tests for the extracted builders in `adapter-core.test.ts`.
- `TauriAdapter.updateTaskNode` still forwards its update as-is rather than calling the shared tri-state builder — that encoding happens in the Rust `#[tauri::command]` handler on that path, which can't literally share TS code. Documented in place; ADR-048 decision 2 (a Rust-side integration test against the Tauri command layer) is the planned way to prove the two encodings agree.

Closes #1500.

## Test plan

- [x] `bun run test` — 148 test files, 3959 tests passed (baseline was 147/3948; +1 new unit test file)
- [x] `bun run test:e2e` — 5 test files, 24 tests passed against a real headless daemon (includes the new adapter-contract.e2e.ts and the fixed insert-position regression test)
- [x] `bunx tsc --noEmit` — clean (only one pre-existing, unrelated error in `backlinks-panel.test.ts`)
- [x] `bun run quality:fix` (desktop-app scope) — 0 errors, 0 warnings
- [x] Full pre-push gate (`test:all` + `cargo build --bin nodespaced` + `test:e2e`) passed
- [x] Addressed pragmatic-code-review feedback (recommendation: APPROVE) — see review comment and follow-up commit

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -195,6 +195,8 @@ IMPORTANT SUB-AGENT INSTRUCTIONS:
    bun run gh:pr <number>    # Creates PR, updates status to "Ready for Review"
    ```
 
+   > ⚠️ **`git push` runs a pre-push gate (`scripts/test-gate.ts`, ADR-047)** that re-runs `test:all`, then `cargo build --bin nodespaced`, then the full `test:e2e` suite — every push, not just the first. On a fresh worktree with no Rust build cache this can take **several minutes** (cold `cargo build` alone can exceed 5 minutes). Give the push command a long timeout (10+ minutes) or run it in the background and wait for completion — a command that times out before the hook finishes looks identical to a real failure but isn't one; check the tail of the actual output for a real test failure vs. an incomplete cold build before concluding the push failed. Do not reach for `--no-verify` to work around slowness — it's reserved for WIP Handoff Commits.
+
 5. **Code Review** — run `/pragmatic-code-review` on every PR before merge. NEVER merge without it.
 
 6. **Merge & Clean Up** — order matters:
--- a/packages/desktop-app/src/lib/services/adapter-core.ts
+++ b/packages/desktop-app/src/lib/services/adapter-core.ts
@@ -1,0 +1,279 @@
+/**
+ * Transport-agnostic backend adapter core.
+ *
+ * The daemon's NodeService contract (packages/proto/proto/node_service.proto) is
+ * consumed by three independent client paths — TauriAdapter (IPC), HttpAdapter
+ * (fetch), and the dev-proxy (REST→gRPC translation) — plus a fourth copy that
+ * used to live in the e2e harness. This module is the single place that encodes
+ * *what the wire contract is*: which fields are required vs. optional-omit vs.
+ * tri-state-clearable, and what each logical operation is named per transport.
+ * TauriAdapter/HttpAdapter/dev-proxy call these builders instead of re-deriving
+ * the encoding by hand, so a field/shape change can no longer drift between them
+ * silently — it becomes one function to update, used everywhere.
+ */
+
+import type { Node, NodeWithChildren, TaskNode, TaskNodeUpdate } from '$lib/types';
+import type { SchemaNode } from '$lib/types/schema-node';
+
+// ============================================================================
+// Shared types (public BackendAdapter surface)
+// ============================================================================
+
+/** Explicit insertion position for new or moved nodes. */
+export type InsertPosition =
+  | { type: 'beginning' }
+  | { type: 'end' }
+  | { type: 'after'; siblingId: string };
+
+/** Factory helpers for building InsertPosition values. */
+export const insertPosition = {
+  beginning: (): InsertPosition => ({ type: 'beginning' }),
+  end: (): InsertPosition => ({ type: 'end' }),
+  after: (siblingId: string): InsertPosition => ({ type: 'after', siblingId }),
+} as const;
+
+export interface CreateNodeInput {
+  id: string;
+  nodeType: string;
+  content: string;
+  properties?: Record<string, unknown>;
+  mentions?: string[];
+  parentId?: string | null;
+  /** Where to insert among siblings. Omit for End (default). */
+  insertPosition?: InsertPosition | null;
+}
+
+export interface UpdateNodeInput {
+  content?: string;
+  nodeType?: string;
+  properties?: Record<string, unknown>;
+  mentions?: string[];
+}
+
+export interface DeleteResult {
+  existed: boolean;
+  deletedCount: number;
+}
+
+export interface EdgeRecord {
+  id: string;
+  in: string;
+  out: string;
+  order: number;
+}
+
+export interface NodeQuery {
+  id?: string;
+  mentionedBy?: string;
+  contentContains?: string;
+  nodeType?: string;
+  limit?: number;
+}
+
+export interface CreateContainerInput {
+  content: string;
+  nodeType: string;
+  properties?: Record<string, unknown>;
+  mentionedBy?: string;
+}
+
+export interface BackendAdapter {
+  // Node CRUD
+  createNode(input: CreateNodeInput | Node): Promise<string>;
+  getNode(id: string): Promise<Node | null>;
+  updateNode(id: string, version: number, update: UpdateNodeInput): Promise<Node>;
+  updateTaskNode(id: string, version: number, update: TaskNodeUpdate): Promise<TaskNode>;
+  deleteNode(id: string, version: number): Promise<DeleteResult>;
+
+  // Hierarchy
+  getChildren(parentId: string): Promise<Node[]>;
+  getDescendants(rootNodeId: string): Promise<Node[]>;
+  getChildrenTree(parentId: string): Promise<NodeWithChildren | null>;
+  moveNode(nodeId: string, version: number, newParentId: string | null, insertPosition: InsertPosition | null): Promise<Node>;
+  moveChildrenToParent(newParentId: string, children: Array<{ id: string; version: number }>): Promise<Node[]>;
+
+  // Mentions
+  createMention(mentioningNodeId: string, mentionedNodeId: string): Promise<void>;
+  deleteMention(mentioningNodeId: string, mentionedNodeId: string): Promise<void>;
+  getOutgoingMentions(nodeId: string): Promise<string[]>;
+  getIncomingMentions(nodeId: string): Promise<string[]>;
+  getMentioningContainers(nodeId: string): Promise<string[]>;
+
+  // Queries
+  queryNodes(query: NodeQuery): Promise<Node[]>;
+  mentionAutocomplete(query: string, limit?: number): Promise<Node[]>;
+
+  // Composite operations
+  createContainerNode(input: CreateContainerInput): Promise<string>;
+
+  // Schema operations (read-only - mutation commands removed in Issue #690)
+  // Returns SchemaNode with typed top-level fields (isCore, schemaVersion, description, fields)
+  getAllSchemas(): Promise<SchemaNode[]>;
+  getSchema(schemaId: string): Promise<SchemaNode>;
+}
+
+// ============================================================================
+// CreateNode — shared request shaping
+// ============================================================================
+
+/**
+ * Normalized CreateNode wire fields, matching CreateNodeRequest in
+ * node_service.proto: parentId/insertPosition are omittable (proto `optional`),
+ * not merely nullable — a `null` sent over IPC/HTTP is a real "no parent" value
+ * on the Rust side, not "field absent."
+ */
+export interface CreateNodeFields {
+  id: string;
+  nodeType: string;
+  content: string;
+  properties: Record<string, unknown>;
+  mentions: string[];
+  parentId: string | null;
+  insertPosition: InsertPosition | null;
+}
+
+export function buildCreateNodeFields(input: CreateNodeInput | Node): CreateNodeFields {
+  return {
+    id: input.id,
+    nodeType: input.nodeType,
+    content: input.content,
+    properties: input.properties ?? {},
+    mentions: (input as CreateNodeInput).mentions ?? [],
+    parentId: (input as CreateNodeInput).parentId ?? null,
+    insertPosition: (input as CreateNodeInput).insertPosition ?? null,
+  };
+}
+
+// ============================================================================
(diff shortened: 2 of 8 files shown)


malibio:
### Code Review Summary

This is a net-positive, well-scoped refactor that does what it says: it collapses four hand-synced adapter implementations into one shared core, and it found two real, previously-shipping bugs while doing it (`POST /api/nodes` silently dropping `insertPosition`, and `POST /api/nodes/:id/parent` sending a dead `insertAfterNodeId` field). The extraction is honest about its own boundaries — the contract test's doc comment explicitly states that true Tauri-IPC round-trip coverage is out of scope here and belongs to ADR-048 decision 2, rather than overclaiming "full parity." Verified against the diff directly (not just the PR description): the tri-state encoding, the route-pattern substitutions, the `HttpAdapter` export, the mock/importOriginal fix, and the live-daemon contract test all check out as described.

**Recommendation: APPROVE**, with one non-blocking process gap worth a fast follow-up (dev-tools has zero static verification in this repo, before or after this PR) and a couple of clarifications worth a comment/doc update.

### Requirements Check

- ✅ **A single transport-agnostic adapter core sourced from nodespace-proto** — `adapter-core.ts` centralizes `buildCreateNodeFields`, `buildTaskNodeUpdatePatch`, `encodeInsertPosition`, `normalizeChildrenTree`, and the HTTP route table, with comments citing the specific `node_service.proto` fields/oneofs each builder encodes. Note: `packages/proto` is a Rust-only crate with no JS/TS codegen, so "sourced from nodespace-proto" here means "hand-encodes the proto's documented contract in one place with citations," not "generated from proto." That's a reasonable interpretation for a TS-side change of this scope — the alternative (standing up JS codegen from `.proto` files) is a much larger, separate undertaking — but the PR body should say this explicitly so a future reader doesn't assume there's generated-binding parity that doesn't exist.
- ✅ **TauriAdapter/HttpAdapter reduced to thin transport shims** — true for `createNode`, `getChildrenTree`, `getDescendants` (both classes now delegate to `buildCreateNodeFields`/`normalizeChildrenTree`/the shared `getDescendantsViaChildren` helper), and for all of `HttpAdapter`'s URL construction (`HTTP_ROUTES.*`). One caveat: `TauriAdapter.updateTaskNode` still does `invoke('update_task_node', { id, version, update })` with the raw `TaskNodeUpdate`, never touching `buildTaskNodeUpdatePatch` — see Critical/Important findings below.
- ✅ **A contract test proves adapter + proxy parity** — `adapter-contract.e2e.ts` does two things: static shape-parity assertions on the shared builders (always run), and a live round-trip of `HttpAdapter → dev-proxy → real daemon` covering the tri-state clear/set/no-change patch and both new-bug-fix insert-position paths. This is real coverage of the HTTP+proxy leg. It is *not* Tauri-IPC parity coverage, and the file's own doc comment says so.
- ✅ **Code passes `bun run quality:fix`; tests pass** — verified independently: `bun run test` 148/3959 passed, `bun run test:e2e` 5/24 passed (including the two new/fixed e2e files), `tsc --noEmit` clean modulo one pre-existing unrelated failure, `quality:fix` (desktop-app scope) 0 errors/0 warnings. Reasonable to not run the workspace-wide `quality:fix` given the known `cargo fmt --all` side-effect on unrelated Rust files.

### Findings

#### Critical Issues

None. No security, correctness-of-shipped-behavior, or architectural-regression issues block this merge.

#### Suggested Improvements

**🟡 `TauriAdapter.updateTaskNode` bypasses the shared tri-state builder — the highest-drift-risk logic in the PR isn't unified across both transports**
📁 `packages/desktop-app/src/lib/services/backend-adapter.ts:90-96`

```ts
async updateTaskNode(id: string, version: number, update: TaskNodeUpdate): Promise<TaskNode> {
  return withDiagnosticLogging(
    'updateTaskNode',
    () => invoke<TaskNode>('update_task_node', { id, version, update }),
    [id, version, update]
  );
}
```

The PR's own framing calls `buildTaskNodeUpdatePatch` "the tri-state clear/set/no-change encoding for UpdateTaskNodeRequest — the highest drift-risk logic," and centralizes it precisely because it's the field most likely to silently diverge. But the Tauri path never calls it — it forwards the raw JS `TaskNodeUpdate` object over IPC, and the equivalent tri-state translation happens in the Rust `#[tauri::command]` handler. That's a legitimate, unavoidable scope boundary (TS and Rust can't share a function), but as written, this PR does not actually reduce the drift surface for this specific field on the Tauri side — it just moves where the un-unified duplicate lives (from a second TS copy to a Rust copy that isn't visible in this diff or checked by this contract test). A future change to the tri-state contract (e.g., adding a field, changing clear semantics) can still update `adapter-core.ts` + the dev-proxy path and silently miss the Rust handler, and nothing here would catch it.

Recommendation: this doesn't need to block the merge, since it can't be closed in TypeScript. But the PR body / a code comment on `TauriAdapter.updateTaskNode` should say explicitly "the tri-state encoding for this path lives in `#[tauri::command] update_task_node` in Rust, not here — see [Rust handler location]" so the boundary is visible rather than implicit, and file a fast-follow issue for a Rust-side test that asserts the Rust handler's clear/set/no-change semantics match `buildTaskNodeUpdatePatch`'s documented contract (this is exactly the gap ADR-048 decision 2 already earmarks — worth cross-referencing there instead of leaving it as a silent asymmetry).

Engineering principle: a "single source of truth" claim is only as strong as its weakest untested seam. Documenting the boundary honestly costs nothing and prevents a future reader from assuming cross-language parity that isn't actually enforced.

**🟡 Cross-package relative import from `dev-tools` into `desktop-app`'s internals is fragile and has zero static verification today**
📁 `packages/dev-tools/src/dev-proxy.ts:9-14` — `import { ... } from '../../desktop-app/src/lib/services/adapter-core.ts'`

Verified: the workspace root's `quality:fix`/`quality:check` scripts only cover `packages/desktop-app`, `packages/skill`, and Rust (`package.json:quality:fix`) — `packages/dev-tools` has no lint or `tsc` script of its own and is not covered by any workspace-level check. This was true before this PR too (dev-proxy.ts was never linted/typechecked), so this isn't a regression introduced here, but the PR adds a new reach-through import into a specific desktop-app internal module (`$lib/services/adapter-core.ts`, reached via a hand-written relative path with an explicit `.ts` extension) precisely in the file with the least static safety net in the repo. If `adapter-core.ts` ever moves, or if desktop-app's directory structure changes, this import breaks silently until someone runs the dev proxy or the e2e harness (which does exercise it as a real subprocess — confirmed via `daemon-harness.ts`'s `spawnDaemonAndProxy`, so it's not undetected at runtime, just undetected in review/CI static checks).

Recommendation: either (a) add `packages/dev-tools` to `quality:check`/`tsc --noEmit` coverage as a small fast-follow so this import path is statically verified, or (b) if the long-term intent is cleaner, promote `adapter-core.ts` (or just the pieces `dev-proxy.ts` needs — `buildTaskNodeUpdatePatch`, `encodeInsertPosition`, `HTTP_ROUTE_PATTERNS`) into a small shared package (e.g. `packages/adapter-contract` or similar) that both `desktop-app` and `dev-tools` depend on via the workspace, rather than a hand-written filesystem reach-across. Given this is a lean/pragmatic-quality codebase and the import does work and is exercised by tests, I would not block merge on this — it's a reasonable "ship it, then harden the seams that don't have a test yet" tradeoff — but it should be a tracked fast-follow, not a silent gap.

Engineering principle: cross-package imports that bypass the package boundary (via relative paths instead of a workspace dependency) work until they don't, and the failure mode (a broken import in an unlint file) is invisible until someone happens to run the exact path that exercises it.

**🟢 `adapter-core.ts`'s `import type { ... } from '$lib/types'` relies on the SvelteKit path alias, which isn't configured for the `dev-tools` package**
📁 `packages/desktop-app/src/lib/services/adapter-core.ts:15-16`

These are `import type` only, so Bun strips them entirely at runtime and this is a non-issue for `bun run dev-proxy` today. But it's an implicit assumption: if any future edit to `adapter-core.ts` turns one of those into a value-level import (e.g. a runtime type-guard helper imported from `$lib/types`), it would break `dev-proxy.ts` at import time with no alias resolution available outside the desktop-app package, and — per the previous finding — nothing would catch that until the e2e harness runs. Worth a one-line comment in `adapter-core.ts` noting that this file's imports must stay `$lib`-alias-free at the value level (or entirely) since it's consumed by a package without that alias configured.

#### Nitpicks

- **Nit:** `packages/desktop-app/src/lib/services/backend-adapter.ts:53` — `EdgeRecord` remains exported from `adapter-core.ts` (re-exported via `backend-adapter.ts`) but has no consumers anywhere in the codebase outside these two files. Pre-existing dead export (it was unused before this PR too, just relocated), not introduced here — fine to leave, but worth a quick grep-and-delete in an unrelated cleanup pass.
- **Nit:** `packages/desktop-app/src/tests/e2e/adapter-contract.e2e.ts:72-79` — the comment on the `UpdateTaskNode` shape-parity test is doing a lot of work to explain what the test *doesn't* prove (that dev-proxy actually calls the shared builder, vs. re-inlining equivalent logic that happens to produce the same output). That's honest and good, but it also means this specific assertion can't catch a regression where dev-proxy silently reverts to a hand-rolled equivalent — only a code-review diff-read can catch that. Consider (non-blocking) a lightweight static check (e.g., a grep-based test asserting `dev-proxy.ts`'s source text contains `buildTaskNodeUpdatePatch(`) if this drift risk is considered high enough to guard mechanically rather than by review discipline alone.
- **Nit:** the PR body's "sourced from nodespace-proto" phrasing invites the reader to assume generated-binding parity; a one-line addition ("hand-encoded from the proto's documented contract, since packages/proto has no JS/TS codegen output") would preempt the question this review had to go verify.

### Answers to the specific scope questions raised

1. **"Sourced from nodespace-proto" AC** — reasonable interpretation given no JS/TS codegen exists; the builders cite the exact proto messages/fields they encode. Not a gap worth blocking on, but worth stating explicitly in the PR body (see nit above) so it isn't mistaken for generated-binding parity.
2. **TauriAdapter as a "thin shim"** — true for every operation except `updateTaskNode`, where the tri-state encoding is un-unified across the IPC/HTTP boundary by necessity (cross-language). This is an acceptable scope boundary, but it should be called out explicitly (comment + fast-follow issue) rather than left implicit, since it's exactly the "highest drift-risk logic" the PR is otherwise centralizing.
3. **Contract test proving "adapter + proxy parity"** — fair reading for this PR's scope: it proves real HttpAdapter+dev-proxy parity live, plus static shape assertions on the shared builders. It explicitly and correctly declines to claim Tauri-IPC parity (the test file's own doc comment defers that to ADR-048 decision 2). This is the right way to scope it — recommend the PR body cross-reference that deferral as explicitly as the test file already does, so a reader of the PR alone (not the test source) gets the same caveat.
4. **Cross-package relative import** (dev-tools → desktop-app) — works and is exercised by tests, but sits in a file with zero static verification (`quality:check` doesn't cover `packages/dev-tools`). Reasonable to ship as-is; recommend a fast-follow to either add dev-tools to the typecheck/lint pipeline or extract the shared pieces into a proper shared package.

### Recommendation

**APPROVE.** The refactor is a genuine net improvement — it collapses real duplication, is caught by tests that exercise the actual runtime paths (not mocks pretending to), and found/fixed two live bugs in the process. The findings above are follow-up hardening (visibility of the Tauri-side boundary, static coverage of dev-tools), not blockers.

malibio (commented):
Review recommendation: APPROVE (see full findings in the review comment above). Note: could not submit a formal GitHub 'Approve' review because GitHub blocks self-approval on your own PR — this comment-type review records the recommendation instead.

malibio:
## Addressed review feedback

Recommendation from the pragmatic-code-review was **APPROVE** — all findings were 🟡/🟢, non-blocking.

**✅ Addressed (2):**
- `TauriAdapter.updateTaskNode` forwarding the raw update instead of calling the shared tri-state builder — documented in place why (Rust-side encoding on that path) and pointed at ADR-048 decision 2 as the planned way to close that gap.
- `adapter-core.ts`'s `$lib`-aliased imports being safe only because they're type-only (dev-proxy.ts imports this file with no alias resolution) — documented the constraint at the import site so a future value-level `$lib` import doesn't silently break dev-proxy.
- Also clarified the PR body's "sourced from nodespace-proto" phrasing to state explicitly that `packages/proto` has no JS/TS codegen, so this is hand-encoded-with-citations, not generated bindings.

**⏭️ Skipped (2):**
- Adding `packages/dev-tools` to the typecheck/lint pipeline — real pre-existing gap, but repo-wide tooling scope beyond this PR; better tracked as its own issue.
- A grep-based static guard against dev-proxy silently re-inlining the tri-state encoding — marginal value given `adapter-contract.e2e.ts`'s live round-trip already exercises this end-to-end against a real daemon.

## Re-Review Decision

**Rationale**: All changes in this follow-up are documentation-only (code comments + PR body wording) — no logic changed, no new surface added. `bunx tsc --noEmit`, `bun run quality:fix`, and `bun run test` (148 files / 3959 tests) all still pass clean after these changes.

PR #1557, opened 2026-07-06
Document that git push runs a slow pre-push gate

malibio:
## Summary

- Adds a note to CLAUDE.md's PR step: `git push` triggers `scripts/test-gate.ts` (ADR-047), which reruns `test:all`, does a `cargo build --bin nodespaced`, and runs the full `test:e2e` suite on every push. On a fresh worktree with a cold Rust build cache this can take several minutes, and a command that times out mid-gate looks identical to a real push failure but isn't one.
- Motivated by repeatedly treating a slow pre-push gate as a failure in a recent session — the gate was still compiling, not actually failing.

## Test plan

- [x] Docs-only change, no code/test impact

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -195,6 +195,8 @@ IMPORTANT SUB-AGENT INSTRUCTIONS:
    bun run gh:pr <number>    # Creates PR, updates status to "Ready for Review"
    ```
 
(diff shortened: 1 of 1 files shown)


malibio:
Folding this into #1556 instead — no need for a separate PR for a one-line doc note.

PR #1559, opened 2026-07-06
Derive ACP's TOOLS_LIST from a shim-tool registry to stop the drift

malibio:
Closes #1509

### DIFF
--- a/packages/agent/src/acp/context_assembly.rs
+++ b/packages/agent/src/acp/context_assembly.rs
@@ -49,9 +49,52 @@ the nodes you (the agent) have been given and the immediate graph context
 around them. Treat the sections below as the authoritative starting point.
 ";
 
-/// Shared tool list included in every agent's activation section.
-const TOOLS_LIST: &str = "Tools available: `nodespace_search_semantic`, `nodespace_get_node`, \
-`nodespace_create_node`, `nodespace_update_node`, `nodespace_get_children`";
+/// A tool exposed to external PTY-spawned agents via the `packages/skill/shims/*`
+/// CLI shims (distinct from the in-process local agent's `Tool` enum in
+/// `local_agent::tools` — external agents call the `nodespace` CLI, not the
+/// local llama.cpp tool executor).
+struct ShimTool {
+    /// Wire name used by every shim (`nodespace-tools.json`, `nodespace-hook.ts`,
+    /// `nodespace-plugin.ts`). Changing this must be mirrored in all four shim files.
+    name: &'static str,
+    description: &'static str,
+}
+
+/// Canonical registry of tools implemented by every shim in `packages/skill/shims/`.
+/// [`tools_list`] and the shim-name assertion test derive from this — keep it in
+/// sync with the shim implementations, not the other way around.
+const SHIM_TOOLS: &[ShimTool] = &[
+    ShimTool {
+        name: "nodespace_search_semantic",
+        description: "Search the NodeSpace knowledge graph using natural language.",
+    },
+    ShimTool {
+        name: "nodespace_get_node",
+        description: "Fetch a single NodeSpace node by its ID.",
+    },
+    ShimTool {
+        name: "nodespace_create_node",
+        description: "Create a new node in the NodeSpace knowledge graph.",
+    },
+    ShimTool {
+        name: "nodespace_update_node",
+        description: "Update the content of an existing NodeSpace node.",
+    },
+    ShimTool {
+        name: "nodespace_get_children",
+        description: "List the direct children of a NodeSpace node.",
+    },
+];
+
+/// Shared tool list included in every agent's activation section, generated
+/// from [`SHIM_TOOLS`].
+fn tools_list() -> String {
+    let mut section = String::from("Tools available:\n");
+    for tool in SHIM_TOOLS {
+        section.push_str(&format!("- `{}`: {}\n", tool.name, tool.description));
+    }
+    section
+}
 
 /// Build the markdown section appended to the context file when NodeSpace tools are registered.
 ///
@@ -73,8 +116,7 @@ fn tools_section(agent_type: AgentType) -> String {
         );
     }
 
-    section.push_str(TOOLS_LIST);
-    section.push('\n');
+    section.push_str(&tools_list());
     section
 }
 
@@ -915,7 +957,7 @@ mod tests {
     // ---- tools section in context file ------------------------------------
 
     #[test]
-    fn tools_section_mentions_all_five_tool_names() {
+    fn tools_section_mentions_every_shim_tool_name() {
         for agent_type in [
             AgentType::ClaudeCode,
             AgentType::Codex,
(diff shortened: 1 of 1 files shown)


malibio:
## Pragmatic Code Review — PR #1559

### Summary

This PR fixes real prose/implementation drift in the ACP context-assembly tool list, but deviates from issue #1509's literal instructions after discovering the issue's premise was factually wrong: `Tool::ALL` (local_agent) and `TOOLS_LIST` (ACP) are disjoint tool surfaces serving different consumers (in-process llama.cpp tool executor vs. externally-spawned CLI coding agents via shims). Deriving `TOOLS_LIST` from `Tool::ALL` as literally specified would have advertised 5+ tools to external agents that the shims don't implement — a functional regression, not a fix. The chosen alternative — a `SHIM_TOOLS` registry as the actual source of truth, with a test tying it to the Gemini JSON shim manifest — is well-reasoned, correctly scoped, and verified against the codebase. Net positive. Approving, with one non-blocking gap noted below.

I independently verified:
- `Tool::ALL`/`local_agent::tools.rs` tool names and `packages/skill/shims/*` tool names are indeed disjoint sets (`search_nodes`, `create_schema`, `delete_node`, etc. vs. `nodespace_search_semantic`, `nodespace_get_node`, etc.) — the premise correction is factually correct, not a rationalization.
- The local-agent-side drift (`Tool` enum ↔ `skill_pipeline.rs` whitelists) is indeed already covered by a pre-existing test, `all_skill_whitelist_tools_are_registered` in `packages/agent/src/skill_pipeline.rs:451`, introduced by PR #1336 well before this issue — this PR correctly avoids re-touching that (confirmed via `git diff origin/main...HEAD --stat`: only `context_assembly.rs` changed).
- `SHIM_TOOLS` in the diff matches `packages/skill/shims/gemini/nodespace-tools.json` byte-for-byte on name+description.
- All three other shims (`claude-code/nodespace-hook.ts`, `codex/nodespace-plugin.ts`, `opencode/nodespace-plugin.ts`) currently implement the identical 5 tool names with identical descriptions (where present) — so today there is zero actual drift across all 4 shims; the new test would catch future drift between the Rust registry and at least the Gemini manifest.
- `cargo test -p nodespace-agent acp::context_assembly` — 28/28 pass, including the two new/modified tests.
- `serde_json` is already a workspace dependency; no new dependency introduced.
- No other module references `TOOLS_LIST`/`tools_section`/`tools_list`, so the prose format change (single-line → bulleted list with per-tool descriptions) is self-contained with no breaking consumers.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- **PR title is stale/misleading.** The PR title reads "Derive the ACP tool list from the Tool enum to stop three-way drift" — this is exactly the approach the implementation correctly rejected. Anyone scanning PR history or `git log --oneline` will be told the opposite of what actually happened. The commit message body gets this right (explains the correction clearly), but the title should be updated to something like "Derive ACP's TOOLS_LIST from a shim-tool registry to stop the drift" (matching the commit's own `messageHeadline`) before merge. This matters more than usual here specifically because the whole point of the PR is fixing a *documentation/drift* problem — leaving the PR's own headline wrong is a bit self-defeating.

- **Coverage gap: only 1 of 4 shims is drift-checked at test time**, and this is worth stating explicitly rather than leaving implicit. `shim_tools_match_gemini_manifest` only parses `gemini/nodespace-tools.json` because it's the one JSON (machine-readable) shim; the other three are hand-authored `.ts` files with no test-time parsing. Today they match (verified above), but nothing stops `claude-code/nodespace-hook.ts`, `codex/nodespace-plugin.ts`, or `opencode/nodespace-plugin.ts` from silently diverging from `SHIM_TOOLS` (and from each other) in a future change — the exact class of bug this issue exists to prevent, just relocated one level down. This is a legitimate, known limitation (correctly flagged in code comments and the commit message), not a blocker — parsing TS at test time from a Rust test would be disproportionate. But given all four `.ts`/`.json` files already duplicate a `runCLI`/`NodespaceCLIError` implementation by hand (per their own existing "no shared module resolution at runtime" comment), a follow-up worth filing as a separate issue: a lightweight build-time or CI script (even a `bun` script, not necessarily Rust) that diffs tool name+description across all 4 shim files against each other, independent of the Rust registry. That would close the loop the Gemini-only test can't reach. Not required for this PR to merge — the issue's stated acceptance criteria are about the ACP list, not full shim-to-shim parity — but worth a tracking issue so it isn't forgotten.

- **`ShimTool` and `SHIM_TOOLS` are not consumed by the shim-writing code path.** `write_shim_files` (unchanged by this PR) still copies the shim files verbatim from `packages/skill/shims/*` rather than generating them from `SHIM_TOOLS`. That's fine and out of scope for this PR — the issue was specifically about the ACP prose list — but it means `SHIM_TOOLS` is a second hand-maintained copy of tool metadata alongside the 4 shim files (5 sources of truth for tool name+description now, down from "prose string with none," which is still a net improvement, just worth being clear-eyed that this isn't a full collapse-to-one-source fix, only a partial one for the prose-generation path).

#### Nitpicks

- Nit: `context_assembly.rs` — `tools_list()` builds its output via repeated `String::from` + `push_str`/`format!` in a loop; fine at this size (5 entries, called per context-assembly, not hot path) — no action needed, just noting it's not zero-alloc-optimized if this ever gets called in a tighter loop.
- Nit: The doc comment on `ShimTool::name` says "Changing this must be mirrored in all four shim files" — good WHY-comment, but consider also cross-referencing `shim_tools_match_gemini_manifest` by name in that comment so a future reader immediately knows which test partially enforces the mirroring claim (currently the connection is implicit).

### Acceptance Criteria Assessment (issue #1509)

- [x] "The ACP tool list is generated from the Tool enum" — **not met literally, and correctly not met.** Investigation (confirmed independently above) showed this criterion was based on a false premise; implementing it as written would have been a regression (advertising unimplemented tools to external agents). The PR substitutes a corrected, narrower goal: the ACP tool list is generated from a canonical registry of the tools the shims actually implement. This was explicitly surfaced to and approved by the user mid-implementation (per task description) — appropriate handling of a bad issue premise, not scope creep.
- [x] "A test asserts whitelist + ACP names resolve against the enum" — the whitelist-half of this was already satisfied pre-existing (`all_skill_whitelist_tools_are_registered`, PR #1336, untouched here) since whitelists and the enum are the same tool surface. The ACP-half is satisfied by the new `shim_tools_match_gemini_manifest`, adapted to check against the correct actual source of truth (the shim manifest) rather than the enum, since the enum has nothing to do with ACP tools.
- [x] "Code passes `bun run quality:fix`; tests pass" — confirmed per task description and independently: `cargo test -p nodespace-agent` 265 tests pass, targeted `context_assembly` tests 28/28 pass, `bun run test` baseline maintained, pre-push gate green. The one failing suite (`ollama_integration.rs`, ~14 tests) is a verified pre-existing flake reproducible on unmodified `main` (confirmed via stash-and-rerun per task description) — correctly excluded from blocking this PR.

Given the issue's literal criteria were built on an incorrect premise, and the deviation was surfaced to and approved by the user rather than made unilaterally, I'm treating the corrected criteria as the ones that matter, and those are met.

### Verdict

**Approve.** This is a real, well-scoped fix for genuine drift risk, backed by verification I could independently reproduce. The one item worth fixing before merge is cosmetic but easy: update the PR title to match the commit's actual headline/intent so the historical record isn't self-contradictory. The shim-parity gap (JSON-only verification) is worth a follow-up issue but is not a blocker — it's a known, disclosed limitation of an already-net-positive change, not a regression this PR introduces.

malibio:
**Review verdict: Approve** (posting as a comment since GitHub blocks self-review approval for this account — the PR author and reviewing token are the same GitHub identity). See the full review above for details. No blockers found; one cosmetic suggestion (PR title is stale) noted before merge.

malibio:
## Address Review — Summary

**✅ Addressed:**
- PR title was stale (described the rejected `Tool::ALL` approach instead of what was actually implemented) — updated to "Derive ACP's TOOLS_LIST from a shim-tool registry to stop the drift".
- `ShimTool::name`'s doc comment now cross-references `shim_tools_match_gemini_manifest` by name, so a future reader can immediately find the test that partially enforces the "must be mirrored in all four shim files" claim.

**⏭️ Skipped (explicitly called out by the reviewer as non-blocking / out of scope):**
- Shim-parity gap (only the Gemini JSON manifest is checked at test time; the other three `.ts` shims are unchecked) — filed as follow-up #1560 per the reviewer's own suggestion, rather than expanding this PR's scope.
- `SHIM_TOOLS` being a second hand-maintained copy of tool metadata — reviewer explicitly noted this is "fine and out of scope for this PR."
- `tools_list()` allocation pattern nitpick — reviewer noted "no action needed."

**Tests:** `cargo test -p nodespace-agent --lib acp::context_assembly` — 28/28 pass. `bun run quality:fix` — clean. Pre-push gate (test:all + cargo build + test:e2e) — all green.

## Re-Review Decision

**Rationale:** The only changes since the reviewed commit are a PR title edit (not code) and a 3-line doc-comment addition with no logic change. No new functionality, no architectural changes, no risk introduced. The reviewer's verdict was already **Approve** with the title fix as the one pre-merge suggestion — that's now done.

PR #1562, opened 2026-07-06
Reconcile chat node-cards incrementally instead of remounting all

malibio:
## Summary
- `chat-markdown.svelte` replaced its entire DOM subtree (`innerHTML =`) and unmounted/remounted every inline `NodeCardInline` component on each content change, so a streaming AI chat message re-mounted all node-cards once per token.
- Now diffs the newly-rendered top-level blocks against the last render and patches only the blocks that actually changed, leaving node-cards in unaffected blocks mounted in place.
- Along the way, isolated `ChatMarkdown` from two pre-existing process-wide singleton landmines that were only surfaced by testing this fix under the full test suite:
  - `marked`'s shared default export is mutated globally by `marked-config.ts`'s `marked.use()` (used for the node editor's inline-only rendering) — now uses a private `new Marked()` instance instead.
  - `dompurify`'s default export binds to whichever `window` existed at first import across the whole test process — now resolved per-call via `createDOMPurify(document.defaultView)`, matching the existing pattern in `mermaid-render.ts`.

## Test plan
- [x] `bun run test` — 149 files / 3966 tests passed (baseline was 148/3963; added 3 new tests)
- [x] `bun run test:all` — frontend + Rust suites both pass
- [x] `bun run quality:fix` — 0 errors, 0 warnings
- [x] New test file `chat-markdown-reconcile.test.ts` verifies:
  - a node-card's DOM node/component identity survives when unrelated trailing content is appended (the streaming case)
  - a node-card in an earlier block survives when a later block is appended
  - a node-card block whose own content changed is still correctly updated

Closes #1520

### DIFF
--- a/packages/desktop-app/src/lib/components/chat/chat-markdown.svelte
+++ b/packages/desktop-app/src/lib/components/chat/chat-markdown.svelte
@@ -130,6 +130,16 @@
    * only touch the trailing block, so unchanged siblings — and any
    * NodeCardInline components mounted inside them — are left untouched
    * instead of being torn down and remounted on every content change.
+   *
+   * This is a positional index diff, not a keyed/LCS diff: block N is only
+   * ever compared against the previous render's block N. That's correct for
+   * the append-only way chat content grows (streaming appends to the last
+   * block or adds new trailing blocks) but would cause needless remounts of
+   * every following block if a block were ever reordered or inserted before
+   * the end — there's no such path today. Node-cards are also reconciled at
+   * block granularity, not individually: if two node-cards share one
+   * top-level block (e.g. the same paragraph) and any part of that block's
+   * text changes, both remount together.
    */
   function patchContent(newHtml: string): void {
     const template = document.createElement('template');
--- a/packages/desktop-app/src/lib/components/chat/chat-markdown.svelte
+++ b/packages/desktop-app/src/lib/components/chat/chat-markdown.svelte
@@ -10,12 +10,29 @@
 -->
 
 <script lang="ts">
-  import { marked, Renderer, type Tokens } from 'marked';
-  import DOMPurify from 'dompurify';
+  import { Marked, Renderer, type Tokens } from 'marked';
+  import createDOMPurify from 'dompurify';
   import { mount, unmount } from 'svelte';
   import NodeCardInline from './node-card-inline.svelte';
 
   let { content = '' }: { content: string } = $props();
+
+  // A private Marked instance, NOT the shared `marked` singleton — the
+  // singleton is mutated process-wide by marked-config.ts's marked.use()
+  // (for the node editor's inline-only rendering), which would otherwise
+  // leak into and corrupt this component's full-markdown chat rendering.
+  const chatMarked = new Marked();
+
+  /**
+   * DOMPurify's default export is a singleton bound to whatever `window`
+   * existed the first time the module was imported anywhere in the process
+   * (see dompurify's createDOMPurify()). Binding explicitly to the live
+   * `window` on every call avoids depending on that import-order-sensitive
+   * global state.
+   */
+  function getDOMPurify() {
+    return createDOMPurify(document.defaultView ?? undefined);
+  }
 
   // Custom renderer that handles nodespace:// URIs
   const chatRenderer = new Renderer();
@@ -49,14 +66,14 @@
   function renderMarkdown(md: string): string {
     if (!md) return '';
     try {
-      const raw = marked(autolinkNodespaceUris(md), {
+      const raw = chatMarked.parse(autolinkNodespaceUris(md), {
         renderer: chatRenderer,
         breaks: true,
         gfm: true,
       });
       if (typeof raw !== 'string') return md;
       // Allow nodespace:// protocol and data attributes for node card placeholders
-      return DOMPurify.sanitize(raw, {
+      return getDOMPurify().sanitize(raw, {
         ADD_ATTR: ['data-node-id', 'data-display-text'],
         ALLOW_UNKNOWN_PROTOCOLS: true,
       });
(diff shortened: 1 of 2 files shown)


malibio:
### Code Review Summary

This is a well-targeted, correctly-scoped fix. It replaces `chat-markdown.svelte`'s full `innerHTML =` remount with a positional top-level-block diff (`isEqualNode` against a pre-hydration snapshot), so unchanged blocks — and the `NodeCardInline` components mounted inside them — survive a content update instead of being torn down every token during streaming. I independently traced the diffing logic line-by-line (including a dedicated edge-case pass on reordering, duplicate node IDs, and hydration-snapshot validity) and verified the two "opportunistic" singleton fixes against the actual `marked-config.ts` / `mermaid-render.ts` source. All of the implementing agent's claims check out. Net positive, no blockers.

Verified independently:
- `bun run test` (frontend, Happy-DOM): **149 files / 3966 tests passed**, matching the PR description.
- `bunx eslint` on the two changed files: clean. `svelte-check`: **0 errors, 0 warnings**.
- `bun run quality:fix` failed, but only on `rust:quality:fix` — a pre-existing Rust build issue in this worktree (`resource path binaries/nodespaced-aarch64-apple-darwin doesn't exist`), unrelated to this PR's TS/Svelte-only diff. Not a regression introduced here.
- Diff is atomic and scoped to exactly 2 files (confirmed via `git diff <merge-base>..HEAD --stat`, since `origin/main` has moved ahead of this branch's fork point): `chat-markdown.svelte` and the new `chat-markdown-reconcile.test.ts`.

### Findings

#### Requirements Validation (Issue #1520)
- "Node-cards reconciled incrementally, not remounted per content change" — met. Verified via the new tests and by tracing `patchContent`: unchanged top-level blocks are `continue`'d past the mount/unmount cycle entirely.
- "Passes `quality:fix`; tests pass" — the frontend half is verified clean; the Rust half fails for an environmental reason orthogonal to this change (see above). Worth a sanity check on a clean checkout before merge, but not attributable to this diff.

#### Suggested Improvements

- **`chat-markdown.svelte:142-160` (`patchContent`)**: The diff is a **positional index comparison**, not a keyed/LCS diff. If a top-level block is reordered or a new block is inserted *before* the last one (rather than appended at the very end), every subsequent block shifts by one index and gets diffed against the wrong old block, causing needless remounts of all node-cards from that point on. This is not a correctness bug — the resulting DOM always converges to the correct final markup — just a missed optimization outside current usage. I confirmed via a targeted trace that the only place `content` is mutated during streaming (`ai-chat-node-viewer.svelte`, `streamingContent += chunk.text`) is strictly append-only, and no edit/regenerate/retry flow exists anywhere in the codebase that would reorder or mid-insert into a message's content. So this is currently unreachable dead-code risk, not a live bug. Worth a one-line code comment noting the diff is positional/append-optimized so a future author adding message-editing doesn't assume it's a general-purpose reconciler.

- **`chat-markdown.svelte:89-124` (block-granularity, not card-granularity)**: Confirmed as described — if two node-cards share one top-level block (e.g. same `<p>`) and any part of that paragraph's text changes, both are remounted together. This is a reasonable, explicitly-tradeoff given there's no DOM-diffing library in the repo. Given typical chat content (one node reference per line/paragraph), this is a low-frequency case. Fine to ship as-is; **Nit:** a one-line code comment near `patchContent` stating this granularity tradeoff explicitly would save a future reader from having to re-derive it (it's currently only in the PR description, which won't survive `git blame` archaeology).

#### Singleton Fixes — Verified Properly Scoped
- Confirmed `marked-config.ts:40` calls `marked.use({...})` at module load on the shared default-exported `marked` singleton — this genuinely does globally mutate parser behavior (disables heading/list processing, overrides `strong`/`em`/`codespan`) for the *entire process*, which would corrupt `ChatMarkdown`'s full-markdown rendering if that file happened to be imported anywhere in the same test run (exactly the "process-wide landmine" the PR describes). The fix (`const chatMarked = new Marked()`) is minimal and correctly scoped — a private instance local to this component, not touching `marked-config.ts` or any other consumer.
- Confirmed `mermaid-render.ts:98-99` already establishes the exact `createDOMPurify(document.defaultView ?? undefined)` per-call pattern this PR now also uses in `chat-markdown.svelte:34`. This is consistency with an existing precedent, not a new pattern — good reuse.
- Confirmed via `grep` that no other file imports from `chat-markdown.svelte`, so both fixes are fully self-contained; nothing else in the codebase could have been relying on the old shared-singleton behavior of this specific component.
- Confirmed `marked@16.4.2` (installed version) exports a `Marked` class, matching the import.

#### Edge Cases (explicitly requested)
- **Duplicate node IDs**: No collision risk. `mountedByEl` is a `Map<Element, ...>` keyed by DOM element object identity, never by `nodeId` (`chat-markdown.svelte:89`). Two placeholder spans referencing the same node ID get fully independent `mount()`/`unmount()` lifecycles. They do share `sharedNodeStore`'s cache by node ID, but that's an intentional, pre-existing cross-instance cache, not something this PR changes or that causes incorrect behavior.
- **`isEqualNode` validity**: Confirmed both sides of every comparison are pre-hydration markup. `newNodes` come from a `<template>` (`template.content.childNodes`, line 137) which is inert — `mount()` is never called against it. `lastRenderedNodes` is snapshotted via `cloneNode(true)` at line 163, *before* `hydratePlaceholders` runs at line 165 on the live tree. So the diff never compares hydrated-vs-unhydrated markup, which would otherwise always mismatch (since `NodeCardInline` injects a child `<a>` into the placeholder span once mounted). This ordering dependency is subtle and correctly implemented; the inline comment above `lastRenderedNodes` explains why.
- **Cleanup/leaks**: No leaks found. `unmountPlaceholders(liveNode)` (line 151) is unconditionally called before any of the replace/remove/append branches for every changed index, so a block is never detached from the DOM while its node-cards are still tracked in `mountedByEl`. The component-destroy effect also clears the full map on teardown.
- **Root-is-placeholder case**: `isPlaceholder(root)` is checked in addition to `root.querySelectorAll(...)` in both `hydratePlaceholders`/`unmountPlaceholders` (lines 101-103, 116-118) — this correctly handles a message that is *only* a bare `nodespace://...` URI with no wrapping paragraph, where the placeholder span is itself the top-level node and wouldn't be found by `querySelectorAll` alone (which only searches descendants). This is exercised by the third new test and is a detail that would have been easy to miss.

#### Nitpicks
- Nit: `chat-markdown.svelte:163` — the `cloneNode(true)` snapshot runs over all top-level blocks on every render (not just changed ones). Negligible for typical chat message sizes, and still a clear net win over the previous full-remount behavior, but if very large rendered messages ever become a concern, only the changed indices need re-snapshotting.
- Nit: consider adding a short comment (as noted above) at `patchContent` documenting the positional-diff assumption, so it's discoverable without reading the PR description or issue history.

### Verdict
Approve. The core fix is correct, appropriately tested (including the previously-easy-to-miss root-placeholder case), and the two singleton fixes are necessary, minimal, and precedented elsewhere in the codebase rather than overreaching. The positional-diff limitation and block-granularity tradeoff are real but currently unreachable/low-impact given the append-only streaming model actually in use, and are reasonable scope boundaries for this issue rather than gaps that should block merge.

malibio (commented):
Review posted above (net positive, no blockers). Unable to formally 'Approve' via gh pr review — GitHub rejects self-approval since this PR was authored under the same account. Recommend merging once the human owner does a final skim, per project policy (never merge without a completed /pragmatic-code-review, which is now done).

malibio:
## Review Response

The review verdict was **Approve** — no 🔴 Critical or 🟡 Important findings, only two nits (both pointing at the same spot).

**Addressed:**
- Added an inline comment above `patchContent` documenting the two things the review flagged as worth making discoverable in-code rather than only in the PR description:
  - The diff is a **positional index comparison**, not keyed/LCS — safe today because chat content only ever grows append-only, but would cause needless remounts if reordering/mid-insertion were ever introduced.
  - Node-cards reconcile at **block granularity**, not individually — two node-cards sharing one paragraph will remount together if that paragraph's text changes.

**Skipped:**
- The `cloneNode(true)` full-snapshot-every-render nit — negligible cost for typical chat message sizes, and re-snapshotting only changed indices would add complexity disproportionate to the actual gain. Can revisit if large-message profiling ever shows this line matters.

No logic changes — comment-only commit. Full test suite re-verified (149/149 files, 3966/3966 tests) plus targeted reconcile tests, and `quality:fix` clean.

## Re-Review Decision

**Rationale**: The only change since the last review is a documentation comment with zero logic impact. All prior findings were nits, not blockers, and the one addressed nit doesn't touch behavior.

PR #1563, opened 2026-07-06
Test the real Tauri-to-daemon seam end-to-end (ADR-048)

malibio:
## Summary

Implements ADR-048: extends the existing Rust `SpawnedDaemon` integration-test harness (previously used only by `daemon_readiness_test.rs`) to drive the real Tauri command layer directly — no webview, no mock-on-mock — against a real headless `nodespaced` and real SQLite.

- **Harness**: `SpawnedDaemon` (real daemon process) + `TauriTestApp` (a `tauri::test::mock_app()` with a connected `GrpcClient` managed as state), giving tests a real `tauri::State<GrpcClient>` via `Manager::state()` — the sanctioned construction path, since `tauri::State` has no public constructor. Extracted into its own crate (`test-support/`) so clippy's dead-code lint sees one compilation shared across every test file instead of flagging whichever subset any single test binary doesn't use.
- **Deleted** `src/tests/integration/tauri-commands.test.ts` — the mock-on-mock file that exercised zero production lines.
- **All 7 priority flows** covered, each against a real daemon:
  - Node CRUD round-trip (create/read/update/delete, hierarchy, properties)
  - Optimistic-edit-vs-echo race — drives the *actual* `watcher::run` (made generic over `tauri::Runtime` for testability) against a real `WatchNodes` stream and a real Tauri event bus. Along the way found `watcher.rs`'s module doc was stale (claimed the gRPC watcher was inert; it's the sole live event path — fixed the comment).
  - Indent/outdent + rapid-Enter ordering
  - ai-chat send→idle, using `ministral-3b-q4km` (smallest catalog model, already downloaded) since no stub inference engine exists anywhere in the codebase
  - Startup readiness at the data-plane level (extends `daemon_readiness_test.rs`'s signal-only coverage)
  - Model-download terminal state incl. cancel
  - Cross-window hierarchy sync (two independent `GrpcClient` connections to one daemon)
- **Adapter contract test** (ADR-048 decision 5): Rust-side counterpart to `adapter-contract.e2e.ts`'s HttpAdapter/dev-proxy parity checks, driving the same 3 scenarios through the real Tauri command layer.
- **Commits-per-edit**: validates the backend half of #1492's fix — one `update_node` call moves the version by exactly one, and a stale pre-confirmation version is OCC-rejected.
- **Pre-push gate**: wired `cargo test -p nodespace-app` into `scripts/test-gate.ts` — the desktop-app Rust integration tests (including the pre-existing `daemon_readiness_test.rs`/`schema_parity_test.rs`) previously ran nowhere at all.

### Bugs found and fixed along the way
- **`SpawnedDaemon` pipe-buffer deadlock**: stdout/stderr were piped but never drained. llama.cpp's own logging overflows the ~64KB OS pipe buffer during any test that loads a real model, silently hanging the daemon (0% CPU) on its next `write()`. Fixed by draining both pipes on background threads (optionally echoing under `E2E_VERBOSE`, matching the TS harness convention). This was latent in the pre-existing tests too.
- **`build-sidecars.ts`**: failed on a fresh checkout/worktree because `binaries/` is `.gitignore`'d and never pre-created before `cp`. Added `mkdirSync`.
- **Filed #1561**: a genuine, intermittent (~1/15 runs) race in `move_node`'s same-parent reorder branch — sibling read, order computation, and write-back aren't wrapped in a transaction, so two concurrent reorders can compute overlapping fractional-order keys. The covering test is split: the RPC-success assertion stays enabled, the strict-final-order assertion is `#[ignore]`'d as a documented regression test for #1561.
- Fixed a stale claim in `e2e-testing.md` (sidecars are `.gitignore`'d/built on demand, not committed).

## Test plan

- [x] `bun run test:all` — 147 frontend files / 3949 tests, 17 skill tests, 966 Rust core tests, all passing
- [x] `bun run quality:fix` — clean, no changes needed
- [x] Full `scripts/test-gate.ts` (the actual pre-push gate) run twice end-to-end, all green
- [x] New Rust seam suite run repeatedly (4-5x) to confirm no flakiness beyond the documented #1561 case
- [x] `/pragmatic-code-review` to follow before merge

Closes #1499.


malibio:
### Code Review Summary

This PR delivers on ADR-048's core mandate: the Tauri command layer — the actual shipping path, previously covered by a mock-on-mock test that exercised zero production lines — now has real integration coverage against a real headless `nodespaced` and real SQLite. I verified this claim rather than taking it on faith: `create_node` et al. in `commands/nodes.rs` are the exact `#[tauri::command]` functions registered in `lib.rs`'s `invoke_handler!`, the tests call them directly via `tauri::State` obtained through `Manager::state()` (the only sanctioned construction path, since `tauri::State` has no public constructor), and I ran several of the new suites directly against the real `nodespaced` binary in this worktree — CRUD, the optimistic-echo-race test, and the indent/outdent suite all pass for real, and the `#[ignore]`'d regression test for #1561 is correctly skipped with the documented reason. This is genuine coverage, not circular.

The pipe-buffer-drain fix is real and correctly implemented — I traced it against `Drop` and the `E2E_VERBOSE` convention and it holds up. The deleted `tauri-commands.test.ts` is confirmed to have tested nothing (`vi.fn()` returning what it was configured to return, on `greet`/`toggle_sidebar` — commands that likely don't even exist in the current command set) — its removal loses zero real coverage. The `#[ignore]`'d #1561 regression test is honestly labeled and does not mask a live problem; it documents a known race with a clear un-ignore condition.

This is a net-positive, well-engineered change that closes a real, previously-acknowledged coverage gap and finds two genuine bugs along the way (the pipe deadlock, and the #1561 reorder race). I have some findings below, none of which should block merge, but the environment-coupling issue (model download) is worth a deliberate decision rather than silent acceptance.

### Findings

#### Critical Issues

None. I did not find a security, correctness, or architectural issue that should block merge.

#### Suggested Improvements

- **`packages/desktop-app/src-tauri/tests/ai_chat_send_to_idle_test.rs`, `model_download_terminal_state_test.rs`**: These tests hard-depend on `ministral-3b-q4km` already being present under `~/.nodespace/models/` on the machine running them. I confirmed `chat_model_download`/`download_model` in `local_agent_service.rs` performs a real network fetch from HuggingFace when the model isn't present — there is no stub or pre-fetch step wired into `bun install`, `test-gate.ts`, or any CI workflow (this repo has no CI runner at all today; `test-gate.ts`'s own doc comment confirms the pre-push hook is the only gate). On a fresh clone, or any teammate's machine that hasn't manually downloaded this ~2GB file, these tests will either hang on a multi-gigabyte download or time out, blocking `git push` for everyone but the author. This is a real gap in "runs headless in CI" per the ADR-048 rationale and issue #1499's acceptance criteria, even though there's currently no CI to expose it. Worth an explicit, documented decision: either (a) add the model to `download:models` / a setup script so a fresh worktree can provision it, (b) skip these two files when the model is absent (e.g., a `Path::exists()` check that prints a clear "skipped: model not present, run `bun run download:models`" rather than hanging on a live download), or (c) explicitly accept this as a known, documented cost of a single-developer gate for now — but that should be a stated decision, not an implicit one baked into a doc comment only future spelunkers will find.

- **`packages/desktop-app/src-tauri/tests/daemon_readiness_test.rs` vs `test-support/src/lib.rs`**: `daemon_readiness_test.rs` defines its own file-local `ENV_MUTEX` for serializing `NODESPACED_SOCKET` mutation, while the new shared crate defines `CONNECT_MUTEX` for the same purpose. Since each `tests/*.rs` file compiles to its own OS process, this doesn't cause a cross-file race (I confirmed this — env vars are process-local, so the two mutexes never actually contend with each other), but it is a duplicated concept with two names that could easily confuse a future reader into thinking they need to coordinate, or into copy-pasting a third one. Now that `EnvGuard` lives in the shared crate, consider moving `CONNECT_MUTEX` (or an equivalent) there too as the one process-wide-env-mutation lock, and having `daemon_readiness_test.rs` use it instead of its own copy — closes the door on a future test file inventing a fourth name for the same lock.

#### Nitpicks

- **Nit**: `test-support/src/lib.rs`'s `TauriTestApp::connect` briefly acquires and releases `CONNECT_MUTEX`/`EnvGuard` internally (via `connected_client`), and `optimistic_echo_race_test.rs`'s two tests then separately re-acquire `hold_connect_mutex_and_socket_env` afterward for the watcher's benefit. I traced this window carefully since it looks at first glance like a gap the mutex doesn't cover — it doesn't cause a bug (the `GrpcClient`'s channel is fixed to whatever socket was resolved at `connect()` time and is immune to later env changes; the watcher's own later, independently-held guard is what actually matters and it's held continuously from before `tokio::spawn` until after the watcher is joined). But the two-mutex-acquisitions-per-test shape is subtle enough that a one-line comment at the `TauriTestApp::connect` call site in that file ("re-acquiring below is for the watcher's env read, not because this connect call is unsafe on its own") would save the next reader the same trace-through.

- **Nit**: `SpawnedDaemon`'s stdout/stderr drain threads (`spawn_drain_thread`) are never joined — harmless today since the process's own exit reclaims them once the child is killed and its pipes close, but a `Drop` impl that joins them (with a short timeout) would make the "no leaked threads" property explicit and verifiable rather than incidental.

- **Nit**: `model_download_terminal_state_test.rs`'s `cancel_of_a_download_with_no_stray_downloading_state_left_behind` exercises cancel-of-a-non-in-progress-download, which is real coverage but the weaker half of the acceptance criterion ("cancel leaves no lingering downloading state or stray control" reads more naturally as a test of cancelling an in-flight download). The doc comment is honest about why (mid-flight cancel of a multi-GB download is nondeterministic/slow), so this is a reasonable, disclosed scope trim rather than a hidden gap — just flagging it so the trade-off is visible in review, not just in the comment.

### Verification performed

- Ran `cargo clippy -p nodespace-app-test-support --all-targets -- -D warnings` — clean, confirming the dead-code-lint fix claim.
- Ran `cargo test -p nodespace-app --test node_crud_tauri_seam_test`, `--test optimistic_echo_race_test`, and `--test indent_outdent_rapid_ordering_test` against the real `nodespaced` binary in this worktree (`NODESPACED_TEST_BIN` pointed at `target/debug/nodespaced`) — all pass, including the real watcher-driven echo-race test, and the #1561 regression test correctly reports `ignored, known race — see #1561` rather than silently passing or being invisible.
- Confirmed `create_node`/`update_node`/etc. in `commands/nodes.rs` are the actual `#[tauri::command]` functions wired into `lib.rs`'s `invoke_handler!`, not a parallel test-only copy.
- Confirmed the deleted `tauri-commands.test.ts` tested `vi.fn()` mocks of `greet`/`toggle_sidebar` — zero production coverage lost.
- Traced the `CONNECT_MUTEX`/`EnvGuard`/`hold_connect_mutex_and_socket_env` interaction against Cargo's per-file-binary process isolation and confirmed no actual race, despite the pattern's first appearance.
- Confirmed the pipe-drain fix's shape against `Drop` and the `E2E_VERBOSE` convention.
- Confirmed `scripts/test-gate.ts` correctly wires `cargo test -p nodespace-app` with `NODESPACED_TEST_BIN` pointed at the binary the gate already builds.

malibio (commented):
Reviewed commit 3e839853920bdc072e76e8b0e88f742a0a6b33f1. Pragmatic code review posted above — net positive, no blockers. Two Improvement-level findings (model-download environment coupling; duplicated env mutex naming) and three nits. See full review comment for details.

malibio:
## Addressed review feedback

**Improvements addressed (2/2):**
- Model-dependent tests (`ai_chat_send_to_idle_test.rs`, `model_download_terminal_state_test.rs`) now skip cleanly with a clear message if `ministral-3b-q4km` isn't provisioned locally, instead of hanging on a live multi-GB download on a fresh machine.
- Consolidated the duplicated `ENV_MUTEX`/`CONNECT_MUTEX` concept — `daemon_readiness_test.rs` now uses the shared crate's `CONNECT_MUTEX` (made `pub`) instead of its own copy.

**Nit addressed (1/3):**
- Added a clarifying comment at the `TauriTestApp::connect` call site in `optimistic_echo_race_test.rs` about the two-mutex-acquisition shape.

**Nits skipped (2/3), per the review's own framing:**
- Joining `SpawnedDaemon`'s drain threads on `Drop` — reviewer confirmed this is harmless today (process exit reclaims them); marginal complexity for a test-only harness.
- Strengthening the cancel test to mid-flight download — reviewer confirmed the current scope is a reasonable, already-disclosed tradeoff (a live in-flight multi-GB download would be nondeterministic/slow to test reliably).

All changes verified: `cargo test -p nodespace-app` (full suite), `bun run test:all`, `bun run quality:fix`, and the full pre-push gate all pass clean.

## Re-Review Decision

**Rationale**: All changes are additive/mechanical — a skip-guard added to existing tests, a rename/consolidation of an internal test-only mutex, and a comment. No architectural changes, no new functionality, no changes to the actual product code being tested. Low risk, straightforward to verify by reading the diff.

Issue #255, opened 2025-10-12
Database initialization race condition in HTTP dev server tests


malibio:
# Bug: Database Initialization Race Condition in HTTP Dev Server Tests

## Summary

Integration tests using the HTTP dev server experience HTTP 500 \"no such table: nodes\" errors due to a race condition in database connection management when rapidly swapping databases across test cases.

## Impact

- **Affects**: All integration tests that use HTTP dev server adapter
- **Current Status**: 7/14 tests in `indent-outdent-operations.test.ts` failing
- **Blocks**: Issue #228 (Indent/Outdent Operations) verification

### Symptom

When running integration tests sequentially:
- **First test**: Usually succeeds
- **Subsequent tests**: Fail with "HTTP 500: Internal Server Error - no such table: nodes"

### Root Cause

**Race condition in libsql connection pooling when dev server swaps databases**

Each test creates a unique database file and calls `/api/database/init?db_path=...` to initialize it. The endpoint:
1. Creates new `DatabaseService::new(db_path)` 
2. Calls `initialize_schema()` (runs `CREATE TABLE IF NOT EXISTS`)
3. Swaps services in `AppState` using `RwLock::write()`

However, **there's no synchronization between the database swap and in-flight connections**:
- libsql maintains an internal connection pool per `Database` instance
- After swapping `AppState`, old connections may still be cached/active
- Subsequent operations may use stale connections pointing to previous test's database

### Test Flow Pattern

```
Test 1: [Init DB A] → [Create Node] → [Update Node] ✓ Works
Test 2: [Init DB B] → [Create Node] → [Update Node in-flight] → 
        ❌ Update uses stale connection to DB A or uninitialized connection to DB B
```

### Dev Server Logs

```
🔄 Database SWAPPED to: /tmp/test-...098-0ve5gm.db
📝 UPDATE request for node: node-2
❌ Node update failed for node-2: QueryFailed("no such table: nodes")
```

### Key Observations

1. Database swap succeeds (confirmed by logs)
2. Node creation typically works (uses fresh connections)
3. Node updates fail (uses cached/stale connections)
4. Pattern is consistent across multiple test runs

### Code Locations

**Backend:**
- `/packages/desktop-app/src-tauri/src/dev_server/node_endpoints.rs:init_database()` - Database swap endpoint
- `/packages/core/src/db/database.rs:DatabaseService::new()` - Creates database and calls `initialize_schema()`
[...]

PR #1568, opened 2026-07-07
database-write-queue.test.ts intermittently hangs/fails in the full suite due to cross-file singleton pollution

malibio:
Closes #1567


malibio:
## Pragmatic Code Review — PR #1568

**Scope:** `git diff origin/main...HEAD` — 3 files, +1 / −567

### Summary
This PR resolves the #1567 full-suite flake by **deleting** `database-write-queue.ts` and its 455-line test, plus a one-line stale doc-comment fix in `shared-node-store.svelte.ts`. Independently verified: the deleted module had **zero production callers**; production write serialization is done entirely by `SimplePersistenceCoordinator`. Deleting the dead module removes the process-global singleton whose self-test leaked state across files under `singleFork: true`. This is a legitimate — and stronger — resolution of the root cause than the issue's proposed reset-based fix. It also aligns squarely with the repo's "delete dead code, don't suppress" standard. **Net-positive, clean, atomic.**

### Requirements Check (issue #1567)
- ✅ **Full suite passes reliably, no hangs** — root cause (the shared singleton) is deleted outright, so there is no cross-file state left to leak. (Pre-merge `test:all` / push gate is the final confirmation.)
- ✅ **Test either resets its singleton or the leaking site is fixed** — satisfied by removal: no singleton, no test, nothing to reset.
- ✅ **Root cause identified and documented** — commit message documents it precisely and corrects the issue's misdiagnosis (there was no in-flight *production* call site; the module's own test was the sole exerciser).
- ✅ **No new failures** — only dead code + its dedicated test removed; no live behavior touched.
- ✅ **Passes quality:fix** — trivially (one comment line changed; two files deleted).

📝 **Deviation from the issue's proposed solution (acceptable):** The issue assumed a real production `queueDatabaseWrite()` call was leaking and asked to find/flush it (and to reconsider `singleFork`). Investigation showed no such call site exists — the only reference outside the test was a stale comment. "Delete the dead module" resolves the *same* root cause (shared mutable module state under `singleFork`) more decisively than a reset helper, and correctly leaves `singleFork` (added per #255 for DB-path-swap tests) unchanged. This does not sidestep a requirement.

### Independent Verification
1. **Zero remaining references** — repo-wide grep for `database-write-queue`, `queueDatabaseWrite`, `getQueueDepth`, `__resetQueueForTesting`, and path/barrel imports: **none**. No orphaned imports.
2. **No lost safety property** — `SimplePersistenceCoordinator` (shared-node-store.svelte.ts:82) is the real per-node serial writer: `executingOperations` set + latest-wins `queuedOperations` map, gated on version-confirmation. It is exercised by live code paths and guarded by `persistence-serial-writer-regression.test.ts` (#1492). The deleted queue was a *global* (not per-node) serializer that nothing called — deleting it removes no guarantee production depends on.
3. **Doc comment accurate** — line 14 now reads "Database writes serialized per-node by SimplePersistenceCoordinator," which matches the actual architecture; the old "queueDatabaseWrite()" reference was stale.
4. **No coverage regression** — the 455-line deleted test exclusively exercised the now-deleted module (real-timer serialization, depth tracking, `__resetQueueForTesting`). No still-live behavior loses coverage; the live serial-writer path retains its own dedicated regression suite.

### Findings
- 🟢 The commit message is exemplary: it states the root cause, corrects the issue's misdiagnosis with evidence, and explicitly notes `singleFork` is intentionally left unchanged. This is the documentation the acceptance criteria asked for.
- 🟢 Change is atomic and scoped — exactly 3 files, no incidental edits (verified via `--stat`).
- 🟢 No security, performance, or maintainability concerns; this strictly reduces surface area and removes a misleading "singleton queue is correct" comment that could have invited future misuse.

No 🔴 or 🟡 findings.

Clean, well-justified dead-code removal that resolves the flake at its root and improves code health. Recommend confirming a green `test:all` / push gate before merge (standard gate, not a blocker on the diff).

PR #1569, opened 2026-07-07
Finish the ADR-049 $effect conversion — 19 residual violations across four clusters

malibio:
Finishes the ADR-049 `$effect` conversion. Every remaining `$effect` that wrote reactive state is either converted to the ADR-conformant primitive or shown to be a legitimate DOM/listener push.

Rebased on #1565, which had already resolved **Cluster 2** (titles are now computed via `computeTabTitle` + plugin `getTitle`, never pushed). The remaining clusters:

### Cluster 1 — data-fetch effects → `onMount` / store methods
- `query-node-viewer`, `collection-node-viewer`, `schema-property-form`, `task-schema-form`: fetch moved to `onMount` (parents remount them via `{#key ...nodeId}` / `{#key node.id-nodeType}`, so the fetch is a discrete per-instance lifecycle load).
- `node-card-inline`: fetch-if-missing moved to `onMount` (cards mount imperatively per nodeId). `fetchAttempted` kept purely as a fetch-settled flag for the "unknown" fallback, written from the fetch callback.
- `collaboration-view`: loads on mount when Pro, else via `proSync.onProConfirmed` (the tier-transition hook) instead of an `isPro`-watching effect.
- **`pane-content`**: hydration no longer keeps a local `hydratedNodeIds` mirror or the `latestRequestedNodeId` staleness guard. `hydrateNode()` only pushes `ensureNode()` into the store; `isNodeHydrated` derives directly from `sharedNodeStore.getNode(nodeId)`. Because state lives per-nodeId on the store — not one shared local mirror — a stale in-flight fetch can only populate its own cell and can never overwrite the current node, so **the guard is deleted**. A not-found response closes the tab only if that tab still points at the fetched nodeId (checked against live tab state). `pane-content-hydration.test.ts` rewritten to assert this event-driven contract.

### Cluster 3 — mirrors, subscribe-in-effect, sequencing
- `calendar`: bind the primitive directly to the `$bindable` `value`; drop the `internalValue` mirror + its prop→state effect.
- `table-view`: clamp the effective page via `$derived` instead of a reset-on-`nodeIds` effect.
- `query-editor`: initialise `jsonText` in the `$state` initializer (`untrack`ed one-shot capture).
- `app-shell`: read `daemonStatus` via `$`-auto-subscription (`$derived`) instead of subscribe-in-effect; load recovered-items from `proSync.onProConfirmed` instead of an `isPro`-guarded effect.
- `pro-sync` store: adds `signedInEpisode` (empty→non-empty `userEmail` edge) and `onProConfirmed` (fires once on tier→pro); tier/userEmail writes routed through `setTier`/`setUserEmail`.
- `pro-sync-pill`: derive first-run inbox-open from `proSync.signedInEpisode` + a dismissed marker instead of a first-run effect.
- `code-block-node`: extract the syntax-highlight / mermaid render pipeline to a new `CodeBlockRenderer` store class (state lives on the store, component reads it). The component triggers renders by calling `renderer.render()`/`renderPreview()`; the two innerHTML-injection effects stay (legitimate DOM push) and now read `renderer.*`.

### Cluster 4 — dead effect
- `SchemaPropertyFormDemo`: deleted the read-state/empty-body effect.

## Testing
- `bun run test:all` green: **150 unit files / 3944 tests**, **17 browser tests**, **971 Rust tests**, plus the pre-push `test:e2e` gate.
- `bun run quality:fix` clean.
- New tests: `pane-content-hydration.test.ts` (rewritten, event-driven contract), `pro-sync.test.ts` (`signedInEpisode` + `onProConfirmed`), `node-card-inline.test.ts` (single-call-site on-mount fetch).

## Note
Hit a **pre-existing timing flake** in `text-node.test.ts:103` on the first push (`modifiedAt === createdAt` when two `new Date()` calls land 1 ms apart). Not touched by this branch; passes on retry. Same class as #1567 — worth a follow-up to seed a fixed timestamp.

### DIFF
--- a/packages/design-system/src/examples/SchemaPropertyFormDemo.svelte
+++ b/packages/design-system/src/examples/SchemaPropertyFormDemo.svelte
@@ -29,10 +29,6 @@
     { value: "charlie", label: "@charlie" }
   ];
 
-  $effect(() => {
-    assigneeSearch;
-  });
-
   // Example task node data
   const taskData = {
     title: "Implement Schema-Driven Property UI",
--- a/packages/desktop-app/src/lib/components/chat/node-card-inline.svelte
+++ b/packages/desktop-app/src/lib/components/chat/node-card-inline.svelte
@@ -7,6 +7,7 @@
 -->
 
 <script lang="ts">
+  import { onMount } from 'svelte';
   import { createLogger } from '$lib/utils/logger';
   import { sharedNodeStore } from '$lib/services/shared-node-store.svelte';
   import { backendAdapter } from '$lib/services/backend-adapter';
@@ -28,20 +29,24 @@
   };
 
   let node = $derived(sharedNodeStore.getNode(nodeId));
+  // True once the on-mount fetch has settled; drives the "unknown" fallback below when
+  // the node still isn't present. Written from the fetch callback, never from an effect.
   let fetchAttempted = $state(false);
 
-  // Fetch from backend if not in store
-  $effect(() => {
-    if (!node && !fetchAttempted) {
+  // Fetch from the backend on mount if the node isn't already in the store. Each card is
+  // mounted imperatively with a fixed nodeId, so this is a one-shot per-card load - not a
+  // reactive watch (ADR-049). Once fetched, `node` updates via the store read above.
+  onMount(() => {
+    if (sharedNodeStore.getNode(nodeId)) return;
+    backendAdapter.getNode(nodeId).then((fetched) => {
+      if (fetched) {
+        sharedNodeStore.setNode(fetched, { type: 'database', reason: 'node-card-fetch' }, true);
+      }
+    }).catch((e) => {
+      log.warn(`Failed to fetch node ${nodeId}:`, e);
+    }).finally(() => {
       fetchAttempted = true;
-      backendAdapter.getNode(nodeId).then((fetched) => {
-        if (fetched) {
-          sharedNodeStore.setNode(fetched, { type: 'database', reason: 'node-card-fetch' }, true);
-        }
-      }).catch((e) => {
-        log.warn(`Failed to fetch node ${nodeId}:`, e);
-      });
-    }
+    });
   });
 
   let title = $derived(node?.content?.split('\n')[0]?.slice(0, 120) || displayText || nodeId.slice(0, 8));
--- a/packages/desktop-app/src/lib/components/collaboration/collaboration-view.svelte
+++ b/packages/desktop-app/src/lib/components/collaboration/collaboration-view.svelte
@@ -10,6 +10,7 @@
   onboarding paths) live in the sibling InvitesPanel (S4).
 -->
 <script lang="ts">
+	import { onMount } from 'svelte';
 	import { membership } from '$lib/stores/membership.svelte';
 	import type { Permission } from '$lib/services/membership-service';
 	import { proSync } from '$lib/stores/pro-sync.svelte';
@@ -53,10 +54,16 @@
 	let approveRole = $state<Record<string, Permission>>({});
 	let busyRequest = $state<string | null>(null);
 
-	$effect(() => {
-		if (proSync.isPro && collectionId) {
+	// Load the roster on mount if Pro is already confirmed (the host tab is Pro-gated, so
+	// this is the common path). If tier confirms after mount, load then - via the store's
+	// transition hook, not a reactive $effect that watches isPro (ADR-049).
+	onMount(() => {
+		if (!collectionId) return;
+		if (proSync.isPro) {
 			membership.loadCollection(collectionId);
+			return;
 		}
+		return proSync.onProConfirmed(() => membership.loadCollection(collectionId));
 	});
 
 	let view = $derived(membership.get(collectionId));
--- a/packages/desktop-app/src/lib/components/layout/app-shell.svelte
+++ b/packages/desktop-app/src/lib/components/layout/app-shell.svelte
@@ -46,15 +46,10 @@
   // covers the brief window before the first daemon-status event arrives (or
   // a short grace period elapses); `unreachable` reflects a terminal
   // `not_running` event.
-  let daemonConnecting = $state(true);
-  let daemonUnreachable = $state(false);
-  $effect(() => {
-    const unsubscribe = daemonStatus.subscribe((s) => {
-      daemonConnecting = s.connecting;
-      daemonUnreachable = s.unreachable;
-    });
-    return unsubscribe;
-  });
+  // Read directly from the store via $-auto-subscription — no effect mirroring the
+  // store into local $state (ADR-049).
+  const daemonConnecting = $derived($daemonStatus.connecting);
+  const daemonUnreachable = $derived($daemonStatus.unreachable);
 
   // First-launch onboarding wizard (Issue #1180).
   let showOnboarding = $state(false);
@@ -78,14 +73,11 @@
   // show a one-time snackbar on first open; the inline badge handles per-node
   // review/restore. Guarded one-shot so it loads exactly once per session; fully
   // inert in community (proSync.isPro is false → load() returns empty → no toast).
-  let recoveredLoadStarted = false;
-  let recoveredNotified = false;
-  $effect(() => {
-    if (!proSync.isPro || recoveredLoadStarted) return;
-    recoveredLoadStarted = true;
+  // Fired from proSync.onProConfirmed (the tier state-transition site) rather than a
+  // guarded $effect (ADR-049); fully inert in community since the callback never fires.
+  function loadRecoveredItems(): void {
     recoveredItems.load().then(() => {
-      if (recoveredNotified || recoveredItems.items.length === 0) return;
-      recoveredNotified = true;
+      if (recoveredItems.items.length === 0) return;
       const n = recoveredItems.items.length;
       conflictNotifications.add({
         nodeId: recoveredItems.items[0].node_id,
@@ -93,7 +85,7 @@
         conflictType: 'recovered-items'
       });
     });
-  });
+  }
 
   async function handleReloginSignIn() {
     if (reloginPending) return;
@@ -249,6 +241,7 @@
     let cleanupMCP: (() => Promise<void>) | null = null;
     let staleNodesInterval: ReturnType<typeof setInterval> | null = null;
     let cleanupProSync: (() => void) | null = null;
+    let unregisterProConfirmed: (() => void) | null = null;
     let unlistenTier: Promise<() => void> | null = null;
 
     if (
@@ -263,6 +256,10 @@
         .start()
         .then((stop) => (cleanupProSync = stop))
         .catch((e) => log.warn('proSync.start failed', { error: e }));
+
+      // Load the recovered-items log once the daemon confirms Pro tier. Registering
+      // here (a listener, not an effect) keeps the sequencing at the transition site.
+      unregisterProConfirmed = proSync.onProConfirmed(loadRecoveredItems);
 
       // Sync theme from backend preferences (overrides localStorage if different)
       invoke<{ activeDatabasePath: string; display: { renderMarkdown: boolean; theme: string } }>('get_settings')
@@ -607,6 +604,7 @@
         (await unlistenTier)();
       }
       cleanupProSync?.();
+      unregisterProConfirmed?.();
       if (cleanupMCP) {
         await cleanupMCP();
       }
--- a/packages/desktop-app/src/lib/components/layout/pane-content.svelte
+++ b/packages/desktop-app/src/lib/components/layout/pane-content.svelte
@@ -31,10 +31,6 @@
   let viewerLoadErrors = $state<Map<string, string>>(new Map());
   let viewerLoading = $state<Set<string>>(new Set());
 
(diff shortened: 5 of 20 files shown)


malibio:
## Code Review — PR #1569: Finish the ADR-049 `$effect` conversion

**Review Type:** Initial Review (no prior review comments)
**Reviewer:** Principal Engineer (Pragmatic Quality framework)
**Verdict:** ✅ **APPROVE**

This is a clean, disciplined finish to the ADR-049 campaign. Every changed `$effect` either disappears (replaced by `onMount` / `$derived` / store-transition hooks) or is demonstrably a legitimate DOM/listener push. The conversions are faithful, the new store abstractions are well-factored, and the tests target the actual behavior contracts. I found no blockers.

---

### Requirements Check (issue #1566)

- ✅ **Cluster 1 — no data-fetching `$effect` in the listed files.** `query-node-viewer`, `collection-node-viewer`, `schema-property-form`, `task-schema-form`, `node-card-inline` → `onMount`; `collaboration-view` → `proSync.onProConfirmed`; `pane-content` hydration → store-derived.
- ✅ **`pane-content`: `latestRequestedNodeId` guard deleted; `hydratedNodeIds` mirror deleted.** No dangling references anywhere in `src/` (only a doc-comment mention in the test header).
- ✅ **Cluster 2 — out of scope, correctly attributed to merged #1565.** Verified: no `onTitleChange`-from-effect site remains in any changed viewer; titles are computed, not pushed.
- ✅ **Cluster 3 — no prop→state mirror or subscribe-into-`$state` effect remains.** calendar mirror gone, table-view clamps via `$derived`, query-editor inits in `$state`, app-shell reads `$daemonStatus` directly, pro-sync-pill derives inbox-open.
- ✅ **Cluster 4 — dead demo effect deleted** (`SchemaPropertyFormDemo`).
- ✅ **No changed `$effect` writes reactive state directly** (see ADR-049 sweep below).
- ✅ **quality:fix / test:all** reported green by author (150 unit files / 3944 tests, 17 browser, 971 Rust, e2e gate).

---

### Focus-area findings

#### 1. `pane-content` hydration correctness (highest risk) — ✅ sound
📁 `packages/desktop-app/src/lib/components/layout/pane-content.svelte:73-91, 113-137`

The race is now **structurally impossible** rather than guarded:
- State lives per-`nodeId` on the store, not in one shared local mirror. A late in-flight `ensureNode()` for a superseded nodeId populates only *its own* store cell (`setNode`) — it cannot clobber the current node. `isNodeHydrated` reads `sharedNodeStore.getNode(activeNodeId) != null`, so it always reflects the current node regardless of resolve order. Correct.
- **Wrong-tab close:** the not-found close is gated on `$tabState.tabs.find(t => t.id === tabId)?.content?.nodeId === nodeId`, read via `untrack`. A superseded navigation whose fetch settles late finds the tab no longer pointing at its nodeId and does nothing. Cannot close the newly-active tab.
- **Genuinely-missing node:** if the tab still points at the missing nodeId, it closes. Correct, and now covered by a dedicated test.
- **Unhandled rejection:** `ensureNode()` is wrapped in try/catch; the promise is awaited inside `hydrateNode`, itself invoked via `untrack(() => hydrateNode(...))` (fire-and-forget) — the internal catch means no unhandled rejection. Covered by the "does not throw" test.
- **Virtual-date path:** `ensureNode` calls `setNode(virtualDateNode, {type:'database'})` for `isValidDateId(nodeId)`, so `getNode()` returns non-null afterward and `isNodeHydrated` flips true. The store read early-return (`if (sharedNodeStore.getNode(nodeId)) return`) correctly short-circuits re-fetch. Correct.

#### 2. `onMount` vs in-place prop change — ✅ verified safe
- `query-node-viewer` / `collection-node-viewer` are registered as **`viewer`** plugins and rendered *only* by `pane-content` inside `{#key `${pane.id}-${content.nodeId}`}` (📁 `pane-content.svelte:165`). Any nodeId change remounts them, so `onMount(loadAndQuery)` re-runs. No in-place-update path exists.
- `task-schema-form` loads a **constant** `'task'` schema — mount-once is correct by construction regardless of keying.
- `schema-property-form` loads by `nodeType`; it is only reachable through `base-node-viewer`, which is itself remounted by `pane-content`'s key on nodeId change. (Note: `schema-property-form.svelte` has **no import/lazy-load site** in `src/` — it appears to be unreferenced/dead, see 🟢 below — but the conversion is internally consistent either way.)

#### 3. `pro-sync` `onProConfirmed` / `signedInEpisode` — ✅ correct, one semantic caveat
📁 `packages/desktop-app/src/lib/stores/pro-sync.svelte.ts:96-137`
- Callback lifecycle is clean: `onProConfirmed` returns a real unregister fn when queued, a no-op when it fires immediately; app-shell calls `unregisterProConfirmed?.()` in cleanup (📁 `app-shell.svelte:607`); collaboration-view returns the unregister fn as the `onMount` cleanup (📁 `collaboration-view.svelte:65`). No leaks.
- `setUserEmail('')` on signout resets `userEmail`, so `signedInEpisode` correctly re-bumps on the next sign-in edge — verified by the "bumps again after sign-out" test.
- See 🟡 below on the one-way `proConfirmed` latch.

#### 4. `code-block-renderer` store — ✅ faithful, no behavior change
📁 `packages/desktop-app/src/lib/design/components/code-block-renderer.svelte.ts`
- Seq-based staleness preserved (`#renderSeq` / `#previewSeq`, checked after each async resolve).
- `destroy()` clears the debounce timer; `onDestroy(() => renderer.destroy())` wired in the component (📁 `code-block-node.svelte:145`).
- The two innerHTML-injection effects read `renderer.mermaidSvg` / `renderer.previewSvg` — right fields, still a legit DOM push.
- Debounce-cancel-on-re-trigger is preserved: `renderPreview()` clears the prior timer at the top, and leaving mermaid/edit mode calls `clearPreview()`. Equivalent to the old effect-cleanup return.

#### 5. `node-card-inline` — ✅ correct, slightly improved
📁 `packages/desktop-app/src/lib/components/chat/node-card-inline.svelte:37-50`
`fetchAttempted` now set in `.finally()` (after the fetch settles) rather than synchronously before it. The `{#if !node && fetchAttempted}` fallback therefore shows the "unknown" state only *after* the fetch has actually failed to produce a node — this removes a pre-existing premature-flash window. When the node is already cached, `onMount` returns early and the truthy `node` branch renders. Correct.

#### 6. ADR-049 conformance sweep — ✅ pass
The only `$effect` sites remaining in changed files:
- `code-block-node.svelte:109/115` — innerHTML DOM injection (blessed).
- `code-block-node.svelte:128/135` — call `renderer.render()/renderPreview()/clearPreview()`; write no reactive state.
- `code-block-node.svelte:290` — click-outside listener register/teardown (blessed, unchanged).
- `pane-content.svelte:126` — the residual. Its body reads tab state (tracked) and calls `loadViewerForNodeType` + `hydrateNode`, **both wrapped in `untrack()`**. It writes no reactive state in its own body; the module-cache maps and the store cell are written *inside* the untracked calls (a lazy module load + an `ensureNode` push into the store — the canonical "push into a non-reactive external system"). This is the honest gray-area residual the issue flagged, and it is **acceptable** under the rule: the effect does not sync/derive/sequence reactive state, it pushes two side-effects into their subsystems on tab change. The code and comment are transparent about it. Not a violation.

#### 7. Test quality — ✅ real coverage
- `pane-content-hydration.test.ts` is a **logic-mirror harness** (re-implements `hydrateNode`, doesn't import it) — a known, documented limitation for BaseNodeViewer-adjacent components. It nonetheless encodes the *new* event-driven contract correctly, including two genuinely new cases (genuinely-not-found closes tab; already-in-store skips fetch) and the reversed assertion (both cells hydrate independently). See 🟢 on its structural limitation.
- `node-card-inline.test.ts` renders the real component and pins all three branches.
- `pro-sync.test.ts` covers the `signedInEpisode` edges thoroughly.

---

### Findings by severity

#### 🟡 Important (non-blocking — worth a follow-up or a confirming manual check)

**🟡 One-way `proConfirmed` latch never resets across a pro→free→pro cycle.**
📁 `pro-sync.svelte.ts:112-120`
`setTier` latches `proConfirmed = true` on the first transition to `pro` and never clears it. After a signout/downgrade and a later re-subscribe to pro, `onProConfirmed` callbacks will **not** re-fire on the second pro transition (a newly-registered callback fires immediately instead). For the current consumer (recovered-items load once per session) this is fine, but if a future consumer needs "on *every* pro confirmation" it will silently get "once per process." Consider either documenting this as intentionally once-per-session, or resetting `proConfirmed` in `signout()`/on tier→non-pro so the semantics match `signedInEpisode`'s re-bump behavior. *Principle: least-surprise — two adjacent transition primitives (`signedInEpisode`, `proConfirmed`) with opposite reset semantics is an easy future footgun.*

**🟡 `calendar` mirror removal depends on bits-ui v1 runtime propagation, not pure logic.**
📁 `packages/desktop-app/src/lib/components/ui/calendar/calendar.svelte:36-42`
The removed workaround explicitly documented a bits-ui bug ("Direct `bind:value` doesn't trigger parent updates"). The change assumes v1.4.7 fixed it. `bits-ui: ^1.4.7` supports this claim and `date-picker.test.ts` exercises the path, but because this is the one change riding on external-library behavior rather than local reasoning, I'd recommend a quick manual confirm that selecting a date in a calendar still propagates to the parent (`onValueChange` + the bound `value`) before merge. *Principle: behavior tied to a third-party runtime quirk deserves an explicit end-to-end check, not just a green unit suite.*

#### 🟢 Suggestions (optional)

**🟢 `table-view` clamp changes reset semantics from "reset to page 0" to "clamp + retain intent."**
📁 `packages/desktop-app/src/lib/components/query/table-view.svelte:32-61`
Previously a `nodeIds` change reset `currentPage = 0`; now `currentPage` is retained and the displayed `page` is clamped. If the result set shrinks then grows again, the old raw `currentPage` re-surfaces rather than staying at 0. This is arguably better UX (preserves intent) and the pager buttons re-anchor `currentPage = page ± 1` on click, so it can't drift out of range interactively. Flagging only so the behavior change is a conscious choice — no action needed if intended.

**🟢 `onProConfirmed`'s registered-before-confirmed path is untested.**
📁 `packages/desktop-app/src/tests/stores/pro-sync.test.ts`
The tests (correctly, given the singleton one-way latch) only exercise the "already pro" path. The transition path — register a callback *before* pro resolves, then confirm pro — is exactly what app-shell relies on at startup and is not covered. A small non-singleton unit (or a `new ProSyncStore()` instance if the class were exported) would close this. Low priority.

**🟢 `pane-content-hydration.test.ts` re-implements `hydrateNode` rather than importing it.**
📁 `packages/desktop-app/src/tests/components/pane-content-hydration.test.ts:24-70`
The harness is a copy of the impl logic, so it can drift from the real function without failing. This is a documented, pragmatic tradeoff (full BaseNodeViewer mount is heavy), and the contract it encodes is correct — but it validates the *shape*, not the shipped code. If the hydration logic is ever extracted into a small importable helper, this test would gain real teeth.

**🟢 `schema-property-form.svelte` appears unreferenced.**
📁 `packages/desktop-app/src/lib/components/property-forms/schema-property-form.svelte`
No `import`/`lazyLoad` site exists for it in `src/` (only comment mentions). Its `onMount` conversion is harmless, but it may be dead code — worth confirming and deleting in a follow-up (CLAUDE.md: delete dead code, don't keep it). Out of scope for this PR.

---

### Summary

A textbook ADR-049 wrap-up: effects converted to the right primitive, the pane-content race eliminated at the substrate level (not papered over with a guard), and a well-factored `CodeBlockRenderer` store. The two 🟡 items are non-blocking — the latch semantics are a future-proofing note, and the calendar change just warrants one manual propagation check given its library dependency. No local files created; no merge performed.

🤖 Reviewed with the Pragmatic Quality framework

malibio (commented):
Review posted as a comment above. Verdict: APPROVE (self-authored PR, so a formal approve is blocked by GitHub). No blockers; two non-blocking 🟡 notes.

malibio:
## Review response

Thanks — addressed the actionable items:

- 🟡 **`proConfirmed` one-way latch** — documented the intentional semantics in `pro-sync.svelte.ts` (Pro tier is a per-process capability probe, never resets on sign-out; contrasted with `signedInEpisode`'s per-session re-bump). Commit `1758f960`.
- 🟡 **calendar bits-ui propagation** — resolved by analysis, no behavior change to worry about: **every real consumer reads the selected date via `onValueChange`, not `bind:value`** (`base-node.svelte:990`, `task-schema-form.svelte`, `generic-schema-form.svelte`). My change preserved `handleValueChange → onValueChange` exactly; `bind:value` is only used by `base-node` to *reset* the value afterward. The channel the app depends on is unchanged, so removing the `internalValue` mirror is safe.

🟢 dispositions:
- **`schema-property-form.svelte` "unreferenced"** — not dead code; it's referenced by `base-node-viewer.svelte`, `plugin-registry.ts`, `types.ts`, and `task-node-viewer.svelte`.
- **table-view page clamp vs reset-to-0** — intentional: clamping retains the user's page position when the result set changes (only correcting out-of-range), which is better UX than snapping to page 0 and is the ADR-conformant `$derived` form.
- **`onProConfirmed` before-confirmed path untested / pane-content harness re-implements** — both stem from the module-singleton latch and the documented full-mount limitation; left as-is per the review's own note that they're structural.

PR #1572, opened 2026-07-07
ADR-053 daemon multi-database: registry scaffold + build_services split (#1532)

mstomar125:
Foundation for ADR-053 (*One Daemon, Multiple Local Databases*). Two incremental, low-risk stages of #1532; the multi-database routing that consumes them lands next.

## Stage 1–2 — proto + registry (dormant scaffold)
- New `packages/proto/proto/database_service.proto`: `DatabaseService { List, Create(name, path?), Register(path), Remove (unregister-only, never deletes files), SetDefault, Rename }` + a `DatabaseStatus` enum (Open/Closed/Missing). Wired into `build.rs` and re-exported from `packages/proto/src/lib.rs`.
- New `packages/daemon/src/services/database_manager.rs`: `DatabaseId` (ULID) + `DatabaseEntry` + a `Registry` persisted to a dedicated `~/.nodespace/databases.toml` (kept separate from `SettingsService`'s `daemon.toml`), and a `DatabaseManager` holding the registry plus an open-databases map. Registry mutations (create/register/remove/set_default/rename) are implemented and persist; `get_or_open` returns a clear "not yet wired" error (no `todo!()`), since the per-database service set it hands back is assembled in Stage 3+.

The scaffold compiles and is **not yet wired** into service assembly or routing — intentional, to keep the sync lock-step green until Stage 5.

## Stage 3 — split `build_services()` into shared + per-database
The seam the routing needs. `build_services()` becomes:
- **`build_shared_services()`** → process-global (`PtySessionManager`, daemon settings) plus the **embedding model loaded once for the whole process** and published over a `tokio::watch` channel.
- **`build_database_services(db_path, &shared)`** → per-database `SqliteStore`, `NodeService` (+seeding), gRPC handlers, `GraphContextAssembler`, import, local-agent watcher, and per-database embedding state — consuming the shared PTY manager rather than making its own.
- The embedding model is **decoupled from stores** (ADR-053: one model backs all databases): the old per-store load task splits into `load_shared_embedding_model_bg` (loads the raw model once, publishes over `watch`) and `wire_database_embeddings_bg` (each database awaits the shared model, then wires its own `NodeEmbeddingService` + `EmbeddingProcessor` + waker).
- All four serve loops (headless/tray × unix/windows) call `build_shared_services` then `build_database_services`; settings now comes from the shared set.

**Behavior-preserving** for the single default database served today.

## Verification
- `cargo build`/`fmt`/`clippy -p nodespace-daemon --all-targets -- -D warnings` clean.
- `cargo test -p nodespace-daemon` **71/71** — including `grpc_round_trip` and `agent_session` end-to-end suites that exercise the assembled services.
- Rebased on current `main`.

## Not in this PR (follow-on, tracked under #1532)
- Hoist the chat `CompositeModelManager` to shared (it's built inside `LocalAgentServiceImpl::new`, per-database — needs a constructor change; behaviorally identical for one database).
- Stage 4: `x-ns-database-id` interceptor + storing `DatabaseServices` in `DatabaseManager.open`.
- Stage 5: `DatabaseService` handler + `DatabaseServiceServer` on `BaseServices` (needs the nodespace-sync lock-step). Stages 6–7: `database_id` on event envelopes; drop `active_database_path`/`NODESPACED_DB_PATH` + Windows pipe parity.

Part of #1532.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3890,6 +3890,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "tray-icon 0.21.3",
+ "ulid",
  "uuid",
 ]
 
@@ -7421,6 +7422,16 @@ dependencies = [
  "memoffset 0.9.1",
  "tempfile",
  "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "ulid"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "470dbf6591da1b39d43c14523b2b469c86879a53e8b758c8e090a470fe7b1fbe"
+dependencies = [
+ "rand 0.9.4",
+ "web-time",
 ]
 
 [[package]]
--- a/packages/daemon/Cargo.toml
+++ b/packages/daemon/Cargo.toml
@@ -24,6 +24,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 serde = { workspace = true }
 serde_json = { workspace = true }
 toml = "0.8"
+ulid = "1"
 uuid = { workspace = true }
 async-trait = { workspace = true }
 chrono = { workspace = true }
--- a/packages/daemon/src/main.rs
+++ b/packages/daemon/src/main.rs
@@ -30,7 +30,7 @@ use nodespace_daemon::{
     SettingsServiceImpl,
 };
 use nodespace_nlp_engine::EmbeddingService;
-use tokio::sync::RwLock;
+use tokio::sync::{watch, RwLock};
 use tonic::transport::Server;
 
 #[cfg(unix)]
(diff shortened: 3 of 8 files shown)


mstomar125:
**APPROVE**

Reviewed the full diff at `b2a434f`. This is well-staged foundational work — a genuinely dormant scaffold plus a behavior-preserving refactor, with the cross-repo-coupled parts correctly deferred.

**Stage 1–2 scaffold — dormant + data-safe:**
- `DatabaseManager`/`Registry` are referenced only in `database_manager.rs` and the `services/mod.rs` re-export — **never wired into `main.rs`'s serve path**, so this can't change current runtime behavior. Good.
- `remove()` is **unregister-only** — it `retain`s the registry, saves, clears `default_database` if it pointed at the removed id, and drops the open-map entry; there is **no `fs::remove_file`**, so no database file is ever deleted. It also releases the registry write lock (block scope) before taking the open-map lock — no two-lock overlap.
- `get_or_open()` returns clear `anyhow` errors (registered-but-not-wired / not-registered) — no `todo!()`/panic, so a premature call fails gracefully.
- Registry persists to a dedicated `~/.nodespace/databases.toml`, deliberately separate from `daemon.toml`, so the two writers can't clobber each other. Load treats `NotFound` as an empty default.

**Stage 3 `build_services` split — behavior-preserving:**
- The embedding-model decoupling is correct: `build_shared_services` loads the model once and publishes over a `tokio::watch`; `wire_database_embeddings_bg` consumes it with the canonical pattern — `loop { if let Some = borrow_and_update() → break; if changed().await.is_err() → return }`. `watch` retains the latest value, so there's **no missed-update race** even if the model is published before a database wires, and the task exits cleanly on sender-drop (no model file) rather than leaking a forever-await.
- All four serve loops (headless/tray × unix/windows) call `build_shared_services` then `build_database_services(&db_path, &shared)`, consuming the shared PTY manager — for the single default database served today this is equivalent to the old assembly.

**Staging discipline (healthy):** Stage 5 (the `DatabaseService` handler + server that needs the nodespace-sync lock-step) and the `x-ns-database-id` routing are explicitly *not* here — so nothing half-wired lands, and there's no cross-repo deploy-order trap yet. The deferred `CompositeModelManager` hoist is documented as behaviorally identical for one database.

CI green — `test:all` + `test:e2e`, and the PR's `71/71` daemon suite includes `grpc_round_trip` + `agent_session`, which exercise the reassembled services end-to-end. No changes requested.

PR #1573, opened 2026-07-07
Implement DatabaseSettingsNode (database-settings schema + behavior)

mstomar125:
## Summary

Implements `DatabaseSettingsNode` as a core node type: the `database-settings` schema entry, `DatabaseSettingsNodeBehavior`, first-launch seeding of the singleton plus its owner `has_role` edge, and registration of `has_role` as a universal built-in relationship.

Per ADR-037, database-level state is separated from identity: `sync_enabled` (user intent) and `auth_status` (system-managed cloud bind state) move onto this singleton, and tenant authority moves onto a `has_role` edge (PersonNode -> DatabaseSettingsNode) instead of living on PersonNode.

### Fixed-ID collision (the key implementation detail)

Schema nodes store their id as the slug, so the `database-settings` schema node already occupies id `"database-settings"`. To keep the instance deterministic without colliding, the singleton **instance** uses a distinct reserved id `"database-settings-singleton"`. Determinism/idempotency (the intent behind the issue's "fixed ID" criterion) is preserved: `create_node` treats a second `database-settings` create as a no-op that returns the existing id.

## Acceptance criteria

- **`database-settings` schema entry with `sync_enabled` and Core-protected `auth_status`** — `packages/core/src/models/core_schemas.rs`: Core `SchemaNode` with `sync_enabled` (boolean, default `false`, Core-protected) and `auth_status` (enum `local`/`connected`, default `local`, Core-protected).
- **`DatabaseSettingsNodeBehavior` implements all `NodeBehavior` methods and is registered** — `packages/core/src/behaviors/mod.rs`: `type_name`, `validate`, `can_have_children` (`false`), `supports_markdown` (`false`), `default_metadata` (`{sync_enabled:false, auth_status:"local"}`), `get_embeddable_content`/`get_parent_contribution` (`None`); registered in `NodeBehaviorRegistry::new()`.
- **`validate` enforces `auth_status` enum and boolean `sync_enabled`** — rejects `auth_status` outside `{local, connected}` and non-boolean `sync_enabled` with `InvalidProperties` (same shape as `AiChatNodeBehavior`).
- **`can_have_children()` -> `false`** — implemented.
- **Fixed ID; second creation is a no-op, not an error** — reserved instance id `database-settings-singleton`; `create_node` returns the existing id on a duplicate `database-settings` create (`packages/core/src/services/node_service/crud.rs`).
- **Seeded at first launch with `sync_enabled:false`, `auth_status:local`, and owner `has_role` edge to local PersonNode** — `seed_database_settings_if_needed` in `packages/core/src/services/node_service/mod.rs`, invoked from `NodeService::new()` after the local person seed; owner edge carries `{role:"owner", status:"active"}`. Idempotent across reopen.
- **`has_role` relationship type registered** — added to the built-in `matches!` set in `create_relationship` and to the built-in-relationships error string (`packages/core/src/services/node_service/relationship.rs`).
- **Unit tests cover singleton, seeding, and validation** — schema-shape test, behavior tests (registration, valid/invalid `auth_status`, non-bool `sync_enabled`, capabilities, defaults, embeddable), and async NodeService tests (seed yields exactly one node with correct defaults + exactly one owner `has_role` edge; second create is a no-op; reopen does not reseed).
- **Passes `cargo clippy`; no lint suppression** — clippy clean with `-D warnings`; no `#[allow]` / suppression added.

## Out of scope (not implemented)

Pro sync writes (`auth_status=connected`, `sync_enabled=true`), RLS enforcement, `auth_identities` bind, sync-toggle UI, and last-admin invariant enforcement.

## Checks

- `cargo test -p nodespace-core -- --test-threads=2` — 985 lib + all integration + 125 doctests pass, 0 failed
- `cargo clippy -p nodespace-core --all-targets -- -D warnings` — clean
- `cargo fmt --all --check` — clean

Note: ADR-037 documents the data-model revision this implements; a docs update lives in the nodespace-docs repo (out of scope for this code PR).

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/packages/core/src/behaviors/mod.rs
+++ b/packages/core/src/behaviors/mod.rs
@@ -2082,6 +2082,74 @@ impl NodeBehavior for PersonNodeBehavior {
     }
 }
 
+/// Built-in behavior for the database-settings singleton node
+///
+/// DatabaseSettingsNode is the single container for database-level configuration:
+/// `sync_enabled` (user intent to sync) and `auth_status` (system-managed cloud
+/// bind state). Tenant roles are NOT stored here — they live on the `has_role`
+/// edge (PersonNode → DatabaseSettingsNode). Per ADR-037, both `role` and
+/// `auth_status` were moved off PersonNode as part of that revision.
+pub struct DatabaseSettingsNodeBehavior;
+
+impl NodeBehavior for DatabaseSettingsNodeBehavior {
+    fn type_name(&self) -> &'static str {
+        "database-settings"
+    }
+
+    fn validate(&self, node: &Node) -> Result<(), NodeValidationError> {
+        // Validate auth_status if present
+        if let Some(auth_status) = node.properties.get("auth_status") {
+            if let Some(auth_status_str) = auth_status.as_str() {
+                match auth_status_str {
+                    "local" | "connected" => {}
+                    _ => {
+                        return Err(NodeValidationError::InvalidProperties(format!(
+                            "Invalid auth_status '{}': must be one of local, connected",
+                            auth_status_str
+                        )));
+                    }
+                }
+            }
+        }
+
+        // Validate sync_enabled is a boolean if present
+        if let Some(sync_enabled) = node.properties.get("sync_enabled") {
+            if !sync_enabled.is_boolean() && !sync_enabled.is_null() {
+                return Err(NodeValidationError::InvalidProperties(
+                    "sync_enabled must be a boolean".to_string(),
+                ));
+            }
+        }
+
+        Ok(())
+    }
+
+    fn can_have_children(&self) -> bool {
+        false // Configuration container; it holds no child content
+    }
+
+    fn supports_markdown(&self) -> bool {
+        false // Settings are not markdown-rendered content
+    }
+
+    fn default_metadata(&self) -> serde_json::Value {
+        serde_json::json!({
+            "sync_enabled": false,
+            "auth_status": "local"
+        })
+    }
+
+    /// Settings carry no semantic content; they are not embedded for search.
+    fn get_embeddable_content(&self, _node: &Node) -> Option<String> {
+        None
+    }
+
+    /// Settings do not contribute to parent embeddings.
+    fn get_parent_contribution(&self, _node: &Node) -> Option<String> {
+        None
+    }
+}
+
 /// Registry for managing node behaviors
 ///
 /// The registry provides thread-safe storage and retrieval of node behaviors.
@@ -2183,6 +2251,7 @@ impl NodeBehaviorRegistry {
         registry.register(Arc::new(SkillNodeBehavior));
         registry.register(Arc::new(ToolNodeBehavior));
         registry.register(Arc::new(PersonNodeBehavior));
+        registry.register(Arc::new(DatabaseSettingsNodeBehavior));
 
         registry
     }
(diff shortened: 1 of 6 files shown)


mstomar125:
**APPROVE** — head `e6cf5e22`. Well-structured, community-safe, and tested for the happy path. This is a clean phase-1 landing of the ADR-037 `database-settings` singleton. A few verify/should-fix items and one cross-layer flag below; none are hard blockers, but see the merge note at the end (core needs the second review pass).

**Verified good:**
- **Community-safe / free-user-inert.** The singleton seeds `sync_enabled:false` + `auth_status:local` and `DatabaseSettingsNodeBehavior` returns `can_have_children:false`, `supports_markdown:false`, and `None` for both `get_embeddable_content`/`get_parent_contribution` (`behaviors/mod.rs`) — so it's a config node, not content, not embedded for search. The schema field description even states `sync_enabled` is "Inert on the free/community tier." Community read/write paths are unchanged.
- **Idempotent seeding.** `seed_database_settings_if_needed` (`node_service/mod.rs`) early-returns when a `database-settings` node exists (backfills existing DBs on next open), and the `create_node` guard (`crud.rs:63-79`) makes a second create a no-op returning the existing id. Runs after `seed_local_person_if_needed`, and fails with `InitializationError` (not a panic) if no PersonNode exists.
- **Fixed-ID collision handled:** schema id `"database-settings"` vs instance id `"database-settings-singleton"` (the `DATABASE_SETTINGS_NODE_ID` constant); `query_nodes_by_type("database-settings")` returns the instance, not the schema node. Good.
- Tests cover the schema fields and the seeded singleton + single owner `has_role` edge.

**Should-fix / verify:**
1. **`DatabaseSettingsNodeBehavior::validate` likely reads the wrong path.** It checks `node.properties.get("auth_status")` / `get("sync_enabled")` at the **top level**, but seeding stores them namespaced under `properties["database-settings"]` (and the test asserts `node.properties["database-settings"]["auth_status"]`). If validation runs on the stored shape, these checks never fire — dead code. The schema enum-validation still enforces `auth_status ∈ {local, connected}`, so it's not a correctness hole, but please align the path or drop the redundant check so it isn't misleading.
2. **Interrupted seed doesn't self-heal.** If `create_node(settings)` commits but `create_relationship(has_role)` then fails, init returns `Err`; on the next launch the guard sees the singleton exists → skips seeding → the owner edge is never created. Consider healing the edge independently (check for the singleton **and** its owner edge), or wrapping both in one transaction.
3. **`has_role` is registered as a universal built-in (`relationship.rs:487`) with no endpoint-type validation** — any node → any node is allowed. Low risk (mostly seed-created), but member_of-style endpoint validation would be tighter.

**Cross-layer flag (not a defect in this PR — it's correctly phased):** ADR-037 moves `auth_status` onto this singleton with vocab `{local, connected}`, but the Pro **sync daemon currently reads/writes `auth_status` on the PersonNode** with vocab `{local, bound}` (see `cloud_sync/identity.rs`). So post-merge the singleton's `auth_status` is **inert** until a companion sync-daemon change adopts it — and that change must reconcile the `local/bound` vs `local/connected` vocabularies and handle that there's **no PersonNode→singleton data migration** (existing bind state won't carry over). Safe to land this alone; just make sure the sync-side follow-up is tracked so the singleton doesn't sit unused.

**Note (pre-existing, not this PR):** `auth_status`/`sync_enabled` are marked `SchemaProtectionLevel::Core`, but Core-protection doesn't appear to be enforced on the write path (no `SchemaProtectionLevel::Core` check in `crud.rs`/behavior validate) — so "Core-protected" is nominal here as elsewhere. Worth confirming whether protected fields are actually write-guarded anywhere.

**Merge note:** per the core policy this is review #1 (of the required multiple) — I'm not treating it as merge-ready on this pass. If you push the should-fixes, the new SHA triggers my re-review (the second pass); if it stays as-is, it still needs a second confirming review before it's pulled to merge. CI is green (both gates).

PR #1574, opened 2026-07-07
Scope browser-mode tests to engine-agnostic DOM and add a tauri-driver smoke suite

malibio:
Closes #1527

### DIFF
--- a/.github/workflows/nightly-webkit.yml
+++ b/.github/workflows/nightly-webkit.yml
@@ -1,0 +1,51 @@
+name: Nightly WebKit smoke
+
+# Non-blocking, nightly smoke coverage of the shipping WebKit engine.
+#
+# The desktop app renders in WKWebView (WebKit / JavaScriptCore), but the
+# browser-mode test tier (`test:browser`) runs in Chromium (Blink). Anything
+# WebKit-specific can pass in Chromium yet break in the real app. This job runs
+# the SAME engine-agnostic browser suite (src/tests/browser/**) against
+# Playwright's bundled WebKit build — the shipping engine's family — so a
+# WebKit-only regression surfaces here.
+#
+# This workflow is deliberately NOT a required status check and is NOT wired into
+# the blocking `ci.yml` gate: it is a nightly signal, not a merge gate. Chromium
+# stays the fast, blocking browser tier (via `test:integration:full` / local
+# runs); WebKit is the nightly delta check.
+
+on:
+  schedule:
+    # 07:00 UTC ≈ midnight PT — nightly.
+    - cron: "0 7 * * *"
+  workflow_dispatch: {}
+
+concurrency:
+  group: nightly-webkit-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  webkit:
+    name: test:webkit (Playwright WebKit)
+    runs-on: macos-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: "1.3.11"
+
+      - name: Install frontend dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Generate SvelteKit types
+        run: bun run --cwd packages/desktop-app sync
+
+      - name: Install Playwright WebKit browser
+        run: bunx --bun playwright install webkit --with-deps
+        working-directory: packages/desktop-app
+
+      - name: Run WebKit smoke suite
+        run: bun run test:webkit
--- a/.github/workflows/nightly-webkit.yml
+++ b/.github/workflows/nightly-webkit.yml
@@ -44,7 +44,13 @@ jobs:
         run: bun run --cwd packages/desktop-app sync
 
       - name: Install Playwright WebKit browser
-        run: bunx --bun playwright install webkit --with-deps
+        # The WebKit revision is pinned transitively by the lockfile's Playwright
+        # version; log it so a nightly red is easy to triage as an engine-version
+        # delta vs. an app regression. (No --with-deps: that installer is
+        # Linux-only and a no-op on macos-latest.)
+        run: |
+          bunx --bun playwright --version
+          bunx --bun playwright install webkit
         working-directory: packages/desktop-app
 
       - name: Run WebKit smoke suite
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "test:coverage": "bun run --cwd packages/desktop-app test:coverage",
     "test:skill": "bun run --cwd packages/skill test",
     "test:e2e": "bun run --cwd packages/desktop-app test:e2e",
+    "test:webkit": "bun run --cwd packages/desktop-app test:webkit",
     "test:all": "bun run test && bun run test:skill && bun run rust:test",
     "test:all:db": "bun run test:db && bun run test:skill && bun run rust:test",
     "check": "bun run --cwd packages/desktop-app check",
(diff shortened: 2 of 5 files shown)


malibio:
## Code Review Summary — PR #1574

**Scope:** Test-infrastructure only (no runtime/product code). 5 files: 2 new configs/workflow, 3 edits.

This is a clean, well-scoped, net-positive change. The design correctly substitutes Playwright's bundled WebKit build for the OS-incompatible literal `tauri-driver`/`WebKitWebDriver` approach — the right call, since tauri-driver has no macOS support and WKWebView exposes no WebDriver, while CI runs on `macos-latest`. Targeting the WebKit/JavaScriptCore engine *family* on the existing runner gets the engine signal the issue actually wants without a new OS runner. The suite verifiably passes under both engines, the nightly job is genuinely non-blocking, and the change carries no lint suppressions, no docs-in-repo, and no issue-number leakage into code/docs content. **Recommendation: APPROVE.**

### Requirements Check (issue #1527)

- ✅ **Browser-mode tests scoped to engine-agnostic behavior.** Addressed via an explicit ENGINE-AGNOSTIC CONTRACT docblock in `vitest.browser.config.ts` plus the empirical proof that the same suite passes under both Blink and WebKit. The default `vitest.config.ts` already excludes `src/tests/browser/**`, so the tier is cleanly partitioned. See the note below on why no test-body changes were required.
- ✅ **A smoke suite exists (nightly, non-blocking).** `nightly-webkit.yml` is `schedule` + `workflow_dispatch`, standalone, not referenced by `ci.yml` or `scripts/test-gate.ts`, and `test:webkit` is deliberately excluded from `test:all`. Genuinely non-blocking.

📝 On criterion #1: the contract-docblock-plus-shared-suite approach is adequate *and* arguably stronger than editing test bodies. The 90 existing browser tests already pass under WebKit unchanged, which is direct evidence they were already engine-agnostic — so there was nothing to rewrite. The docblock converts that implicit property into an enforced, documented contract, and the WebKit config is the executable check that guards it going forward. Scoping here means "constrain what the tier is allowed to assert," which a contract + dual-engine run enforces better than any one-time edit would.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

🟡 **Near-duplicate configs risk silent drift.** `vitest.webkit.config.ts` and `vitest.browser.config.ts` are byte-for-byte identical except `browser.name` ('webkit' vs 'chromium') and the docblock. Today the duplication is trivial, but the two files must stay in lockstep on `include`, `setupFiles`, timeouts, `pool`/`singleFork`, and aliases — the engine-agnostic contract depends on both configs running the *exact same suite the same way*. If someone later bumps a timeout or adds a setup file in one and not the other, the tiers silently diverge and the "both must pass" guarantee quietly weakens.
📁 `packages/desktop-app/vitest.webkit.config.ts` / `vitest.browser.config.ts`

Recommendation (optional): extract a shared factory, e.g. `makeBrowserConfig(engine: 'chromium' | 'webkit')` in a small `vitest.browser.base.ts`, with each file reducing to a one-line call. Engineering principle: DRY applied to the *invariant* (the shared suite + run semantics), keeping only the *variant* (engine name) per file.

Counterpoint — reasonable to defer: with only two thin files and the identity currently guaranteed by verification (both run green), a base-config abstraction is borderline over-engineering (YAGNI). If you keep them separate, the pragmatic mitigation is a one-line comment in each config pointing at the other as its lockstep twin, so the coupling is discoverable. Either resolution is fine; flagging so the drift risk is a conscious choice, not an accident.

#### Nitpicks

🟢 **Nit: `--with-deps` is a no-op on macOS.** `bunx --bun playwright install webkit --with-deps` — the `--with-deps` flag drives an OS-level system-dependency installer that Playwright only implements for Linux (apt-based). On `macos-latest` it is silently skipped, so it's harmless but misleading — it reads as if macOS system deps are being provisioned when nothing happens. Dropping `--with-deps` makes the step's intent match its effect. (Leave it if you anticipate ever moving this job to a Linux runner; in that case a one-word comment noting "Linux-only, inert on macOS" would remove the head-scratch.)
📁 `.github/workflows/nightly-webkit.yml:47`

🟢 **Nit: WebKit build is unpinned — nightly may drift with the Playwright version.** The local worktree needed a specific WebKit build installed, and CI resolves the WebKit revision from whatever `playwright` version `bun install --frozen-lockfile` locks. That's actually the correct coupling (lockfile pins Playwright → Playwright pins its WebKit revision), so there's no separate version to pin here. The only residual risk: a Playwright bump changes the WebKit revision under you and a nightly failure could reflect an engine-version delta rather than an app regression. No action needed now — just worth knowing that when this job goes red, "did Playwright's WebKit revision move?" is a valid first triage question. Consider echoing `bunx playwright --version` as a diagnostic step if triage friction ever bites.

🟢 **Nit: "blocking browser tier" phrasing is slightly generous.** The workflow docblock says "Chromium stays the fast, blocking browser tier." Strictly, neither `ci.yml` nor the pre-push gate runs `test:browser` — `test:all` is `test && test:skill && rust:test` (Happy-DOM unit + skill + Rust), so the Chromium *browser* suite is only enforced via `test:integration:full` / local runs, which the parenthetical does correctly note. The word "blocking" over-claims relative to CI. Minor wording; the qualifier saves it. Consider "the fast, local/pre-integration browser tier" for precision.
📁 `.github/workflows/nightly-webkit.yml:15`

### Standards & Security
- No lint suppressions, no `any`, no `{@html}`, no raw `console.*` — N/A for config/YAML but clean regardless.
- No `docs/` dir or nested `.md` added to the repo; testing-guide lives in `nodespace-docs` as required.
- No issue numbers leaked into code or doc *content*; `(closes #1527)` in the commit message is process mechanics and explicitly allowed.
- Workflow uses least-surprise, pinned tool versions consistent with `ci.yml` (checkout@v7, setup-bun@v2, bun 1.3.11, frozen-lockfile). `concurrency` with `cancel-in-progress` is appropriate for a nightly. No secrets, no elevated permissions, no untrusted input. No security surface.

### Verdict
Net positive, low risk, criteria met. None of the findings block merge — the config-duplication item is the only one worth a deliberate decision (refactor vs. accept-with-a-pointer-comment), and either choice is defensible.

**Recommendation: APPROVE**

malibio (commented):
Pragmatic code review complete — recommendation: **APPROVE** (posted as a self-authored PR, so GitHub blocks a formal Approve; recording as a review comment). Full report is in the PR comment above. Net-positive, low-risk, both acceptance criteria met. Only non-blocking findings: 1 Improvement (config-duplication drift risk) + 3 Nits.

malibio:
## Review feedback addressed (commit `e4104993`)

The `/pragmatic-code-review` was **APPROVE** with no Critical or Important findings — all items were 🟢 Suggestions/Nitpicks. Disposition:

| Finding | Disposition |
|---------|-------------|
| Near-duplicate configs risk silent drift | ✅ Added `LOCKSTEP TWIN` pointer comments to both `vitest.browser.config.ts` and `vitest.webkit.config.ts` so the required coupling (identical except `browser.name`) is discoverable. Skipped the shared-factory refactor — borderline over-engineering / YAGNI for two thin files, per the review's own note. |
| `--with-deps` no-op on macOS | ✅ Dropped it (Linux-only installer, inert on `macos-latest`) with a rationale comment. |
| WebKit build unpinned by design | ✅ Added `playwright --version` log to the install step so a nightly red is easy to triage as an engine-version delta vs. an app regression. |
| "blocking browser tier" phrasing slightly generous | ⏭️ Skipped — marginal wording; the existing parenthetical qualifier already covers it. Not worth a churn commit. |

**Verification:** `test:webkit` still 90/90 under WebKit; changes are comment/CI-only. Full pre-push gate (test:all + cargo build + test:e2e) passed on push.

## Re-Review Decision

**Decision: NO RE-REVIEW NEEDED** — the changes are comment-only + a CI-step cleanup with no logic or config-behavior change. PR is ready for merge pending your approval.

PR #1575, opened 2026-07-07
Harden local-agent silent-failure guards: empty response and narrated-not-executed tool calls slip through

malibio:
Closes #1531

### DIFF
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -55,6 +55,13 @@ const SYSTEM_PROMPT_BUDGET: u32 = 4_000;
 /// Tokens available for conversation history.
 const HISTORY_TOKEN_BUDGET: u32 = TOTAL_TOKEN_BUDGET - SYSTEM_PROMPT_BUDGET;
 
+/// Last-resort user-facing message when a turn produces nothing usable — no
+/// tool executions to summarize and no (non-empty, non-pseudo-code) text from
+/// the model. Guarantees the chat UI always shows an honest failure notice
+/// rather than an empty bubble.
+const EMPTY_RESPONSE_FALLBACK: &str =
+    "⚠️ I wasn't able to produce a response for that. Please try again.";
+
 fn session_prompt_override(session: &AgentSession) -> Option<&str> {
     session.system_prompt_override.as_deref()
 }
@@ -120,6 +127,32 @@ fn contains_action_claim(text: &str) -> bool {
         "has been deleted",
     ];
     ACTION_PHRASES.iter().any(|p| lower.contains(p))
+}
+
+/// Detect a tool call that the model emitted as plain text instead of invoking.
+///
+/// Some locally-hosted models (e.g. `mistral:7b` via Ollama) don't reliably use
+/// the structured `tool_calls` response field — they instead print the call as
+/// prose, e.g. `execute_query(target_type='task', ...)`. No tool ever executes,
+/// yet the pseudo-code is persisted verbatim as the assistant's answer, so the
+/// user sees a raw code snippet with no indication anything went wrong.
+///
+/// This is deliberately narrow: it only matches a registered tool name
+/// immediately followed by `(` (allowing whitespace between). Matching a general
+/// `snake_case(` shape would false-positive on legitimate prose that references
+/// functions. Tool names are taken from the registry ([`crate::local_agent::tools::Tool::ALL`])
+/// so the detector stays in sync as tools are added or removed.
+fn looks_like_narrated_tool_call(text: &str) -> bool {
+    let lower = text.to_ascii_lowercase();
+    crate::local_agent::tools::Tool::ALL.iter().any(|tool| {
+        let name = tool.name();
+        // Find each occurrence of the tool name and check the next
+        // non-whitespace character is an opening paren.
+        lower.match_indices(name).any(|(idx, _)| {
+            let after = &lower[idx + name.len()..];
+            after.trim_start().starts_with('(')
+        })
+    })
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


malibio:
## Pragmatic Code Review — PR #1575 (closes #1531)

### Summary

**Net positive — recommend merge after considering two minor improvements below.**

This is a focused, well-scoped defensive change entirely within `packages/agent/src/local_agent/agent_loop.rs` (+276/−8). It closes exactly the two silent-failure gaps described in the issue: (A) empty final responses yielding blank chat bubbles, and (B) narrated-but-not-executed tool calls leaking raw pseudo-code to the user. The guard logic is placed at all three exit paths, the heuristic is deliberately narrow and registry-driven, the tracing is structured and correct, and the test coverage maps directly onto the issue's three scope items. No lint suppression, no dead code, no `console`-equivalent logging violations — clippy and the targeted tests are green locally.

The change matches the framework's "durable safety net over per-model fixes" rationale: it makes the failure honest regardless of which model/tool-calling quirk triggered it.

### Verification performed
- `cargo test -p nodespace-agent` narrated + empty suites: **8 passed, 0 failed**
- `cargo clippy -p nodespace-agent --lib`: **clean**
- Traced all three guard placements and their interaction with the pre-existing anti-fabrication and tool-failure guards.

---

### Findings

#### Critical Issues
None.

#### Suggested Improvements

**[Improvement] Max-iteration `all_tool_executions.is_empty()` fallback doesn't re-check for a narrated call (~line 872-881).** In the `else` branch, when there are no tool executions to summarize, the code does `normalize_response(&response_text)` and only substitutes `EMPTY_RESPONSE_FALLBACK` when that is *empty*. It does **not** re-run `looks_like_narrated_tool_call` on `response_text` here, so a leaked pseudo-call sitting in the last iteration's text could in principle be returned verbatim through this specific path. In practice this is very hard to reach — a zero-tool-execution turn whose final text is a narrated call is caught earlier by the main-branch guard at line 522 and returned before the loop can fall through here — which is why I'm marking it Improvement, not Blocker. But since scenario B is precisely "raw pseudo-code must never reach the user," making this net symmetric with the others (`if normalized.is_empty() || looks_like_narrated_tool_call(&normalized) { EMPTY_RESPONSE_FALLBACK }`) would make the guarantee airtight and self-documenting rather than relying on an upstream branch. Principle: defense-in-depth nets should each independently hold their invariant.

**[Improvement] Tail-path narrated call with real executions synthesizes a "completed" summary (~line 940-965).** When `looks_like_narrated_tool_call(&normalized_tail)` fires and `all_tool_executions` is non-empty, the output is a bullet list like `• Execute Query completed`. If those executions actually *failed* (the exact scenario the issue reproduces: broken `execute_query` looping), "completed" is slightly optimistic phrasing for what was really a failed loop. This is pre-existing behavior for the empty-tail case (the narrated-call condition just newly routes into it), and the sibling test `empty_final_inference_after_tool_failure_still_surfaces_response` documents that a bullet summary is the accepted outcome, so this is not a regression. Worth a follow-up thought: the bullet synthesizer could distinguish error vs. success executions (`• Execute Query failed (2×)`) for a more honest diagnostic. Out of scope for this PR; noting for the durable-honesty goal.

#### Nitpicks

- **Nit:** The narrated-call replacement string at line 532 is identical to the anti-fabrication string at line 509 ("I'd like to help with that. Could you confirm..."). That's a reasonable UX choice (same user-facing intent), but the two are now duplicated literals. A shared `const CONFIRM_ACTION_FALLBACK` would keep them in sync and make the intent that "these are the same message" explicit. Optional.

- **Nit:** `looks_like_narrated_tool_call` lowercases the whole input on every call and iterates all 14 tool names with `match_indices`. This runs at most a couple of times per turn on short strings, so performance is a non-issue — but if it ever moved onto a hot path, note the allocation. No action needed now.

- **Nit:** The heuristic will fire on a *legitimate* explanatory answer that happens to write a tool name immediately followed by `(`, e.g. "You'd call `get_node(id)` to fetch it." The tests correctly cover the no-paren case as a non-match, and the issue explicitly accepts the narrow-over-general tradeoff ("narrow is safer/lower-risk to ship first"), so this is the intended, documented risk envelope — flagging only so it's on record as a known false-positive shape.

---

### Scope & acceptance-criteria validation (issue #1531)

1. **Close the empty-response gap** — Done. `EMPTY_RESPONSE_FALLBACK` added at the max-iteration `all_tool_executions.is_empty()` fallback (line 877) and as a final safety net (line 983) covering every tail branch, including the `Err`/empty-synthesis case. Matches the requested `"⚠️ I wasn't able to produce a response..."` string.
2. **Detect narrated-but-not-executed tool calls** — Done, via the narrow registry-driven `looks_like_narrated_tool_call` (registered tool name + `(`), wired into all three paths. Correctly implements the issue's recommended "narrow is safer" option and stays in sync with `Tool::ALL`.
3. **Regression coverage** — Done. Both scenarios have integration tests asserting the persisted message is never empty and never contains `execute_query(`; plus detector unit tests (positive/negative/prose/empty) and a constant-invariant test.

All three scope items satisfied. Out-of-scope items (root `execute_query` filter bug, Ollama tool-calling reliability, OpenAI-compat backend) are correctly left untouched.

### Standards compliance (CLAUDE.md)
- No lint suppression, no `#[allow(...)]`, no dead code. ✅
- Logging via `tracing` with structured fields (`session_id`, `model`, `iteration`, `response_preview`) — no raw stdout. ✅
- No backward-compat scaffolding; direct implementation. ✅
- Detector sourced from the tool registry rather than a hardcoded name list — good DRY/maintainability choice that self-updates as tools change. ✅

malibio (commented):
Initial pragmatic code review posted as a PR comment. Net positive — recommend merge after considering two minor (non-blocking) Improvement items. No blocking issues; using --comment rather than approve since two defense-in-depth refinements are worth a look. Full report in the PR comment above.

malibio:
## Review addressed — commit `9358a708`

Addressed both suggested improvements plus the DRY nit from the review:

**✅ Improvement 1 — Max-iteration narrated-call net now airtight.** The empty-executions fallback (`~line 875`) now re-checks `looks_like_narrated_tool_call` in addition to `is_empty()`, so a leaked pseudo-call can never return verbatim through that path. The scenario-B guarantee no longer depends on the upstream main-branch guard catching it first.

**✅ Improvement 2 — Honest failure summaries.** Extracted the two identical bullet-synthesis blocks into `summarize_executions()`, made error-aware: a tool whose calls all errored is reported `• property query failed (2×)` instead of an optimistic `completed` — the exact looping-`execute_query` scenario. A tool that succeeded at least once still reads `completed`.

**✅ Nit — DRY.** Extracted `CONFIRMATION_REQUEST` const so the anti-fabrication and narrated-call guards share one string and stay in sync.

**⏭️ Skipped (documented):** the heuristic firing on prose that writes a registered tool name immediately followed by `(` — this is the issue's explicitly-accepted narrow-over-general tradeoff. Left as-is.

**Tests:** +3 unit tests for `summarize_executions` (all-failed / success / partial-failure); scenario-A integration test now asserts the summary says `failed`, not `completed`. Full agent suite: **273 passed**; clippy clean.

## Re-Review Decision

**Decision: NO RE-REVIEW NEEDED.** The changes are low-risk: a dedup refactor (behavior-preserving except the intended failed-vs-completed labelling), a one-condition symmetry fix on an existing guard, and a string-constant extraction — all covered by new/updated tests. No new functionality or architectural change.

PR #1576, opened 2026-07-07
Cap the unbounded per-line buffer in the Ollama NDJSON readers

malibio:
Closes #1510

### DIFF
--- a/packages/agent/src/local_agent/mod.rs
+++ b/packages/agent/src/local_agent/mod.rs
@@ -4,6 +4,7 @@ pub mod inference;
 pub mod model_manager;
 pub mod ollama_inference;
 pub mod ollama_model_manager;
+pub mod ollama_ndjson;
 pub mod otlp_tracer;
 pub mod prompt_dump;
 pub mod prompt_templates;
--- a/packages/agent/src/local_agent/ollama_inference.rs
+++ b/packages/agent/src/local_agent/ollama_inference.rs
@@ -2,6 +2,7 @@ use crate::agent_types::{
     ChatInferenceEngine, ChatModelSpec, InferenceError, InferenceRequest, InferenceUsage,
     ModelFamily, StreamingChunk,
 };
+use crate::local_agent::ollama_ndjson::NdjsonLineBuffer;
 use async_trait::async_trait;
 use futures::StreamExt;
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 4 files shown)


malibio:
## Pragmatic Code Review — PR #1576 (closes #1510)

### Requirements Check

- ✅ **Inter-newline buffer is bounded with a clear error past the cap** — `NdjsonLineBuffer` checks the unterminated tail against `MAX_LINE_BYTES` (1 MiB) after draining complete lines and returns `NdjsonError::LineTooLong { limit }` with a descriptive `Display` message (`ollama_ndjson.rs:76`). Both callers propagate it as `InferenceError::Engine` / `ModelError::DownloadFailed`.
- ✅ **UTF-8 decoding happens per complete line, not per chunk** — bytes accumulate raw in a `Vec<u8>`; each complete line is decoded with strict `std::str::from_utf8` (`ollama_ndjson.rs:67`). The old `from_utf8_lossy(&chunk)` per-chunk pattern is removed from both readers.
- ✅ **Passes `quality:fix`; tests pass** — `cargo fmt --check` clean, `cargo clippy` reports no warnings for the module, and all 6 unit tests pass.

### Code Review Findings

🟢 **Sound memory bound.** The unbounded-growth path (accumulate forever between newlines) is closed: any chunk that pushes the *unterminated tail* past 1 MiB errors out. A newline drains and releases the buffer, so the count resets per line — correctly captured by the `newline_resets_overflow_count` test.

🟢 **Correct multi-byte UTF-8 handling.** A partial multi-byte sequence at the end of the buffer is never drained (no `\n`), so it stays buffered until the line completes, then decodes strictly. `decodes_multibyte_utf8_split_across_chunks` proves the fix for the exact bug the old lossy decode caused.

🟢 **`\r\n` handled.** `.trim()` strips a trailing `\r`, preserving the prior readers' behavior.

🟢 **Clean refactor / no duplication.** One shared buffer replaces two copies of the hand-rolled `String::find('\n')` + slice-reassign loop (DRY). No lint suppressions, no dead code, `tracing` used for the parse-failure warning. Error mapping in both callers is identical and correct (`ollama_inference.rs:246`, `ollama_model_manager.rs:230`).

🟡 **The cap does not bound a single *completed* line (`ollama_ndjson.rs:64-80`).** The overflow check runs only on the tail *after* draining. If one `bytes_stream()` chunk delivers, say, a 5 MiB line ending in `\n`, the whole 5 MiB is buffered and decoded into a `String` before any check — the cap never sees it. In practice this is transport-bounded (a single completed line can't be larger than the chunk that terminated it, and it's released immediately after draining), so it is not an unbounded-growth vector and does not regress the stated goal. Worth a one-line doc note that the cap bounds the *unterminated* tail, not an individual completed line — or, if strictly bounding per-line size is desired, check `pos + 1 > MAX_LINE_BYTES` inside the drain loop before allocating. Not a blocker.

🟢 **Nit — test coverage is good but could add two cases:** (1) a completed line that itself exceeds the cap, to pin down the intended semantics above; (2) an unterminated tail that grows across *multiple* `push` calls to just over the cap (current overflow test trips it in a single push). Optional.

🟢 **Performance.** `Vec::drain(..=pos)` is O(remaining) shifting per line, but NDJSON lines are small and the remaining buffer is typically tiny, so this is negligible. No concern.

### Recommendation

**APPROVE** (self-approval disallowed here — recording as a comment). The change is a clear net improvement: it closes the unbounded-buffer vector and fixes the split-multibyte corruption bug, with focused tests and clean style. The single 🟡 item is a doc-clarity / optional-hardening note, not a correctness or security defect.

malibio:
Addressed the 🟡 review note in `53f6b30f`: the cap now bounds **every** line, not just the unterminated tail — an oversized completed line is rejected inside the drain loop before it's allocated. Added the two suggested tests (completed line over cap; tail growing past cap across multiple pushes). ndjson module now has 8 tests, all passing; clippy `-D warnings` clean; full push gate green.

PR #1579, opened 2026-07-07
Unify the frontend store layer on Svelte 5 runes

malibio:
Closes #1513


malibio:
## Pragmatic Code Review — PR #1579

**Scope**: Full diff `origin/main...HEAD` (4 commits, 58 files, +1345 / −1628)
**Issue**: #1513 — Unify the frontend store layer on Svelte 5 runes (ADR-049)

---

### Summary

This is a clean, disciplined, **net-positive** refactor. Six legacy `writable()` stores (collections, schemas, layout, settings, navigation, status-bar) are migrated to rune-based class stores that match the existing `model-store.svelte.ts` precedent. The change is mechanical where it should be and thoughtful where it matters (persistence side-effects, cross-store getters, e2e rune globals). Old `.ts` files are deleted outright — no compat shims, no dual-paradigm bridge — exactly the greenfield posture the repo mandates. Reactivity is preserved throughout: every mutation reassigns `this.state` immutably, and no method mutates state in place. Verification (unit 3940 pass, svelte-check 0 errors, e2e 24 pass, pre-push gate, live smoke) is thorough and I independently re-ran the store suites (260 pass) and confirmed the key invariants below.

**Recommendation: APPROVE.** No blocking issues. A few optional nits only.

---

### Requirements Check

- ✅ **Legacy `writable()` stores migrated to rune stores** — All 6 stores are now `$state`-backed classes exported as singletons; `derived()` exports became computed getters (`collectionsTree`, `selectedCollection`, `builtInSchemas`, etc.). Old `.ts` files deleted; all production imports use the `.svelte` suffix.
- ✅ **Imperative `get(store)` reads eliminated** — No `get()` from `svelte/store` remains in the migrated surface. (The one residual `get` import in `src/lib/design/theme.ts` is an out-of-scope local writable store, not part of this issue.) The daemon-readiness e2e was converted from `get(builtInSchemas)` to `schemasStore.builtInSchemas`.
- ✅ **Passes `quality:fix`; tests pass** — No lint suppressions, no `any`, no `console.*` in the new store files; store suites pass locally; full verification green per the PR description.

---

### Verified Invariants (review focus areas)

- **Reactivity of getters**: `$derived(store.getter)` consumers (navigation-sidebar) re-track correctly because getters re-read `$state` on every access. Confirmed for `collectionsTree`, `builtInSchemas`, `customSchemas`, `selectedCollection`, `selectedCollectionMembers`.
- **Cross-store getters stay reactive**: `collectionsState.selectedCollection` / `selectedCollectionMembers` read `collectionsData.state.collections|members`. Since `collectionsData.state` is `$state`, a `$derived` wrapping these getters re-runs when the data store changes. Sound.
- **Immutable reassignment**: All 13/15/8/4/1 state writes across the stores use `this.state = { ...this.state, ... }`. Grep for in-place mutation (`push/splice/set/delete`/direct prop assign on `this.state`) returns **zero** hits.
- **`#persist()` fidelity vs old `.subscribe()`**: The old navigation store persisted on *every* `tabState.set()`; the new store calls `#persist()` at the tail of every mutating method (createPane, closePane, setActivePane, resizePane, setActiveTab, closeTab, addTab, updateTabContent, reorderTab, moveTabBetweenPanes, clearAllTabs). The 500ms debounce, the `#isInitialized` guard, and the `NodeExpansionCoordinator.getExpandedNodeIds` enrichment are all preserved identically. Layout `#persist()` likewise mirrors the old subscribe. `setActiveNavItem` correctly does *not* persist — matching old behavior where `navigationItems` was a separate, unpersisted writable.
- **`activePane` not-in-persisted-shape**: Both old and new hardcode `activePane: 'today'` on load; `setActivePane` triggering a persist of the other fields is unchanged behavior, not a regression.
- **Delegator completeness**: Every previously free-exported mutation function is still exported as a thin delegator. Removed exports (`tabState`, `layoutState`, `navigationItems`, `appSettings`, `settingsInitialCategory`, `statusBarVisible`, `builtInSchemas`/`customSchemas` as derived) have zero dangling references in production — the only remaining textual matches are code comments and correctly-migrated field accessors (`layoutStore.navigationItems`, `settingsStore.appSettings`).

---

### Findings

#### 🟢 Suggestions / Nits (all optional)

**1. `pruneEmptyCollections` mutates its input in place**
📁 `src/lib/stores/collections.svelte.ts:300-306`

`item.children = keptChildren` mutates the `CollectionItem` objects passed in. This is **safe today** because the items are freshly built from a local `itemMap` inside `buildCollectionsTree` on every getter access, and never touch the reactive `$state` (which holds raw `CollectionInfo[]`). But a function named `prune…` returning a new array while also mutating its argument is a hidden side-effect that could bite a future caller who passes shared/reactive data.
*Nit*: either rename to signal the in-place edit, or build children immutably. Principle: least-surprise / referential transparency.

**2. `setup-svelte-mocks.ts` for e2e is the right call — worth a one-line note**
📁 `vitest.e2e.config.ts:37`

This is **not** a hack: it reuses the exact same setup file the main `vitest.config.ts` already loads. The `$state` mock is a non-reactive passthrough, which is fine for the e2e tests here because they either import a store only for its module-load side-effect (`onDaemonReconnect` registration) or **poll** `schemasStore.builtInSchemas` in a loop rather than relying on reactive re-render. Good. The config comment already explains the "why" — no change needed; flagging only to confirm the approach is sound.

**3. `schemasData = { loadSchemas }` is a slightly redundant surface**
📁 `src/lib/stores/schemas.svelte.ts:57`

Now that `schemasStore.loadSchemas()` exists, the standalone `schemasData` namespace is somewhat duplicative. It's justified — it preserves the `collectionsData`/`schemasData` symmetry that app-shell, navigation-sidebar, and collection-refresh consume, and it's the object the daemon-reconnect wiring targets — so keeping it is reasonable. Optional future cleanup, not for this PR.

---

### Test Quality Assessment

The removed `layout.test.ts` tests were the three **"Store Reactivity"** cases (`notifies subscribers when layoutState changes`, `…when navigationItems changes`, `multiple subscribers receive updates`) plus a `.subscribe()`-based "update navigation items" case. These asserted the Svelte-store `.subscribe()` contract, which **no longer exists** on a rune class — they tested the removed paradigm, not behavior. Equivalent behavioral coverage (state actually changes on mutation; `setActiveNavItem` marks a single item active; persistence fires post-init) is **retained** via direct `.state` assertions and the rewritten persistence tests. No real coverage was lost. The unit-count delta of −4 is fully accounted for.

---

### ADR-049 Compliance

✅ No state-to-state `$effect`. ✅ Persistence is an explicit `#persist()` method call, not an effect. ✅ Derivation via getters, not `derived()`. ✅ Immutable `$state` reassignment throughout. ✅ Matches the `model-store.svelte.ts` precedent.

### Project Standards

✅ No `console.*` (uses `createLogger`). ✅ No `any`. ✅ No lint suppression / `@ts-expect-error` / deprecation / dead-code allow. ✅ No backward-compat shims — old files deleted.

malibio (commented):
Review recommendation: **APPROVE** (posted as a comment because GitHub blocks self-approval on your own PR). Clean, net-positive ADR-049 rune migration — no blocking issues, three optional nits. Full report in the review comment above.

malibio:
## Review feedback addressed

The review returned **APPROVE** with 3 optional 🟢 suggestions and no Critical/Important findings.

**✅ Addressed (1):**
- **`pruneEmptyCollections` mutating its input in place** (`collections.svelte.ts`) — rewrote as a pure function that builds new `CollectionItem` objects instead of reassigning `item.children`. This getter runs on every reactive read, so a side-effect-free contract is safer. Behavior unchanged; all 40 collections tests pass. (commit cf862b11)

**⏭️ Skipped (2), with reasoning:**
- **e2e `setupFiles` fix** — reviewer confirmed it's sound (reuses the exact setup file the main vitest config already loads), explicitly *not a hack, no change needed*.
- **`schemasData = { loadSchemas }` wrapper** — kept for symmetry with `collectionsData` and the daemon-reconnect wiring; reviewer marked it optional future cleanup, *not for this PR*.

Quality: eslint + svelte-check clean (0 errors / 1729 files). Pre-push gate (test:all + cargo build + e2e + Rust) passed.

PR #1582, opened 2026-07-08
Add pro_join_collection Tauri command (open-collection self-join)

mstomar125:
Completes the desktop half of the membership command surface: wires the daemon's `JoinCollection` gRPC — the open-collection complement to `RequestJoin` — through to a Tauri command. The daemon half shipped in nodespace-sync#287 (`JoinCollection` rpc on `CloudSyncService`); this adds the matching client binding + command in core.

## Changes
- **`proto/nodespace_pro.proto`** (vendored Pro contract): add `rpc JoinCollection (JoinCollectionRequest) returns (MembershipAck)` + `message JoinCollectionRequest { string collection_id = 1; }`, mirroring the daemon proto.
- **`commands/pro_sync.rs`**: `pro_join_collection(app, collection_id)` → forwards to `client.join_collection(...)`. Mirrors `pro_request_join` (returns `()` since the ack is just `"ok"`). The open-vs-restricted gate is enforced server-side (the cloud RPC rejects a restricted collection with 42501), so no client-side check.
- **`lib.rs`**: register the command in the invoke handler.

## Notes
- No UI yet — this parallels the other `pro_*` membership commands (`pro_request_join`, `pro_set_member`, …), which exist for tests and a future UI.
- Community build unaffected: this is a Pro-only command on `nodespace-app`; the `pb` bindings regenerate from the vendored proto at build time.

## Verification
`cargo check -p nodespace-app` clean (regenerates the `pb` bindings from the updated proto — `join_collection` method + `JoinCollectionRequest` type resolve). Backend-only change; no frontend surface touched.

### DIFF
--- a/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
+++ b/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
@@ -50,6 +50,10 @@ service CloudSyncService {
   rpc CreateInvite (CreateInviteRequest) returns (CreateInviteResponse);
   rpc AcceptInvite (AcceptInviteRequest) returns (MembershipAck);
   rpc RequestJoin (RequestJoinRequest) returns (RequestJoinResponse);
+  // Open-collection self-join — the complement to RequestJoin (which is for
+  // restricted collections). The cloud RPC rejects a restricted collection with
+  // 42501, so the open-vs-restricted gate is enforced server-side.
+  rpc JoinCollection (JoinCollectionRequest) returns (MembershipAck);
   rpc ApproveRequest (ApproveRequestRequest) returns (MembershipAck);
 
   // --- Invite / request listings + revocation (S2, #239). These three cloud
@@ -188,6 +192,9 @@ message RequestJoinRequest {
 }
 message RequestJoinResponse {
   string request_id = 1;
+}
+message JoinCollectionRequest {
+  string collection_id = 1;
 }
 message ApproveRequestRequest {
   string request_id = 1;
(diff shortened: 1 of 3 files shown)


mstomar125:
**APPROVE**

Verified against the diff at `c456d8a`:

- **Proto matches the daemon exactly.** The vendored `nodespace_pro.proto` blob goes `0c0ebe56..91c65c95` — identical hashes to nodespace-sync#288's daemon proto — with the same `rpc JoinCollection (JoinCollectionRequest) returns (MembershipAck)` (positioned after `RequestJoin`) and `message JoinCollectionRequest { string collection_id = 1; }` (after `RequestJoinResponse`). Same field number, same types → the regenerated `pb` client method and the daemon's server method share one wire contract; no drift.
- **Command is a faithful mirror.** `pro_join_collection(app, collection_id)` acquires the same `membership_client`, calls `client.join_collection(JoinCollectionRequest { collection_id })`, maps the error, and returns `()` — correct, since the daemon's `JoinCollection` returns a static `MembershipAck{detail:"ok"}` with no payload (unlike `pro_request_join`, which returns a `request_id`). Open-vs-restricted stays server-side (cloud rejects restricted with 42501); no redundant client gate. `JoinCollectionRequest` added to the `pb::{…}` import.
- **Registered + gated correctly.** Added to the `invoke_handler!` list in `lib.rs` (right after `pro_request_join`), so it's actually invokable. Pro-only command on `nodespace-app`, `pb` regenerates from the vendored proto at build time → **community build unaffected**.
- **CI green** — `test:all` + `test:e2e (headless daemon)` both SUCCESS.

This closes the JoinCollection command surface (daemon #288 + this) as designed — no UI yet, parallel to the other `pro_*` membership commands. Clean.

PR #1583, opened 2026-07-08
Add open-collection self-join UI to the invitations modal

mstomar125:
Adds the first UI surface for `pro_join_collection` (merged in #1582): a direct **Join** for open collections, the complement to the existing "request to join" flow for restricted ones. This is the JoinCollection UI follow-up noted when the command landed.

## Changes
- **`membership-service.ts`** — `joinCollection(collectionId)` over the `pro_join_collection` command (void; the daemon returns a static ack).
- **`membership.svelte.ts` store** — `joinCollection()`, Pro-gated via `requirePro()` exactly like `requestJoin` (a no-op path throws in community mode rather than faking success).
- **`invitations-inbox.svelte`** — the "Request to join a collection" section becomes **"Join a collection"**: one collection-id input, a **Join** button (direct self-join, open collections) next to **Request** (restricted). A hint explains the difference. The open-vs-restricted gate stays server-side (the cloud RPC rejects a restricted collection), so a wrong choice surfaces the server's error instead of a client-side guess.

## Community build
Unaffected — the invitations modal is a Pro onboarding surface, and the store method is `requirePro()`-gated (inert without a `ProClient`), covered by the community-mode test.

## Tests
`bun run test` on the two affected suites — **16 pass** (store: `joinCollection` forwards to the service when Pro / throws in community mode; modal: the Join button calls `joinCollection`, not `requestJoin`, and confirms). `bun run quality:fix` clean (eslint/prettier + clippy).

## Scope
Discovering joinable collections (search/browse) remains the documented Phase-2 follow-up — the collection id is still shared out of band. This PR only adds the direct-join action alongside the existing request action.

### DIFF
--- a/packages/desktop-app/src/lib/components/collaboration/invitations-inbox.svelte
+++ b/packages/desktop-app/src/lib/components/collaboration/invitations-inbox.svelte
@@ -48,6 +48,7 @@
 
 	let joinId = $state('');
 	let requesting = $state(false);
+	let joining = $state(false);
 	let requestMsg = $state<string | null>(null);
 	let requestErr = $state<string | null>(null);
 
@@ -90,6 +91,24 @@
 			requesting = false;
 		}
 	}
+
+	async function joinCollection() {
+		const id = joinId.trim();
+		if (!id) return;
+		joining = true;
+		requestMsg = null;
+		requestErr = null;
+		try {
+			await membership.joinCollection(id);
+			requestMsg = 'Joined — it will appear in your sidebar as it syncs.';
+			joinId = '';
+		} catch (e) {
+			log.warn('joinCollection failed', { error: e });
+			requestErr = friendly(e);
+		} finally {
+			joining = false;
+		}
+	}
 </script>
 
 {#if open}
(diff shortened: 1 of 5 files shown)


mstomar125:
**APPROVE** (code); one gate before merge below.

Verified against the diff at `d52b152`:

- **Wiring correct.** `membership-service.joinCollection` → `invoke<void>('pro_join_collection', { collectionId })` — matches the command registered in #1582, and `{ collectionId }`→`collection_id` follows the exact camelCase↔snake_case pattern of the working sibling `pro_request_join`. Returns void (daemon's static ack).
- **Community build stays inert.** The store's `joinCollection()` calls `this.requirePro()` *before* touching the service — identical to `requestJoin` — so community mode throws rather than faking success. Covered by the community-mode test.
- **Server-side gate surfaced correctly.** The Svelte `joinCollection()` `await`s the store call and, on rejection (e.g. a restricted collection's `42501`), sets `requestErr = friendly(e)` and logs — so a wrong open/restricted choice shows the server's real error instead of a client-side guess. Success sets a confirm message + clears the input.
- **UI robustness.** `joining` `$state`, both Join and Request buttons disabled while `requesting || joining || !joinId.trim()` (no double-submit / no empty submit), `finally` resets the loading flag.
- **Tests are the right ones** — modal test asserts clicking **Join** calls `joinCollection` and **not** `requestJoin` (the key distinction) + shows the confirmation; store test covers Pro-forward vs community-throw.
- Scope note (discovery/search deferred to Phase-2, id shared out-of-band) is reasonable and documented.

**Merge gate:** `test:all (frontend + skill + Rust)` is currently **IN_PROGRESS** (test:e2e already SUCCESS) — hold the merge until it's green, since this touches the Svelte store/component suites it's meant to exercise. No code changes requested.

Issue #293, opened 2025-10-19
Implement OrderedListNode (Auto-Numbered Lists)


malibio:
## Overview

Implement **OrderedListNode** to support auto-numbered ordered lists using markdown \`1. \` syntax. This completes basic markdown document support alongside existing text, header, quote-block, code-block, and task nodes.

## Design Reference

**IMPORTANT**: Follow the established design specifications:
- 🎨 **[Component Design](docs/design-system/components.html)** - See "Ordered List Nodes" section for complete UI design
- 🎨 **[Icon System](docs/design-system/patterns/foundations/icon-systems.html)** - See "OrderedListNode Icon System" section for icon specifications

## Reference Documentation

**IMPORTANT**: Before implementing, read these comprehensive guides:
- 📚 **[How to Add a New Node Type](docs/architecture/development/how-to-add-new-node-type.md)** - Complete implementation checklist
- 📚 **[Pattern Detection and Content Templates](docs/architecture/development/pattern-detection-and-templates.md)** - Pattern detection & template system guide
- 📚 **[Node Types Architecture](docs/architecture/components/node-types.md)** - Core node type architecture

## Scope

**In Scope:**
- Auto-numbered ordered lists with \`1. \` prefix pattern
- Pattern detection for auto-conversion
- Slash command \`/ordered-list\` or \`/ol\`
- CSS-based auto-numbering in view mode (1, 2, 3...)
- Enter key creates new ordered list sibling
- Backspace at start converts to text node

**Out of Scope (Future Enhancement):**
- Nested/hierarchical ordered lists (\`canHaveChildren: false\`)
- Starting at arbitrary numbers (e.g., \`5. \`)
- Mixed numbering formats (e.g., \`a.\`, \`i.\`, \`I.\`)

## Markdown Specification

**CommonMark Standard:**
- Pattern: \`1. \` (number, period, space)
- Auto-numbering: All items use \`1. \` in source, renderer provides sequential numbers
- Reference: [CommonMark Lists Tutorial](https://commonmark.org/help/tutorial/06-lists.html)

---

#### Step 1.1: Update ALLOWED_NODE_TYPES

**File:** \`packages/desktop-app/src-tauri/src/constants.rs\`

**⚠️ CRITICAL**: Forgetting this step causes HTTP 400 errors when creating nodes!

\`\`\`rust
/// Allowed node types in the system
///
/// Currently supported node types:
/// - \"text\": Plain text nodes
[...]

Issue #72, opened 2025-08-15
Phase 1.3: Implement EventBus System


malibio:
## Parent Issue
Part of Epic #69: Text Editor Architecture Refactor - Phase 1

## Summary
Implement a centralized event system to decouple components and enable clean communication between services and UI.

## Current Problem
Components are tightly coupled through prop drilling and direct function calls, making the system hard to extend and test.

## Proposed Solution
Create an EventBus system that enables:

### Core Functionality
- Type-safe event emission and subscription
- Event filtering and routing
- Async event handling
- Event debugging and logging

### EventBus Interface
```typescript
interface EventBus {
  // Content events
  emit(event: 'content:changed', data: ContentChangedEvent): void;
  emit(event: 'node:created', data: NodeCreatedEvent): void;
  emit(event: 'node:deleted', data: NodeDeletedEvent): void;
  emit(event: 'node:combined', data: NodeCombinedEvent): void;
  
  // Selection events
  emit(event: 'selection:changed', data: SelectionChangedEvent): void;
  emit(event: 'cursor:moved', data: CursorMovedEvent): void;
  
  // Hierarchy events
  emit(event: 'node:indented', data: NodeIndentedEvent): void;
  emit(event: 'node:outdented', data: NodeOutdentedEvent): void;
  
  // Subscription methods
  on<T>(event: string, handler: (data: T) => void): () => void;
  off(event: string, handler: Function): void;
  once<T>(event: string, handler: (data: T) => void): void;
}
```

## Implementation Tasks
- [ ] Create EventBus interface and implementation
- [ ] Define TypeScript event schemas
- [ ] Implement subscription management
- [ ] Add event filtering and routing
- [ ] Create event debugging utilities
- [ ] Add comprehensive unit tests
- [ ] Update services to emit events
- [ ] Update components to listen to events

## Acceptance Criteria
- [ ] EventBus supports type-safe event emission
- [ ] Components can subscribe/unsubscribe to events
- [ ] Services emit appropriate events for all operations
- [ ] Event system handles async operations correctly
- [ ] Event debugging shows clear event flow
- [ ] No memory leaks from event subscriptions

## Dependencies
Should be implemented after Phase 1.1 and 1.2 so services can emit events

## Estimated Effort
2-3 days

## Definition of Done
- [ ] EventBus implemented with full test coverage
- [ ] All services emit appropriate events
- [ ] Components listen to relevant events
- [ ] Event flow documented
- [ ] Code review completed

[...]

PR #1584, opened 2026-07-08
Collection discovery (browse & join) in the invitations modal

mstomar125:
## Summary

Phase 2, Slice 3 (core) of **collection discovery (browse & join)**. Wires the discovery RPC end-to-end into the onboarding modal.

Before this, the invitations modal could only redeem a share code or join a collection whose **id was pasted in by hand** — there was no way to see which collections you could actually join. This adds a real discovery list: the collections you're *not* a member of yet, each with a per-row **Join** (open) / **Request** (restricted).

## Changes

- **proto**: `ListJoinableCollections` rpc + `JoinableCollection { id, name, restricted }` (mirrors the sync daemon proto).
- **pro_sync.rs**: `pro_list_joinable_collections` Tauri command (registered in `lib.rs`).
- **membership-service.ts**: `listJoinable()` + `JoinableCollection` type.
- **membership store**: `joinable` list + `loadJoinable()`; joining an open collection drops it from the loaded list; `reset()` clears discovery state.
- **invitations-inbox.svelte**: replaces the manual collection-id entry with a browse list (loaded on open), per-row Join/Request with in-flight + status states, a Refresh control, and loading / empty / error states. Open vs. restricted shown as a tag.

## Testing

- `cargo check -p nodespace-app` — clean
- `bun run quality:fix` (eslint + svelte-check) — **0 errors / 0 warnings / 0 problems** across 1729 files
- `bun run test` (service + store + component): **33 passed** (11 service, 13 store, 9 component), incl. new discovery tests: row mapping, load/error/inert, drop-on-join, request-keeps, and per-row Join/Request UI + empty state.

## Notes

- The daemon forwards the caller's JWT, so the cloud RPC filters to what the caller can see and excludes their own memberships **server-side**; the open-vs-restricted gate is enforced server-side too (restricted → request → admin approval).
- Pairs with the sync daemon RPC (nodespace-sync #293) and the cloud `list_joinable_collections` RPC (nodespace-cloud #72).
- **Community build unchanged** — the Pro-gated store is inert without a Pro client, so the modal shows no discovery list in community mode.

### DIFF
--- a/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
+++ b/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
@@ -65,6 +65,13 @@ service CloudSyncService {
   rpc ListRequests (ListRequestsRequest) returns (ListRequestsResponse);
   rpc RevokeInvite (RevokeInviteRequest) returns (MembershipAck);
 
+  // Collection discovery (browse & join). Lists the collections the signed-in user
+  // could join but isn't a member of yet (open + restricted), so the client can
+  // offer Join / Request without the collection id being shared out of band. Thin
+  // adapter over the cloud `list_joinable_collections` RPC (JWT-forwarded; the RPC
+  // filters to what the caller can see and excludes their memberships server-side).
+  rpc ListJoinableCollections (ListJoinableCollectionsRequest) returns (ListJoinableCollectionsResponse);
+
   // Caller identity (#238/#239). Returns THIS device's bound PersonNode id (+
   // the signed-in email) so the UI can tell which roster row is "me" and gate
   // admin controls on the caller's own per-collection role. A dedicated one-shot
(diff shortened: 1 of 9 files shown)


mstomar125:
**APPROVE**

Verified against the diff at `9ac6c81` — this completes Phase-2 discovery end-to-end (cloud#72 RPC → sync#293 daemon → this command+UI):

- **Proto + command** — vendored `JoinableCollection { id=1, name=2, restricted=3 }` matches the sync#293 daemon proto (wire-compatible); `pro_list_joinable_collections` maps rows to `JoinableCollectionDto` passing `restricted` straight through (the safe-side default already lives in the daemon), registered in `lib.rs`.
- **Community build stays inert** — the passive `loadJoinable()` early-returns on `if (!this.isPro)` (no throw), and the modal gates the call on `open && membership.isPro && !loadedOnce`. Correct distinction from the mutating `joinCollection`/`requestJoin`, which keep the `requirePro()` throw. So community mode simply shows no list.
- **Per-row actions correct** — open → `joinCollection(id)` then the store drops that id from `joinable` (guarded on `joinableLoaded`); restricted → `requestJoin(id)` and the row is intentionally **kept** (a pending request isn't membership). The `restricted` flag drives both the "Restricted" tag and the Join-vs-Request choice; `actingId` disables only the in-flight row and guards against concurrent/double submits (`if (actingId) return`).
- **State handling** — distinct loading / error (`friendly()`) / empty / list states; `joinableLoaded` separates "not fetched" from "fetched-empty"; `reset()` clears discovery state on sign-out; a Refresh control re-loads.
- **CI green** — `test:all` + `test:e2e` both SUCCESS; 33 tests including row-mapping, load/error/inert, drop-on-join, request-keeps, and the per-row Join/Request + empty-state UI.

The open-vs-restricted flag stays advisory (server enforces via `join_collection`/`request_join` + RLS), consistent with #72/#293. Clean — no changes requested.

PR #1585, opened 2026-07-08
dev:tauri extracts an invalidly-signed nodespaced sidecar → CODESIGNING SIGKILL crash-loop → LLM turn hangs

malibio:
Closes #1580

### DIFF
--- a/packages/desktop-app/src-tauri/src/daemon_setup.rs
+++ b/packages/desktop-app/src-tauri/src/daemon_setup.rs
@@ -443,8 +443,13 @@ pub async fn wait_for_daemon(socket_path: &Path, max_wait: Duration) -> DaemonSt
 }
 
 /// Extract a sidecar binary from the Tauri bundle to `~/.nodespace/bin/`,
-/// but only if the destination is missing or has a different file size than
-/// the bundled source. Returns true if the binary was updated.
+/// but only if the destination is missing, has a different file size than
+/// the bundled source, or fails code-signature verification. Returns true
+/// if the binary was (re-)extracted.
+///
+/// Writes to a temp path in the same directory, re-seals the ad-hoc signature
+/// (macOS only), then renames into place — so a concurrently-launched launchd
+/// `KeepAlive` daemon can never `mmap` a partially-written or unsigned image.
 async fn extract_sidecar_if_changed(app: &AppHandle, name: &str, bin_dir: &Path) -> Result<bool> {
     let src = resolve_sidecar_path(app, name)?;
     let dest = bin_dir.join(name);
@@ -455,13 +460,19 @@ async fn extract_sidecar_if_changed(app: &AppHandle, name: &str, bin_dir: &Path)
         .len();
 
     if let Ok(dest_meta) = tokio::fs::metadata(&dest).await {
-        if dest_meta.len() == src_size {
+        if dest_meta.len() == src_size && verify_signature(&dest) {
             tracing::debug!(
-                "{} is up-to-date (size={}), skipping extraction",
+                "{} is up-to-date (size={}) and validly signed, skipping extraction",
                 name,
                 src_size
             );
             return Ok(false);
+        }
+        if dest_meta.len() == src_size {
+            tracing::warn!(
+                "{} matches bundled size but has an invalid code signature — re-extracting",
+                name
+            );
         }
     }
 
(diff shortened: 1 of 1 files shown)


malibio:
## Code Review Summary

This PR correctly fixes the root cause of issue #1580: the extracted daemon sidecar losing its ad-hoc code signature, causing macOS to `SIGKILL` it with `CODESIGNING`/`Invalid Page` and crash-loop under launchd `KeepAlive`. The fix is well-targeted, minimal in surface area (single file, +117/-7), and addresses both defects called out in the issue: no re-signing after copy, and a size-only staleness check that could never self-heal a corrupted install. The temp-file + rename pattern also closes the concurrency window where a `KeepAlive` relaunch could `mmap` a partially-written binary. This is a clear net positive and I recommend **approval**, with a couple of low-severity items noted below (not blocking).

Verified locally: `cargo build`/`clippy -p nodespace-app --lib` clean, and the two new macOS tests (`resign_then_verify_succeeds`, `tampering_after_sign_invalidates_signature`) pass, correctly reproducing the sign → tamper → invalidate cycle described in the bug report.

## Findings

### Critical Issues
None.

### Suggested Improvements

- **`daemon_setup.rs:486-497` — temp file is not cleaned up if `set_executable`/`resign_binary` fails.** If `resign_binary` fails (e.g. `codesign` binary missing/misconfigured, or the underlying `Command::status()` I/O errors), `extract_sidecar_if_changed` returns `Err` and the `<name>.tmp-<pid>` file is left behind in `bin_dir` permanently — nothing ever revisits it. Worst case is cosmetic (a stray few-hundred-MB file per failed attempt in `~/.nodespace/bin/`), but over repeated failed dev-loop restarts (e.g. a broken `codesign` toolchain) these will accumulate silently with no visibility. Consider a `defer`-style cleanup, e.g. wrap the fallible steps and `let _ = tokio::fs::remove_file(&tmp_dest).await;` on any early return, or at minimum log a `tracing::warn!` when this error path is hit so it's discoverable. Low severity since this only triggers on `codesign` itself failing, which is already an abnormal environment.

- **`kill_stale_daemon_sync` (line ~106) retains a pure size-only staleness check, independently of the fix in this PR.** This function runs *before* `ensure_daemon_running`/`extract_sidecar_if_changed`, deciding whether to SIGKILL a currently-running daemon so the frontend doesn't connect to a stale one. It still only compares `bundled_size == installed_size`. Concretely: if `~/.nodespace/bin/nodespaced` is corrupted-but-same-size (the exact scenario this issue is about), `kill_stale_daemon_sync` will *not* kill it, and the app may briefly connect to (or the OS may attempt to relaunch) the invalidly-signed binary before `ensure_daemon_running`'s corrected check runs later in startup and re-extracts it. I believe this is **not a correctness gap** for the steady-state fix — `extract_sidecar_if_changed` will still self-heal moments later — but there's a window during app startup where the old corrupted binary could still be launched/crash once before the fix kicks in. Worth a one-line comment on `kill_stale_daemon_sync` noting that its size-only check is intentionally a cheap pre-check and the authoritative self-heal lives in `extract_sidecar_if_changed`, so a future reader doesn't "fix" one without the other or assume this function is also signature-aware.

- **No integration test exercises `extract_sidecar_if_changed` itself.** The two new tests validate `resign_binary`/`verify_signature` in isolation via a stand-in Mach-O binary (`/bin/cat`), which is a reasonable substitute given the function is private and needs a live `AppHandle`. But nothing exercises the full copy → temp-file → resign → rename → staleness-recheck sequence, e.g. "given a same-size-but-corrupted dest, does the function actually re-extract and end up validly signed." Given this file had zero tests before this PR, the bar met here is already an improvement, but a follow-up integration test (extracting `resolve_sidecar_path`/`bin_dir` seams to allow constructing a fake `AppHandle`-free version of this function, or an `#[cfg(test)]` helper that takes `src`/`dest` directly instead of `app`) would give confidence beyond "the primitives work" to "the orchestration is correct." Not blocking — flagging for a fast-follow rather than this PR.

### Nitpicks

- **Nit:** `daemon_setup.rs:486` — `format!("{}.tmp-{}", name, std::process::id())` uses only the PID for uniqueness. Two different app launches will always have distinct PIDs so this can't collide in practice, but consider also appending a monotonic counter or timestamp if this function is ever called twice concurrently for the *same* sidecar name within one process (it currently isn't — `ensure_daemon_running` calls it sequentially with `.await` for the daemon binary then the CLI binary, both with different `name`s). No action needed today; just noting the uniqueness invariant is "one call per (name, pid) at a time," which happens to hold.

- **Nit:** `resign_binary`/`verify_signature` shell out to `codesign` with `path` as the only argument, and `path` is always a locally-constructed path under `bin_dir` (never derived from user/network input) — no injection risk. Confirmed via `Command::arg(path)` (not shell interpolation), so even a path containing spaces or shell metacharacters would be passed as a single argv element safely.

## Acceptance Criteria (issue #1580)

- [x] Extracted `~/.nodespace/bin/nodespaced` passes `codesign --verify --strict` after every extraction — `resign_binary` runs unconditionally on every extraction before the rename.
- [x] An already-corrupted install self-heals instead of being skipped on size match — `verify_signature(&dest)` is now required in addition to the size check.
- [x] Daemon no longer SIGKILLs with `CODESIGNING` during startup or the sqlite-vec path — addressed at the source (the binary is now validly signed on disk); this is a static-analysis-level fix, consistent with the reproduction the issue documented (a plain valid-signature `cp` was already known to be stable).
- [x] AI-chat turns complete end-to-end without a mid-turn daemon restart — follows from the above; not separately re-verified end-to-end in this review (would require a live `dev:tauri` run), but the fix directly targets the documented root cause.
- [x] Same guarantees applied to the `nodespace` CLI sidecar — confirmed both calls in `ensure_daemon_running` (lines 238-239) route through the same corrected `extract_sidecar_if_changed`.

All five acceptance criteria are met by the diff.

malibio (commented):
Review posted above. Recommendation: APPROVE (net positive, no blocking issues) — GitHub disallows self-approval on this PR since it's authored under the same account, so recording as a comment-only review instead of a formal approval.

malibio:
## Addressed review feedback

✅ **Addressed** (2/3 recommendations):
- Temp file (`<name>.tmp-<pid>`) is now removed if `set_executable`/`resign_binary` fails partway, instead of being left behind silently in `~/.nodespace/bin/`.
- Added a clarifying comment on `kill_stale_daemon_sync` explaining it's an intentional cheap size-only pre-check, not a signature-aware gate — the real self-heal for a corrupted same-size install happens moments later in `extract_sidecar_if_changed`.

⏭️ **Skipped** (1/3):
- Full end-to-end integration test for `extract_sidecar_if_changed` itself. The reviewer explicitly flagged this as a fast-follow rather than blocking — the function is private and needs a live `AppHandle`, and this file had zero tests before this PR. The new `resign_binary`/`verify_signature` unit tests already cover the actual bug mechanism (sign → tamper → re-verify fails).

Verified: `cargo clippy -p nodespace-app --all-targets` clean, `cargo test -p nodespace-app --lib daemon_setup` (2/2 new tests pass), `bun run test` (150 files / 3950 tests, matches baseline), full pre-push gate (test:all + cargo build + e2e) passed on push.

## Re-Review Decision

**Rationale**: Both changes are small, low-risk, and mechanical — an error-path cleanup (`remove_file` on failure) and a doc-comment addition. Neither changes the core logic already reviewed (re-sign + verify-before-skip + atomic rename), introduces new architecture, or touches any of the concurrency/atomicity guarantees the original review scrutinized. No new surface area for bugs.

PR #1589, opened 2026-07-09
ADR-053 Stage 4: DatabaseManager lazy open + service-assembly extraction

mstomar125:
## ADR-053 Stage 4 — DatabaseManager lazy open + service-assembly extraction

Part of the ADR-053 epic ("One Daemon, Multiple Local Databases", #1532). Builds on the Stage 1–3 scaffold (registry skeleton + `build_shared_services`/`build_database_services` split).

This stage makes the daemon's `DatabaseManager` a working component: it can register a default, resolve a request's routing header, and **lazily open and cache** a database's service set. It stays deliberately off the live request path — wiring per-request routing into the serve loops changes the cross-repo `BaseServices`/`build_base_router` contract (ADR-043 lock-step with `nodespaced-pro`) and lands in the following, sync-coordinated stage.

### What changed

- **Hoist service assembly into the library** (`services/assembly.rs`). Moves `SharedServices`, `DatabaseServices`, `build_shared_services`, `build_database_services` (and their private helpers) out of `main.rs` into a library module so the `DatabaseManager` can build per-database service sets. `main.rs` shrinks to boot/serve wiring only. Pure relocation — no behavior change.
- **Split shared services into a router-owned part and a build context.** `SharedServices { settings, context }`, where the cloneable `SharedContext { pty_manager, model, has_model }` carries exactly the process-global inputs a per-database build needs. `build_database_services` now takes `&SharedContext`.
- **DatabaseManager routing resolution.** `ensure_default_registered(name, path)` (idempotent first-boot default registration) and `resolve_database_id(header)` (`x-ns-database-id` → id; absent → default; unknown → error).
- **DatabaseManager lazy open + cache.** `get_or_open(id)` resolves the registry entry's path, assembles its `DatabaseServices` from the `SharedContext` outside the open lock, then caches and returns a shared `Arc<DatabaseServices>` (double-checked so concurrent first-opens of the same id converge on one handle). Replaces the prior "not yet wired" placeholder.

### Behavior

Single-default-database behavior is unchanged — the serve loops still build the default database directly. Nothing here is on the live request path yet, so community and Pro runtime behavior is identical to before.

### Tests

`cargo test -p nodespace-daemon` green: daemon lib **46 passed** (adds a `get_or_open` build-and-cache test proving first-open assembles + reports `Open` and a second open returns the same cached `Arc`, and that an unregistered id errors), plus the existing resolution and integration suites (main / grpc_round_trip / agent_session / import). `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.


mstomar125:
Stage-4 boundary note: this PR delivers the `DatabaseManager` foundation (register-default / resolve-header / lazy-open-and-cache) and hoists service assembly into the daemon library. It is intentionally **not** on the live request path — wiring per-request routing into the serve loops changes the cross-repo `BaseServices` / `build_base_router` contract (ADR-043 lock-step with `nodespaced-pro`), so that lands in the next, sync-coordinated stage.

mstomar125:
**APPROVE**

Reviewed the full diff at `409bd19`. Stage 4 correctly activates the #1572 scaffold while staying off the live path.

**`get_or_open` — concurrency is right.** Fast-path reads the `open` map; on a miss it resolves the path (read lock), then **assembles `DatabaseServices` outside the `open` lock** so opening database A never serializes behind B, then takes the write lock and **double-checks** — if a concurrent caller inserted first, it returns theirs and drops its own set, so every request converges on one shared `Arc`. This is the correct standard pattern, and the new test asserts exactly it (first open builds + `Open`, second returns the same cached `Arc`, unregistered id errors).

**`resolve_database_id` fails closed.** An `x-ns-database-id` naming an unregistered database → `Err` (never a silent fall-through to the default, which would be a cross-database leak); absent header → the default, so single-database clients keep working. `ensure_default_registered` uses proper double-checked locking (read fast-path → write-lock re-check → adopt-first-or-register) and is idempotent — no duplicate default.

**Dormant + behavior-preserving.** The serve loops still build the default database directly; `resolve_database_id`/`get_or_open` aren't wired into request handling yet, and the `BaseServices`/`build_base_router` change + `x-ns-database-id` interception (ADR-043 sync lock-step) are explicitly deferred to the next stage — so nothing half-wired or cross-repo-coupled lands here. The `SharedServices → { settings, SharedContext { pty_manager, model, has_model } }` split is a mechanical threading refactor; the `46/46` daemon suite (incl. `grpc_round_trip`/`agent_session`/`import`, which exercise the assembled services end-to-end) plus green `test:all`/`test:e2e` corroborate the assembly move preserved behavior.

One thing to watch when this goes live (not a change request now, since it's dormant): two concurrent *first*-opens of the same id will each run `build_database_services` before the double-check discards one — i.e. two `SqliteStore` opens + idempotent schema seeds on the same file momentarily. Benign in-process (multiple SQLite connections, `IF NOT EXISTS` seeding), but if the open path ever takes an exclusive lock or runs a non-concurrency-safe migration, a per-id "loading" guard would avoid the transient. Fine as-is for Stage 4.

PR #1591, opened 2026-07-09
ADR-053 Stage 5 (part 2): route import, embeddings, and agent-session requests

mstomar125:
## ADR-053 Stage 5 (part 2) — route import, embeddings, and agent-session requests

Part of the ADR-053 epic ("One Daemon, Multiple Local Databases", #1532). Follows part 1 (#1590), which added the `DbManagerLayer` routing middleware and routed `NodeService`. This applies the same, now-reviewed pattern to the remaining cleanly per-database services.

### What changed

Each service gains a `route(&request)` helper — identical in shape to `NodeServiceImpl::route` — that reads the `DatabaseManager` injected by `DbManagerLayer`, resolves `x-ns-database-id` (absent → default; **unregistered → rejected**, never silently served from the default), and returns the target database's service. With no manager present (the Pro daemon installs no layer; unit tests construct impls directly) it returns `self`, so behavior is unchanged.

- **ImportService** — both handlers route, so an import writes to the targeted database. New test `import_rejects_unregistered_database_header`.
- **EmbeddingsService** — all seven handlers route (search / queue / regenerate / stale-count operate on the targeted database's vectors). The `stale_count_inner` helper deliberately stays on `self`: it runs on the already-routed impl.
- **AgentSessionService** — only `launch_session` routes. The `PtySessionManager` is process-global (one `Arc` backs every database), so stream / write / resize / terminate / list act on the shared session set regardless of database. `launch_session` is the sole handler that reads the per-database context assembler and node service (for context + capture), so it follows the targeted database.

### Cross-repo

No impl constructor, `BaseServices` field, or `build_base_router` signature is touched — no cross-repo change; `nodespaced-pro` is unaffected (it inherits the extension-aware handler bodies via its pinned core rev and, installing no layer, behaves identically).

### Verification

The routing wiring is enforced by the type-checker: a missed `self`→`this` swap surfaces as an unused-`this` error under `clippy -D warnings`, and returning the wrong service field is a type error. On top of that, `import_rejects_unregistered_database_header` exercises the fail-closed path end-to-end for a non-`NodeService` service (part 1's `routes_requests_by_database_header` already proves data isolation for the routing mechanism itself).

`cargo test -p nodespace-daemon` green (lib **48**); `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.

### Deferred

**LocalAgentService** routing. Its model manager is process-global (with a pending hoist-to-shared refactor) while only its chat-turn processing is per-database; routing it now would prematurely couple to that unresolved global/per-database split. It will be handled together with the model-manager-to-shared work.

### DIFF
--- a/packages/daemon/src/services/agent_session_service.rs
+++ b/packages/daemon/src/services/agent_session_service.rs
@@ -24,6 +24,7 @@ use tokio::sync::broadcast::error::RecvError;
 use tonic::{Request, Response, Status};
 use uuid::Uuid;
 
+use crate::db_routing::DATABASE_ID_HEADER;
 use crate::nodespace::{
     agent_session_service_server::AgentSessionService, AgentAvailability, CheckAvailabilityRequest,
     CheckAvailabilityResponse, LaunchSessionRequest, LaunchSessionResponse, ListSessionsRequest,
@@ -32,6 +33,7 @@ use crate::nodespace::{
     WriteInputResponse,
 };
 use crate::services::capture_service::{finalize_capture, CompletedSession};
+use crate::services::database_manager::DatabaseManager;
 use crate::services::settings_service::{read_capture_settings, CaptureConfig};
 
 /// gRPC adapter that owns shared handles to the PTY engine.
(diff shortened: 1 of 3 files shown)


mstomar125:
**APPROVE**

Reviewed the full diff at `50e228c`. This is a faithful, consistent application of the routing pattern reviewed in #1590 to the remaining cleanly-per-database services.

- **The three `route()` helpers are identical in shape to the approved `NodeServiceImpl::route`** — each reads the injected `Arc<DatabaseManager>` and, absent it, `return Ok(self.clone())` (Pro daemon / direct-constructed tests unchanged), else resolves `x-ns-database-id` (absent → default, unregistered → `NotFound`, never silent-default) and returns that database's service. Same fail-closed semantics across all three.
- **EmbeddingsService — `stale_count_inner` is on the right impl.** Every handler does `let this = self.route(&request).await?` and then calls `this.stale_count_inner()` (e.g. lines 171→174, 256→258), so the helper — which keeps `&self` — runs on the *routed* impl, operating on the targeted database's vectors. All seven handlers route.
- **AgentSessionService — the un-routed handlers are correct, not an oversight.** Only `launch_session` routes (it reads `this.assembler` / node service to build context + capture). stream/write/resize/terminate/list act on the process-global `PtySessionManager` (one `Arc` for the whole process), which is right: ADR-053 is one-user-multiple-*local*-databases, and a PTY session is addressed by session id, not database — a launched session already carries its origin database's context. No cross-database concern (single-user model).
- **ImportService** — both handlers route.
- **Completeness is compiler-enforced:** a handler that binds `this` but then reaches for `self.<field>` leaves `this` unused → fails `clippy -D warnings`, and returning the wrong service field is a type error. Combined with the new `import_rejects_unregistered_database_header` (fail-closed on a non-NodeService) and part 1's isolation test, the mechanical routing is well-covered. `test:all` + `test:e2e` green.
- **Cross-repo untouched** — no `BaseServices`/`build_base_router`/constructor change, so `nodespaced-pro` is unaffected (inherits extension-aware bodies, installs no layer → identical behavior). The `LocalAgentService` deferral (global model manager + pending hoist) is a sensible flag-and-defer.

Clean — no changes requested.

PR #1592, opened 2026-07-09
ADR-053: DatabaseService gRPC server (local-database registry)

mstomar125:
Part of the ADR-053 epic ("One Daemon, Multiple Local Databases", #1532). This adds the **client-facing `DatabaseService`** — the reachable path that makes the per-request database routing built in Stages 1–5 user-visible. Until now a client had no way to register a second database, so routing was exercised only in tests; production always resolved to the single default.

## What changed

- **`DatabaseServiceImpl`** (`services/database_service.rs`) wraps the existing `DatabaseManager` and implements the already-generated proto: `List` / `Create` / `Register` / `Remove` / `SetDefault` / `Rename`. The manager already had every method these need — this is the gRPC adapter over it.
- **Process-global, not routed.** Unlike the per-database services (nodes, import, embeddings, agent-session), this service operates on the registry itself, so it deliberately does *not* consult the `x-ns-database-id` header. Mutating RPCs re-read the registry snapshot to build the `DatabaseInfo` response, so `status`/`is_default` reflect the write just applied. Unknown-id mutations map to `NotFound`; create/register IO failures map to `Internal`.
- **Wired into `BaseServices` + `build_base_router` + all four serve loops**, cloning the `Arc<DatabaseManager>` already opened for the default database (no double-open).

## Cross-repo (ADR-043)

Adding `database` to `BaseServices` is the intended compile-time contract: `nodespaced-pro` must provide the impl when it bumps its core pin. That is a coordinated follow-up on the `nodespace-sync` side (bump pin + seed a manager with the active database); the Pro daemon installs no routing layer, so its data/sync path is unchanged and multi-database-under-one-cloud-session stays an explicit deferral.

## Verification

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo test -p nodespace-daemon` — **51 lib** (+3 new `DatabaseService` unit tests) + all integration suites green.
- **New e2e** (`tests/database_service_e2e.rs`) drives the *real* tonic transport: the `DbManagerLayer` injects the manager over the wire → `Create` registers a second database → a header-routed `NodeService` write lands in it (visible **with** the header, `NotFound` **without** it — real isolation), the second database flips to `Open`, and an unregistered header is rejected.

### DIFF
--- a/packages/daemon/src/lib.rs
+++ b/packages/daemon/src/lib.rs
@@ -40,18 +40,19 @@ pub use nodespace_proto::nodespace;
 pub use nodespace_proto::{
     AgentAvailability, AgentSessionServiceClient, AgentSessionServiceServer, CaptureContentLevel,
     CaptureSettingsResponse, CheckAvailabilityRequest, CheckAvailabilityResponse,
-    EmbeddingsServiceClient, EmbeddingsServiceServer, GetCaptureSettingsRequest,
-    ImportServiceClient, ImportServiceServer, LaunchSessionRequest, LaunchSessionResponse,
-    ListSessionsRequest, ListSessionsResponse, LocalAgentServiceClient, LocalAgentServiceServer,
-    NodeData, NodeServiceClient, NodeServiceServer, ResizeRequest, ResizeResponse, SessionInfo,
-    SettingsServiceClient, SettingsServiceServer, StreamOutputRequest, TerminateSessionRequest,
-    TerminateSessionResponse, UpdateCaptureSettingsRequest, WriteInputRequest, WriteInputResponse,
+    DatabaseServiceClient, DatabaseServiceServer, EmbeddingsServiceClient, EmbeddingsServiceServer,
+    GetCaptureSettingsRequest, ImportServiceClient, ImportServiceServer, LaunchSessionRequest,
+    LaunchSessionResponse, ListSessionsRequest, ListSessionsResponse, LocalAgentServiceClient,
+    LocalAgentServiceServer, NodeData, NodeServiceClient, NodeServiceServer, ResizeRequest,
+    ResizeResponse, SessionInfo, SettingsServiceClient, SettingsServiceServer, StreamOutputRequest,
+    TerminateSessionRequest, TerminateSessionResponse, UpdateCaptureSettingsRequest,
+    WriteInputRequest, WriteInputResponse,
 };
 
 pub use db_routing::{DbManagerLayer, DATABASE_ID_HEADER};
 pub use router::{build_base_router, BaseServices};
 pub use services::{
     build_database_services, build_shared_services, AgentSessionHandler, DatabaseManager,
-    DatabaseServices, EmbeddingsServiceImpl, ImportServiceImpl, LocalAgentServiceImpl,
-    NodeServiceImpl, SettingsServiceImpl, SharedContext, SharedServices,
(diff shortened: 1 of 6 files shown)


mstomar125:
**APPROVE** — with a hard cross-repo lock-step flag (below); the code itself is correct.

### What I verified
- **`DatabaseServiceImpl` is a thin, correct adapter** over the already-reviewed `DatabaseManager` (#1572/#1589):
  - `list` → `manager.list()` snapshot; `create(name, Option<path>)` / `register(path)` map IO failures → `Status::internal`; `remove` / `set_default` / `rename` map unknown-id → `Status::not_found`. Correct: creation failures are genuinely internal, missing-id lookups are genuinely not-found.
  - `remove` calls `manager.remove(&id)`, which per #1572 is **unregister-only** (retain + save, no `fs::remove_file`) — no data-destruction path introduced here. Good.
  - `info_for` re-reads the registry snapshot after a write and returns `Status::internal("… missing from registry after write")` if absent — a sane post-write invariant.
  - Process-global by design (operates on the registry, not a routed DB) — correct not to route it through `DbManagerLayer`.
- **Core still compiles:** `main.rs` supplies `database: DatabaseServiceImpl::new(manager.clone())` at all 4 `BaseServices` construction sites; `build_base_router` registers `DatabaseServiceServer::new(services.database)`.
- **e2e is real, not a smoke test:** `database_service_e2e.rs` drives actual tonic transport — creates a second DB, routes a write to it via the `x-ns-database-id` header, asserts isolation from the default DB, and asserts unknown-id → `Code::NotFound`. Strong coverage of the routing contract.

### ⚠️ Cross-layer lock-step (the reason to gate the merge/deploy order)
This PR adds a **required public field** `pub database: DatabaseServiceImpl` to `BaseServices` — the struct that **`nodespaced-pro` (nodespace-sync) also constructs** under the ADR-043 shared-router contract. This is a **compile-time breaking change**, not a runtime one:

> The moment nodespace-sync bumps its core pin to a rev containing this PR, its `BaseServices { … }` construction fails to build with `missing field 'database'`. The Pro daemon will not compile until sync provides its own `DatabaseServiceImpl` (seed a `DatabaseManager` with the active DB via `ensure_default_registered`) in that same pin bump.

The PR body already flags this ("intended compile-time contract … coordinated follow-up on the nodespace-sync side") — that's the healthy call, and making the field compile-required (rather than an `Option` that silently no-ops) is the *right* design. But operationally:

1. **No paired nodespace-sync PR exists yet** (checked: zero open sync PRs). Until one is ready, **do not advance the sync-side core pin past this commit**, or the Pro build breaks.
2. This is exactly the **DatabaseService-gRPC-server boundary the ADR-053 epic was documented as "parked" at** (cross-repo + user check-in). Proceeding is fine, but the sync-side DatabaseServiceImpl construction should land as a coordinated pair.
3. **Pro-side semantics note:** once sync adds the field, the Pro daemon will *expose* DatabaseService but installs **no `DbManagerLayer`** (#1590) — so registering a 2nd DB on Pro is **inert** (writes still route to the default). That matches the epic's "multi-DB-under-one-cloud-session = explicit deferral," just worth stating so it isn't mistaken for a bug when a Pro-registered DB doesn't route.

### Merge gate
CI `test:all` + `test:e2e` are **QUEUED** at this SHA (9385e32e) — must go green before merge. (Core policy: I re-review on every new head SHA.)

**Net:** approve the code; hold the merge/pin-advance until the paired nodespace-sync DatabaseServiceImpl change is ready so the two land as a lock-step pair.

PR #1593, opened 2026-07-09
ADR-053: stamp WatchNodes events with the serving database id

mstomar125:
Part of the ADR-053 epic ("One Daemon, Multiple Local Databases", #1532). One of the two remaining acceptance criteria: **event streams carry `database_id`** so a client watching a specific database can attribute the events it receives when the daemon serves several.

## What changed
- **Proto:** `NodeEvent` gains an additive `string database_id = 7;` (sibling of the event oneof). Backward-compatible — existing consumers (the tauri watcher, the frontend) ignore it, so the community app is behaviorally unchanged.
- **`NodeServiceImpl`** gains a `database_id`, set via a **`with_database_id` builder**. `build_database_services` calls it with the registry id when a database is opened through the manager. **`NodeServiceImpl::new` is unchanged**, so the single-database Pro daemon (`nodespaced-pro` constructs it directly) and direct-construction unit tests keep an empty id and carry no cross-repo break.
- **Watch handler** stamps `this.database_id` (the routed database, resolved by the existing `route()`) onto every emitted event. `convert_domain_event` now returns the inner `NodeEventKind` so the id is applied at a single construction site.

Header-less/default watch streams carry the default database's id; a stream routed by `x-ns-database-id` carries that database's id.

## Verification
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo test -p nodespace-daemon` — **51 lib + all integration suites** green.
- **e2e** (`database_service_e2e.rs`): opens watch streams on the default and a second database over the real transport, then routes a write to each and asserts each stream's `Created` event carries **its own** database id (default id for the header-less stream, the second id for the routed stream).

## Cross-repo
No constructor or `BaseServices` change — `nodespaced-pro` is unaffected and inherits the additive proto field transparently on its next pin bump.

### DIFF
--- a/packages/daemon/src/services/assembly.rs
+++ b/packages/daemon/src/services/assembly.rs
@@ -119,6 +119,7 @@ pub async fn build_shared_services() -> Result<(SharedServices, Option<tokio::ta
 pub async fn build_database_services(
     db_path: &std::path::Path,
     shared: &SharedContext,
+    database_id: &str,
 ) -> Result<(DatabaseServices, Option<tokio::task::JoinHandle<()>>)> {
     if let Some(parent) = db_path.parent() {
         tokio::fs::create_dir_all(parent).await.with_context(|| {
@@ -145,7 +146,8 @@ pub async fn build_database_services(
 
     let node_service = Arc::new(node_service);
 
-    let node_service_grpc = NodeServiceImpl::new(node_service.clone(), embedding_state.clone());
+    let node_service_grpc = NodeServiceImpl::new(node_service.clone(), embedding_state.clone())
+        .with_database_id(database_id.to_string());
 
     // EmbeddingsService is only registered when a model file exists at startup
     // (the shared model). If the model appears later, the endpoint is absent
--- a/packages/daemon/src/services/database_manager.rs
+++ b/packages/daemon/src/services/database_manager.rs
@@ -389,7 +389,7 @@ impl DatabaseManager {
         // Assemble the service set (opens SQLite, seeds schema) without holding
         // the `open` lock. The per-database embedding-wiring task is detached,
         // matching the boot path. `last_opened_at` bookkeeping is deferred.
-        let (services, _embed_task) = build_database_services(&path, &self.context)
+        let (services, _embed_task) = build_database_services(&path, &self.context, id.as_str())
             .await
             .with_context(|| format!("opening database {id}"))?;
         let services = Arc::new(services);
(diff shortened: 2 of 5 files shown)


mstomar125:
**APPROVE** — clean additive Stage 6; verified against the code.

### Verified
- **Proto is truly additive, no field collision.** On `upstream/main`, `NodeEvent` uses field numbers **1–6** (all inside `oneof event`); `string database_id = 7;` is the next free number and a sibling of the oneof, not a oneof arm — so it accompanies every event regardless of kind, and proto3 unknown-field handling means existing consumers (tauri watcher, frontend) ignore it. Wire-backward-compatible. Community app behaviorally unchanged. ✅
- **No cross-repo break (deliberate, unlike #1592).** `NodeServiceImpl::new` keeps its 2-arg signature — the struct gains `database_id: String` defaulted to `String::new()` inside `new`, and the id is applied via a separate `with_database_id(self, String) -> Self` builder. `nodespaced-pro` constructs `NodeServiceImpl` directly, so it keeps an empty id and needs no coordinated change (it inherits the additive proto field on its next pin bump). Confirmed no `BaseServices`/ctor-signature change in the diff. ✅
- **Id is stamped from the routed target, at a single site.** The watch loop reads `this.database_id` (`node_service.rs:75`) — and `this` is the service that `route()` resolved to, i.e. the target DB's `NodeServiceImpl` (whose id was set by `with_database_id` when `build_database_services` opened it through the manager). So a header-routed stream carries that DB's id and a header-less stream carries the default's. `convert_domain_event` was refactored to return the inner `NodeEventKind`, so the wrapping `NodeEvent { event, database_id }` is built at one construction site (`:86`) — no risk of a kind that forgets to stamp the id. ✅
- **e2e proves the attribution**, not just compilation: `database_service_e2e.rs` opens watch streams on the default and a second DB over real transport, writes to each, and asserts each stream's `Created` event carries **its own** id. Strong.

### Note (not a defect)
Per sync#294, the Pro daemon opens its single DB **directly** (not through the manager, no `DbManagerLayer`), so Pro watch events will carry an **empty** `database_id`. That's exactly what the proto comment documents ("Empty when the daemon serves a single unregistered database (Pro daemon)") and the sync consumer already tolerates it (ignores field 7). Consistent — flagging only so the empty-id-on-Pro stream isn't later mistaken for a bug.

### Merge gate
CI `test:all` + `test:e2e` **QUEUED** at `2db9f060` — must go green before merge. (Core: I re-review on each new head SHA.)

This is one of the two remaining #1532 acceptance criteria (event streams carry `database_id`); the other is Stage 7 (drop `active_database_path`). Approve.

PR #1594, opened 2026-07-09
ADR-053: stop injecting NODESPACED_DB_PATH from the daemon service definitions

mstomar125:
Part of the ADR-053 epic (#1532) — the first half of the final acceptance criterion ("remove `active_database_path` + plist `NODESPACED_DB_PATH` injection").

## What changed
The macOS launchd plist and the Linux systemd unit (`daemon_setup.rs`) injected `NODESPACED_DB_PATH` = `$HOME/.nodespace/database/nodespace.db` — the **exact path the daemon already derives in `resolve_db_path()` when the variable is unset**. Under ADR-053 the database registry is the source of truth for which database the daemon opens, so hard-coding the path in the service definition is redundant. This removes the injection from both service definitions.

## Safety
Verified the daemon's default (`resolve_db_path()` → `$HOME/.nodespace/database/nodespace.db`) is identical to the removed injected value, so **the database location does not change**. `NODESPACED_DB_PATH` remains honored when set (integration tests / alternate deployments still redirect storage via `test-support`), it is simply no longer injected by the packaged service definition.

## Scope note
The other half of the criterion — removing the vestigial `active_database_path` setting (the daemon never reads it for DB selection; its `select/reset-database` commands are effectively no-ops) and re-sourcing the Settings DB-path display from `DatabaseService.List` — is a separate focused change tracked on #1532, since it touches the settings proto + the tauri command layer + the frontend.

### DIFF
--- a/packages/desktop-app/src-tauri/src/daemon_setup.rs
+++ b/packages/desktop-app/src-tauri/src/daemon_setup.rs
@@ -610,7 +610,6 @@ fn write_plist(home: &Path, plist_path: &Path, daemon_bin: &Path) -> Result<()> 
     let home_str = home.to_string_lossy();
     let bin_str = daemon_bin.to_string_lossy();
     let socket_path = xml_escape(&format!("{}/{}", home_str, daemon_socket_relative()));
-    let db_path = xml_escape(&format!("{}/{}/nodespace.db", home_str, DAEMON_DB_DIR));
     let log_out = xml_escape(&format!("{}/{}/nodespaced.log", home_str, DAEMON_LOG_DIR));
     let log_err = xml_escape(&format!(
         "{}/{}/nodespaced-error.log",
@@ -663,8 +662,6 @@ fn write_plist(home: &Path, plist_path: &Path, daemon_bin: &Path) -> Result<()> 
     <dict>
         <key>NODESPACED_SOCKET</key>
         <string>{socket}</string>
-        <key>NODESPACED_DB_PATH</key>
-        <string>{db}</string>
         <key>NODESPACE_UI_BINARY</key>
         <string>{ui_binary}</string>
 {pro_env}    </dict>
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE** — the data-safety claim is independently verified; this is a safe, well-scoped removal.

### The one thing that mattered: does removing the injection move the DB? No — verified.
The risk in dropping `NODESPACED_DB_PATH` injection is that the daemon's fallback path might differ from the injected one, silently relocating every community user's database (their notes would "vanish" into a fresh empty DB). I checked both sides at head `bf6bc858`:
- **Removed injected value** (`daemon_setup.rs`): `$HOME/{DAEMON_DB_DIR}/nodespace.db`, and `DAEMON_DB_DIR = ".nodespace/database"` (`:28`) → `$HOME/.nodespace/database/nodespace.db`.
- **Fallback** (`packages/daemon/src/lib.rs:23` `resolve_db_path()`): when `NODESPACED_DB_PATH` is unset it returns `home/.nodespace/database/nodespace.db` (`:32-34`; the doc-comment at `:22` states this verbatim).

**Byte-identical.** The daemon falls through to exactly the path that was being injected, so the database location does not change. ✅

### Other checks
- **`NODESPACED_DB_PATH` still honored when set** (`lib.rs:24` env check retained) — integration tests / `test-support` / alternate deployments keep working; only the packaged launchd/systemd service definition stops hard-coding it. Correct per ADR-053 (the registry, not the service def, is the source of truth).
- **Dir creation retained** — `daemon_setup.rs:219 let db_dir = home.join(DAEMON_DB_DIR);` is untouched, so the storage directory is still created; the diff is purely the two env-injection blocks (`+0/-7`).
- Community behavior unchanged (this only removes a redundant env export that resolved to the default anyway).

### Scope / follow-up
This is only the **first half** of the final #1532 criterion (stop the plist/systemd injection). The **second half** — removing the vestigial `active_database_path` setting (daemon never reads it for DB selection; `select/reset-database` are no-ops) and re-sourcing the Settings→Database/Diagnostics path display from `DatabaseService.List` — is still a separate pending PR (touches settings proto + tauri commands + frontend). So **#1532 does not close on this merge alone**; expect one more Stage-7 PR.

### Merge gate
`test:e2e` COMPLETED/SUCCESS; `test:all` IN_PROGRESS at `bf6bc858` — let it finish green before merge. (Core: re-review on each new head SHA.)

Approve.

PR #1595, opened 2026-07-09
Remove vestigial active_database_path daemon setting (ADR-053)

mstomar125:
## Summary

Final cleanup for the ADR-053 multi-database epic (#1532). The daemon never read `active_database_path` from `daemon.toml` — it resolves its database via the `DatabaseService` registry — so this field was vestigial and the Tauri commands that wrote it were no-ops. With the prior change that stopped injecting `NODESPACED_DB_PATH` from the service definitions (#1594), those commands became fully inert.

This removes the dead field and the dead command/UI surface while preserving the read-only database-path display, now sourced from the registry (the actual source of truth).

## Changes

- **Daemon** — remove `active_database_path` from `DaemonConfig`, `DaemonConfigResponse`, and the `UpdateDaemonConfig` handler (`settings_service.rs`) and from the `SettingsService` proto messages. `SettingsService` now owns only `grpc_address` + capture settings.
- **Tauri** — delete the no-op `select_new_database` / `reset_database_to_default` commands, their `invoke_handler` registrations, and the File-menu items/handler that only ever called them. `get_settings` now sources `SettingsResponse.active_database_path` from `DatabaseService.List`'s default entry via a new `GrpcClient::database_service_client()` accessor — the response field name is unchanged, so the Settings and Diagnostics path displays render exactly as before.
- **Frontend** — remove the "Change Location…" / "Reset to Default" buttons from the Database settings section and the `menu-select-database` event listener that invoked them. The read-only "Active Database Path" display is kept.

## Validation

- `cargo fmt --check` clean; `cargo clippy` clean on `nodespace-daemon` and `nodespace-app`.
- `cargo test -p nodespace-daemon` — all pass, including the DatabaseService gRPC e2e (`create_second_database_then_route_a_write_to_it`, `watch_events_are_stamped_with_the_serving_database_id`) and the `settings_service` unit tests.
- `cargo check` on the Tauri crate (`nodespace-app`) compiles with sidecars built.
- Frontend `bun run test`: 3952/3952 pass (incl. `settings.test.ts`).
- Pre-push gate (`test:all` + `cargo build --bin nodespaced` + `test:e2e`) green.

Community build is behaviorally unchanged: the path display still works (registry-sourced); the only visible difference is the removal of the already-broken select/reset affordances.

This completes the final acceptance criterion of the ADR-053 epic.

### DIFF
--- a/packages/daemon/src/services/settings_service.rs
+++ b/packages/daemon/src/services/settings_service.rs
@@ -25,7 +25,6 @@ fn default_socket_path() -> String {
 /// On-disk representation of `~/.nodespace/daemon.toml`.
 #[derive(Debug, Default, Serialize, Deserialize)]
 struct DaemonConfig {
-    active_database_path: Option<String>,
     grpc_address: Option<String>,
     #[serde(default)]
     capture: CaptureConfig,
@@ -145,7 +144,6 @@ impl SettingsServiceImpl {
 
     fn config_to_response(config: &DaemonConfig) -> DaemonConfigResponse {
         DaemonConfigResponse {
-            active_database_path: config.active_database_path.clone().unwrap_or_default(),
             grpc_address: config
                 .grpc_address
                 .clone()
@@ -181,9 +179,6 @@ impl GrpcSettingsService for SettingsServiceImpl {
 
         let mut config = self.read_config().await?;
 
-        if !req.active_database_path.is_empty() {
-            config.active_database_path = Some(req.active_database_path);
-        }
         if !req.grpc_address.is_empty() {
             config.grpc_address = Some(req.grpc_address);
         }
--- a/packages/desktop-app/src-tauri/src/commands/settings.rs
+++ b/packages/desktop-app/src-tauri/src/commands/settings.rs
@@ -1,15 +1,18 @@
 //! Settings commands for reading and updating app preferences.
 //!
-//! Daemon config (database path, gRPC address) is owned by `nodespaced` and
-//! fetched/updated via the `SettingsService` gRPC RPC.
+//! Daemon config (gRPC address) is owned by `nodespaced` and fetched/updated
+//! via the `SettingsService` gRPC RPC.
+//!
+//! The active database path is not daemon config — it is read from the
+//! `DatabaseService` registry (ADR-053: one daemon, multiple local
+//! databases), which is the source of truth for which database is default.
 //!
 //! Display preferences (theme, render_markdown) are UI-only state that remain
 //! in Tauri local storage and are never sent to the daemon.
 
 use crate::services::GrpcClient;
 use nodespace_proto::nodespace::{
-    GetCaptureSettingsRequest, GetDaemonConfigRequest, UpdateCaptureSettingsRequest,
-    UpdateDaemonConfigRequest,
+    GetCaptureSettingsRequest, ListDatabasesRequest, UpdateCaptureSettingsRequest,
 };
 use tauri::{AppHandle, Manager};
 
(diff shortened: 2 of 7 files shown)


mstomar125:

Independently re-verified the whole diff against the ADR-053 intent — clean, cohesive removal.

Verified:
- `active_database_path` fully removed from the daemon `DaemonConfig` struct, `config_to_response`, and the `UpdateDaemonConfig` handler, and from the `SettingsService` proto get/update messages.
- Both no-op Tauri commands (`select_new_database`, `reset_database_to_default`) deleted along with their `invoke_handler` registrations, the File-menu items, and the menu match-arm that handled their event ids — no orphaned menu id left behind. The frontend `menu-select-database` listener + its cleanup are gone.
- `get_settings` now sources `SettingsResponse.active_database_path` from `DatabaseService.List`'s default entry via the new `GrpcClient::database_service_client()`; the response field name is unchanged so the Settings/Diagnostics path display renders identically (and now always reflects the resolved default rather than a manually-picked value). `ensure_default_registered()` runs before the gRPC server serves, so `List` always returns a default entry while the daemon is up.
- Imports correct (swaps to `ListDatabasesRequest`, drops the two now-unused request types; daemon still uses them for its still-defined RPCs). `get_default_database_path()` is still live (used by `migrate_legacy_database_if_needed`). No greenfield/`console.*`/`any`/lint-suppression violations.

Validation: `cargo fmt`/`clippy` clean on `nodespace-daemon` + `nodespace-app`; `cargo test -p nodespace-daemon` green (incl. DatabaseService gRPC e2e + `settings_service` unit tests); Tauri crate compiles; `bun run test` 3952/3952; CI `test:all` + `test:e2e` green.

Non-blocking follow-ups (out of scope for this PR): with these deletions, the `GetDaemonConfig`/`UpdateDaemonConfig` RPCs now have no callers and `grpc_address` is unused by the socket bind — worth a separate cleanup ticket. Optional proto hygiene: `reserved 1;` on the two messages.

mstomar125:
**APPROVE** (post-merge — this opened and merged between watcher polls; reviewing for the record). Correct, well-scoped, and it correctly closes the ADR-053 epic. One minor proto-hygiene follow-up below.

### Verified
- **The removal is genuinely dead code.** `active_database_path` was never read by the daemon for DB selection (the registry/`DatabaseService` is the source of truth), and after #1594 stopped injecting `NODESPACED_DB_PATH` the `select_new_database` / `reset_database_to_default` commands were fully inert. Removed cleanly from `DaemonConfig`/`DaemonConfigResponse`/`UpdateDaemonConfig` handler + proto, the two Tauri commands + their `invoke_handler` registrations + the File-menu items, and the frontend "Change Location…/Reset to Default" buttons + `menu-select-database` listener.
- **Display is preserved — community-unaffected holds.** The app→frontend `SettingsResponse.active_database_path` **field name is kept**, now sourced from the registry (`settings.rs`): `client.list(ListDatabasesRequest{})` → `databases.iter().find(|db| db.is_default).map(|db| db.path.clone()).unwrap_or_default()`. So the read-only Active Database Path in Settings **and** Diagnostics renders exactly as before; the only visible change is the removal of the already-broken select/reset affordances. Matches the "free users unaffected" invariant.
- **Graceful degradation:** no default entry → empty string (documented on the accessor). A failed `List` makes `get_settings` error — but that's the same failure domain as the pre-existing daemon connection this command already needed, so no new fragility.
- CI green at merge (`test:all` + `test:e2e` both SUCCESS); frontend `bun test` 3952/3952 including `settings.test.ts`.

### Minor follow-up (proto hygiene, non-blocking)
Both `active_database_path = 1;` removals (in `UpdateDaemonConfigRequest` and `DaemonConfigResponse`) drop field number **1 without a `reserved 1;` / `reserved "active_database_path";`**. Best practice on field deletion is to reserve the number+name so a future field can't silently reuse `1` with a different type. **Live risk is low** here — these are local, co-bundled Tauri↔daemon IPC messages (client and daemon are always the same build, not persisted to disk, no rolling/mixed-version wire skew) — so it's hygiene, not a current hazard. But since both are small config messages likely to gain fields, a future dev filling the number-1 hole is a plausible footgun; a one-line `reserved 1;` in each message in a fast-follow would close it.

**Net:** correct final piece of #1532 (now closed). Approve; suggest the two `reserved` lines as a trivial fast-follow.

PR #1597, opened 2026-07-09
CLI: client-side multi-database selection (ADR-053)

mstomar125:
## Summary

Client-side (CLI) half of ADR-053 "One Daemon, Multiple Local Databases" (#1533). Now that the daemon serves N registered databases behind its socket (#1532), this lets the `nodespace` CLI target, create, and manage local databases — the first real client to exercise `DatabaseService` and `x-ns-database-id` routing end to end.

This PR is the **CLI slice only**; the desktop switcher UI / stores / event filtering follow in separate PRs, so it does not close #1533.

## Changes

- **Routing header** — a concrete `DatabaseIdInterceptor` (tonic `Interceptor`) stamps `x-ns-database-id` on every data-plane request. It's applied uniformly (even with no selection, where it stamps nothing and the daemon serves its default), so client types stay nameable via aliases (`NodeClient`, `ImportClient`, `SessionClient`, `LocalAgentClient`). The `connect*` helpers build clients through the interceptor.
- **`--database <name|id>` / `NODESPACE_DATABASE`** — a global selector resolved once per invocation, before dispatch, against `DatabaseService.List`. The daemon resolves the header as an **id only**, so a name is resolved to its id client-side (exact-id match first, then a unique name match; clear errors on no-match and on ambiguous duplicate names). A `--database` that doesn't resolve fails before the command runs.
- **`nodespace database list|create|register|remove|rename|use`** — registry management over `DatabaseService`. `remove` unregisters only and says so (never deletes the file). `use` maps to `SetDefault` (the CLI is stateless; the registry's default is the single source of truth). These subcommands operate on the registry globally and are never routed by `--database`.
- **`nodespace diagnostics`** rewritten — enumerates the registry via `List` (name, path, default marker, status per database) and runs the node/root/schema counts + size-on-disk against the targeted database (default or `--database`), instead of computing one path from `resolve_db_path()`.
- **SKILL.md** documents the flag, env var, and subcommands.
- **Integration tests** — a new in-process routing harness (`DbManagerLayer` + `DatabaseService` + `NodeService` over UDS) covering the registry round-trip, name→id resolution, unregistered-selection error, and cross-database write isolation (a write routed to database B is invisible against the default database — no bleed).

## Validation

- `cargo fmt --check` clean; `cargo clippy -p nodespace-cli --all-targets -- -D warnings` clean.
- `cargo test -p nodespace-cli` — 5 unit + 14 integration, 0 failures.
- No Rust crate outside `packages/cli` consumes the `nodespace_cli` library API, so the changed `connect*`/`diagnostics` signatures affect nothing else; the desktop app consumes the CLI only as a sidecar binary.

Community build behavior is unchanged for existing invocations: with no `--database`/`NODESPACE_DATABASE`, every command routes to the daemon's default database exactly as before.


mstomar125:

Independently reviewed the full diff against the merged daemon side. The implementation matches the ADR-053 design intent and I verified the contract against the daemon: `resolve_database_id` (`database_manager.rs:346`) confirms id-only header resolution, and `insert_entry` (`database_manager.rs:422`) confirms the daemon allows duplicate names — so the CLI's client-side name→id resolution and its ambiguity branch are both genuinely necessary.

Verified:
- **Interceptor** stamps `x-ns-database-id` per-request (streaming included — the session/import aliases carry the interceptor, and `session.rs` clones it onto the input client); `None` stamps nothing so the daemon serves its default. `for_id` propagates the non-ASCII error path.
- **Resolver** is id-first, then unique-name, with clear no-match and ambiguity errors; a single implementation is reused by `--database`, `remove`, `rename`, and `use` (no divergent copies). `resolve_routing` makes no `List` call when `--database` is absent, and an invalid selection fails before any command half-runs.
- **`database` subcommands**: `remove` is unregister-only and says the file is not deleted; `use` → `SetDefault`; `list` marks the default and shows status. All network paths use `?`/`.context`.
- **diagnostics** enumerates the registry and counts/sizes against the targeted database; graceful zero-databases / no-default handling; the size helpers + their unit tests are intact; the old single-`database_path` shape and `resolve_db_path` are fully gone from the CLI.
- **Mechanical `<Channel>`→alias swap** across node/mention/schema/search/model/import/session with all now-unused `Channel` imports removed.

Two non-blocking review findings addressed in `5c4f67eb`:
- Added `ambiguous_name_selection_errors_but_id_still_resolves` — the duplicate-name ambiguity branch is reachable (daemon doesn't enforce unique names) and was previously uncovered; the test creates two `work` databases, asserts the name selection errors with the colliding ids listed, and that selecting by id still resolves.
- SKILL.md no longer lists `model` among routed commands — `LocalAgentService` doesn't read the routing header (model management is daemon-global); the flag is accepted but has no effect there.

Validation: `cargo fmt --check` + `cargo clippy -p nodespace-cli --all-targets -- -D warnings` clean; `cargo test -p nodespace-cli` = 5 unit + 15 integration, 0 failures. Community behavior unchanged (no `--database` → daemon default, as before).

mstomar125:
**APPROVE** — the first real client of `DatabaseService`/`x-ns-database-id`, and it's correct end-to-end. Verified the two things most likely to bite here.

### Verified
- **No-selection stamps an ABSENT header, not an empty one.** `DatabaseIdInterceptor::call` is `if let Some(id) = &self.database_id { metadata.insert(DATABASE_ID_HEADER, id) }` — with `database_id: None` it inserts nothing, so the daemon's `resolve_database_id` sees an absent header → default DB. This is the right distinction: an empty-string `x-ns-database-id` would have risked resolving to `NotFound` instead of default. ✅
- **Header key is the shared const**, `nodespace_daemon::DATABASE_ID_HEADER` — no hand-copied string that could drift from the daemon's resolver. ✅
- **Name→id resolution is unambiguous and fail-loud** (`resolve_database_id_by_selection`): exact **id** match first (preferred, unambiguous), else filter by name → `[]` bails "no database named or with id '{selection}' is registered", `[one]` returns its id, `[many]` bails "'{selection}' is ambiguous — N databases share that name". Resolved once before dispatch, so a bad `--database` fails before the command runs. ✅
- **The one daemon-unrouted client is harmless here.** `local_agent` routing is deferred on the daemon (#1591), and the interceptor stamps the header on `LocalAgentClient` too — but `LocalAgentClient` is used **only** by `model.rs` (`model list|load|recommended`), and model management is process-global (not per-database). So the daemon ignoring the header there is *correct behavior* — `--database` is simply an inert no-op on `model` subcommands, not a silent wrong-database write. Every content command routes through a client the daemon **does** honor: node/mention → `NodeClient`, import → `ImportClient`, session → `SessionClient` (all routed per #1590/#1591). No `--database` → wrong-DB mismatch exists. ✅
- **Registry subcommands are correctly un-routed** (`list/create/register/remove/rename/use` operate on the process-global registry, not via `--database`); `remove` is unregister-only (matches `DatabaseManager::remove`, never deletes the file); `use` → `SetDefault` (CLI stateless, registry is the source of truth). ✅
- **Community-unaffected:** with no `--database`/`NODESPACE_DATABASE`, every command routes to the daemon default exactly as before.

### Tests
Strong: the in-process routing harness (`DbManagerLayer` + `DatabaseService` + `NodeService` over UDS) covers registry round-trip, name→id resolution, unregistered-selection error, and **cross-database write isolation** (a write routed to DB B is invisible against the default — proves no bleed). 5 unit + 14 integration, 0 failures; clippy `-D warnings` + fmt clean.

### Note (not a change request)
`--database` is accepted (global flag) on `model` subcommands but has no effect there since models are process-global — that's correct, just worth a one-line doc mention in SKILL.md if you want to preempt "why didn't `--database` change my model list" questions.

### Merge gate
CI `test:all` + `test:e2e` IN_PROGRESS at `5c4f67eb` — let both go green before merge. This is the CLI slice of #1533 (desktop switcher/stores/event-filtering follow separately), so it correctly does **not** close #1533. Approve.

PR #1601, opened 2026-07-09
Drop in-flight database reads after an active-database switch (ADR-053)

mstomar125:
## Summary

Follow-on hardening from the ADR-053 desktop switcher (#1533 / #1598). When the active database is switched, a read dispatched against the *previous* database before the switch — whose response resolves *after* the switch's `clearAll()` — could `setNode` the old database's rows into the now-active store. Those nodes are unreferenced by the new tree (invisible in the outliner) but linger in `sharedNodeStore` and can surface via global search / mention resolution until the next reload.

This closes that window using the issue's **Option A**: a monotonic database epoch on `SharedNodeStore`, bumped by `clearAll()` (already called by `switchTo`). Every path that reads from the daemon and then writes into the store captures the epoch before awaiting and drops the write if it has advanced.

## Changes

- **`shared-node-store.svelte.ts`** — `private databaseEpoch`, `currentEpoch()` getter, `clearAll()` bumps it. Guards on `ensureNode`, `loadChildrenForParent` (restructured so both daemon reads complete before a single staleness check gates the parent, children, and structureTree-edge writes together), `doLoadChildrenTree`, `resyncNodeFromServer`.
- **`tauri-sync-listener.ts`** — guards on `fetchAndUpdateNode` (per-event hydration) and `flushPendingNodeFetches` (Pro reconnect-replay coalescer, whole-burst drop). `isActiveDatabaseEvent` gates on event *arrival*, before these async fetches dispatch, so it cannot close the in-flight window on its own.
- **`browser-sync-service.ts`** — `fetchAndUpdateNode` guard (HTTP-adapter parity; browser mode can't switch databases today, so defensive/future-proof).
- **`navigation-service.ts`** — link-click resolution now aborts (returns `null`) when the epoch advances, rather than opening a tab for a node absent from the now-active store.
- Also guarded: `node-card-inline`, `query-node-viewer`, `ai-chat-node-viewer`, `app-shell` MCP node-updated listener.
- `switchTo` comment updated to describe the epoch guard that closes this window.

## Tests

- `shared-node-store-database-epoch.test.ts` (new, 3) — store-level: a read resolving after the epoch advances is dropped; within the same epoch it is applied.
- `tauri-sync-listener.test.ts` (+2) — the community per-event path and the Pro coalescer path each drop an in-flight read across a switch.
- Full file green (31/31); eslint + svelte-check 0/0 (1738 files).

## Acceptance

- [x] A read issued against database A that resolves after a switch to B does not leave A's nodes in the B store (regression tests above).
- [x] No behavior change in the common (no in-flight read across a switch) case — the guard is a same-epoch no-op.

Closes #1599.

### DIFF
--- a/packages/desktop-app/src/lib/components/chat/node-card-inline.svelte
+++ b/packages/desktop-app/src/lib/components/chat/node-card-inline.svelte
@@ -38,8 +38,11 @@
   // reactive watch (ADR-049). Once fetched, `node` updates via the store read above.
   onMount(() => {
     if (sharedNodeStore.getNode(nodeId)) return;
+    // ADR-053: capture the database generation so a switch mid-fetch drops this
+    // read instead of writing the previous database's node into the now-active store.
+    const epoch = sharedNodeStore.currentEpoch();
     backendAdapter.getNode(nodeId).then((fetched) => {
-      if (fetched) {
+      if (fetched && sharedNodeStore.currentEpoch() === epoch) {
         sharedNodeStore.setNode(fetched, { type: 'database', reason: 'node-card-fetch' }, true);
       }
     }).catch((e) => {
--- a/packages/desktop-app/src/lib/components/layout/app-shell.svelte
+++ b/packages/desktop-app/src/lib/components/layout/app-shell.svelte
@@ -130,8 +130,12 @@
       async (event) => {
         log.debug('[MCP] Node updated:', event.payload.node_id);
         try {
+          // ADR-053: capture the database generation so a switch mid-fetch drops
+          // this write instead of writing the previous database's node into the
+          // now-active store.
+          const epoch = sharedNodeStore.currentEpoch();
           const node = await invoke<Node>('get_node', { id: event.payload.node_id });
-          if (node) {
+          if (node && sharedNodeStore.currentEpoch() === epoch) {
             sharedNodeStore.setNode(node, { type: 'mcp-server' }, false);
 
             // Invalidate collection member caches since node title may have changed
--- a/packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte
@@ -340,8 +340,12 @@
       if (cancelled || attempts >= MAX_ATTEMPTS) return;
       attempts++;
       try {
+        // ADR-053: drop this poll's write if the active database switches while
+        // the fetch is in flight, so the previous database's node isn't written
+        // into the now-active store.
+        const epoch = sharedNodeStore.currentEpoch();
         const fetched = await backendAdapter.getNode(nodeId);
-        if (fetched && !cancelled) {
+        if (fetched && !cancelled && sharedNodeStore.currentEpoch() === epoch) {
           sharedNodeStore.setNode(fetched, { type: 'database', reason: 'poll' }, true);
           // If SSE is down, nudge it to reconnect.
           if (!browserSyncService.isConnected()) {
--- a/packages/desktop-app/src/lib/components/viewers/query-node-viewer.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/query-node-viewer.svelte
@@ -71,6 +71,9 @@
 
   async function loadAndQuery(schemaId: string) {
     const loadId = ++currentLoadId;
+    // ADR-053: capture the database generation so a switch mid-load drops these
+    // rows instead of writing the previous database's nodes into the now-active store.
+    const epoch = sharedNodeStore.currentEpoch();
     queryState = 'loading';
     error = null;
     schemaNode = null;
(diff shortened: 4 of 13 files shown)


mstomar125:
**APPROVE**

Reviewed independently (adversarial pass over every daemon read-then-`setNode` site).

**Caught and fixed during review (was a blocker):** the first pass guarded the `SharedNodeStore` methods but missed the busiest desktop store-population paths — the Tauri domain-event hydration in `tauri-sync-listener.ts` (`fetchAndUpdateNode` and the Pro reconnect-replay coalescer `flushPendingNodeFetches`). `isActiveDatabaseEvent` gates on event *arrival*, synchronously, before the async fetch dispatches — so it does not close the in-flight-read window on its own. Both are now epoch-guarded (whole-burst drop for the coalescer), plus `browser-sync-service.fetchAndUpdateNode` for HTTP-adapter parity, and `navigation-service` now aborts the navigation (`return null`) rather than opening a tab for a node absent from the now-active store.

**Verification of the final state:**
- Every path capturing `currentEpoch()`/`databaseEpoch` *before* the await and re-checking *after*, before any write: `ensureNode`, `loadChildrenForParent` (restructured so both daemon reads precede a single check that gates parent + children + `structureTree` edges together), `doLoadChildrenTree`, `resyncNodeFromServer`, the two `tauri-sync-listener` paths, `browser-sync-service`, `navigation-service`, `node-card-inline`, `query-node-viewer`, `ai-chat-node-viewer`, and the app-shell MCP node-updated listener. No missed call site; no guard captures the epoch after the await.
- No TOCTOU: `databaseEpoch` is mutated only in synchronous `clearAll()`, and no `await` sits between any check and its `setNode`.
- Tests genuinely exercise the drop (would fail with the guard removed): store-level (`shared-node-store-database-epoch.test.ts`, 3) plus the community per-event and Pro coalescer domain-event paths (`tauri-sync-listener.test.ts`, +2). `isActiveDatabaseEvent(undefined)` returns `true`, so the test events reach the fetch rather than being dropped early.
- Targeted suites green (34 across the two files), eslint + svelte-check 0/0 (1738 files), full `cargo test -p nodespace-app` pre-push gate green.

Non-blocking follow-up (out of scope here): the app-shell MCP `NODE_CREATED` listener writes `event.payload.node` with no fetch and no active-database guard — an event-*delivery* race (not an in-flight *read*), pre-existing. Worth closing separately.

test:e2e passed; test:all failed only on the environmental `llama-cpp-sys-2` `'openssl/err.h' file not found` runner flake (frontend 151/151 passed) — re-running.

mstomar125:
**APPROVE** — this correctly closes the in-flight-read race I flagged on #1598 (tracked as #1599). The mechanism is right and the coverage is thoughtfully broad.

### The guard pattern is textbook-correct
An epoch/generation counter is the right tool for invalidating stale async writes, and it's implemented properly:
- `shared-node-store.svelte.ts`: `private databaseEpoch = 0`; `clearAll()` does `databaseEpoch++` (and `switchTo` already calls `clearAll()`); every fetch-then-write path does `const epoch = this.databaseEpoch` **before** the `await`, then `if (this.databaseEpoch !== epoch) return …` **after** it, before writing. Capture-before-await / check-after-await / drop-on-mismatch — exactly the ordering that closes the window. Same-epoch is a cheap integer compare, so the no-switch common case is a genuine no-op (matches the acceptance box). ✅
- **`loadChildrenForParent` restructure is a real correctness improvement, not just a guard:** both daemon reads complete, then a *single* staleness check gates the parent + children + `structureTree`-edge writes together. A per-write check would have risked partial application (parent dropped, children applied, or vice-versa) — gating them as a unit is correct. ✅

### Coverage is well-reasoned
The key insight is explicit in the PR: `isActiveDatabaseEvent` gates at event **arrival**, so it can't close a window that opens on the subsequent async fetch — the guard has to live at the store **write** boundary. From there the sweep is comprehensive: store internals (`ensureNode`, `loadChildrenForParent`, `doLoadChildrenTree`, `resyncNodeFromServer`), both hydration paths (`tauri-sync-listener` per-event `fetchAndUpdateNode` **and** the Pro reconnect-replay `flushPendingNodeFetches` whole-burst drop), the HTTP adapter (`browser-sync-service`, defensive), navigation link-resolution (aborts to `null` rather than opening a tab for a node absent from the new store), and the independently-fetching viewers (`node-card-inline`, `query-node-viewer`, `ai-chat-node-viewer`, `app-shell` MCP listener).

### Tests
Good: `shared-node-store-database-epoch.test.ts` proves a read resolving after the epoch advances is dropped and a same-epoch read is applied; `tauri-sync-listener.test.ts` adds drop-across-switch coverage for **both** the community per-event path and the Pro coalescer. 31/31 file green; eslint + svelte-check 0/0.

### Caveat (non-blocking)
Correctness of the *mechanism* is proven; **exhaustive** coverage rests on the enumeration being complete — i.e. that no other `await <daemon read>` → `setNode/batchSetNodes` path exists outside the touched files. I couldn't fully prove that from here (method-call code search is unreliable), but any missed path would be a narrow residual of an already-narrow, non-corrupting race (orphan rows are invisible in the tree, only surfacing in search until reload). The manual 2-database click-test already recommended for #1598 would also exercise this — worth doing switch-with-a-child-load-in-flight during that pass.

### Merge gate
CI `test:all` + `test:e2e` both green at `29392f79`. Closes #1599. Approve.

PR #1600, opened 2026-07-09
Remove vestigial SettingsService daemon-config RPCs (ADR-053)

mstomar125:
## Summary
Follow-on cleanup after ADR-053 removed `active_database_path` (#1595). The `SettingsService.GetDaemonConfig`/`UpdateDaemonConfig` RPCs and their sole remaining field `grpc_address` are vestigial — zero callers, and `grpc_address` is never consumed (the daemon binds its UDS via `socket_path()`/`NODESPACED_SOCKET`, not this field).

Deletes the two RPCs + their `GetDaemonConfigRequest`/`UpdateDaemonConfigRequest`/`DaemonConfigResponse` messages from the proto, and the matching handler impls, `config_to_response`, the now-dead `default_socket_path` helper, the `grpc_address` field on `DaemonConfig`, and the now-unused imports. Keeps `DaemonConfig.capture` and the live `GetCaptureSettings`/`UpdateCaptureSettings` RPCs.

Net: 2 files, +5 / −66.

## Validation
- Tree-wide grep (incl. the `nodespace-sync` repo) confirms zero callers of any removed symbol — only generated build artifacts referenced them.
- `cargo fmt --check` + `cargo clippy -p nodespace-daemon --all-targets -- -D warnings` clean.
- `cargo test -p nodespace-daemon` green (51+... 0 failed).
- `cargo check -p nodespace-cli` and `cargo check -p nodespace-app` (against the regenerated proto) both clean.

Closes #1596

PR #1602, opened 2026-07-10
Pin OPENSSL_ROOT_DIR in CI so the llama.cpp cold build finds OpenSSL

mstomar125:
## Summary

Kills the recurring `test:all` flake where the Rust build fails with:

```
llama.cpp/vendor/cpp-httplib/httplib.h:389:10: fatal error: 'openssl/err.h' file not found
```

## Root cause

`llama-cpp-sys-2` builds vendored `llama.cpp`, whose `CMakeLists.txt` sets `option(LLAMA_OPENSSL "..." ON)` by default. With that on, the vendored `cpp-httplib` compiles with `CPPHTTPLIB_OPENSSL_SUPPORT` and `#include <openssl/err.h>`. On the `macos-latest` runners, Homebrew's `openssl@3` is present but **not on CMake's default search path**, so `find_package(OpenSSL)` intermittently fails to locate it and the compile errors on the missing header.

It presents as *intermittent* because `swatinem/rust-cache` skips the compile entirely on a warm cache — the failure only surfaces on a **cold** cache (eviction, or a fork PR, which can't read the base repo's cache), which is exactly when a re-run "fixes" it. The crate's `build.rs` only forwards `CMAKE_*`-prefixed env vars as CMake defines, so `LLAMA_OPENSSL=OFF` can't be passed to disable it — the correct fix is to make OpenSSL discoverable.

## Fix

In both the `test:all` and `test:e2e` jobs, install `openssl@3` and pin `OPENSSL_ROOT_DIR` to its Homebrew prefix so `find_package(OpenSSL)` succeeds deterministically:

```yaml
brew install protobuf openssl@3
echo "OPENSSL_ROOT_DIR=$(brew --prefix openssl@3)" >> "$GITHUB_ENV"
```

No product code changes; runtime behavior is unaffected (the daemon does not use cpp-httplib — this only makes the vendored build compile reliably).

## Validation

This PR is raised from a fork, so its own CI runs on a **cold** cache and rebuilds `llama-cpp-sys-2` — a green `test:all` here directly exercises and confirms the fix (the very build that previously flaked).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,11 +39,17 @@ jobs:
       - name: Rust cache
         uses: swatinem/rust-cache@v2
 
-      - name: Install Protobuf
+      - name: Install Protobuf and OpenSSL
         run: |
           export HOMEBREW_NO_REQUIRE_TAP_TRUST=1
           brew untap aws/tap azure/bicep 2>/dev/null || true
-          brew install protobuf
+          brew install protobuf openssl@3
+          # llama.cpp's vendored cpp-httplib compiles with LLAMA_OPENSSL=ON, so a
+          # cold build needs OpenSSL headers. Homebrew's openssl@3 is not on CMake's
+          # default search path, so find_package(OpenSSL) intermittently misses and
+          # the build fails on <openssl/err.h>. Pin OPENSSL_ROOT_DIR so it is found
+          # deterministically on every runner.
+          echo "OPENSSL_ROOT_DIR=$(brew --prefix openssl@3)" >> "$GITHUB_ENV"
 
       - name: Setup Bun
         uses: oven-sh/setup-bun@v2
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE** — correct root cause, canonical fix, and it's self-validated by its own cold-cache run. This is the right permanent kill for the flake that's been costing re-runs on #1600/#1601.

### Root cause is precisely diagnosed
`llama-cpp-sys-2` vendors `llama.cpp`, whose `CMakeLists.txt` defaults `LLAMA_OPENSSL ON`, so the vendored `cpp-httplib` compiles with `CPPHTTPLIB_OPENSSL_SUPPORT` and `#include <openssl/err.h>`. Homebrew's `openssl@3` is installed on the runners but not on CMake's default search path, so `find_package(OpenSSL)` misses and the compile dies on the missing header. The "intermittent" behavior is explained correctly: `swatinem/rust-cache` skips the compile on a **warm** cache, so it only surfaces on a **cold** one — cache eviction or a **fork PR** (which can't read the base repo's cache). That matches exactly what I observed: the flake hit fork PRs (#1600, #1601) and "cleared" on re-run once artifacts warmed. And the note that `build.rs` only forwards `CMAKE_*` env → can't pass `LLAMA_OPENSSL=OFF`, so making OpenSSL discoverable is the correct lever, is accurate.

### The mechanism is canonical and correctly scoped
- `OPENSSL_ROOT_DIR` is precisely the hint CMake's `find_package(OpenSSL)` consults — pinning it to `$(brew --prefix openssl@3)` makes discovery deterministic, and computing it dynamically means it's right on both `/opt/homebrew` (ARM) and `/usr/local` (Intel) runners.
- Written via `>> "$GITHUB_ENV"` (not `export` in the run block), so it persists to the **subsequent** cargo build/test steps rather than dying with the step — correct.
- **No collateral:** `OPENSSL_ROOT_DIR` is a CMake variable; Rust's `openssl-sys` reads `OPENSSL_DIR` (a different var), so this doesn't perturb other OpenSSL linkage — it's narrowly confined to the vendored llama.cpp CMake build.

### Applied to both jobs, self-validating
The identical block lands in **both** `test:all` and `test:e2e` (both compile the native llama-cpp code, both need it). And because this PR is raised from a fork, its own CI runs on a **cold** cache and rebuilds `llama-cpp-sys-2` — the very build that flaked. Both checks are `COMPLETED/SUCCESS`, which is direct proof the fix works on the failing path, not just an assertion.

### Scope
`.github/workflows/ci.yml` only, +16/−4; no product code, runtime unaffected (the daemon doesn't use cpp-httplib). Community build behavior unchanged.

**Net:** correct diagnosis + minimal, deterministic, self-proven fix. This ends the re-run-until-green waste. Approve.

PR #1603, opened 2026-07-10
Scope daemon compute per database (ADR-053): active-first embedding priority, close + idle eviction

mstomar125:
## Summary

ADR-053 §1/§4 compute follow-through: with N databases open in one daemon, the store-bound background machinery is now scoped per database while sharing the single loaded embedding model.

Most per-database scoping already landed with the ADR-053 daemon core — each open database already has its own `EmbeddingProcessor` (draining its own store's stale queue), its own `LocalAgentServiceImpl` event watcher (subscribed to that database's event bus, filtered to `ai-chat`), its own `GraphContextAssembler`, and its own `AgentSessionHandler` over the shared `PtySessionManager`; the embedding model is process-global (published once over a `watch` channel, shared text-keyed LRU). This change adds the remaining lifecycle and priority behavior.

## What's new

- **Active-first embedding priority** — a process-global `EmbeddingScheduler` (in `nodespace-core`, keyed on the database ULID as a plain `String` so core stays free of any daemon type) that all per-database `EmbeddingProcessor`s share. Each processor acquires a permit around every batch; while the active database has batches queued, non-active databases defer at batch granularity, so a background bulk import into B cannot starve foreground semantic-search freshness in A. The active-database signal is the live `WatchNodes` stream — the desktop already opens exactly one on the database it displays and re-subscribes on switch, so no protocol change is needed. A non-active database is never permanently starved: it resumes the moment the active database goes idle, and with no active database (`None`) every database runs immediately.
- **Per-database close + task cancellation** — `DatabaseManager::close(id)` drops only that database's `EmbeddingProcessor` and cancels its agent event watcher (a new `CancellationToken` on `LocalAgentServiceImpl`). It never calls `release_gpu_context()`/`shutdown()` on the shared nlp-engine model — that is one-way and global, reserved for daemon shutdown.
- **Shutdown drains all open databases** — previously only the boot database's processor was drained; shutdown now iterates every open database (`shutdown_all`), then releases the shared GPU context exactly once.
- **Idle eviction** — a reaper closes non-default, non-active, idle (default 5 min, overridable via `NODESPACE_DB_IDLE_SECS`) local-only databases that have no pending embeddings, releasing their store connection; the next request reopens transparently via the existing lazy `get_or_open`.

## Acceptance criteria

- [x] Embedding backlogs drain independently per database; the active database's queue takes priority — `EmbeddingScheduler` + a unit test that distinguishes active-first from a plain FIFO gate (a non-active batch queued before a later active batch is still granted second).
- [x] Agent events and chat sessions in database A never leak into database B — per-database event watcher + session handler; covered by `per_db_compute` test.
- [x] Idle local-only databases close and release their store connection; next request reopens transparently — reaper + `close` + `get_or_open` rebuild; covered by tests.
- [x] One embedding model instance serves all databases — the model stays process-global; no second load.
- [x] `bun run test:all` green; `cargo fmt` + `cargo clippy --workspace -- -D warnings` clean.

## Non-goals (unchanged)

Per-database fairness beyond simple active-first; sync residency pinning (the sync issue pins the active synced database). Community single-database behavior is unchanged — one open database, always effectively active, never evicted.

## Review

Independently reviewed (adversarial pass over the scheduler concurrency, GPU-teardown safety, eviction guards, and watcher cancellation) — no deadlock, no lost-wakeup, no permanent starvation, teardown once-only-on-shutdown. Follow-up fixes from that review: strengthened the active-first unit test to genuinely distinguish priority from FIFO, RAII-guarded the active-waiter count so a cancelled acquire can't leak it, and made the reaper treat a not-yet-recorded database as freshly used rather than idle.

Closes #1534.

### DIFF
--- a/packages/cli/tests/cli_integration.rs
+++ b/packages/cli/tests/cli_integration.rs
@@ -56,7 +56,11 @@ async fn spawn_test_daemon() -> (PathBuf, oneshot::Sender<()>, TempDir) {
             .await
             .expect("failed to build NodeService"),
     );
-    let service = NodeServiceImpl::new(node_service, Arc::new(tokio::sync::RwLock::new(None)));
+    let service = NodeServiceImpl::new(
+        node_service,
+        Arc::new(tokio::sync::RwLock::new(None)),
+        Arc::new(nodespace_core::services::EmbeddingScheduler::new()),
+    );
 
     let listener = UnixListener::bind(&sock_path).expect("failed to bind test UDS socket");
     let incoming = UnixListenerStream::new(listener);
@@ -96,6 +100,7 @@ fn routing_test_context() -> SharedContext {
         pty_manager: Arc::new(PtySessionManager::new()),
         model,
         has_model: false,
+        scheduler: Arc::new(nodespace_core::services::EmbeddingScheduler::new()),
     }
 }
 
--- a/packages/core/src/services/embedding_processor.rs
+++ b/packages/core/src/services/embedding_processor.rs
@@ -96,21 +96,35 @@ pub struct EmbeddingScheduler {
 }
 
 /// A held scheduler slot. Dropping it releases the shared-model gate and, for an
-/// active-database batch, decrements the active-waiter count and wakes any
-/// deferred non-active waiters so priority is re-evaluated.
+/// active-database batch, drops the [`HighPriorityReservation`] which decrements
+/// the active-waiter count and wakes any deferred non-active waiters so priority
+/// is re-evaluated.
 pub struct SchedulerPermit<'a> {
     _permit: SemaphorePermit<'a>,
-    scheduler: &'a EmbeddingScheduler,
-    is_active: bool,
+    /// Present only for an active-database batch.
+    _reservation: Option<HighPriorityReservation<'a>>,
 }
 
-impl Drop for SchedulerPermit<'_> {
+/// RAII active-waiter reservation. Bumping `high_waiters` *before* the gate is
+/// acquired is what makes concurrent non-active batches defer; guarding it in a
+/// drop type ensures an `acquire` future cancelled while awaiting the gate can
+/// never leak the count (a leak would defer every non-active database forever).
+struct HighPriorityReservation<'a> {
+    scheduler: &'a EmbeddingScheduler,
+}
+
+impl<'a> HighPriorityReservation<'a> {
+    fn new(scheduler: &'a EmbeddingScheduler) -> Self {
+        scheduler.high_waiters.fetch_add(1, Ordering::SeqCst);
+        Self { scheduler }
+    }
+}
+
+impl Drop for HighPriorityReservation<'_> {
     fn drop(&mut self) {
-        if self.is_active {
-            self.scheduler.high_waiters.fetch_sub(1, Ordering::SeqCst);
-            // Wake deferred non-active waiters so they re-check the backlog.
-            self.scheduler.notify.notify_waiters();
-        }
+        self.scheduler.high_waiters.fetch_sub(1, Ordering::SeqCst);
+        // Wake deferred non-active waiters so they re-check the backlog.
+        self.scheduler.notify.notify_waiters();
     }
 }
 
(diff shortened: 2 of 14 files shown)


mstomar125:
**APPROVE**

Reviewed independently (my own read plus an adversarial pass focused on the new concurrency primitive) with special scrutiny on the four highest-risk areas:

- **`EmbeddingScheduler` concurrency** — traced line by line: no deadlock (the 1-permit gate is held only for one batch; the `active` mutex is never held across an await), no lost-wakeup (non-active waiters use the `Notified::enable()` register-before-check ordering), and no permanent starvation (an active batch drops its permit and `yield_now`s each batch, opening a real `high_waiters == 0` window for deferred non-active waiters). Critically, the **sole/community database never defers forever** — with no active database `high_waiters` stays 0, so a lone database's non-active path breaks out on the first check.
- **GPU teardown safety** — `release_gpu_context()` has exactly one call site (daemon shutdown, once, after `shutdown_all`); per-database `close`/`remove`/`evict`/`shutdown_all` only drop that database's processor and cancel its watcher, never touching the shared model.
- **Idle eviction guards** — the reaper skips the default, the scheduler-active, and any pending-embedding database, and collects candidates before releasing locks (never holds `open` across an await); evicted databases reopen transparently via `get_or_open`.
- **Watcher cancellation** — `close()` calls `shutdown()`, whose `CancellationToken` breaks the watcher's `select!` loop; the task's captured `Arc` is released on exit so it can't outlive the database.

Follow-up fixes applied from the review (commit `14a3ab01`): (1) the active-first unit test previously registered the active batch before the deferred one, so a plain FIFO gate would produce the same ordering — rewrote it so a non-active batch queues *before* a second active batch and asserted the active one is still granted first (genuinely distinguishes priority from FIFO); (2) RAII-guarded the `high_waiters` increment so an `acquire` future cancelled while awaiting the gate releases the count instead of leaking it; (3) the reaper now treats a not-yet-recorded database as freshly used, not idle.

Validated green locally before pushing: full `test:all` + `test:e2e` + `cargo test -p nodespace-app` via the pre-push gate; `cargo test -p nodespace-core` (1259) and `-p nodespace-daemon` (87, incl. the 5 `per_db_compute` acceptance tests); `cargo fmt` + `cargo clippy --workspace -- -D warnings` clean.

mstomar125:
**APPROVE** — I independently ground-verified the two hardest parts (scheduler concurrency + eviction guards); both are correct and defensively coded. This closes #1534 cleanly.

### EmbeddingScheduler — lost-wakeup-safe, leak-free, no permanent starvation
The non-active defer loop is the textbook tokio pattern and it's exactly right:
```rust
loop {
    let notified = self.notify.notified();
    tokio::pin!(notified);
    notified.as_mut().enable();                       // register BEFORE the check
    if self.high_waiters.load(SeqCst) == 0 { break; }
    notified.await;
}
```
`enable()` registers the waiter before reading `high_waiters`, so a `notify_waiters()` firing between the check and the await (which stores no permit) cannot be lost. Verified the surrounding invariants too:
- **`HighPriorityReservation` is genuine RAII** — `Drop` does `high_waiters.fetch_sub(1)` **then** `notify.notify_waiters()` (correct order: woken waiters observe the decremented count), and it's held across the active path's `gate.acquire().await`, so a cancelled active acquire releases the count instead of leaking it (the exact bug the adversarial pass fixed).
- **No permanent starvation** — when the active DB goes idle its last reservation drops → `high_waiters → 0` → notify → deferred DBs proceed. `set_active(None)`/switch routes the former-active DB through the non-active path, so `high_waiters` naturally drains; with `active = None` nobody takes the active path so every DB proceeds immediately. Active-first at batch granularity (permit dropped between batches) is the intended priority, and "fairness beyond active-first" is an explicit non-goal.

### Idle reaper — every dangerous eviction is guarded
`evict_idle_databases` excludes, in order: the **default** (`default_id != id`), the **active** (`scheduler.active_id() != id`), anything **not idle beyond the window**, a **just-opened DB with no recorded activity** (treated as freshly-used, so it can't be reaped between its open and first request — the other adversarial-pass fix), and any DB with **pending embeddings** (`has_pending_embeddings → continue`, so a mid-drain processor is never torn down). It collects candidates under the read lock and calls `close` **after** releasing it, so `open`/`registry` locks aren't held across the per-DB write awaits. And the open-request-races-eviction case is safe by construction: `close` drops only the map's `Arc`, so any in-flight caller holding an `Arc` to the services completes normally, and the next request lazily reopens via `get_or_open`. `close()` additionally refuses the default/active itself (defense-in-depth) and never calls the one-way `release_gpu_context()` on the shared model.

### Model/lifecycle scoping is correct
One process-global embedding model (published over `watch`, shared LRU) — no second load; `shutdown_all` now drains **every** open DB's processor (not just the boot DB) and then releases the GPU context exactly once; per-DB agent watcher cancellation via a fresh `CancellationToken`. Session/event isolation (A's chat events never reach B) is covered by `per_db_compute`.

### Community-unaffected
Single database is always effectively active and never evicted (default + active guards both apply to it) — behavior unchanged. ✅

### Merge gate
`test:e2e` COMPLETED/SUCCESS; `test:all` IN_PROGRESS at `14a3ab01` — let it finish green (the new #1602 OpenSSL pin should keep it deterministic). Nice touch that the adversarial pass strengthened the active-first unit test to actually distinguish priority from a plain FIFO gate. Approve.

Issue #1535, opened 2026-07-05
Per-database cloud sync: runtime tenant binding and single-active session lifecycle


malibio:
## Overview

Sync layer of ADR-053 §4–6 (`../nodespace-docs/decisions/053-one-daemon-multiple-databases.md`): each database binds independently to a cloud tenant, identity and sign-in are per database, and the daemon runs at most ONE live sync session — the active database's. Implementation lands mostly in the private nodespace-sync repo (`nodespaced-pro`); this issue tracks the core-side surface and coordinates the whole slice.

## Problem Statement

- The Pro tenant schema (`NODESPACED_PRO_SCHEMA`/`COLLECTION`) is baked at build time into the service file environment (`daemon_setup.rs`) — one build is hard-bound to one tenant, incompatible with per-database tenants and already contradicting the documented runtime provisioning flow.
- Sync/auth state is a single global session: one `CloudSyncService` task, one keychain identity, one global `proSync` frontend store fed by an unscoped `sync:status` stream.

## Proposed Solution

Tenant binding becomes runtime state recorded per database (DatabaseSettingsNode + registry mirror). The daemon holds one sync session with an activate/deactivate lifecycle. Billing follows tenant = database = billable unit.

### Tenant binding (runtime, per database)
- Remove build-time `NODESPACED_PRO_SCHEMA`/`COLLECTION` injection from `daemon_setup.rs`; deployment constants (Supabase URL, anon key) may stay build-time.
- Record the bound tenant schema in each database's DatabaseSettingsNode with a registry mirror for pre-open display. Bind/unbind RPCs on the Pro surface.
- Enforce the ADR-053 invariant at bind time: a tenant schema binds to at most one local database per machine.

### Single-active session lifecycle
- Activate (database becomes active): cursor-based catch-up (`sync_seq > cursor`), then real-time sync. Reuses the existing reconnect/reconciliation flow — an idle synced database is indistinguishable from an offline device.
- Deactivate (switch away): drain pending outbound pushes, then stop the session and Realtime channel.
- "Active" is daemon-level and persisted — the last-active database keeps syncing with the app window closed.
- Writes routed to a non-active database (e.g. CLI) do not push until that database activates; decide during implementation whether sustained CLI activity should count as activation.

### Identity and sign-in (per database)
- Sign-in runs in the context of a database; different databases may use different provider accounts. Keychain entries per `(tenant_schema, user_id)` — the existing key layout.
[...]

PR #1605, opened 2026-07-10
Drop build-time tenant baking from the Pro daemon plist (ADR-053 per-database cloud sync, S2 core)

mstomar125:
## Summary

S2 (core half) of ADR-053 per-database cloud sync (#1535): stop baking the **tenant** (schema + collection) into the Pro daemon's launchd plist at build time. Only the deployment-wide Supabase **endpoint** — project URL + publishable anon key — stays build-time; the tenant a database syncs to is bound per database at runtime and is no longer an install-time constant.

- Removed the `PRO_SCHEMA` / `PRO_COLLECTION` `option_env!` constants (`NODESPACE_PRO_SCHEMA` / `NODESPACE_PRO_COLLECTION` build-time env).
- Removed the `NODESPACED_PRO_SCHEMA` / `NODESPACED_PRO_COLLECTION` keys from the Pro `EnvironmentVariables` block the app writes into `~/Library/LaunchAgents`. `NODESPACED_PRO_SUPABASE_URL` + `NODESPACED_PRO_ANON_KEY` remain (deployment-wide, not tenant-specific).

The tenant model for this epic is a Postgres **schema in one shared Supabase project**, so the URL + anon key are legitimately deployment-wide and correct to keep build-time; only schema/collection are per-tenant. With this change a Pro install no longer carries a single baked tenant — the daemon takes its tenant from its runtime configuration (the per-database bound tenant added in S1, wired into the daemon's session path in a later slice).

## Community safety

Fully inert for the community build: these env keys were only ever emitted when `is_pro_build()` (i.e. a Pro build with `NODESPACE_PRO_SUPABASE_URL` baked in). The community `nodespaced` daemon never received them. No behavior change to the community edition.

## Tests

- `cargo fmt -p nodespace-app -- --check` clean; `cargo clippy -p nodespace-app --all-targets -- -D warnings` clean; `cargo test -p nodespace-app` green (no test asserts on the plist tenant keys — the plist writer has no snapshot coverage, and no other code referenced `PRO_SCHEMA` / `PRO_COLLECTION`).
- Change is isolated to the Pro plist generation in `daemon_setup.rs` — no core-lib, frontend, or e2e surface touched.

The matching sync-repo change (remove the `.pkg` plist template placeholders + the `build-pro-pkg.sh` / `build-pro-dmg.sh` tenant substitutions) lands as a follow-up PR in nodespace-sync and must merge after this one, so a Pro build never emits an empty schema.

Part of #1535 (S2 of the S1–S8 decomposition on the issue).

### DIFF
--- a/packages/desktop-app/src-tauri/src/daemon_setup.rs
+++ b/packages/desktop-app/src-tauri/src/daemon_setup.rs
@@ -42,8 +42,6 @@ const SYSTEMD_SERVICE_NAME: &str = "nodespace.service";
 // project's PUBLISHABLE key (safe to bake into the binary).
 const PRO_SUPABASE_URL: Option<&str> = option_env!("NODESPACE_PRO_SUPABASE_URL");
 const PRO_ANON_KEY: Option<&str> = option_env!("NODESPACE_PRO_ANON_KEY");
-const PRO_SCHEMA: Option<&str> = option_env!("NODESPACE_PRO_SCHEMA");
-const PRO_COLLECTION: Option<&str> = option_env!("NODESPACE_PRO_COLLECTION");
 
 /// True when this binary was built as the Pro edition (cloud env baked in).
 fn is_pro_build() -> bool {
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE** — the diff is correct and community-inert. Two deploy-order flags on the epic (the code is fine; these are about *when things can ship together*).

### Verified
- **Community-safe:** the removed `NODESPACED_PRO_SCHEMA` / `NODESPACED_PRO_COLLECTION` plist keys (and their `PRO_SCHEMA`/`PRO_COLLECTION` `option_env!` constants + substitutions) live entirely inside the `let pro_env = if is_pro_build() { … }` block. Community builds (`is_pro_build() == false`) never emitted them, so removal is a true no-op for the community edition. ✅
- **Correct split:** the deployment-wide endpoint (`NODESPACED_PRO_SUPABASE_URL` + `NODESPACED_PRO_ANON_KEY`) legitimately stays build-time; only the per-tenant schema/collection moves to runtime. The updated comment states the ADR-053 rationale accurately. +7/−10, one file, CI green (`test:all` + `test:e2e`).

### ⚠️ Flag 1 — cross-repo lock-step (paired sync S2 not open yet)
The PR states the matching **nodespace-sync** change (remove the `.pkg` plist template placeholders + the `build-pro-pkg.sh` / `build-pro-dmg.sh` tenant substitutions) is a separate PR that **must merge after this one**. I checked: that sync PR is **not open yet** (branch `feature/adr053-s2-drop-buildtime-tenant` exists locally). The stated order (core first, then sync) is right — it ensures no path ever substitutes an *empty* schema string. Tracking: S2 isn't complete until that sync PR lands; until then the `.pkg` build path still bakes a tenant while the app-written plist no longer does.

### ⚠️ Flag 2 — half-wired window (the one to watch)
This PR removes the build-time tenant, but its own text says the daemon "takes its tenant from its runtime configuration (the per-database bound tenant added in S1, **wired into the daemon's session path in a later slice**)." So the sequence is: **S1 stores** `bound_tenant` (inert) → **S2 removes** the build-time tenant → **a later slice wires** the daemon's sync session to read `bound_tenant`. Between S2 and that wiring slice, a Pro daemon has **no tenant source at all** — build-time removed, runtime not yet consumed. 

**Consequence:** a Pro `.pkg`/DMG built from a main commit in that window would ship a daemon that can't resolve which tenant to sync to. Since release CI is disabled and Pro builds are cut manually, this is a transient main-branch state, not an automatic shipping risk — but **no Pro build should be cut until the runtime-wiring slice lands**, else that build's sync is dead. Worth stating explicitly on the epic so it isn't missed when someone next builds a Pro installer.

### Reminder — still inside the user-gated #1535
S2 is the first slice that actually changes the **Pro tenant mechanism** (build-time → runtime). It doesn't yet touch keychain / per-DB OAuth / billing, but it's on the path there. Per the batch plan's carve-out, keep the user + Anoop in the loop before the cross-repo behavioral wiring (and the sync-side tenant/keychain slices) ship.

**Net:** approve the core change; land the paired sync S2 after it; hold Pro installer builds until the runtime-tenant-wiring slice closes the S2→wiring gap.

PR #1606, opened 2026-07-10
Read/write API for the per-database bound cloud tenant (ADR-053 per-database cloud sync, S3 core)

mstomar125:
## Summary

Core read/write API for the per-database bound cloud tenant (ADR-053 per-database cloud sync), consumed by the Pro sync daemon in a paired nodespace-sync change. S1 added the `bound_tenant_schema` / `bound_tenant_collection` fields on the `DatabaseSettingsNode` singleton; this adds the two `NodeService` methods to read and write them so the daemon can resolve its tenant at runtime instead of from a build-time constant.

- `get_bound_tenant() -> Option<(schema, collection)>` — reads the singleton via the existing id-free `query_nodes_by_type("database-settings", …)` path and the `database-settings` property namespace. Returns `Some` only when **both** fields are present and non-empty; `None` means the database is unbound (fresh install, or tenant never set) — callers treat that as "no cloud sync target yet".
- `set_bound_tenant(schema, collection)` — merges both fields into the existing `database-settings` namespace (sibling fields like sync state / auth status preserved) and persists via the version-checked `update_node`. A missing singleton is an error, not a silent no-op (it is seeded on database open).

## Community safety

Purely additive — two new `pub` methods. Nothing in the community build calls the setter, and the getter returns `None` on an unbound database (the only state a community install is ever in). No behavior change to the community edition.

## Tests

- `test_get_set_bound_tenant_roundtrip`: fresh install is unbound (`None`); bind → `get` returns exactly what was set; sibling `sync_enabled` / `auth_status` survive the merge; re-bind overwrites; empty schema/collection reads back as unbound.
- `cargo test --lib -p nodespace-core -- --test-threads=2` green; `cargo clippy -p nodespace-core --lib -- -D warnings` clean; `cargo fmt --check` clean.

Part of the S1–S8 decomposition of ADR-053 per-database cloud sync (tracked on #1535). The paired nodespace-sync change wires the Pro daemon to resolve its tenant from `get_bound_tenant()` at boot (seeding from the CLI/env tenant on first run), replacing the build-time tenant removed in S2.

### DIFF
--- a/packages/core/src/services/node_service/mod.rs
+++ b/packages/core/src/services/node_service/mod.rs
@@ -1011,6 +1011,90 @@ impl NodeService {
         Ok(())
     }
 
+    /// Read the cloud tenant (schema + collection) this database is bound to, from
+    /// the DatabaseSettingsNode singleton (ADR-053 per-database cloud sync). Returns
+    /// `Some((schema, collection))` only when both are present and non-empty; `None`
+    /// when the database is unbound — a fresh install, or one whose tenant has never
+    /// been set. Callers treat `None` as "no cloud sync target yet".
+    pub async fn get_bound_tenant(&self) -> Result<Option<(String, String)>, NodeServiceError> {
+        let Some(node) = self
+            .query_nodes_by_type("database-settings", None)
+            .await?
+            .into_iter()
+            .next()
+        else {
+            return Ok(None);
+        };
+        let settings = node.properties.get("database-settings");
+        let schema = settings
+            .and_then(|s| s.get("bound_tenant_schema"))
+            .and_then(|v| v.as_str())
+            .unwrap_or_default();
+        let collection = settings
+            .and_then(|s| s.get("bound_tenant_collection"))
+            .and_then(|v| v.as_str())
+            .unwrap_or_default();
+        if schema.is_empty() || collection.is_empty() {
+            return Ok(None);
+        }
+        Ok(Some((schema.to_string(), collection.to_string())))
+    }
+
+    /// Bind this database to a cloud tenant (schema + collection) by writing the
+    /// authoritative fields on the DatabaseSettingsNode singleton (ADR-053). Merges
+    /// into the existing `database-settings` namespace so sibling fields (sync state,
+    /// auth status) are preserved. Once set, `get_bound_tenant` returns these values
+    /// until re-bound. The singleton is seeded on database open, so a missing one is
+    /// an error rather than a silent no-op.
+    pub async fn set_bound_tenant(
+        &self,
+        schema: &str,
+        collection: &str,
+    ) -> Result<(), NodeServiceError> {
+        let node = self
+            .query_nodes_by_type("database-settings", None)
+            .await?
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE** (self-review to unblock the paired nodespace-sync e2e; both CI checks green).

Reviewed critically:

- **`get_bound_tenant`** reads the singleton via the id-free `query_nodes_by_type("database-settings", …)` (the same path the seed/idempotency code uses) and the `database-settings` property namespace — matching how the seed persists `sync_enabled`/`auth_status`. Returns `Some` only when both fields are present and non-empty; a malformed/absent `database-settings` object yields `None` via `unwrap_or_default()`. Correct.
- **`set_bound_tenant`** clones properties, `entry("database-settings").or_insert(json!({}))` then inserts both keys — so sibling fields (sync state, auth status) survive (merge, not overwrite), and a missing namespace is created rather than panicking. Persists via the version-checked `update_node(id, node.version, …)`; a missing singleton is an explicit error. Correct.
- **Validation interaction:** the `DatabaseSettingsNode` behavior `validate` reads these keys at the *top* level while they persist *namespaced* (a pre-existing quirk shared by `sync_enabled`) — so the write isn't rejected; values are always `String`, so even if validated they'd pass. The green roundtrip test confirms `update_node` accepts the write.
- **Community-safe:** two additive `pub` methods; the community build never calls the setter, and the getter returns `None` on the only state a community install is in (unbound). No behavior change.
- **Test** covers unbound→`None`, bind→roundtrip, sibling preservation, re-bind overwrite, and empty→unbound. `cargo test --lib -p nodespace-core -- --test-threads=2` green (989 passed), clippy/fmt clean.

Net: correct, additive, community-inert. Merging to unblock the sync-side daemon boot read + its live validation.

mstomar125:
**APPROVE** (post-merge — opened+merged between watcher polls). Clean additive read/write API; correct semantics.

- **`get_bound_tenant`** reads the `database-settings` singleton via the id-free `query_nodes_by_type` path and returns `Some((schema, collection))` **only when both are non-empty** (`if schema.is_empty() || collection.is_empty() { return Ok(None) }`). That's the right call — a half-set binding reads as unbound, so the daemon never targets a partial tenant. `None` = unbound (fresh install / never set).
- **`set_bound_tenant`** clones properties, takes the `database-settings` namespace via `entry(...)` (so sibling fields — sync state, auth status — are preserved, not clobbered), and persists through the **version-checked `update_node(node.version, …)`** (OCC). A missing singleton is an `InitializationError`, not a silent no-op — correct, since the singleton is seeded on database open.
- **Community-safe:** two additive `pub` methods; nothing in the community build calls the setter, and the getter returns `None` for the only state a community install is ever in. No behavior change.
- Test covers roundtrip + sibling preservation (`sync_enabled`/`auth_status` survive) + re-bind overwrite + empty→unbound. CI green (`test:all` + `test:e2e`). +126/−0.

Approve. The paired nodespace-sync#296 consumes this to resolve the daemon's tenant at boot.

Issue #1537, opened 2026-07-05
Registry-driven Pro UI surfaces resolved from daemon capability and DatabaseSettingsNode state


malibio:
## Overview

Pro-edition GUI surfaces (sync pill, re-login modal, collaboration tab) are hardcoded into shared components behind scattered `{#if proSync.isPro}` guards. Replace this with a small UI-extension registry whose contributions resolve their rendered variant from two signals: the daemon build-variant capability (app-global) and the active database's `DatabaseSettingsNode` (node-held, per ADR-053). Shared components import nothing Pro; community builds render upsell teasers in the same slots.

## Problem Statement

- `app-shell.svelte` imports `ProSyncPill`/`ProReloginModal` directly and owns Pro-specific `$effect` orchestration (relogin state, recovered-items load-once).
- `collection-node-viewer.svelte` hardcodes the collaboration tab behind `proSync.isPro` (lines ~185, ~206).
- The single boolean `proSync.isPro` conflates two independent questions that ADR-053 separates: "can this daemon sync?" (build variant, app-global, static per session) and "is sync enabled/authenticated for the active database?" (held on that database's `DatabaseSettingsNode`, changes on database switch or settings edit). A Pro daemon with a non-synced active database currently cannot be represented.
- Community edition hides Pro surfaces entirely instead of showcasing upgrade paths (ADR-039 calls for upgrade CTAs in place of sync controls but provides no mechanism).

## Proposed Solution

A sibling registry to the existing node-type `PluginRegistry` — do NOT extend `PluginDefinition` (different keys, lifecycles, and consumers). One `UiExtensionDefinition` for Pro sync with a `resolveVariant()` state machine over the two signals; the registry exposes contributions through a `$derived`-friendly reactive wrapper.

### Variant state machine

| Condition | Variant | Component |
|---|---|---|
| `proSync.tier !== 'pro'` | `teaser` | `pro-teaser-pill.svelte` (new) — upgrade CTA |
| Pro + `sync_enabled: false` | `enable-prompt` | `enable-sync-pill.svelte` (new) — enable sync for this database |
| Pro + enabled + `auth_status: 'local'` | `sign-in` | sign-in pill state |
| Pro + enabled + `auth_status: 'connected'` | `connected` | existing `pro-sync-pill.svelte`, unchanged internally |

[...]

PR #1607, opened 2026-07-10
Add UI-extension registry for two-axis Pro-sync surfaces (ADR-053)

mstomar125:
## Summary

Adds a declarative UI-extension registry that resolves the active Pro-sync surface from two independent signals and renders it through generic extension slots, replacing the hard-coded pill/modal wiring in the app shell.

Part of the ADR-053 multi-database epic (#1537): the Pro-sync surface must react per active database, not just per daemon build.

## Two-axis variant resolution

- **Axis 1 — build tier** (`proSync.tier`, app-global): a non-Pro daemon resolves to the static upgrade `teaser`.
- **Axis 2 — active database** (`DatabaseSettingsNode.sync_enabled` / `auth_status`, per-database): a Pro daemon resolves to `enable-prompt` → `sign-in` → `connected`.

The resolver (`lib/plugins/ui-extensions.svelte.ts`) maps the two signals to one variant and returns only the chrome contributions / viewer extensions matching it. Both reads are reactive, so a tier change or a `DatabaseSettingsNode` update (including a database switch) re-resolves the surface.

## What changed

- New `UiExtensionRegistry` (`lib/plugins/ui-extensions.ts`) + reactive resolver + built-in Pro extension definition (`pro-plugin.ts`) + `ExtensionOutlet`.
- App-shell overlay/modal slots and the collection-node viewer render registry contributions via `ExtensionOutlet` instead of importing pill/modal components directly.
- Re-login orchestration moves out of the app shell into a registry slot component (`pro-relogin-slot.svelte`); the presentational `ProReloginModal` is unchanged.
- Dependent stores (`membership`, `recovered-items`) gate on the two-axis `isProSyncActive()` instead of the tier-only `proSync.isPro`.
- `database.svelte.ts` hydrates the `DatabaseSettingsNode` on load and after a database switch so the variant re-resolves for the active database.

## Community build unchanged

Non-Pro builds resolve to the static teaser only; every daemon-backed Pro surface stays gated. The `test:free-users` guardrail covers this.

## Testing

- `bun run test` — 3989 passed
- `bun run test:free-users` — 11 passed
- `bun run quality:check` (eslint + svelte-check) — 0 errors, 0 warnings
- Full pre-push pyramid (frontend + core-lib + headless e2e + Tauri-seam) green

## Follow-up

- Backend gap: there is no per-database "enable sync" intent command yet, so `enable-sync-pill` opts a database into sync via the existing interactive sign-in (`pro_initiate_oauth`), which flips `sync_enabled` / `auth_status` on bind. A dedicated set-`sync_enabled` command is a backend follow-up.
- Live desktop UI validation of the variant transitions is pending.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/packages/desktop-app/src/lib/components/collaboration/collaboration-locked.svelte
+++ b/packages/desktop-app/src/lib/components/collaboration/collaboration-locked.svelte
@@ -1,0 +1,53 @@
+<!--
+  Collaboration locked placeholder.
+
+  Shown in the collection viewer's Collaboration tab for the `enable-prompt`
+  variant — a Pro daemon whose active database has `sync_enabled: false`. People &
+  roles live in the cloud tenant a database is bound to, so there is nothing to
+  show until sync is enabled. Purely static: no membership calls, no daemon
+  commands. Accepts the host collection's `nodeId` for a uniform viewer-extension
+  signature (surfaced as a data attribute for debugging).
+-->
+<script lang="ts">
+  let { nodeId }: { nodeId: string } = $props();
+</script>
+
+<div class="collab-locked" data-collection-id={nodeId}>
+  <div class="lock" aria-hidden="true">🔒</div>
+  <p class="headline">Collaboration is off for this database</p>
+  <p class="detail">
+    Enable cloud sync for this database to invite people, manage roles, and share
+    collections.
+  </p>
+</div>
+
+<style>
+  .collab-locked {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    gap: 8px;
+    padding: 48px 24px;
+    text-align: center;
+    color: hsl(var(--muted-foreground));
+  }
+
+  .lock {
+    font-size: 28px;
+    opacity: 0.8;
+  }
+
+  .headline {
+    margin: 0;
+    font-size: 15px;
+    font-weight: 600;
+    color: hsl(var(--foreground));
+  }
+
+  .detail {
+    margin: 0;
+    max-width: 360px;
+    font-size: 13px;
+    line-height: 1.5;
+  }
+</style>
--- a/packages/desktop-app/src/lib/components/collaboration/collaboration-tab.svelte
+++ b/packages/desktop-app/src/lib/components/collaboration/collaboration-tab.svelte
@@ -1,0 +1,15 @@
+<!--
+  Collaboration tab.
+
+  Registry wrapper that mounts the unchanged `CollaborationView` for the
+  sync-enabled variants (`sign-in` / `connected`). Exists only to move the view's
+  mounting behind the UI-extension registry — `CollaborationView` keeps its own
+  internal `proSync.tier` gate and all its logic.
+-->
+<script lang="ts">
+  import CollaborationView from '$lib/components/collaboration/collaboration-view.svelte';
+
+  let { nodeId }: { nodeId: string } = $props();
+</script>
+
+<CollaborationView collectionId={nodeId} />
--- a/packages/desktop-app/src/lib/components/enable-sync-pill.svelte
+++ b/packages/desktop-app/src/lib/components/enable-sync-pill.svelte
@@ -1,0 +1,84 @@
+<!--
+  Enable-sync prompt.
+
+  Rendered in the app-shell overlay slot for the `enable-prompt` variant — a Pro
+  daemon whose active database has `sync_enabled: false`. Offers to turn cloud sync
+  on for this database.
+
+  The concrete "turn sync on" action available today is starting the interactive
+  sign-in (`pro_initiate_oauth`): a Pro user opts a database into sync by signing
+  in to the cloud, after which the daemon binds the tenant and flips
+  `sync_enabled` / `auth_status` on the DatabaseSettingsNode, advancing the variant.
+  (A dedicated per-database "enable sync" intent command is backend follow-up.)
+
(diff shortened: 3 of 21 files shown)


mstomar125:
**APPROVE** — well-structured, community-safe two-axis Pro-UI registry. The failing `test:e2e` is an infra flake unrelated to this PR (analysis below); re-run to green the gate.

### Community-safety verified at the source, not just the test
The one invariant that matters for a Pro-UI PR — a community build must never render a daemon-backed Pro surface — holds by construction. `resolveProSyncVariant()` gates on the tier **first** and short-circuits:
```ts
if (proSync.tier !== 'pro') return 'teaser';   // non-Pro (and pre-probe 'unknown') → teaser, unconditionally
const settings = activeDatabaseSettings();
if (settings?.sync_enabled !== true) return 'enable-prompt';
if (settings.auth_status === 'connected') return 'connected';
return 'sign-in';
```
So the per-database axis-2 (`sync_enabled`/`auth_status`) is never even read for a non-Pro build — no `DatabaseSettingsNode` value can flip a community install into a Pro surface. Backed by: every contribution is lazy `() => import(...)` (Pro components aren't pulled unless their variant is active), and the `free-user-guardrail` test asserts community resolves to `teaser` only, modal + collaboration slots contribute `[]`, and the pre-probe `unknown` tier is teaser-only too (no Pro flash). `test:free-users` 11/11, `test:all` (frontend 3989 + skill + Rust) green.

### The state machine + gating change are correct
- teaser (non-Pro) → enable-prompt (Pro, sync off) → sign-in (Pro, sync on, not connected) → connected — coherent, and the slot table (overlay/modal/collab-tab per variant) maps cleanly.
- `isProSyncActive()` = `sign-in || connected` (Pro **and** `sync_enabled`) now gates `membership`/`recovered-items`, replacing the tier-only `proSync.isPro`. That's a deliberate *tightening* — those features no longer light up on a Pro daemon whose active database hasn't enabled sync. Correct per the epic's per-database intent; just flagging it's a real behavior change (a Pro-but-sync-disabled DB now hides membership/recovered-items until enabled).
- `database.svelte.ts` re-hydrates the `DatabaseSettingsNode` on load + after a switch, so the surface re-resolves per active database. Consistent with #1598's switch flow.

### ⚠️ CI: `test:e2e` FAILURE is an infra flake, not this PR
All 3 failures are in `adapter-contract.e2e.ts` (HttpAdapter → dev-proxy → daemon round-trips), each with the identical root error: `connect ENOENT …/daemon.sock` — **the headless daemon never bound its socket** in the runner, so 3 daemon-dependent tests couldn't connect (21/24 passed). This PR is **frontend-only** (Svelte/TS UI-extension registry, stores, components) and touches no daemon/adapter/socket/Rust code, so it cannot have caused a daemon-startup failure. Consistent with the impl's "pre-push e2e green locally." **Re-run `test:e2e` to clear the merge gate** — no code change needed. (If the daemon-socket-ENOENT e2e flake recurs across PRs, it's a candidate for the same infra hardening the OpenSSL flake got in #1602.)

### Follow-up (documented, fine)
No per-database "enable sync" intent command yet, so `enable-sync-pill` opts in via the existing interactive sign-in (`pro_initiate_oauth`), which flips `sync_enabled`/`auth_status` on bind — a dedicated setter is a backend follow-up. Reasonable to defer.

**Net:** approve the code; the only thing between it and merge is a `test:e2e` re-run for the daemon-socket flake. (Core = I re-review on any new head SHA.)

mstomar125:
**APPROVE**

Reviewed as an independent correctness pass over the full diff (registry data class, reactive resolver, variant↔component↔slot mapping, app-shell rewire, per-database settings hydration, and the two-axis store gates).

Two issues surfaced and were fixed before this approval:

1. **(High) Recovered-items regression — gate/trigger signal mismatch.** The store's `load()` gate had been widened to the two-axis `isProSyncActive()`, but its only runtime trigger is the tier-only one-shot `proSync.onProConfirmed`. At boot, tier resolves (often synchronously) before the active database's `DatabaseSettingsNode` hydrates on its separate, strictly-longer chain — so the one-shot fired while the gate was still unsatisfied, `load()` returned empty, and the recovery log never loaded for the whole session. Fixed by reverting the gate to the tier-only `proSync.isPro` (the recovery log is per-user, not per-database, so the two-axis narrowing bought nothing but the race) and adding a pro-path regression test (`tests/stores/recovered-items.test.ts`).

2. **(Latent) Re-login dismissal lost on variant remount.** Dismissal state lived inside `pro-relogin-slot`, which the modal slot mounts via `{#each … (c.variant)}`; a `sign-in`↔`connected` flip (both map to the same slot) remounts it and reset the dismissed-episode to `-1`, re-arming an already-dismissed modal. This is state the shell used to hold outside the remounting subtree. Fixed by hoisting it to `proSync.dismissedReloginEpisode` so dismissal survives the remount.

What checked out: complete/correct variant→component→slot mapping (all four variants; app-shell requests exactly the two slots pro-plugin registers); reactive `SharedNodeStore` reads (not cached snapshots); correct re-hydration after `clearAll()` on switch (epoch guard captured post-bump); no `$derived`/`$effect` loops; membership gate sense preserved.

Verification after fixes: `bun run test` 3991 passed (156 files) · `test:free-users` 11 passed · `quality:check` 0 errors / 0 warnings · full pre-push pyramid (frontend + core-lib + e2e + Tauri-seam) green.

PR #1609, opened 2026-07-10
Delete the dead SurrealDB DDL generator (SchemaTableManager)

malibio:
Closes #1496


malibio:
### Code Review Summary

Clean, well-scoped deletion. This is a textbook "delete dead weight" PR: `SchemaTableManager` generated SurrealDB DDL (`DEFINE TABLE ... SCHEMAFULL TYPE RELATION`) that was computed, threaded through `create_schema_node_atomic`/`update_schema_node_atomic` as `_ddl_statements`, and then silently discarded by the SQLite store (`self.create_node(node, source, None).await` — the DDL param was never used). Removing it is a strict net positive: ~1250 lines gone, zero behavior change, one fewer drift hazard between a `#[cfg(test)]` DDL re-implementation and the real generator.

All 4 acceptance criteria from #1496 are met:
- `SchemaTableManager` + its 320-line integration test file are fully deleted
- `_ddl_statements` params and the `if node.node_type == "schema" { ... } else { ... }` DDL branches in `crud.rs` and `node_service/mod.rs` (`seed_core_schemas_if_needed`) are gone, collapsed to a single unconditional `create_node`/`update_node` call
- Schema-node creation/update now goes through the plain `NodeService` → `SqliteStore::create_node`/`update_node` path, same as any other node type
- `cargo clippy -p nodespace-core --lib --tests -- -D warnings` and `cargo build -p nodespace-core --lib` are clean (spot-checked independently)

**Verification performed beyond the stated test run:**
- Repo-wide grep for `SchemaTableManager`, `schema_table_manager`, `create_schema_node_atomic`, `update_schema_node_atomic`, `generate_relationship_ddl_statements`, `ddl_statements` — zero hits outside git history. No dangling references, no orphaned imports/`pub use`.
- The `else`-branch tracing/error-handling that survived the collapse in `crud.rs` (`NodeServiceError::query_failed` wrapping, `tracing::debug!` timing log) is exactly what schema nodes now get too — no regression, since the schema branch's own error handling (`NodeServiceError::query_failed(e.to_string())`) was strictly less informative than what remains.
- `client_id`/`execution_context` threading verified intact: create path still passes `self.client_id.clone(), self.execution_context.clone()` into `store.create_node`; update path still passes `self.client_id.clone()` into `store.update_node`. Since the schema branch previously discarded `execution_context` entirely (`create_schema_node_atomic` didn't accept it), schema-node creation actually gains correct execution-context propagation as a side effect of the collapse — a small correctness improvement, not just a wash.
- **`compute_edge_table_name`/`edge_table` retention judgment call: confirmed correct.** `relationship_cache.rs:218` has a live call site (`edge_table: relationship.compute_edge_table_name(source_type)`) building `InboundRelationship` entries for the in-memory reverse-relationship index — this is unrelated to DDL/table creation, it's just a naming convention borrowed for cache-key construction. Issue #1496 did not list this for removal ("Deliberately KEPT" was the right call — removing it would have broken `relationship_cache.rs` and was never required).
- `_ddl_statements` params removed cleanly from `sqlite_store/nodes.rs`; the two `*_atomic` wrapper methods that only forwarded to `create_node`/`update_node` are gone rather than kept as pointless indirection — correct, since nothing else called them (confirmed via grep).
- Doc comments in `lib.rs` and `services/mod.rs` updated to drop `SchemaTableManager` mentions; `pub mod`/`pub use` removed.
- Remaining "DDL" string hits in `sqlite_store/mod.rs` are unrelated (SQLite's own idempotent embedding-column migration) — not in scope and correctly untouched.

**Test coverage:** No new tests added, and that's the right call here. This is a pure branch-collapse of dead/no-op code — the SQLite behavior for schema-node create/update is unchanged (it always ignored the DDL). Existing integration coverage (`skill_updater_test.rs`, `schema_hydration_test.rs`) already exercises schema-node create/update through `NodeService`, and that now runs through the same path as regular nodes. Adding tests would just be re-testing generic `create_node`/`update_node`, which is YAGNI here.

### Findings

#### Critical Issues
None.

#### Suggested Improvements
None — this is as clean as a deletion PR gets.

#### Nitpicks
None worth noting; the diff is tight and the judgment calls (keeping `compute_edge_table_name`, deleting the atomic wrapper methods entirely rather than aliasing) are both correct.

### Recommendation
**Approve.** Net positive with no regressions found: acceptance criteria fully met, no dead code left behind, no dangling references, and the create/update collapse is behavior-preserving (plus a minor correctness gain in execution_context propagation for schema nodes). Safe to merge.

malibio (commented):
Pragmatic code review: RECOMMEND APPROVE (self-approval blocked by GitHub — see full findings in the PR comment above). Net positive dead-code deletion, all acceptance criteria met, no dangling references, create/update collapse is behavior-preserving with a minor correctness gain (execution_context now propagates for schema-node creation).

PR #1614, opened 2026-07-10
Remove the dead lifecycle_status=deleted soft-delete filters

malibio:
Closes #1505

### DIFF
--- a/packages/core/src/agent_params.rs
+++ b/packages/core/src/agent_params.rs
@@ -58,8 +58,7 @@ pub struct SearchSemanticParams {
     pub include_markdown: Option<usize>,
 
     /// Include archived nodes in search results (default: false).
-    /// By default, search only returns active nodes. Nodes with
-    /// lifecycle_status = "deleted" are never included.
+    /// By default, search only returns active nodes.
     #[serde(default)]
     pub include_archived: Option<bool>,
 
--- a/packages/core/src/db/sqlite_store/embeddings.rs
+++ b/packages/core/src/db/sqlite_store/embeddings.rs
@@ -562,7 +562,7 @@ impl SqliteStore {
             .join(" OR ");
 
         let sql = format!(
-            "SELECT n.id FROM node n JOIN node_fts f ON f.id = n.id WHERE node_fts MATCH ?1 AND n.lifecycle_status != 'deleted' ORDER BY rank LIMIT {}",
+            "SELECT n.id FROM node n JOIN node_fts f ON f.id = n.id WHERE node_fts MATCH ?1 ORDER BY rank LIMIT {}",
             candidate_limit
         );
 
@@ -599,18 +599,13 @@ impl SqliteStore {
 
             if let Some(row) = rows.next().await? {
                 let root_id: String = row.get(0)?;
-                // Verify root is not deleted
-                if let Ok(Some(n)) = self.get_node(&root_id).await {
-                    if n.lifecycle_status != "deleted" {
-                        root_ids.insert(root_id);
-                    }
+                if self.get_node(&root_id).await.is_ok_and(|n| n.is_some()) {
+                    root_ids.insert(root_id);
                 }
             } else {
                 // node_id is itself the root
-                if let Ok(Some(n)) = self.get_node(&node_id).await {
-                    if n.lifecycle_status != "deleted" {
-                        root_ids.insert(node_id);
-                    }
+                if self.get_node(&node_id).await.is_ok_and(|n| n.is_some()) {
+                    root_ids.insert(node_id);
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


malibio:
### Code Review Summary

Clean, well-scoped cleanup that removes three unreachable `lifecycle_status == "deleted"` / `!= "deleted"` predicates. I independently verified the core premise across the whole workspace (Rust + TypeScript): local deletion is a hard `DELETE FROM node`, no writer anywhere sets `lifecycle_status = "deleted"`, and a DB trigger (`node_fts_delete`) removes the FTS shadow row on hard delete — so the BM25 query's `JOIN node_fts` and the "root still exists" checks could never observe a `"deleted"` node in the first place. The change is a net positive: it deletes dead code and an inaccurate doc comment without altering observable behavior. Diff is atomic (three files, all part of the same removal), small, and self-verifying by inspection.

One structural gap is worth flagging (non-blocking): `lifecycle_status` is a free-form, unvalidated `String` throughout the write path (`NodeUpdate.lifecycle_status: Option<String>`, playbook `update_node` action reads it straight from JSON params, `ops/node_ops.rs` passes `input.lifecycle_status` through unchecked). So "deleted" being unreachable today is an *empirical* fact about current callers, not a *structural* guarantee — nothing stops a future playbook action or API caller from writing `lifecycle_status: "deleted"` tomorrow, at which point the newly-permissive code in this PR would silently start showing "deleted" nodes in FTS/BM25 and semantic search. That's consistent with the issue's own framing (dead code now; revisit if the lifecycle is ever actually implemented), so I'm not blocking on it, but it's the one thing I'd want a follow-up issue for if this repo continues to allow arbitrary lifecycle_status strings.

### Requirements Check (issue #1505)

- [x] Dead `lifecycle_status = "deleted"` filters removed — all three call sites from the issue's reference list (`sqlite_store` filters, `search_ops.rs` branch) are removed; equivalent embeddings.rs guards simplified to plain existence checks.
- [x] Code passes `bun run quality:fix`; tests pass — `cargo build -p nodespace-core --lib` compiles clean; no lint suppressions added. (Full `test:all` / `quality:fix` run not independently re-executed in this review pass, see note below.)
- [x] Atomic change — touches exactly the three files implicated by the dead-code claim, nothing unrelated bundled in.

### Findings

#### Critical Issues
None.

#### Suggested Improvements
- **packages/core/src/ops/node_ops.rs:174-183, packages/core/src/playbook/actions.rs:629-631**: `lifecycle_status` remains an unvalidated `String` end-to-end (no enum, no allow-list check before the `UPDATE node SET lifecycle_status = ?1` in `nodes.rs:256/373`). This PR is correct that nothing *currently* writes `"deleted"`, but the removed filters were arguably load-bearing as defense-in-depth against exactly this kind of latent risk. Since the project's stated policy is hard-delete-only with no soft-delete lifecycle, consider a small follow-up to constrain `lifecycle_status` writes to `{"active", "archived"}` (e.g., validate in `NodeService::update_node` or convert to an enum) so the invariant this PR is now relying on is enforced rather than just currently true. Not a blocker for this PR — it's a pre-existing gap this PR merely stops papering over.

#### Nitpicks
- **Nit:** packages/core/src/db/sqlite_store/embeddings.rs:602-609: the two branches are now identical except for which variable (`root_id` vs `node_id`) is checked/inserted — could be a tiny helper (`async fn insert_if_exists(&self, id: &str, set: &mut HashSet<String>)`) to remove the duplication, but the current form is only 4 lines each and clearly readable, so this is optional polish only.
- **Nit:** packages/core/src/agent_params.rs:58-60: doc comment update is accurate and minimal — no notes.

### Recommendation: APPROVE

The change does exactly what the issue asked, is verified correct against the actual write paths and DB trigger behavior, is atomic, and improves code health by removing misleading dead code. The unvalidated-string gap noted above is pre-existing and out of scope for this PR; recommend opening a fast-follow issue rather than blocking here.

malibio:
## Review Response

The pragmatic code review returned **APPROVE** with 0 Critical and 0 Important findings — only two optional Suggestions/Nitpicks, both explicitly flagged by the reviewer as non-blocking and out of scope for this PR.

- ⏭️ **Skipped: constrain `lifecycle_status` writes to `{"active", "archived"}`** — reviewer explicitly called this a pre-existing gap unrelated to this PR's scope, recommending "a fast-follow issue rather than blocking here." Filed as #1615.
- ⏭️ **Skipped: extract `insert_if_exists` helper in `embeddings.rs`** — reviewer called this "optional polish only," noting the current 4-line duplicated branches are "clearly readable." Not worth the churn in a minimal cleanup PR.

No code changes made — the diff is unchanged from what was reviewed.

## Re-Review Decision

**Rationale**: No code was changed in response to review feedback; both suggestions were deferred (one to a follow-up issue, one skipped as unnecessary polish). The reviewed diff is exactly what will merge.

PR #1616, opened 2026-07-10
Make ADR-048 Tauri-seam e2e tests tolerant of machine load (closes #1610)

malibio:
## Summary
- Centralize the daemon-connect timeout as `DAEMON_CONNECT_TIMEOUT` (60s, up from a duplicated 30s literal at every call site) in `test-support/src/lib.rs`
- Raise `ai_chat_send_to_idle_test`'s turn-completion wait (180s → 300s) — a real inference call, same load-sensitivity class
- Serialize `cargo test -p nodespace-app` to `--test-threads=1` in the pre-push gate (`scripts/test-gate.ts`)

## Why `--test-threads=1`, not `=2`
Every test in this suite spawns a real `nodespaced` process that loads a real embedding model (Metal shader compilation included) before its socket binds. I initially tried `--test-threads=2` (matching `nodespace-core`'s existing cap) but reproduced the exact failure from the issue against real contention — a concurrent `cargo build -p nodespace-daemon` in a sibling worktree — with the daemon reporting `NotRunning` after the *full* 60s wait, not just running close to it. `--test-threads=1` passed cleanly and repeatedly under the same live contention. Serializing costs almost no wall-clock: the suite's total runtime is dominated by the one real-inference test (~25-40s) regardless of thread count, since every other test is sub-second.

## Test plan
- [x] `cargo test -p nodespace-app -- --test-threads=1` passes on an idle machine
- [x] Same command passes repeatedly against a live competing `cargo build -p nodespace-daemon` (reproducing the issue's background-load scenario)
- [x] Full `bun run scripts/test-gate.ts` (pre-push gate) passes end-to-end
- [x] `bun run quality:fix` — 0 errors, 0 warnings, clippy clean

Closes #1610.

### DIFF
--- a/packages/desktop-app/src-tauri/test-support/src/lib.rs
+++ b/packages/desktop-app/src-tauri/test-support/src/lib.rs
@@ -31,6 +31,22 @@ use std::time::Duration;
 use nodespace_app_lib::services::GrpcClient;
 use tauri::Manager;
 
+/// Default timeout for `TauriTestApp::connect`'s daemon-health wait, shared
+/// by every `tests/*.rs` call site instead of each hardcoding its own
+/// duplicate literal.
+///
+/// 60s, not the 30s this used to be: spawning a whole OS process (daemon
+/// binary load + socket bind) is CPU-bound work, and within a single
+/// `tests/*.rs` binary `#[tokio::test]`s run concurrently by default, so
+/// several real daemon spawns can contend for the same cores at once —
+/// worse on a machine that also has an unrelated `cargo test` running
+/// elsewhere (see #1610). 30s was tuned for an unloaded machine and had no
+/// headroom for that; 60s gives slack without materially raising the
+/// wall-clock cost of a clean, unloaded run, since `wait_for_daemon` returns
+/// as soon as the socket is healthy rather than sleeping out the full
+/// duration.
+pub const DAEMON_CONNECT_TIMEOUT: Duration = Duration::from_secs(60);
+
 /// A running headless `nodespaced` pointed at a temp-dir socket and database.
 /// Killed on drop so a panicking test never leaks the process.
 pub struct SpawnedDaemon {
--- a/packages/desktop-app/src-tauri/test-support/src/lib.rs
+++ b/packages/desktop-app/src-tauri/test-support/src/lib.rs
@@ -22,6 +22,14 @@
 //! flagging whichever subset any single test binary doesn't happen to call
 //! as unused — `-D warnings` in `rust:lint` would otherwise hard-fail on
 //! that per-binary view.
+//!
+//! Running this crate's tests directly with a bare `cargo test -p
+//! nodespace-app` uses cargo's default per-binary test concurrency, NOT the
+//! `--test-threads=1` the pre-push gate (`scripts/test-gate.ts`) applies —
+//! see that file's comment for why serializing matters here. Prefer `cargo
+//! test -p nodespace-app -- --test-threads=1` when iterating locally on more
+//! than one test in the same file to avoid the same daemon-spawn contention
+//! #1610 fixed in the gate.
 
 use std::io::{BufRead, BufReader, Read};
 use std::path::PathBuf;
@@ -44,7 +52,10 @@ use tauri::Manager;
 /// headroom for that; 60s gives slack without materially raising the
 /// wall-clock cost of a clean, unloaded run, since `wait_for_daemon` returns
 /// as soon as the socket is healthy rather than sleeping out the full
-/// duration.
+/// duration. This is a ceiling on the failure case, not an added cost on
+/// the success path — a healthy daemon returns in well under a second, so
+/// raising this value does not add 60s to every test's typical run time,
+/// only to how long a genuinely stuck run takes to fail.
 pub const DAEMON_CONNECT_TIMEOUT: Duration = Duration::from_secs(60);
 
 /// A running headless `nodespaced` pointed at a temp-dir socket and database.
--- a/packages/desktop-app/src-tauri/tests/adapter_contract_test.rs
+++ b/packages/desktop-app/src-tauri/tests/adapter_contract_test.rs
@@ -16,13 +16,11 @@
 //! a change that breaks the contract on one path and not the other shows up
 //! as exactly one of the two suites failing.
 
-use std::time::Duration;
-
 use nodespace_app_lib::commands::nodes::{
     create_node, get_children, move_node, update_task_node, CreateNodeInput, InsertPositionInput,
 };
 use nodespace_app_lib::types::{TaskNodeUpdate, TaskStatus};
-use nodespace_app_test_support::{SpawnedDaemon, TauriTestApp};
+use nodespace_app_test_support::{SpawnedDaemon, TauriTestApp, DAEMON_CONNECT_TIMEOUT};
 use serde_json::json;
 
 fn text_input(id: &str, content: &str, parent_id: Option<String>) -> CreateNodeInput {
@@ -52,7 +50,7 @@ fn task_input(id: &str) -> CreateNodeInput {
 #[tokio::test]
 async fn task_tri_state_update_clear_set_no_change_matches_the_http_adapter_contract() {
     let daemon = SpawnedDaemon::spawn();
-    let harness = TauriTestApp::connect(&daemon, Duration::from_secs(30)).await;
+    let harness = TauriTestApp::connect(&daemon, DAEMON_CONNECT_TIMEOUT).await;
     let state = harness.client_state();
 
     let id = uuid::Uuid::new_v4().to_string();
@@ -108,7 +106,7 @@ async fn task_tri_state_update_clear_set_no_change_matches_the_http_adapter_cont
 #[tokio::test]
 async fn create_node_insert_position_matches_the_http_adapter_contract() {
     let daemon = SpawnedDaemon::spawn();
-    let harness = TauriTestApp::connect(&daemon, Duration::from_secs(30)).await;
+    let harness = TauriTestApp::connect(&daemon, DAEMON_CONNECT_TIMEOUT).await;
     let state = harness.client_state();
 
     let parent_id = uuid::Uuid::new_v4().to_string();
@@ -146,7 +144,7 @@ async fn create_node_insert_position_matches_the_http_adapter_contract() {
 #[tokio::test]
 async fn move_node_insert_position_matches_the_http_adapter_contract() {
     let daemon = SpawnedDaemon::spawn();
-    let harness = TauriTestApp::connect(&daemon, Duration::from_secs(30)).await;
+    let harness = TauriTestApp::connect(&daemon, DAEMON_CONNECT_TIMEOUT).await;
     let state = harness.client_state();
 
     let parent_a_id = uuid::Uuid::new_v4().to_string();
--- a/packages/desktop-app/src-tauri/tests/ai_chat_send_to_idle_test.rs
+++ b/packages/desktop-app/src-tauri/tests/ai_chat_send_to_idle_test.rs
@@ -28,7 +28,9 @@ use std::time::Duration;
 
 use nodespace_app_lib::commands::nodes::{create_node, get_node, update_node, CreateNodeInput};
 use nodespace_app_lib::types::NodeUpdate;
-use nodespace_app_test_support::{model_file_available, SpawnedDaemon, TauriTestApp};
+use nodespace_app_test_support::{
+    model_file_available, SpawnedDaemon, TauriTestApp, DAEMON_CONNECT_TIMEOUT,
+};
 use nodespace_proto::nodespace::EnsureModelReadyRequest;
 use serde_json::json;
 use tokio_stream::StreamExt;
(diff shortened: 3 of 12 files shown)


malibio:
### Code Review Summary

This is a well-scoped, well-documented fix for a real, reproducible flake (#1610). I independently re-derived the full diff and read the entire `test-support/src/lib.rs` concurrency model rather than trusting the provided summary, and everything in the summary checks out — no discrepancies found. The two commits tell an honest story: the author tried `--test-threads=2` first (matching the `nodespace-core` precedent), empirically found it insufficient against a live repro, and moved to `=1` rather than shipping the weaker fix to match precedent. That's the right instinct.

The core engineering judgment — `--test-threads=1` for this specific crate — is correct and proportionate once you look at where the concurrency actually lives: `SpawnedDaemon::spawn()` (a real OS process + Metal shader compile) happens **before** `CONNECT_MUTEX` is acquired, so multiple `#[tokio::test]`s in one binary (up to 5, in `node_crud_tauri_seam_test.rs`) race to spawn full daemon processes concurrently today. `CONNECT_MUTEX` only serializes the health-wait/connect/env-mutation step, not the expensive part. `--test-threads=1` closes exactly that gap. I also verified the two tests that exist to cover concurrent-client scenarios (`cross_window_hierarchy_sync_test.rs`, `optimistic_echo_race_test.rs`) achieve their concurrency via sequential/independent client handles and `tokio::spawn` **within** a single `#[tokio::test]` function, not via separate test functions racing — so `--test-threads=1` does not suppress any coverage. This is a case where the fix is more surgical than "blunt instrument" once you trace the actual synchronization boundaries; it looks blunt (a workspace-wide CLI flag) but its effect is precisely targeted at the uncoordinated resource it needed to fix.

Net positive. Recommending **approve**, with one non-blocking follow-up worth tracking as an issue (the raw-`cargo test` bypass gap) and a couple of nits.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- **`scripts/test-gate.ts` / raw `cargo test -p nodespace-app` bypass gap** — `--test-threads=1` is only applied by the `test-gate.ts` wrapper. I confirmed there is no `.cargo/config.toml` or workspace-level default that would apply it to a bare `cargo test -p nodespace-app` invocation, and no other reference to that command exists in the repo outside the wrapper. A developer running the crate's tests directly (which is the natural thing to do while iterating on a single test, rather than running the full `test-gate.ts` pipeline) gets the *old*, unfixed concurrency behavior and can still hit the #1610 flake locally, with no signal that the fix even exists for that invocation path. This isn't a reason to block — the pre-push gate is the actual enforcement point and this PR fixes that correctly — but it's worth a short-lived follow-up issue (or even just a one-line comment in a README/CLAUDE.md near "how to run this suite") pointing developers at the gate's flag when running the crate directly. A `[profile.test]`-adjacent or `.cargo/config.toml` `[alias]` wouldn't easily encode `--test-threads`, so the realistic options are either a doc pointer or accepting the gap — but it should be a conscious choice, not a silent one.

- **`CONNECT_MUTEX` overlap, not conflict, with `--test-threads=1`** — Confirmed `CONNECT_MUTEX` (a `tokio::sync::Mutex<()>` held across `wait_for_daemon` + `EnvGuard::set` + `GrpcClient::connect()`) exists to serialize mutation of the process-global `NODESPACED_SOCKET` env var across concurrently-running `#[tokio::test]`s in one binary. Under `--test-threads=1` there is no longer any concurrency for it to arbitrate within that binary, so it becomes a no-op lock (cheap, harmless, still correct). This is not a bug and not something to remove now — if `--test-threads` is ever raised back (e.g. once daemon spawn itself is made cheaper/mocked), the mutex is exactly the safety net that would be needed again. Just flagging that its raison d'être is currently dormant; no action needed.

#### Nitpicks

- **Nit:** `test-gate.ts`'s new comment (lines 45-60) is long and does a lot of work (explains the mechanism, the empirical falsification of `=2`, and the wall-clock justification) — genuinely valuable given there's no CI to cross-check against, but consider whether the empirical claim ("confirmed empirically: `--test-threads=2` still reproduced #1610's exact daemon-health timeout... on this machine") would be better placed in the PR description / issue thread than baked permanently into the source comment, since "on this machine" is a claim that can't be re-verified by a future reader and will age awkwardly. Not blocking — the reasoning itself (spawn-before-mutex, CPU/GPU contention) stands on its own without the anecdote.
- **Nit:** `DAEMON_CONNECT_TIMEOUT`'s doc comment is accurate and good, but could note explicitly that it does not (and by design should not) bound `wait_for_daemon`'s per-test-binary total cost, since `wait_for_daemon` returns early on success — a future reader unfamiliar with the polling loop might assume 60s is added to every test's fixed cost.

### Section-by-section coverage

**Architectural fit** — Good. `DAEMON_CONNECT_TIMEOUT` as a single `pub const` in the shared `test-support` crate, replacing ~11 duplicated literals, is the correct DRY fix and matches how the crate already centralizes other shared fixture logic (`SpawnedDaemon`, `CONNECT_MUTEX`, `EnvGuard`). The `--test-threads=1` change lives in the one place (`test-gate.ts`) that actually needs to know about it. Nothing here reaches into business logic or changes any production code path — this is a pure test-infrastructure change, exactly matching its stated scope.

**Correctness of the migration** — Verified directly (not just via grep on the summary): all ~11 call sites across `adapter_contract_test.rs`, `ai_chat_send_to_idle_test.rs`, `commits_per_edit_test.rs`, `cross_window_hierarchy_sync_test.rs`, `daemon_readiness_test.rs`, `indent_outdent_rapid_ordering_test.rs`, `model_download_terminal_state_test.rs`, `node_crud_tauri_seam_test.rs`, `optimistic_echo_race_test.rs`, `startup_readiness_data_plane_test.rs` now use `DAEMON_CONNECT_TIMEOUT`. Zero remaining `Duration::from_secs(30)` daemon-connect literals anywhere in `tests/` or `test-support/` (confirmed by grep). `schema_parity_test.rs` doesn't spawn a daemon and is correctly untouched. Import hygiene is correct in both directions: the 3 files that keep `use std::time::Duration` (`ai_chat_send_to_idle_test.rs`, `optimistic_echo_race_test.rs`, `startup_readiness_data_plane_test.rs`) each have genuine other direct `Duration::from_secs`/`from_millis` uses I confirmed by grep; the other ~8 files correctly drop the now-unused import. `DAEMON_CONNECT_TIMEOUT`'s doc comment is factually accurate against the code it describes (mutex scope, per-binary concurrency, early-return behavior of `wait_for_daemon`). `cargo check -p nodespace-app-test-support` passes clean.

**`--test-threads=1` vs `=2` tradeoff** — This is the crux of the PR and I traced the actual concurrency boundaries rather than taking the commit message's word for it (see task item 2-4). Key fact: `SpawnedDaemon::spawn()` is called by each test *before* any mutex is acquired — `CONNECT_MUTEX` only wraps `wait_for_daemon` + `GrpcClient::connect()`. So today, absent `--test-threads=1`, up to 5 real `nodespaced` processes (in `node_crud_tauri_seam_test.rs` alone) can be mid-spawn — loading a real embedding model with Metal shader compilation — fully concurrently, fully uncoordinated by any lock. That is precisely the load-inducing behavior #1610 describes, and it exists *within a single clean, unloaded run*, not just under external contention — external load (a concurrent `cargo build -p nodespace-daemon`) just makes the existing self-inflicted contention worse. Given that, `--test-threads=1` isn't papering over a latent bug elsewhere; it's removing a genuine, avoidable source of resource contention that the test suite itself creates. I judge this a reasonable, well-justified choice for a pre-push reliability gate, not a lazy blunt instrument — the "blunt instrument" framing undersells how precisely targeted the actual effect is once you look at where `SpawnedDaemon::spawn()` sits relative to `CONNECT_MUTEX`. I confirmed (via `cross_window_hierarchy_sync_test.rs` and `optimistic_echo_race_test.rs`) that no concurrency-coverage is lost: both files' "concurrent client" scenarios are constructed via multiple `TauriTestApp::connect` calls or `tokio::spawn` **inside** one `#[tokio::test]`, which `--test-threads` (a knob for inter-test-function scheduling only) does not affect at all.

**Timeout-raise DX tradeoff (30s→60s, 180s→300s)** — Acceptable given the constraints stated in the issue and confirmed in the repo: `scripts/test-gate.ts`'s own header comment ("This repo has no CI runner for tests — this hook is the only gate. See ADR-047.") does exist, exactly as claimed — I read it directly. That raises the stakes of a local dev's wait time but also means there's no second, faster signal elsewhere to fall back on, so a generous local timeout is the safer failure mode (a real hang still fails, just slower, vs. a false negative that ships silently). Both `wait_for_daemon` and `poll_until_idle_with_new_assistant_reply` return as soon as their condition is met rather than sleeping out the full duration, so the raise only affects worst-case/failure latency, not typical-case wall-clock — that claim in the doc comments is accurate and I confirmed it against `wait_for_daemon`'s polling-loop implementation in `daemon_setup.rs`. A genuinely hung daemon now takes up to 60s (was 30s) and a genuinely hung ai-chat turn up to 300s (was 180s) to fail locally before push — worse, but proportionate to the problem being fixed, and `git push --no-verify` remains available (per the file's own documented bypass policy) for anyone blocked by a real regression they need to investigate separately.

**Consistency/completeness across ~11 call sites** — Complete and consistent; verified directly, not just via grep count matching. No stray literals, no partial migration, no test file missed.

**Anything unrelated bundled in** — None found. Diff stat is exactly the 12 files described, all within the stated scope (timeout constant, its call sites, import cleanup, and the test-gate.ts threading change). `.mcp.json` seen in `git status` is untracked local tooling, not part of this diff, correctly not touched.

**Issue #1610 acceptance criteria — gap check**

1. *"`cargo test -p nodespace-app` passes reliably under representative background load without raising wall-clock cost of a clean run significantly"* — **Unverified by me, as instructed.** I did not run the suite (correctly out of scope for this review — it's slow/flaky by design and the exact thing being fixed). The architectural analysis above supports that the fix targets the right root cause, and the author's own commit message describes a live empirical repro-and-fix cycle (⇒ `=2` insufficient, `=1` verified repeatedly under real contention), which is credible but not something I can independently confirm from a static review. Flagging as unverified, not as failing.
2. *"Daemon-health wait has headroom for slow spawn under load"* — **Met.** 30s→60s via `DAEMON_CONNECT_TIMEOUT`, applied consistently.
3. *"Other fixed-timeout assertions with the same load-sensitivity reviewed"* — **Met.** The ai-chat turn-completion wait (180s→300s) is the one other fixed-timeout assertion called out in the issue by name, and it's addressed with the same reasoning.
4. *"`bun run quality:fix` passes; tests pass"* — **Unverified by me, as instructed** (told not to run quality:fix or the mutating/slow suite). `cargo check -p nodespace-app-test-support` compiles cleanly, which is a partial positive signal for the Rust side, but not a substitute for the full gate.

### Recommendation

**Approve.** The change is atomic, correctly scoped, well-reasoned, and the migration is complete and mechanically verified. The `--test-threads=1` decision is the one judgment call in this PR with real tradeoffs, and on inspection of the actual synchronization boundaries (spawn-before-mutex) it's the right call, not an overcorrection. The one real gap — raw `cargo test -p nodespace-app` bypassing the fix entirely — is worth a follow-up issue but does not block merge, since the pre-push gate (the actual enforcement mechanism this repo relies on per its own "no CI" comment) is fixed correctly.

malibio (commented):
Review posted above. Recommendation: APPROVE (net positive, no blockers) — GitHub blocks self-approval since this PR's author is the authenticated account, so recording as a comment-review instead. See the full review comment for the complete analysis and one non-blocking follow-up.

malibio:
## Addressed review feedback

Thanks for the thorough review. Summary of what changed in commit 4e55fe65:

✅ **Addressed:**
- **Nit — `test-gate.ts`'s "confirmed on this machine" anecdote**: replaced with the durable structural reason `--test-threads=1` is needed. `SpawnedDaemon::spawn()` happens *before* `CONNECT_MUTEX` is acquired, so concurrent daemon spawns are self-inflicted contention this suite creates on its own (not just an external-load sensitivity) — that's the claim that holds up for a future reader, not a one-time empirical anecdote tied to this machine.
- **Nit — `DAEMON_CONNECT_TIMEOUT` doc comment**: clarified explicitly that 60s is a ceiling on the failure path, not an added cost on the success path, since `wait_for_daemon` returns as soon as the socket is healthy.
- **Suggested — raw `cargo test -p nodespace-app` bypass gap**: added a doc pointer in `test-support/src/lib.rs`'s crate-level doc comment (the natural place a developer landing in this crate would look) directing them to run `cargo test -p nodespace-app -- --test-threads=1` directly when iterating locally, since the gate's flag doesn't apply to a bare invocation.

⏭️ **Skipped:**
- **`CONNECT_MUTEX` now dormant under `--test-threads=1`** — you explicitly flagged this as informational only ("not a bug... no action needed"), so no code change made. It remains the safety net if thread count is ever raised again.

Verified after changes: `cargo check -p nodespace-app-test-support` compiles clean, `bun run quality:fix` passes (0 errors/warnings, clippy clean), and the full `scripts/test-gate.ts` pre-push gate passed end-to-end before pushing.

## Re-Review Decision

**Rationale**: All changes in this follow-up commit are doc-comment-only — no logic, no behavior change, no new code paths. The one skipped item was explicitly marked non-actionable by the reviewer. Low risk, straightforward, and directly traceable to the review's own findings.

PR #1620, opened 2026-07-11
BM25 root-resolution is an N+1 on the search hot path

malibio:
Closes #1504

### DIFF
--- a/packages/core/src/db/sqlite_store/embeddings.rs
+++ b/packages/core/src/db/sqlite_store/embeddings.rs
@@ -581,36 +581,55 @@ impl SqliteStore {
             return Ok(HashSet::new());
         }
 
-        // Resolve each match to its root using recursive CTE
-        let mut root_ids = HashSet::new();
+        // Resolve every match to its root in one set operation: seed the recursive
+        // CTE with the whole match set (depth 0), walk `has_child` parents, and for
+        // each seed keep the deepest ancestor reached (its root, or itself if it has
+        // no parent).
+        let placeholders: Vec<String> = (1..=matching_ids.len())
+            .map(|i| format!("?{}", i))
+            .collect();
+        let sql = format!(
+            r#"WITH RECURSIVE ancestors(seed_id, node_id, depth) AS (
+                SELECT id, id, 0 FROM node WHERE id IN ({})
+                UNION ALL
+                SELECT a.seed_id, r.in_node, a.depth + 1 FROM relationship r
+                JOIN ancestors a ON r.out_node = a.node_id
+                WHERE r.relationship_type = 'has_child' AND a.depth < 100
+            )
+            SELECT seed_id, node_id FROM ancestors a
+            WHERE a.depth = (SELECT MAX(depth) FROM ancestors WHERE seed_id = a.seed_id)"#,
+            placeholders.join(", ")
+        );
 
-        for node_id in matching_ids {
-            let mut rows = self.db.query(
-                r#"WITH RECURSIVE ancestors(node_id, depth) AS (
-                    SELECT in_node, 1 FROM relationship WHERE out_node = ?1 AND relationship_type = 'has_child'
-                    UNION ALL
-                    SELECT r.in_node, a.depth + 1 FROM relationship r
-                    JOIN ancestors a ON r.out_node = a.node_id
-                    WHERE r.relationship_type = 'has_child' AND a.depth < 100
-                )
-                SELECT node_id FROM ancestors ORDER BY depth DESC LIMIT 1"#,
-                libsql::params![node_id.clone()],
-            ).await.context("Failed to get root for BM25 match")?;
+        let params: Vec<libsql::Value> = matching_ids
+            .iter()
+            .map(|id| libsql::Value::Text(id.clone()))
+            .collect();
 
-            if let Some(row) = rows.next().await? {
-                let root_id: String = row.get(0)?;
-                if self.get_node(&root_id).await.is_ok_and(|n| n.is_some()) {
-                    root_ids.insert(root_id);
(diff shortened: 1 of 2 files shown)


malibio:
### Code Review Summary

Solid, correct fix for the N+1 described in #1504. The rewrite collapses root resolution from `1 + 2N` queries down to a flat 3 (FTS5 match, batched CTE, batched existence check), and I verified the multi-seed CTE semantics directly with a throwaway test covering the exact scenario the issue targets (multiple simultaneous matches, some sharing a root, some in separate trees, one matching a root directly) — all resolved correctly with no cross-seed contamination. Build, clippy (`-D warnings -D clippy::unused-async`), `quality:fix`, and all 6 existing BM25 tests pass clean. One real gap: none of the shipped tests actually exercise the multi-match batch case this PR is architecturally about — every existing test drives a single match term through a single tree.

### Requirements Check

- ✅ Root resolution runs as one query over the match set, not per-match — confirmed: single CTE seeded via `WHERE id IN (...)` over the whole `matching_ids` set.
- ✅ No per-result `get_node` round-trips in the search loop — confirmed: replaced with one `get_nodes_by_ids` batched call.
- ✅ Code passes `bun run quality:fix`; tests pass — confirmed locally (`cargo fmt --all`, `cargo clippy --all-targets -D warnings -D clippy::unused-async`, `cargo test -p nodespace-core --test embedding_service_test` all clean; working tree unchanged after `quality:fix`).

### Correctness Verification (CTE rewrite)

I traced the SQL and additionally validated it empirically rather than by inspection alone. Wrote a scratch test (not committed) with three trees in one call: root_a with two matching children (shared-root case), root_b with a matching grandchild (separate tree, deeper path), and root_c matching directly at the root (no-parent case). `bm25_search_roots("zzzmarker", 50)` returned exactly `{root_a, root_b, root_c}` — no children/grandchildren leaked through, no cross-seed interference.

Why this holds architecturally: the recursive member carries `seed_id` down every path (`a.seed_id` propagated unchanged through the `JOIN`), so each generated row is tagged with the seed it originated from even when two seeds' ancestor chains merge into the same physical node higher up the tree. The final filter `a.depth = (SELECT MAX(depth) FROM ancestors WHERE seed_id = a.seed_id)` is correlated on `seed_id`, so one seed's max-depth computation can never be polluted by another seed's deeper/shallower chain — they're disjoint partitions of the same result set. A `has_child` cycle would behave identically to the old code (both guard with `depth < 100`), so no regression there either.

Existence re-check is semantically equivalent to the old per-root `get_node` gate: `get_nodes_by_ids` returns a `HashMap<String, Node>` for whatever subset of the candidate IDs still exist, and the final filter (`existing.contains_key(id)`) reproduces "only keep roots that still exist" with one round trip instead of N. Empty-candidate short-circuit (`if candidate_roots.is_empty()`) correctly avoids issuing a pointless `IN ()` query.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- 📁 `packages/core/tests/embedding_service_test.rs` (BM25 test block, ~1240-1353): **Missing test coverage for the multi-match batch scenario.** All 6 existing BM25 tests drive exactly one query term through exactly one document tree per call. This PR's entire value proposition is batching root resolution across *multiple simultaneous* FTS matches — including the case where two matches share a root (exercising the correlated-subquery partitioning) and the case where matches land in disjoint trees in the same call. That path is currently unverified by the test suite; a regression there (e.g., someone "simplifying" the correlated subquery to a plain `GROUP BY` without the `seed_id` correlation) would not be caught by CI. I verified correctness manually during this review with a throwaway test and it passed, but recommend adding a version of it as a permanent regression test.
  Engineering principle: test coverage should track the specific risk a change introduces, not just re-confirm pre-existing behavior. A refactor's tests should target the new code paths it creates.

#### Nitpicks
- Nit: `packages/core/src/db/sqlite_store/embeddings.rs:588-590` and `:627-628` — the placeholder-building pattern (`(1..=n).map(|i| format!("?{}", i))`) is duplicated from `get_nodes_by_ids` in `nodes.rs:177`. Not blocking (three lines, low churn risk), but if a third call site appears, worth extracting a small `bind_in_clause(ids) -> (String, Vec<Value>)` helper.

### Performance

Confirmed: query count is now flat regardless of match count — 1 (FTS5) + 1 (CTE over full match set) + 1 (batched existence check) = 3, versus the old `1 + 2N`. This is the intended fix for the hot-path scaling issue in #1504.

### Security

No injection risk. All dynamic SQL construction only builds placeholder counts (`?1, ?2, ...`); actual `node_id` values are always passed through bound `libsql::Value::Text` parameters, never string-interpolated into the query text. Same pattern as the pre-existing `get_nodes_by_ids`.

**Recommendation**: APPROVE

The fix is architecturally correct, verified empirically against the exact multi-seed scenario it's meant to solve, and meets all three acceptance criteria. The missing multi-match test is a real gap worth closing but isn't a blocker — the existing 6 BM25 tests plus clippy plus my manual verification give reasonable confidence, and the fix is small and easily revertible if a regression surfaces.

malibio:
**Review recommendation: APPROVE** (posted as comment — GitHub blocks self-approval of your own PR via `gh pr review`). See full review above for details.

malibio:
Addressed the reviewer's coverage gap: added `test_bm25_search_roots_batches_multiple_matches_across_trees` covering the multi-match batch scenario (two matches sharing a root, one nested via grandchild in a separate tree, one match that is itself a root) in a single `bm25_search_roots` call — exactly the seed-partitioning risk surface the correlated `MAX(depth)` subquery introduces. All 35 tests in `embedding_service_test.rs` pass, full pre-push gate (unit + browser + rust workspace + e2e) green.

PR #1622, opened 2026-07-11
SQLite DDL migration runner (pre-launch requirement)

malibio:
Closes #1320

### DIFF
--- a/packages/core/src/db/migrations/mod.rs
+++ b/packages/core/src/db/migrations/mod.rs
@@ -1,0 +1,90 @@
+//! Versioned DDL migration runner.
+//!
+//! Schema changes are tracked via SQLite's `PRAGMA user_version` (an integer stored
+//! in the database file header). On startup, `run` compares the database's current
+//! version against the known migrations and applies any with a higher version, in
+//! order, each in its own transaction that also bumps `user_version` — so a crash
+//! mid-migration leaves the database at a consistent, re-resumable version.
+//!
+//! Adding a schema change: add a new `vNNN_description` module below, add its
+//! version to `LATEST_VERSION`, and add a matching arm to `apply_migration`. Never
+//! edit a past migration's `apply` fn once it has shipped — existing databases will
+//! already have applied it under its old definition, and rerunning a changed
+//! version won't happen (each migration runs at most once per database).
+
+mod v001_initial_schema;
+mod v002_embedding_origin;
+
+use anyhow::{Context, Result};
+
+/// Highest migration version known to this build. Bump when adding a migration.
+pub const LATEST_VERSION: i64 = 2;
+
+async fn apply_migration(tx: &libsql::Transaction, version: i64) -> Result<()> {
+    match version {
+        1 => v001_initial_schema::apply(tx).await,
+        2 => v002_embedding_origin::apply(tx).await,
+        _ => unreachable!("no migration defined for version {version}"),
+    }
+}
+
+fn migration_name(version: i64) -> &'static str {
+    match version {
+        1 => "initial_schema",
+        2 => "embedding_origin",
+        _ => "unknown",
+    }
+}
+
+/// Read `PRAGMA user_version` (0 on a brand-new database).
+async fn current_version(conn: &libsql::Connection) -> Result<i64> {
+    let mut rows = conn
+        .query("PRAGMA user_version", ())
+        .await
+        .context("Failed to read PRAGMA user_version")?;
+    let version: i64 = rows
+        .next()
+        .await?
+        .map(|row| row.get(0))
+        .transpose()?
+        .unwrap_or(0);
+    Ok(version)
+}
+
+/// Apply all migrations with `current_version < version <= target_version`, in
+/// ascending order. Each migration runs in its own transaction that also bumps
+/// `user_version` to that migration's version, so a database is never left
+/// between two versions.
+pub async fn run_up_to(conn: &libsql::Connection, target_version: i64) -> Result<()> {
+    let start_version = current_version(conn).await?;
+
+    for version in (start_version + 1)..=target_version {
+        let tx = conn
+            .transaction()
+            .await
+            .with_context(|| format!("Failed to begin transaction for migration {version}"))?;
+
+        apply_migration(&tx, version).await.with_context(|| {
+            format!(
+                "Migration {:03}_{} failed",
+                version,
+                migration_name(version)
+            )
+        })?;
+
+        tx.execute(&format!("PRAGMA user_version = {version}"), ())
+            .await
+            .with_context(|| format!("Failed to bump user_version to {version}"))?;
+
+        tx.commit()
+            .await
+            .with_context(|| format!("Failed to commit migration {version}"))?;
+    }
+
+    Ok(())
+}
+
+/// Apply all pending migrations, bringing the database to [`LATEST_VERSION`].
+pub async fn run(conn: &libsql::Connection) -> Result<()> {
+    run_up_to(conn, LATEST_VERSION).await
+}
--- a/packages/core/src/db/migrations/mod.rs
+++ b/packages/core/src/db/migrations/mod.rs
@@ -88,3 +88,25 @@ pub async fn run_up_to(conn: &libsql::Connection, target_version: i64) -> Result
 pub async fn run(conn: &libsql::Connection) -> Result<()> {
     run_up_to(conn, LATEST_VERSION).await
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Guards against the panic in `apply_migration`'s `unreachable!` arm: if
+    /// `LATEST_VERSION` is bumped without adding a matching `apply_migration` arm,
+    /// `run` would panic against a real user database instead of failing a build.
+    /// This test converts that mistake into a compile-time-adjacent, always-run
+    /// test failure instead.
+    #[tokio::test]
+    async fn every_version_up_to_latest_has_a_defined_migration() {
+        let temp_dir = tempfile::TempDir::new().unwrap();
+        let db_path = temp_dir.path().join("version_coverage.db");
+        crate::db::ensure_sqlite_vec_registered().await;
+        let database = libsql::Builder::new_local(&db_path).build().await.unwrap();
+        let conn = database.connect().unwrap();
+
+        // Panics via `unreachable!` if any version in range lacks an arm.
+        run_up_to(&conn, LATEST_VERSION).await.unwrap();
+    }
+}
--- a/packages/core/src/db/migrations/v001_initial_schema.rs
+++ b/packages/core/src/db/migrations/v001_initial_schema.rs
@@ -1,0 +1,138 @@
+//! Baseline schema: `node`, `relationship`, `embedding` tables, their indexes, the
+//! `node_fts` full-text index (and its sync triggers), and the `vec_embeddings`
+//! vector index. This is the schema every NodeSpace database started life with,
+//! canonicalized as migration 1.
+//!
+//! Connection-level PRAGMAs (`journal_mode`, `foreign_keys`, `synchronous`,
+//! `busy_timeout`) are NOT part of this migration — SQLite forbids changing
+//! `synchronous` inside a transaction, and `foreign_keys`/`synchronous`/
+//! `busy_timeout` are per-connection session settings, not persisted schema
+//! state, so they're set unconditionally on every connection open in
+//! `SqliteStore::new`, independent of migration state.
+//!
+//! Frozen as of this migration's creation — do not edit to match later schema
+//! changes (those are later-numbered migrations, e.g. `v002_embedding_origin`
+//! adds `embedding.origin`, which this baseline intentionally omits). See
+//! `db/schema.sql` for a human-readable reference of the CURRENT schema (after
+//! all migrations), not this baseline.
+
+use anyhow::{Context, Result};
+
+const BASELINE_SQL: &str = r#"
+CREATE TABLE IF NOT EXISTS node (
+    id               TEXT    PRIMARY KEY,
+    node_type        TEXT    NOT NULL,
+    content          TEXT    NOT NULL DEFAULT '',
+    properties       TEXT    NOT NULL DEFAULT '{}',
+    title            TEXT,
+    lifecycle_status TEXT    NOT NULL DEFAULT 'active',
+    version          INTEGER NOT NULL DEFAULT 1,
+    sync_seq         INTEGER,
+    created_at       TEXT    NOT NULL,
+    modified_at      TEXT    NOT NULL
+) STRICT;
+
+CREATE INDEX IF NOT EXISTS idx_node_type      ON node (node_type);
+CREATE INDEX IF NOT EXISTS idx_node_modified  ON node (modified_at);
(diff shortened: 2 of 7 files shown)


malibio:
### Code Review Summary

**Review Type**: Initial Review (PR #1622, no prior reviews found)

This PR replaces the old `CREATE TABLE IF NOT EXISTS` + ad-hoc `migrate_embedding_origin` approach with a proper `PRAGMA user_version`-tracked, numbered migration runner. I traced the frozen migration-1 baseline against the exact pre-PR `schema.sql` + `initialize_schema` code path (via `git show` on the commits that introduced each piece of schema), ran the new `migrations_test.rs` suite and the existing `sqlite_store` unit tests, and checked `cargo clippy`. Overall this is a well-executed, net-positive change: the highest-risk item (baseline fidelity) checks out exactly, the two bugs found during manual testing are fixed soundly, and the atomicity/ordering model is correct. There is one real (minor) test-coverage regression and a couple of documentation/defensive-code nits worth addressing, but nothing blocking.

### Requirements Check (issue #1320)

- [x] **`PRAGMA user_version` is used to track applied migration version** — `migrations/mod.rs:40-52` reads it via `current_version`, `run_up_to` bumps it per-migration inside the same transaction as the DDL (`migrations/mod.rs:75-77`). Met.
- [x] **Current schema is canonicalized as migration `001`** — `v001_initial_schema.rs` freezes the exact pre-PR `schema.sql` DDL (verified table-by-table below). Met.
- [x] **New DDL changes are added as numbered migrations, not edits to `schema.sql`** — `v002_embedding_origin.rs` adds the `origin` column as migration 2; `schema.sql` is no longer executed (`include_str!` removed) and is explicitly documented as a reference snapshot. Met.
- [x] **A user on migration `001` who upgrades gets all subsequent migrations applied in order** — `run_up_to` iterates `(start_version+1)..=target_version` ascending, each in its own transaction; covered by `database_on_migration_one_upgrades_through_all_subsequent_migrations` (migrations_test.rs:130). Met.
- [x] **Tests cover: fresh DB, already-current DB, multi-step upgrade path** — `fresh_database_ends_up_at_latest_version_with_full_schema`, `already_current_database_is_a_no_op`, `database_on_migration_one_upgrades_through_all_subsequent_migrations` all present and passing. A bonus regression test (`pre_runner_database_with_origin_already_applied_upgrades_without_error`) covers the real-world pre-runner-database bug found during manual testing. Met, and then some.

All acceptance criteria are met. I ran `cargo test -p nodespace-core --test migrations_test` (4/4 pass) and `cargo test -p nodespace-core --lib db::sqlite_store` (19/19 pass, including the pre-existing `test_concurrent_writers_retry_instead_of_erroring_on_busy`), plus `cargo clippy -p nodespace-core --lib --tests` (clean).

### Verification of the specific risk areas called out

**(a) Connection-level PRAGMAs pulled out of the migration transaction — correct and safe.** `apply_connection_pragmas` (`sqlite_store/mod.rs:166-180`) runs via `conn.query(...)` outside any transaction, called once per `SqliteStore::new()` (i.e., once per connection open — there is exactly one production call site, the daemon's store construction), *before* `initialize_schema`/migrations run (`sqlite_store/mod.rs:146-147`). This is correct: SQLite does forbid changing `synchronous` inside a transaction, and `foreign_keys`/`synchronous`/`busy_timeout` are per-connection session state that resets on every new connection, so re-applying them every open is required, not wasteful — the "overhead" is four trivial PRAGMA round-trips, negligible relative to opening a DB file. Setting `busy_timeout` before migrations run is also the right order: it makes a second process racing on migrations retry through `SQLITE_BUSY` instead of hard-failing (see concurrency note below). No ordering hazard found — `journal_mode = WAL` before DDL execution is the standard, safe idiom.

**(b) `PRAGMA table_info` guard in `v002_embedding_origin.rs` before `ALTER TABLE` — sound, and the regression test genuinely exercises the scenario.** `has_origin_column` (`v002_embedding_origin.rs:15-27`) mirrors the old ad-hoc check exactly. The new test `pre_runner_database_with_origin_already_applied_upgrades_without_error` (`migrations_test.rs:97-127`) applies migration 1 for real via the runner, then manually runs the exact `ALTER TABLE ... ADD COLUMN origin` the old ad-hoc code used to simulate a pre-runner database, confirms `user_version` is still 1, then calls `migrations::run` and asserts it succeeds and lands at `LATEST_VERSION`. This is a faithful reproduction of the real bug (verified against `git log` — the old `migrate_embedding_origin` ran outside `user_version` tracking, so exactly this "column present, version stale" state was reachable in production). Fix and test are both sound.

**(c) Atomicity — correct.** Each migration runs in its own `libsql::Transaction`; `user_version` is bumped via `tx.execute(...)` *inside* that same transaction before `tx.commit()` (`migrations/mod.rs:62-81`), so a crash mid-migration can never leave the schema changed with the old `user_version`, or vice versa. A failure on migration N aborts that transaction (auto-rollback on `Transaction` drop) and returns `Err` before advancing further, so the database is left at `N-1`'s `user_version` with `N-1`'s schema fully applied — consistent and re-resumable on next startup. This is the correct granularity (per-migration, not one giant transaction for the whole batch) since it allows partial progress across restarts without redoing already-applied migrations.

**(d) HIGHEST RISK — migration 1 baseline fidelity: verified exact match.** I diffed `v001_initial_schema.rs`'s `BASELINE_SQL` plus its FTS5/vec0 block against the `schema.sql` + `initialize_schema` code that existed immediately before the `origin` column was introduced (commit `de414e35^`, i.e. right after PR #1416/"Migrate local DB from SurrealDB to SQLite"). Table-by-table:
- `node`: identical columns, types, defaults, `STRICT`, and all three indexes (`idx_node_type`, `idx_node_modified`, `idx_node_lifecycle`).
- `relationship`: identical columns/defaults/FKs/`ON DELETE CASCADE`, all four indexes including the composite unique constraint.
- `embedding`: identical (correctly *without* `origin`, which is deferred to migration 2 as intended).
- FTS5 `node_fts` virtual table + all three sync triggers (insert/update/delete): byte-for-byte identical to the old inline code.
- `vec0` virtual table (`vec_embeddings`) with `FLOAT[{DEFAULT_EMBEDDING_DIMENSION}] distance_metric=cosine`: identical.
- Connection PRAGMAs (`journal_mode`, `foreign_keys`, `synchronous`, `busy_timeout=5000`): correctly excluded from the migration and moved to `apply_connection_pragmas`, matching the documented rationale.

No discrepancy found. This is the correct, careful outcome for the highest-risk item in the PR.

**(e) Test coverage gaps** — real gaps exist, though none rise to Critical:
- No test for a migration failing mid-batch (e.g., migration 2 throwing partway through) and verifying the DB is left at a resumable, non-corrupt state and that a subsequent successful `run` recovers. The transactional design (per (c)) makes this very likely to work, but it is asserted nowhere.
- No test for concurrent migration attempts (two processes/connections calling `migrations::run` against the same file simultaneously) — relevant given `busy_timeout` is explicitly there to support the dual-daemon scenario documented at `sqlite_store/mod.rs:1142-1146`. See concurrency note below.
- No downgrade-scenario test (calling `run_up_to` with a `target_version` below the current version) — behavior is a safe no-op by construction (empty range), but that's implicit, not asserted.
- The "pre-runner database" regression test doesn't insert data before altering, so it doesn't re-verify existing-row defaulting/index-shape the way the deleted test did (see finding below).

**(f) Doc accuracy — good, appropriately discoverable.** `schema.sql`'s new header (`schema.sql:1-8`) clearly states it is "NOT executed directly" and names the migration directory as the source of truth. `v001_initial_schema.rs`'s module doc explicitly says "Frozen as of this migration's creation — do not edit" and points forward to `v002` and to `schema.sql` for the current-state view. `migrations/mod.rs`'s module doc explains the "never edit a past migration" invariant clearly. A future developer skimming `schema.sql` first (the most likely entry point) will not mistake it for live code.

**(g) `ensure_sqlite_vec_registered` becoming `pub` — justified, correctly scoped.** The only external consumer is `packages/core/tests/migrations_test.rs:7`, which is a separate integration-test crate and therefore cannot see `pub(crate)` items — `pub` is the minimum visibility that makes this compile. The doc comment (`sqlite_store/mod.rs:84-86`) accurately explains why. Not overly broad.

**(h) Deleted `test_migrate_embedding_origin_upgrades_legacy_db` — coverage gap, not fully replaced.** See Suggested Improvements below; the "duplicate column" crash itself is now well covered, but two assertions from the deleted test (existing-row data survives with the correct default, and the rebuilt index's column order) are not re-verified on the pre-runner-simulated path.

### Findings

#### Critical Issues

None found.

#### Suggested Improvements

- **`packages/core/tests/migrations_test.rs:97-127`** (relates to point h/e): `pre_runner_database_with_origin_already_applied_upgrades_without_error` doesn't insert a row into `embedding` before the simulated legacy `ALTER TABLE`, and doesn't assert the rebuilt `idx_emb_modified` index shape on this path (only the separate `fresh_database_...` test checks the index, on a fresh DB). The deleted `test_migrate_embedding_origin_upgrades_legacy_db` checked both: that a pre-existing row got `origin = 'local'` by default, and that the index was rebuilt with `origin` leading. Recommend adding an `INSERT INTO embedding (...)` before the simulated legacy ALTER and asserting its `origin` value plus the index SQL in this test, so the full behavior the old test protected is still covered on the *exact* upgrade path this migration exists for. **Principle:** a replaced test should preserve at least the same assertions as what it superseded, not just the crash-repro; otherwise a future edit to the ALTER/index-rebuild logic in `v002` could silently drop index coverage or data-preservation and nothing would fail.

- **`packages/core/src/db/migrations/mod.rs:58-85`** (relates to point e): no test exercises a migration failing partway through a multi-migration `run` (e.g., mocking migration 2 to fail) and asserting `user_version` stays at 1 and a subsequent `run` recovers cleanly. Given this is explicitly called out as a documented guarantee ("a crash mid-migration leaves the database at a consistent, re-resumable version" — `migrations/mod.rs:6-7`), it's worth a test that actually forces a failure (e.g., a temporary migration 3 that always errors, gated behind `#[cfg(test)]`, or directly testing `run_up_to` with an out-of-range/bogus target that trips the `unreachable!` in a controlled way) to prove the claim rather than relying on transactional semantics being correct by inspection alone. **Principle:** documented invariants about failure recovery are exactly the kind of thing that should be pinned by a test — they're easy to break silently in a future refactor (e.g., someone batching multiple migrations into one transaction "for performance").

- **`packages/core/src/db/sqlite_store/mod.rs:1142-1146`** and **`migrations/mod.rs`** (concurrency, point e): the comment above `test_concurrent_writers_retry_instead_of_erroring_on_busy` documents that two `SqliteStore`s (dev + prod daemon) can open the same file concurrently, and `busy_timeout` is relied on to make the *second* writer retry rather than fail. That test covers ordinary writes, but there is no test for two processes both calling `SqliteStore::new` (and therefore both racing to run migrations) against the same fresh/stale file. In practice `busy_timeout=5000` combined with per-migration transactions makes this very likely to be safe (the loser blocks up to 5s then retries, and each migration's `user_version` bump is atomic so there's no "half-applied migration 2" state to observe), but a concurrent-migration-open test would directly validate the exact scenario the `busy_timeout` PRAGMA was introduced for, rather than leaving it as an inference from two separate tests.

#### Nitpicks

- **Nit:** `packages/core/src/db/migrations/mod.rs:27`: `apply_migration`'s `_ => unreachable!(...)` will panic if `LATEST_VERSION` is ever bumped without a matching match arm being added — the module doc does warn about this (`mod.rs:9-10`), but consider a `debug_assert!` at module load or a unit test that asserts `apply_migration` has a defined arm for every version `1..=LATEST_VERSION` (trivially done by calling `migration_name` and checking it isn't `"unknown"`), so this class of mistake fails fast in CI rather than at runtime against a real user database.
- **Nit:** `packages/core/src/db/migrations/mod.rs:31-37`: `migration_name`'s `_ => "unknown"` arm is dead in practice (any version reaching it would have already panicked in `apply_migration`'s `unreachable!`) — harmless, but slightly inconsistent messaging between the two match statements for the same invalid-version condition.
- **Nit:** `packages/core/src/db/migrations/v001_initial_schema.rs:81-89`: the baseline SQL is split on `;` and each statement executed individually, same pattern as the old `initialize_schema`. This is fine given the baseline is frozen and known not to contain semicolons inside string literals/triggers, but it's worth a one-line comment noting *why* this naive split is safe here specifically (no BEGIN...END trigger bodies in `BASELINE_SQL` itself, unlike the FTS triggers which are added via separate `tx.execute` calls) so a future migration author doesn't copy the same `.split(';')` pattern for a migration that *does* need multi-statement trigger bodies inline.

### Overall Recommendation

**APPROVE**

The core deliverable — a versioned, transactional migration runner replacing the ad-hoc `CREATE IF NOT EXISTS` approach — is implemented correctly and meets every acceptance criterion in issue #1320. The highest-risk item (baseline fidelity) was independently verified against git history and is exact. Both bugs found during manual testing (synchronous-in-transaction, duplicate-column on pre-runner DBs) are fixed soundly with a real regression test for the latter. The gaps noted above (partial-batch-failure test, concurrent-migration-open test, and restoring the deleted test's full assertion set) are genuine but non-blocking — they harden an already-correct implementation rather than fix a defect. Recommend a fast follow-up (or addressing now if convenient) rather than blocking this PR on them.

malibio (commented):
Review posted — see comment above for full report. Recommendation: APPROVE (self-approval blocked by GitHub; author and reviewer account are the same).

malibio:
## Address Review Summary

Addressed 2 of the 3 suggested improvements from the review; skipped 1 nitpick as not worth the churn.

### ✅ Addressed
- **Restored deleted test's assertions** — `pre_runner_database_with_origin_already_applied_upgrades_without_error` now inserts a row before the simulated legacy `ALTER`, and asserts both the `origin` column's default value survives and the rebuilt index still leads with `origin` — matching what the deleted `test_migrate_embedding_origin_upgrades_legacy_db` unit test verified.
- **Added `every_version_up_to_latest_has_a_defined_migration` test** — guards against `LATEST_VERSION` being bumped without a matching `apply_migration` arm, which would otherwise panic (`unreachable!`) against a real user database instead of failing in CI.
- **Documented the `.split(';')` safety assumption** in `v001_initial_schema.rs` so a future migration author doesn't copy the pattern for DDL containing multi-statement trigger bodies.

### ⏭️ Skipped
- **Mid-batch failure/recovery test** — the consistency guarantee here is structural (each migration runs in its own SQLite transaction, and `user_version` is bumped inside that same transaction before commit), not runner-specific logic. Testing it would mean simulating a crash mid-transaction, which SQLite's own transaction semantics already prove correct; disproportionate effort for this PR's scope.
- **Concurrent `SqliteStore::new()` race test** — the exact `busy_timeout` retry mechanism this would exercise is already covered by `test_concurrent_writers_retry_instead_of_erroring_on_busy` in `sqlite_store/mod.rs`. A migration-specific duplicate would be redundant.
- **Removing the dead `migration_name` `"unknown"` arm** — harmless, unreachable defensive code; not worth the churn.

Full test suite (976 lib tests, all `packages/core/tests/*.rs` integration binaries, `quality:fix`) passes clean. Pushed as commit 6ddfcf8f.

## Re-Review Decision

**Rationale**: All changes are additive test coverage plus one doc comment — no logic changes to the migration runner itself. The original review's APPROVE recommendation stands.

PR #1623, opened 2026-07-11
Add PTY agents (Claude Code, Gemini, Codex) to AI Chat model selector

malibio:
Closes #1450

### DIFF
--- a/packages/desktop-app/src/lib/components/viewers/ai-chat-model-selector.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/ai-chat-model-selector.svelte
@@ -80,9 +80,12 @@
 
   // PTY agents (Claude Code, Gemini CLI, Codex, ...) — excludes agentStore's
   // "local:" entries, which are in-process llama.cpp models already surfaced
-  // in the "Local" section above.
+  // in the "Local" section above. The full list (available + unavailable) is
+  // needed to render unavailable agents disabled rather than hiding them.
   const ptyAgents = $derived(agentStore.agents.filter((a) => !isLocalAgent(a.id)));
-  const availablePtyAgents = $derived(ptyAgents.filter((a) => a.available));
+  const availablePtyAgents = $derived(
+    agentStore.availableAgents.filter((a) => !isLocalAgent(a.id))
+  );
 
   function isReady(m: ChatModelEntry): boolean {
     return m.status?.status === 'ready' || m.status?.status === 'loaded';
--- a/packages/desktop-app/src/lib/components/viewers/ai-chat-model-selector.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/ai-chat-model-selector.svelte
@@ -37,6 +37,7 @@
   import { addTab, navigationStore, setActiveTab } from '$lib/stores/navigation.svelte';
   import { createLogger } from '$lib/utils/logger';
   import type { OpenAiCompatConfig } from '$lib/types/ai-chat-node';
+  import { agentStore, isLocalAgent } from '$lib/stores/agent-store.svelte';
 
   const log = createLogger('AiChatModelSelector');
 
@@ -45,7 +46,7 @@
   // Sentinel values used in the <select> value attribute.
   const SETUP_SENTINEL = '__setup__';
   const HEADER_SENTINEL_PREFIX = '__header__';
-  const PTY_SENTINEL = '__pty__';
+  const PTY_PREFIX = 'pty:';
 
   let {
     nodeId,
@@ -76,6 +77,12 @@
   const nativeModels = $derived(models.filter((m) => m.backend === 'gguf'));
   const ollamaModels = $derived(models.filter((m) => m.backend === 'ollama'));
   const ramTooLow = $derived(ramGb > 0 && ramGb < MIN_RAM_GB);
+
+  // PTY agents (Claude Code, Gemini CLI, Codex, ...) — excludes agentStore's
+  // "local:" entries, which are in-process llama.cpp models already surfaced
+  // in the "Local" section above.
+  const ptyAgents = $derived(agentStore.agents.filter((a) => !isLocalAgent(a.id)));
+  const availablePtyAgents = $derived(ptyAgents.filter((a) => a.available));
 
   function isReady(m: ChatModelEntry): boolean {
     return m.status?.status === 'ready' || m.status?.status === 'loaded';
@@ -136,6 +143,9 @@
 
   onMount(async () => {
     await refresh();
+    if (agentStore.agents.length === 0) {
+      await agentStore.refreshAgents();
+    }
 
     if (isTauri) {
       try {
@@ -231,12 +241,16 @@
       return;
     }
 
-    if (value === PTY_SENTINEL) {
-      log.debug('PTY agent session selected', { nodeId });
-      // No model — the specific harness is chosen inside AiChatPtySession.
-      onSelect?.({ provider: 'pty', modelId: '' });
+    if (value.startsWith(PTY_PREFIX)) {
+      const agentId = value.slice(PTY_PREFIX.length);
+      log.debug('PTY agent selected', { provider: 'pty', agentId, nodeId });
+      onSelect?.({ provider: 'pty', modelId: agentId });
       return;
     }
+  }
+
+  function ptyValue(agentId: string): string {
+    return `${PTY_PREFIX}${agentId}`;
   }
 
   function downloadBadgeText(m: ChatModelEntry): string {
@@ -314,10 +328,16 @@
         </optgroup>
       {/if}
 
-      <!-- ── Agent CLI (PTY) ── -->
-      <optgroup label="Agent CLI">
-        <option value={PTY_SENTINEL}>Launch agent session…</option>
-      </optgroup>
+      <!-- ── PTY Agents ── -->
+      {#if availablePtyAgents.length > 0}
+        <optgroup label="PTY Agents">
+          {#each ptyAgents as agent (agent.id)}
+            <option value={ptyValue(agent.id)} disabled={!agent.available}>
+              {agent.name}{agent.available ? '' : ' (not installed)'}
+            </option>
+          {/each}
+        </optgroup>
+      {/if}
 
       <!-- ── Action: Set up… ── -->
       <optgroup label="─────────────────" disabled></optgroup>
--- a/packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte
@@ -93,7 +93,9 @@
         ? `openai-compat:${model}`    // model = config UUID
         : provider === 'ollama'
           ? model                      // model = full daemon ID "ollama:<name>"
-          : `native:${model}`
+          : provider === 'pty'
+            ? `pty:${model}`           // model = agent id, e.g. "claude-code"
+            : `native:${model}`
       : ''
   );
 
@@ -161,12 +163,10 @@
     }
 
     if (selection.provider === 'pty') {
-      // PTY sessions store no messages and no model — the conversation lives
-      // in the external harness. AiChatPtySession owns harness selection and
-      // capture:* properties from here. Explicitly clear `model` with `null`:
-      // the backend's namespaced-properties merge is additive (omitted keys
-      // are left as-is), so a prior native/ollama selection would otherwise
-      // linger on the node after switching to PTY.
+      // PTY sessions store no messages — the conversation lives in the
+      // external harness. `model` holds the chosen agent id (e.g.
+      // "claude-code") so AiChatPtySession can pre-select it in the launch
+      // config; capture:* properties are filled in separately once launched.
       const current = sharedNodeStore.getNode(nodeId) as unknown as AiChatNode | undefined;
       sharedNodeStore.updateNode(
         nodeId,
@@ -175,7 +175,7 @@
             messages: current?.messages ?? [],
             status: current?.status ?? 'active',
             provider: 'pty',
-            model: null,
+            model: selection.modelId || null,
           },
         },
         { type: 'viewer', viewerId: 'ai-chat-viewer' }
--- a/packages/desktop-app/src/lib/components/viewers/ai-chat-pty-session.svelte
+++ b/packages/desktop-app/src/lib/components/viewers/ai-chat-pty-session.svelte
@@ -151,6 +151,14 @@
   let availabilityLoading = $state(true);
 
   onMount(async () => {
+    // Pre-select the agent chosen in the header AiChatModelSelector, when one
+    // was already stored on the node (issue #1450); otherwise keep the
+    // 'claude-code' default.
+    const nodeModel = node?.properties?.model as string | undefined;
+    if (nodeModel) {
+      selectedAgent = nodeModel;
+    }
+
     try {
       const [settings, availResult] = await Promise.all([
         getCaptureSettings(),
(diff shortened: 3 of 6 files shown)


malibio:
### Code Review Summary

This PR replaces the generic "Launch agent session…" PTY entry point with a per-agent "PTY Agents" section in the model selector, sourced from `agentStore`, and threads the chosen agent id through to `AiChatPtySession` so it pre-selects the right harness. The diff is small, atomic, well-tested (8 focused unit tests, all passing), and the code itself is clean and consistent with the surrounding patterns (sentinel/prefix encoding, `$derived` filtering, additive-merge-safe null handling).

However, tracing the data source the new UI depends on (`agentStore.refreshAgents()` → `tauriCommands.acpRefreshAgents()`) shows it is a **hardcoded stub that always returns `[]`**, with no registered Tauri/gRPC command backing it (`acp_refresh_agents` doesn't exist anywhere in `src-tauri` or the daemon — only the unrelated, already-wired `check_agent_availability` RPC does, which `AiChatPtySession`'s own separate hardcoded `AGENT_OPTIONS` list already uses). In the shipped Tauri app, `agentStore.agents` will only ever contain `local:`-prefixed model entries; the PR's `ptyAgents` filter (`!isLocalAgent(a.id)`) will always evaluate to an empty array, so the new "PTY Agents" section this PR was written to add will not actually render for real users — only in non-Tauri/mock dev mode, where `MOCK_AGENTS` fills in. This is a **functional gap for the PR's primary purpose**, not a pre-existing issue this PR merely inherits passively — the PR is specifically titled/scoped to make PTY agents selectable from this dropdown, and that outcome does not occur outside of mocked dev data.

The change is a net positive over the status quo (a single opaque "Launch agent session…" entry) *for the code paths it touches directly* (type threading, node-viewer routing, PTY-session pre-selection), and the plumbing is correct and will "just work" the moment `acpRefreshAgents`/`agentStore` is wired to a real backend. But as it stands, the acceptance criteria are not met end-to-end in the actual application.

### Findings

#### Critical Issues

- **`packages/desktop-app/src/lib/stores/agent-store.svelte.ts:168-170` (via `tauri-commands.ts`)** — `agentStore.refreshAgents()` populates `this.agents` from `tauriCommands.acpRefreshAgents()`, which is a permanent no-op stub (`return [];`, no `invoke()` call, no `isTauri()` branch at all). There is no `acp_refresh_agents` command registered in `src-tauri/src/lib.rs` or implemented in the daemon — the real, working agent-availability RPC in this codebase is `check_agent_availability` (`CheckAvailabilityRequest`/`Result`), already consumed separately by `AiChatPtySession` via `ptyCheckAgentAvailability()`. Net effect: in a real Tauri build, `agentStore.agents` will only ever contain `local:`-prefixed model entries (from `chatModelList()`), so `ptyAgents` in `ai-chat-model-selector.svelte:84` (which filters out `isLocalAgent`) is always empty, and the "PTY Agents" `<optgroup>` never renders. **AC 1–3 and 7 ("PTY Agents section shown when ≥1 agent available", "each available agent selectable", "unavailable hidden/disabled", "agent detection reads from `agentStore.availableAgents`") are not satisfied in the shipped app** — only in non-Tauri/mock dev mode.
  - Fix options: (a) wire `agentStore`'s ACP path to `ptyCheckAgentAvailability()` (the data source `AiChatPtySession` already trusts) instead of the dead `acpRefreshAgents()`/`acp_list_agents` stub, using the existing hardcoded `AGENT_OPTIONS` id list (or centralize that list) as the candidate set; or (b) if landing the real wiring is out of scope for this PR, the issue/PR should say so explicitly and either scope this PR down to "wire the UI, defer live detection" or block on the detection fix landing first. Right now the PR description and commit message read as if this is fully functional end-to-end.

#### Suggested Improvements

- **`packages/desktop-app/src/lib/components/viewers/ai-chat-model-selector.svelte:84`** — Issue #1450 AC explicitly says "Agent detection reads from `agentStore.availableAgents`" (the store already exposes this getter, filtering `a.available`). The implementation instead reads `agentStore.agents` and filters locally into two derived lists (`ptyAgents`, `availablePtyAgents`). That's a reasonable technical choice since the UI needs *both* available and unavailable agents (to render disabled options) and `availableAgents` alone can't provide the unavailable set — but it diverges from what the AC names, and duplicates filtering logic that already exists on the store (`get availableAgents()`). Consider deriving `availablePtyAgents` from `agentStore.availableAgents.filter(a => !isLocalAgent(a.id))` for consistency with the store's public API, even though `ptyAgents` (the full list) still needs to come from `.agents`.

- **`packages/desktop-app/src/lib/components/viewers/ai-chat-pty-session.svelte`** — This component maintains its own independent, hardcoded `AGENT_OPTIONS` array (`claude-code`, `codex`, `gemini-cli`, `pi`, `open-code`) for its config-step dropdown, entirely separate from `agentStore`/`ai-chat-model-selector`'s agent list. The two lists happen to use consistent ids today, but nothing enforces that — a new agent added to one won't automatically appear in the other, and `ai-chat-model-selector`'s dropdown (sourced from the dead `acpRefreshAgents` stub) is already diverging from this list in practice. Once the detection-source issue above is fixed, consider having both components source the agent catalog from one place (e.g. `agentStore`, backed by `check_agent_availability`) to avoid this drift risk long-term. Not blocking for this PR's diff, but worth flagging since this PR increases the number of places that need to stay in sync.

#### Nitpicks

- **Nit:** `packages/desktop-app/src/lib/components/viewers/ai-chat-model-selector.svelte:144-147` — `onMount` calls `agentStore.refreshAgents()` only `if (agentStore.agents.length === 0)`. If a user has already visited a chat node this session (populating `agentStore.agents` with only local models, since ACP is a stub), reopening the selector on a different node will skip the refresh and never retry ACP detection even after e.g. installing a new CLI agent mid-session. Minor since `agentStore` is a session-lived singleton already used elsewhere this way, but worth a `refreshAgents()` call site review once the stub is fixed and refresh actually matters.
- **Nit:** `ai-chat-model-selector.svelte:337` — the unavailable-agent option text appends `" (not installed)"` even for agents whose `available: false` might stem from missing auth rather than missing binary (per `AiChatPtySession`'s richer `agentStatus()` — `binary_missing` vs `auth_missing` vs both). The model selector's simpler boolean doesn't distinguish these, which could mislead a user who has the CLI installed but not authenticated. Not blocking — the richer diagnostic already exists one click away in the PTY config screen — but a `title` tooltip reusing that distinction would improve clarity.

### Testing

- 8/8 new/updated unit tests pass (`bun run --cwd packages/desktop-app test src/tests/components/ai-chat-model-selector.test.ts`), covering: PTY-agent selection emitting `{provider:'pty', modelId}`, sentinel/header no-ops, local-agent exclusion from the PTY list, unavailable agents retained (not filtered) for disabled rendering, and the node-viewer's `pty` branch (model persisted vs. nulled).
- Full frontend suite: 4025/4027 passing; the 2 failures are in `architecture-benchmarks.test.ts`, a known machine-load-flaky performance test unrelated to this change.
- Gap: no test exercises the actual `agentStore.refreshAgents()` → `acpRefreshAgents()` path against a real/mocked Tauri `invoke`, which is exactly where the critical issue above lives — the tests validate the derivation logic assuming `agentStore.agents` is populated correctly, not that it ever will be in production.

### Recommendation

**Request changes.** The code that was written is clean, well-tested, and correctly threads the PTY agent id through node persistence and pre-selection (AC 4, 5, 6 are genuinely satisfied). But the feature's headline behavior — PTY agents appearing as selectable options in the model dropdown — does not occur in the real app because it's built on `agentStore`'s ACP detection path, which is a dead stub left over from #1117 with no backing Tauri/gRPC command. This should either be wired to the real `check_agent_availability` RPC (already used successfully by `AiChatPtySession`) before merge, or the PR scope/description should be revised to make clear this only works with mock data today and file a fast-follow issue for live detection.

malibio (commented):
Review recommendation: REQUEST CHANGES (GitHub blocks formal 'request changes' review on one's own PR — recorded here instead). See the full review posted as a PR comment above for details: the PTY-agent list added to the model selector is sourced from agentStore, whose ACP path (acpRefreshAgents) is a hardcoded stub returning [] with no backing Tauri/gRPC command, so the new 'PTY Agents' section will not render in the real app — only with mock dev data. Recommend wiring to the existing check_agent_availability RPC (already used by AiChatPtySession) before merge, or explicitly rescoping this PR.

malibio:
## Address Review Summary

**Decision**: Addressed the Critical finding directly (commit fb964423) rather than delegating — the fix was already scoped and understood while writing this reply.

### ✅ Addressed

- **🔴 Critical — dead ACP detection stub**: `agentStore.refreshAgents()` now calls `ptyCheckAgentAvailability()` (`check_agent_availability`, the RPC `AiChatPtySession`'s own config screen already uses successfully) instead of the permanent `acpRefreshAgents()` stub that always returned `[]`. Added `availabilityToAgent()` mapping (binary/auth found → `available`, with a display-label lookup mirroring `AiChatPtySession.AGENT_OPTIONS`) plus 4 new unit tests covering the mapping (known label, unknown-agent fallback, binary-missing, auth-missing). This makes the "PTY Agents" section in the model selector actually populate for real Tauri users, not just mock/dev mode.
- **🟢 Suggestion — derive from `agentStore.availableAgents`**: `availablePtyAgents` in `ai-chat-model-selector.svelte` now filters `agentStore.availableAgents` (the store's public getter) instead of re-filtering `.agents` locally, matching the wording of acceptance criterion 7 in issue #1450.

### ⏭️ Skipped

- **🟢 Nitpick — refresh-skip-if-already-populated staleness**: `onMount` still only calls `refreshAgents()` `if (agentStore.agents.length === 0)`. Left as-is — `agentStore` is a session-lived singleton already used this way elsewhere in the app, and re-running detection on every dropdown open is a separate UX/polling concern beyond this PR's scope.
- **🟢 Nitpick — "not installed" label conflates missing-binary vs missing-auth**: Left as-is. The richer breakdown (`binary_missing` / `auth_missing` / both) already exists one click away in `AiChatPtySession`'s config screen via `agentStatus()`; duplicating that distinction into the top-level dropdown's option text is a UX nicety, not a correctness issue, and out of scope for this PR.
- **🟡 Suggested improvement — unify `AiChatPtySession`'s hardcoded `AGENT_OPTIONS` with `agentStore`**: Left as-is per the reviewer's own note that this is "not blocking for this PR's diff" — both lists are now sourced from the same underlying daemon catalog (`check_agent_availability`), so the drift risk it flagged is reduced, but consolidating the two hardcoded label lists into one shared source is a larger refactor better done separately.

### Testing

- `bun run test` — 159 files / 4031 tests passing (baseline 4025 + 6 net new across both commits).
- `bun run --cwd packages/desktop-app quality:check` — 0 errors, 0 warnings.
- `bun run test:all` (frontend + browser + Rust) passed in full via the pre-push gate on both commits.

## Re-Review Decision

**Rationale**: The Critical finding was addressed with a real architectural change — swapping the data source an entire UI section depends on (from a dead stub to a live RPC) and changing the shape/derivation of `agentStore.agents`. This is exactly the kind of change the address-review guidance flags as warranting verification: it's not a trivial or purely cosmetic fix, and I want independent confirmation that the new `ptyCheckAgentAvailability()` wiring is correct and that no other consumer of `agentStore` was broken by the mapping change.

malibio:
### Code Review Summary (Re-Review — delta only, `review-20260711-073320`...`fb964423`)

This re-review covers only the follow-up commit (`fb964423`) that addressed the prior Critical finding. Scope: `agent-store.svelte.ts` (+21/-3), `ai-chat-model-selector.svelte` (+5/-2), and new unit tests in `agent-store.test.ts` (+66). I did not re-litigate the rest of the PR.

**Verdict: the Critical issue is genuinely fixed, correctly, with no regressions.**

### Findings

#### Critical Issues
None. The previously-blocking issue is resolved.

- **`agent-store.svelte.ts:141-151`** — `refreshAgents()` now calls `tauriCommands.ptyCheckAgentAvailability()` (`invoke('check_agent_availability')`) instead of the dead `acpRefreshAgents()` stub. I confirmed `check_agent_availability` is a real, registered command: it's listed in the `generate_handler!` macro in `packages/desktop-app/src-tauri/src/lib.rs:544`, and its implementation in `packages/desktop-app/src-tauri/src/commands/agent_session.rs:343-365` makes a real gRPC call (`c.check_agent_availability(Request::new(CheckAvailabilityRequest {}))`) through `GrpcClient`, mapping the daemon response 1:1 into `CheckAvailabilityResult`. This is the exact same command `AiChatPtySession`'s config screen already calls successfully — not a new/unverified code path.
- **Field mapping (`AgentAvailabilityInfo` → `AcpAgentInfo`)** — verified end-to-end: the Rust struct uses `#[serde(rename_all = "camelCase")]` (`agent_type`→`agentType`, `binary_found`→`binaryFound`, `auth_found`→`authFound`, etc.), which matches the TS `AgentAvailabilityInfo` interface in `tauri-commands.ts:302-309` exactly, field-for-field. `availabilityToAgent()` (`agent-store.svelte.ts:95-104`) then maps `agentType`→`id`, looks up a display label with a sane fallback to the raw id, and correctly computes `available: info.binaryFound && info.authFound` (both binary presence *and* auth are required — matches the original review's own framing of "binary_missing vs auth_missing", both now folded correctly into a single boolean).
- **No regression for other `agentStore` consumers** — I grepped all non-test usages of `agentStore.agents` / `agentStore.availableAgents`: `agent-selector.svelte:8` and `ai-onboarding-dialog.svelte:20-21`. Both only read `agent.id`, `agent.name`, and `agent.available` off list entries — all three fields are populated identically in shape/type by the new mapping vs. the old stub/mock path, so nothing downstream breaks.

#### Suggested Improvements
None new. (The three items carried over from the original review — refresh-skip staleness, "not installed" label granularity, and unifying `AiChatPtySession`'s hardcoded `AGENT_OPTIONS` with `agentStore` — were all explicitly marked non-blocking/"not blocking for this PR's diff" in the original review text, and the reasoning given in the Address Review Summary for leaving them out of scope is consistent with that framing. Confirmed by re-reading the original review comment verbatim. No objection to deferring them.)

#### Nitpicks
- **Nit:** `agent-store.svelte.ts:86-92` — `PTY_AGENT_LABELS` is now a second hardcoded id→label map living alongside `AiChatPtySession`'s `AGENT_OPTIONS`. This is the same drift risk the original review flagged as a non-blocking suggestion, just now instantiated in a second location rather than resolved. Not a new problem introduced by this diff, and already acknowledged in the Address Review Summary as an accepted follow-up; just noting it didn't shrink.

### Testing

- `bun run --cwd packages/desktop-app test src/tests/stores/agent-store.test.ts src/tests/components/ai-chat-model-selector.test.ts` — **26/26 passing** (18 store tests including 4 new `availabilityToAgent` cases: known label, unknown-agent fallback, binary-missing, auth-missing; 8 model-selector tests unchanged and still green).
- `bun run --cwd packages/desktop-app quality:check` — **0 errors, 0 warnings** (eslint + svelte-check clean).
- The new `availabilityToAgent` tests directly exercise the mapping logic that was the crux of the fix, including the two distinct "unavailable" paths (missing binary vs. missing auth) — good targeted coverage for exactly what changed.

### Recommendation

**Approve.** The follow-up commit correctly replaces the dead `acpRefreshAgents()` stub with the real, already-proven `check_agent_availability` RPC, the field mapping is verified correct against the actual Rust struct definitions (not just assumed from naming), no other `agentStore` consumer is affected, and the change is covered by new unit tests. The PR now delivers on its stated acceptance criteria end-to-end, not just in mock/dev mode.

malibio (commented):
Re-review recommendation: APPROVE (GitHub blocks formal approval on one's own PR — recorded here instead). See full report posted as a PR comment above: the follow-up commit (fb964423) correctly replaces the dead acpRefreshAgents() stub with the real check_agent_availability RPC, field mapping verified correct, no downstream regressions, 26/26 targeted tests + quality:check clean.

PR #1624, opened 2026-07-11
Batch bulk embedding inserts instead of per-row execute loops

malibio:
Closes #1524

### DIFF
--- a/packages/core/src/db/sqlite_store/embeddings.rs
+++ b/packages/core/src/db/sqlite_store/embeddings.rs
@@ -62,41 +62,98 @@ impl SqliteStore {
         .context("Failed to delete existing embeddings")?;
 
         let now = Utc::now().to_rfc3339();
-        for emb in embeddings {
-            let id = uuid::Uuid::new_v4().to_string();
-            let dimension = emb.vector.len() as i64;
-            let vector_blob: Vec<u8> = emb.vector.iter().flat_map(|f| f.to_le_bytes()).collect();
-            let model_name = emb
-                .model_name
-                .unwrap_or_else(|| "nomic-embed-text-v1.5".to_string());
+        let rows: Vec<(String, Vec<u8>, Vec<libsql::Value>)> = embeddings
+            .into_iter()
+            .map(|emb| {
+                let id = uuid::Uuid::new_v4().to_string();
+                let dimension = emb.vector.len() as i64;
+                let vector_blob: Vec<u8> =
+                    emb.vector.iter().flat_map(|f| f.to_le_bytes()).collect();
+                let model_name = emb
+                    .model_name
+                    .unwrap_or_else(|| "nomic-embed-text-v1.5".to_string());
 
-            tx.execute(
-                "INSERT INTO embedding (id, node_id, vector, dimension, model_name, chunk_index, chunk_start, chunk_end, total_chunks, content_hash, token_count, stale, error_count, last_error, origin, created_at, modified_at) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10, ?11, 0, 0, NULL, ?12, ?13, ?14)",
-                libsql::params![
-                    id.clone(),
-                    emb.node_id.clone(),
-                    vector_blob.clone(),
-                    dimension,
-                    model_name,
-                    emb.chunk_index as i64,
-                    emb.chunk_start as i64,
-                    emb.chunk_end as i64,
-                    emb.total_chunks as i64,
-                    emb.content_hash,
-                    emb.token_count as i64,
-                    origin.to_string(),
-                    now.clone(),
-                    now.clone(),
-                ],
-            ).await.context("Failed to insert embedding")?;
+                let params = vec![
+                    libsql::Value::Text(id.clone()),
+                    libsql::Value::Text(emb.node_id.clone()),
+                    libsql::Value::Blob(vector_blob.clone()),
+                    libsql::Value::Integer(dimension),
+                    libsql::Value::Text(model_name),
+                    libsql::Value::Integer(emb.chunk_index as i64),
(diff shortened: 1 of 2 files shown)


malibio:
## Code Review — PR #1624: Batch bulk embedding inserts

**Review Type:** Initial review (no prior reviews/comments on the PR).

### Requirements Check (issue #1524)

- [x] **Bulk embedding inserts use batched statements** — Both target functions now build chunked multi-row `INSERT`s instead of per-row `tx.execute` loops:
  - `upsert_embeddings_with_origin` (covers both `upsert_embeddings` and `apply_remote_embeddings`): `embedding` table batched at `EMBEDDING_CHUNK = 60` (14 params/row → 840/statement), `vec_embeddings` mirror batched at `VEC_CHUNK = 400` (2 params/row → 800/statement).
  - `create_stale_embedding_markers_bulk`: batched at `ID_CHUNK = 180` (5 params/row → 900/statement).
- [x] **Code passes quality/tests** — `cargo fmt --check -p nodespace-core` clean; `cargo clippy --lib -p nodespace-core` clean; both new `chunk_boundary` tests pass. (Skipped the workspace-wide `quality:fix` deliberately since it reformats unrelated files and fmt is already clean here.)

Both acceptance criteria are met.

### Code Review Findings

#### 🟢 Suggestions (all optional)

- 🟢 **`embeddings.rs:65-95` — upfront `rows` Vec clones each `vector_blob` twice.** Each row keeps `vector_blob` in the tuple *and* clones it into `params` (`Value::Blob(vector_blob.clone())`), then the vec-mirror loop clones it a third time. For 768-dim f32 vectors (~3KB each) this is a few extra copies per embedding. This is a bounded per-node write path (not a hot unbounded loop), so the readability win of the two-pass structure is a reasonable trade — **nit only, do not block**. YAGNI applies.

- 🟢 **`embeddings.rs:99, 135, 728` — chunk sizes are correct but the headroom rationale lives only in a comment.** The math (60×14=840, 400×2=800, 180×5=900, all < 999) is sound with comfortable headroom. Since params-per-row is coupled to the exact column list in the SQL string, a future column addition would silently shrink headroom. Not worth a guard today given the arithmetic is obvious and tests would catch a ceiling breach — noting for awareness only.

- 🟢 **Test coverage — boundary-exact cases not covered.** Tests exercise "comfortably over" the boundary (130 embeddings vs chunk 60 → 3 batches; 200+1 markers vs chunk 180 → 2 batches) rather than the exact boundary (e.g. exactly 60/61, or 0). Given the chunking is a straightforward `chunks()` iterator and the empty case is guarded by early returns (`embeddings.is_empty()` / `node_ids.is_empty()`), current coverage is pragmatically adequate. A single extra assert at N=EMBEDDING_CHUNK would be cheap insurance but is not required.

- 🟢 **Pre-existing (not introduced here): `create_stale_embedding_markers_bulk` returns `Ok(node_ids.len())`, not rows actually inserted.** With `INSERT OR IGNORE`, the return count over-reports when duplicates are deduped (the new test asserts `created == 201` while only 200 rows land). This semantic is **unchanged from `origin/main`** — verified via `git show origin/main:...embeddings.rs`. Flagging only so it's a conscious choice; out of scope for this PR.

### Verification Notes (independently confirmed)

- **Placeholder index math is correct, no off-by-one.** 1-indexed params, `base = i*N`, offsets `base+1..=base+N`. Boundary row i=0 → `?1..?14`; last row of a full 60-chunk → base 826, `?827..?840`. Param flattening (`flat_map(|(_,_,p)| p.clone())`) preserves the same row order as placeholder generation. Vec mirror `(?{i*2+1}, ?{i*2+2})` correct.
- **Transaction atomicity preserved.** Single `tx` created once, single `tx.commit()` after all chunks. An error mid-batch propagates via `?`, the function returns early, `tx` drops uncommitted → libsql rolls back. No explicit rollback needed; consistent with the DELETE-then-INSERT pattern already in this function. DELETE ordering (vec_embeddings before embedding) unchanged and correct.
- **Cross-chunk dedup is a genuinely strong test.** In `test_create_stale_embedding_markers_bulk_across_chunk_boundary` (mod.rs:955-988): 200 distinct ids + 1 duplicate of index 0. With `ID_CHUNK = 180`, the original (index 0) lands in **chunk 1** (indices 0–179) and the duplicate (index 200) lands in **chunk 2** (indices 180–200). So `INSERT OR IGNORE` dedupes against a row committed by a *prior batch statement in the same transaction* — a real cross-chunk dedup, not the trivial same-statement case. Good.
- **No collateral breakage.** All other callers of the three functions (grepped across `services/`, `node_service/`, `nodes.rs`) go through unchanged public signatures and behavior (ordering, transaction boundaries, return types all preserved).
- **CLAUDE.md compliance:** old per-row loops fully deleted (not commented/dead), no lint suppression, no `#[deprecated]`/`#[allow(dead_code)]`, no dual-code-path or backward-compat cruft. Change is atomic and single-purpose.

### Recommendation

**APPROVE.** Clean, net-positive change that directly satisfies #1524. Batching is correct, atomicity preserved, and the two new tests meaningfully exercise cross-batch-boundary behavior including cross-chunk `INSERT OR IGNORE` dedup. All findings are 🟢 optional nits — none block merge.

malibio (commented):
REVIEW VERDICT: APPROVE (submitted as comment because GitHub blocks approving one's own PR). See the full review in the PR comment above. Batching is correct, transaction atomicity is preserved, and the two new tests meaningfully exercise cross-batch-boundary behavior including cross-chunk INSERT OR IGNORE dedup. All findings are optional 🟢 nits — none block merge.

malibio (commented):
Review verdict: APPROVE (self-review restriction prevents a formal GitHub 'Approve' status — see full findings in the review comment above). Net-positive change; correctly implements #1524's batching requirement with atomicity and INSERT OR IGNORE dedup semantics preserved across chunk boundaries. All findings are optional nits — none blocking.

PR #1626, opened 2026-07-11
Sync-in-async bridge in the CEL graph resolver

malibio:
Closes #1523

### DIFF
--- a/packages/core/benches/performance.rs
+++ b/packages/core/benches/performance.rs
@@ -695,7 +695,11 @@ fn bench_graph_resolver(c: &mut Criterion) {
                     for _ in 0..iters {
                         let mut resolver = GraphResolver::new(Arc::clone(&svc_clone));
                         let start = std::time::Instant::now();
-                        black_box(resolver.resolve_path(&root, std::slice::from_ref(&rel_name)));
+                        black_box(
+                            resolver
+                                .resolve_path(&root, std::slice::from_ref(&rel_name))
+                                .await,
+                        );
                         total += start.elapsed();
                     }
                     total
@@ -717,7 +721,7 @@ fn bench_graph_resolver(c: &mut Criterion) {
                     for _ in 0..iters {
                         let mut resolver = GraphResolver::new(Arc::clone(&svc_clone));
                         let start = std::time::Instant::now();
-                        black_box(resolver.resolve_path(&root, &segments));
+                        black_box(resolver.resolve_path(&root, &segments).await);
                         total += start.elapsed();
                     }
                     total
@@ -742,7 +746,7 @@ fn bench_graph_resolver(c: &mut Criterion) {
                     for _ in 0..iters {
                         let mut resolver = GraphResolver::new(Arc::clone(&svc_clone));
                         let start = std::time::Instant::now();
-                        black_box(resolver.resolve_path(&root, &segments));
+                        black_box(resolver.resolve_path(&root, &segments).await);
                         total += start.elapsed();
                     }
                     total
@@ -764,10 +768,10 @@ fn bench_graph_resolver(c: &mut Criterion) {
                     for _ in 0..iters {
                         let mut resolver = GraphResolver::new(Arc::clone(&svc_clone));
                         // First call populates cache
-                        resolver.resolve_path(&root, &segments);
+                        resolver.resolve_path(&root, &segments).await;
                         // Second call should hit cache
                         let start = std::time::Instant::now();
-                        black_box(resolver.resolve_path(&root, &segments));
+                        black_box(resolver.resolve_path(&root, &segments).await);
                         total += start.elapsed();
                     }
                     total
@@ -899,7 +903,9 @@ fn bench_cel_evaluation_e2e(c: &mut Criterion) {
 
         group.bench_function("simple_property_no_resolver", |b| {
             b.iter(|| {
-                black_box(cel::evaluate_conditions(&conditions, &node, &evt, None));
+                rt.block_on(async {
+                    black_box(cel::evaluate_conditions(&conditions, &node, &evt, None).await);
+                });
             });
         });
     }
@@ -921,12 +927,10 @@ fn bench_cel_evaluation_e2e(c: &mut Criterion) {
                     for _ in 0..iters {
                         let mut resolver = GraphResolver::new(Arc::clone(&svc_clone));
                         let start = std::time::Instant::now();
-                        black_box(cel::evaluate_conditions(
-                            &conditions,
-                            &node,
-                            &evt,
-                            Some(&mut resolver),
-                        ));
+                        black_box(
+                            cel::evaluate_conditions(&conditions, &node, &evt, Some(&mut resolver))
+                                .await,
+                        );
                         total += start.elapsed();
                     }
                     total
@@ -952,12 +956,10 @@ fn bench_cel_evaluation_e2e(c: &mut Criterion) {
                     for _ in 0..iters {
                         let mut resolver = GraphResolver::new(Arc::clone(&svc_clone));
                         let start = std::time::Instant::now();
-                        black_box(cel::evaluate_conditions(
-                            &conditions,
-                            &node,
-                            &evt,
-                            Some(&mut resolver),
-                        ));
+                        black_box(
+                            cel::evaluate_conditions(&conditions, &node, &evt, Some(&mut resolver))
+                                .await,
+                        );
                         total += start.elapsed();
                     }
                     total
--- a/packages/core/benches/performance.rs
+++ b/packages/core/benches/performance.rs
@@ -648,8 +648,8 @@ fn bench_path_extraction(c: &mut Criterion) {
 /// relationship chains. Also benchmarks cache effectiveness by resolving
 /// the same path twice and comparing latency.
 ///
-/// Requires a multi-threaded tokio runtime since GraphResolver uses
-/// `block_in_place` internally.
+/// Uses a multi-threaded runtime to match production, not because
+/// GraphResolver requires one — it resolves paths with plain async/.await.
 fn bench_graph_resolver(c: &mut Criterion) {
     let rt = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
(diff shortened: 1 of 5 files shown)


malibio:
## Pragmatic Code Review — PR #1626

### Review Type
**Initial Review.** `gh pr view 1626` shows `reviews: []` and `comments: []` — no prior review activity exists on this PR.

### Requirements Check

Issue #1523 acceptance criteria (quoted verbatim):

> - [ ] The resolver no longer relies on `block_in_place` (or the requirement is guarded/documented)
> - [ ] Code passes `bun run quality:fix`; tests pass

| Criterion | Status |
|---|---|
| Resolver no longer relies on `block_in_place` | ✅ Verified. `fetch_related_nodes`, `resolve_path`, `resolve_collection`, `enrich_context` are now plain `async fn`s calling `NodeService` via `.await`. Repo-wide grep for `block_in_place`, `Handle::block_on`, `.block_on(` in `packages/core/src/playbook/` returns zero hits. |
| Code passes `quality:fix`; tests pass | ✅ `cargo fmt --check -p nodespace-core` is clean. `cargo test -p nodespace-core --lib graph_resolver` (23 passed), `--lib playbook::cel::` (27 passed), `--lib playbook::actions::` (35 passed) — all green, 0 failures. `cargo check -p nodespace-core --benches` compiles cleanly. |

Both stated acceptance criteria are met. See findings below for gaps not captured by the criteria as written (test *strength*, not just test *presence*).

### Code Review Findings

#### 🔴 Critical
None. The core architectural fix is sound and the sync/async boundary is correctly relocated to the CEL-evaluation edge (per the updated module doc comment in `graph_resolver.rs`). Nothing here blocks merge.

#### 🟡 Important

**1. No regression test actually exercises the deadlock scenario this PR fixes.**
`packages/core/src/playbook/graph_resolver.rs:651,667,683,721,794,834,881,927,943,963,1040,1089` — every single test in this file uses `#[tokio::test(flavor = "multi_thread")]`, unchanged by this PR (only `.await` was mechanically threaded through). None use the default `#[tokio::test]` (which is `current_thread`) or an explicit `flavor = "current_thread"`.

Under the *old* `block_in_place` implementation, `flavor = "multi_thread"` was **required** for these tests to pass at all — `block_in_place` panics/is a no-op-hang risk on a current-thread runtime. So these tests already existed as an implicit guard against running on the wrong flavor, but they prove the old code needed multi-threaded; they don't prove the new code works *without* it. A test that pins `flavor = "current_thread"` and calls `resolve_path`/`resolve_collection`/`enrich_context` is the one piece of evidence that would directly demonstrate the bug is fixed (it would have deadlocked/panicked before this change and now passes). As it stands, the fix is verified only by code inspection + the fact that `multi_thread` tests still pass — which they always did, before and after.

Recommendation: add at least one test like:
```rust
#[tokio::test(flavor = "current_thread")]
async fn resolve_path_works_under_current_thread_runtime() { ... }
```
exercising `resolve_path` (and ideally `enrich_context`, since that's the call site actually invoked from `evaluate_conditions`/`engine.rs`). This is cheap to add and is the single highest-value test this PR is missing.

**2. `cel.rs` / `actions.rs` tests run under `current_thread` by coincidence, but never touch the resolver path, so they don't corroborate the fix either.**
`packages/core/src/playbook/cel.rs:628` onward and `packages/core/src/playbook/actions.rs` (`BindingContext` tests) use bare `#[tokio::test]`, which defaults to `current_thread`. This looked promising at first — bare `#[tokio::test]` is exactly the runtime flavor that used to deadlock. However:
- Every `evaluate_conditions(...)` call in `cel.rs` tests passes `None` for the resolver (e.g. `cel.rs:628-799`).
- Every `BindingContext::new(&node, &event, None)` call in `actions.rs` tests (e.g. `actions.rs:848` onward) passes `None` for `graph_resolver`.

Since `resolve_trigger_path`'s resolver branch (`actions.rs:209`, `if let Some(ref mut resolver) = self.graph_resolver`) is only reached when a resolver is actually present, these `current_thread` tests never execute a code path that used to call `block_in_place`. They validate the mechanical `.await`-threading compiles and preserves existing non-graph behavior, nothing more. Not a blocker by itself, but combined with finding #1, it means **the whole diff has zero tests that would have failed against the pre-fix code on a single-threaded runtime.**

**3. Stale doc comment in the benchmark file.**
`packages/core/benches/performance.rs:651-652`:
```rust
/// Requires a multi-threaded tokio runtime since GraphResolver uses
/// `block_in_place` internally.
fn bench_graph_resolver(c: &mut Criterion) {
```
This comment is now factually wrong — `GraphResolver` no longer uses `block_in_place`, and (per finding #1) the whole point of this PR is that it no longer *requires* a multi-threaded runtime. Leaving stale rationale in a comment is exactly the kind of drift this PR is trying to eliminate elsewhere (the module doc comment in `graph_resolver.rs` was properly updated). The benchmark harness itself still legitimately uses `Builder::new_multi_thread()` (that's an orthogonal, reasonable choice for a benchmark driver — nothing wrong with using multi-thread for realistic perf measurement), but the comment's stated *reason* is stale and should either be deleted or reworded (e.g. "uses a multi-threaded runtime to match production" rather than implying it's still a correctness requirement).

#### 🟢 Suggestions / Nits

**Nit:** `packages/core/src/playbook/actions.rs:330-350` (`resolve_bindings_in_value`) — the `Box::pin(...).await` recursion pattern is correctly applied to both the `Object` and `Array` branches and compiles/passes as verified by the test run. `Box::pin` is the right call here over pulling in `async-recursion` (not currently a dependency — adding a macro crate for two call sites would be over-engineering per YAGNI) and over restructuring to an iterative worklist (would meaningfully complicate the code for a traversal that's bounded by action-param JSON depth, typically shallow, and runs once per action per rule evaluation — not a tight inner loop). The one-`Box`-allocation-per-recursion-level overhead here is a non-issue: this runs per playbook action execution, not in a hot loop, and JSON param nesting in practice is a handful of levels deep at most.

**Nit:** `packages/core/src/playbook/actions.rs:334` — the array branch was changed from an iterator `.map(...).collect()` to an explicit `for` loop with `Vec::with_capacity`. This is a necessary and correct adaptation (you can't easily `.await` inside a `.map` closure over `Result` collection), not a style regression — flagging only to note it's intentional, not an oversight.

**Nit:** Benchmark correctness (verification area 4) — confirmed `bench_graph_resolver` and `bench_cel_evaluation_e2e` use the correct pattern for benchmarking async code with criterion: `b.iter_custom(|iters| { rt.block_on(async { ... }) })`, with `std::time::Instant::now()`/`.elapsed()` measurement taken *inside* the async block and scoped tightly around just the `resolve_path`/`evaluate_conditions` call (setup — `GraphResolver::new`, in the 1-hop/3-hop/5-hop cases — happens before `start = Instant::now()`, and `total` accumulates only the measured span). This is a legitimate variant of criterion's async support (manual `iter_custom` + `rt.block_on`, as opposed to `Criterion::to_async(&rt).iter(...)`) and the measured scope is not accidentally including setup/teardown. No regression in benchmark fidelity from this change.

### Summary of the Four Requested Verification Areas

1. **Completeness of async conversion** — Complete. No remaining `block_in_place`/`Handle::block_on`/`.block_on(` in the playbook module or any of its callers (`engine.rs`, `actions.rs`, benches). Repo-wide search found only the benchmark harness's own `rt.block_on` (a legitimate, pre-existing pattern for driving async setup/measurement from a sync bench function, not a sync-bridge-inside-async-code smell) and unrelated top-level runtime bootstrapping in `desktop-app/src-tauri/src/main.rs` and `daemon/src/main.rs` (process entry points, correctly out of scope).
2. **`Box::pin` recursion pattern** — Correctly implemented, compiles, and passes tests for both object and array traversal. Right tradeoff versus `async-recursion` (avoids a new dependency) or an iterative worklist (unnecessary complexity for this call frequency/depth).
3. **Test coverage** — This is the real gap (see Important #1 and #2 above): the diff mechanically converts existing tests to `.await` without adding any test that would have caught the original bug (a `current_thread`-flavored test that actually exercises `resolve_path`/`resolve_collection`/`enrich_context`). The fix is correct by inspection and by the removal of the anti-pattern, but there's no regression test proving it.
4. **Benchmark correctness** — Correct. Uses `iter_custom` + `rt.block_on`, a valid criterion async pattern, with measurement scoped tightly to the operation under test.

### Recommendation

**COMMENT.** The architectural fix itself is a clean, atomic, net-positive change that directly addresses the issue's acceptance criteria, and both stated criteria pass. This is not a blocker. However, I'd strongly encourage adding the missing `current_thread`-flavor regression test (Important #1) before merge — it's cheap, and it's the difference between "we changed the code so it no longer contains the anti-pattern" and "we have proof the anti-pattern is gone." The stale doc comment (Important #3) is a one-line fix. Given the "Net Positive > Perfection" philosophy and that this is pre-release/no-users code, I would not block merge on these, but they're worth addressing either in this PR or as fast immediate follow-up before this pattern is trusted as "fixed" long-term.

malibio:
Addressed review feedback:
- Added `resolve_path_and_enrich_context_work_under_current_thread_runtime` — a `#[tokio::test(flavor = "current_thread")]` regression test that would have deadlocked/panicked under the old `block_in_place` bridge, now passes.
- Fixed the stale doc comment on `bench_graph_resolver` that claimed it still required a multi-threaded runtime because of `block_in_place`.

Both stated acceptance criteria were already met; these address the reviewer's Important-severity gaps. `test:all` + `quality:fix` clean, pushed in 1369fae9.

malibio:
Rebased onto main to resolve a conflict with #1625 (CEL program caching), which changed `evaluate_conditions`'s signature from `&[String]` to `&[CompiledCondition]` and removed `ConditionResult::CompileError`. Reconciled: kept the new signature/type, added `async`/`.await` on top, updated test call sites to the `conds()` helper, and dropped the one test that exercised the now-removed `CompileError` variant.

No new logic — mechanical rebase only. `cargo clippy -D warnings -D clippy::unused-async`, `quality:fix`, and full `test:all` (979 Rust + 4031 frontend + 17 browser tests) all clean. Force-pushed.

PR #1627, opened 2026-07-11
Make the agent scenario matrix (aichat-matrix.ts) an assertable CI eval, not just an observation harness

malibio:
Closes #1364

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "install:all": "bun install",
     "chat": "bun run scripts/aichat.ts",
     "eval:routing": "bun run scripts/routing-eval.ts",
+    "eval:agent": "bun run scripts/aichat-matrix.ts",
     "prompt:dump": "bun run scripts/prompt-dump.ts",
     "build:sidecars": "bun run scripts/build-sidecars.ts",
     "build:pkg:macos": "bash scripts/build-pkg.sh",
--- a/scripts/aichat-matrix.ts
+++ b/scripts/aichat-matrix.ts
@@ -1,24 +1,37 @@
 #!/usr/bin/env bun
 /**
- * aichat-matrix.ts — drive the issue #1329 A/B scenario matrix through aichat.ts.
+ * aichat-matrix.ts — assertable end-to-end agent-behavior eval (issue #1364).
  *
- * Runs the 8 scenarios from #1329 once against the model the daemon currently has
- * loaded, capturing per-scenario tools-offered, tools-called, the assistant reply,
- * and wall-clock latency into a JSON results file. Run it once per model to A/B/C
- * (E4B vs 12B vs 31B).
+ * Runs the #1329 scenario matrix (greeting, schema creation, instance creation,
+ * list/query, update, empty-result, multi-type CRUD) through aichat.ts and asserts
+ * a structured, machine-checkable expectation per scenario — not just captures
+ * prose for a human to read.
+ *
+ * This is the third eval layer, distinct from the other two:
+ *   - chat::parser::tests            — tool-call PARSING shape (fixtures)
+ *   - scripts/routing-eval.ts        — skill ROUTING accuracy (which skill fires)
+ *   - scripts/aichat-matrix.ts (this)— END-TO-END behavior (right tool, right
+ *                                      number of times, right effect)
+ *
+ * Under ADR-038 the model calls search_skills before acting, so assertions check
+ * for the TARGET tool tolerating search_skills, never raw tool count.
  *
  * The daemon, socket, and DB are managed by the caller (see #1329 reproduce steps).
  * This script only talks to the already-running test daemon via scripts/aichat.ts,
- * reusing its log-scraping for tool calls.
+ * reusing its log-scraping for tool calls. Skills must be seeded before running.
  *
  * Usage:
  *   NS_BIN=target/release/nodespace \
  *   NODESPACED_SOCKET=/tmp/nodespaced-test/daemon.sock \
  *   NS_LOG=/tmp/nodespaced-test/daemon.log \
  *   NS_MODEL=<id> NS_TIMEOUT_MS=240000 \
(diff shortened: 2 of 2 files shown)


malibio:
## Pragmatic Code Review — PR #1627

### Requirements Check (issue #1364)

- ✅ Each scenario has a structured, machine-checkable `Expectation` (not prose) — `scripts/aichat-matrix.ts:74-85`.
- ✅ Runner asserts per-scenario pass/fail and exits non-zero on failure — `assertExpectation` (154-222), exit codes at 355 (baseline regression) and 367 (any scenario failure).
- ✅ Assertions check for the target tool, tolerating `search_skills`, not raw tool count — `actionTools()` filters `ROUTING_TOOLS` before every check (150-152, 158).
- ✅ Empty-result (`noRetry`, scenario 7) and no-extra-types (`noExtraTypes`, scenario 3) are explicitly asserted — 251-258 (scenario 7), 233-238 (scenario 3).
- ✅ Exposed as a bun script: `"eval:agent": "bun run scripts/aichat-matrix.ts"` — `package.json:35`.
- ✅ Baseline-comparison mode mirroring `routing-eval.ts`'s `ROUTING_BASELINE` — `AICHAT_MATRIX_BASELINE` (59, 335-361), same regression/exit-code shape.
- ✅ Documented in `nodespace-docs/development/agent-eval.md` — accurate, describes the three-layer eval model, usage, and all 8 scenarios; committed separately to nodespace-docs per project convention.
- ✅ `bun run quality:fix` — ran it; only touches `packages/desktop-app` and Rust, doesn't cover top-level `scripts/*.ts` at all (confirmed against `package.json`'s `quality:fix` definition), so there's nothing for that gate to flag here. See note below on `bunx prettier` noise.

All 8 acceptance criteria are met.

### Code Review Findings

#### 🟢 Suggestions

**`scripts/aichat-matrix.ts` formatting vs. raw `prettier`**
Running `bunx prettier --check` directly (not through the project's `quality:fix`, which doesn't touch top-level `scripts/*.ts`) flags long lines in this file. I checked the sibling file this PR mirrors — `scripts/routing-eval.ts` — and it has the identical characteristic, along with `scripts/aichat.ts` and `scripts/prompt-dump.ts`. This is a pre-existing repo convention (top-level scripts aren't Prettier-gated), not a regression from this change, and consistent with the pattern being mirrored. No action needed.

**`toolSequence` matching is a greedy subsequence check, not "consecutive order"**
`scripts/aichat-matrix.ts:182-195` — the loop advances `idx` on any match and never resets, so e.g. `["search_nodes", "update_node"]` would also pass against actions like `["update_node", "search_nodes", "update_node"]` (an earlier stray `update_node` doesn't invalidate the match). For scenario 6 this is very unlikely to produce a false pass in practice (the agent has no reason to call `update_node` before searching), but worth knowing this asserts "these appear in this order somewhere," not "these are the first two action calls." The doc comment (`aichat-matrix.ts:79-80`) already discloses this accurately ("other tools may appear between/around them"), so this is working as designed and documented — flagging only as a note in case a future scenario needs stricter adjacency.

**No unit test for `assertExpectation`**
The assertion logic (`scripts/aichat-matrix.ts:154-222`) is pure and the entire point of this issue — a bug in the `toolSequence` subsequence walk or the `noRetry` run-length counter would silently make the eval useless while still "passing" review by inspection. That said, `routing-eval.ts` (the explicit pattern this issue asks to mirror) also has zero unit tests for its own assertion logic, and this repo's `scripts/` directory has no test files at all today. Given the project's stated pattern-mirroring intent and that this is a dev-tool harness (not shipped product code), I'm not blocking on this — but a follow-up issue adding a lightweight `assertExpectation`-only unit test (no daemon required, pure function, cheap to test) would meaningfully de-risk this eval layer without violating YAGNI, since the fixture data (arrays of tool-call strings) needs no infrastructure at all. Recommend filing this as a small follow-up rather than blocking this PR on it.

**`MUTATING_TOOLS` — not present, no issue**
I checked for a leftover `MUTATING_TOOLS` list possibly copied from `routing-eval.ts` and left unused. It does not exist anywhere in this diff or in `aichat-matrix.ts` — only `ROUTING_TOOLS` (`search_skills`) is defined, and it's used consistently everywhere assertions run (`actionTools()`). No dead code here.

**Header comment / docs redundancy**
`scripts/aichat-matrix.ts:1-39` header and `nodespace-docs/development/agent-eval.md` cover similar ground (scenario list, ADR-038 routing-tolerance rule, baseline flag) but at different altitudes — the header is a quick in-file usage reference for someone reading the script, the doc page is the discoverable reference for someone browsing `nodespace-docs/development/`. Redundant in content but each serves a different reader; not worth trimming.

All 8 acceptance criteria are satisfied, the implementation correctly mirrors `routing-eval.ts`'s conventions (baseline env var, exit codes, per-case assertion shape), the subsequence/run-length assertion logic is correct for the scenarios it's asserting today, and the one deferred item (unit-testing `assertExpectation`) is consistent with the existing precedent in this codebase and reasonable to defer to a follow-up rather than block on.

malibio:
## Review Feedback Addressed

The initial review came back with **zero blocking findings** (🔴/🟡) — all 8 acceptance criteria for #1364 were confirmed met, and the recommendation was **APPROVE**.

One 🟢 non-blocking suggestion was raised: unit-test `assertExpectation` directly, since it's the core logic this issue is about. Filed as a standalone follow-up rather than an in-PR change, since:
- The reviewer's own note was "recommend filing a follow-up issue," not a requested PR change.
- `scripts/routing-eval.ts` — the explicit pattern this PR mirrors — has zero tests for its equivalent `assertFixture` function, so adding tests here would be scope creep beyond what #1364 asked for and inconsistent with the sibling harness.
- It's a clean, small, standalone unit — pure function, no daemon/socket/DB needed.

Follow-up: #1630

No code changes were needed to address this review. PR is unchanged since the review and ready for merge at the user's discretion.

PR #1631, opened 2026-07-11
Unit-test assertExpectation in scripts/aichat-matrix.ts

malibio:
Closes #1630

### DIFF
--- a/scripts/aichat-matrix.test.ts
+++ b/scripts/aichat-matrix.test.ts
@@ -1,0 +1,147 @@
+/**
+ * Unit tests for assertExpectation() in scripts/aichat-matrix.ts.
+ *
+ * scripts/ sits outside the desktop-app Vitest project (no Svelte/DOM
+ * dependency here), so this uses Bun's native test runner directly:
+ *
+ *   bun test scripts/aichat-matrix.test.ts
+ *
+ * Do NOT use `bun run test` for this file — that command is scoped to
+ * packages/desktop-app's Happy-DOM Vitest suite and won't discover it.
+ */
+
+import { describe, expect, test } from "bun:test";
+import { assertExpectation, type Expectation } from "./aichat-matrix";
+
+describe("assertExpectation", () => {
+  describe("noTools", () => {
+    const expect_: Expectation = { kind: "noTools" };
+
+    test("passes with no action tools", () => {
+      expect(assertExpectation(expect_, []).passed).toBe(true);
+    });
+
+    test("tolerates search_skills", () => {
+      expect(assertExpectation(expect_, ["search_skills"]).passed).toBe(true);
+    });
+
+    test("fails with any action tool present", () => {
+      const result = assertExpectation(expect_, ["create_node"]);
+      expect(result.passed).toBe(false);
+      expect(result.failure).toContain("create_node");
+    });
+
+    test("fails with an action tool even alongside search_skills", () => {
+      const result = assertExpectation(expect_, ["search_skills", "create_node"]);
+      expect(result.passed).toBe(false);
+    });
+  });
+
+  describe("toolOnce", () => {
+    const expect_: Expectation = { kind: "toolOnce", tool: "create_node" };
+
+    test("passes with exactly one call", () => {
+      expect(assertExpectation(expect_, ["create_node"]).passed).toBe(true);
+    });
+
+    test("passes with exactly one call interleaved with search_skills", () => {
+      expect(assertExpectation(expect_, ["search_skills", "create_node"]).passed).toBe(true);
+    });
+
+    test("fails with zero calls", () => {
+      const result = assertExpectation(expect_, []);
+      expect(result.passed).toBe(false);
+      expect(result.failure).toContain("got 0");
+    });
+
+    test("fails with multiple calls", () => {
+      const result = assertExpectation(expect_, ["create_node", "create_node"]);
+      expect(result.passed).toBe(false);
+      expect(result.failure).toContain("got 2");
+    });
+  });
+
+  describe("toolSequence", () => {
+    const expect_: Expectation = { kind: "toolSequence", tools: ["search_nodes", "update_node"] };
+
+    test("passes when the sequence appears in order", () => {
+      expect(assertExpectation(expect_, ["search_nodes", "update_node"]).passed).toBe(true);
+    });
+
+    test("passes with other tools interleaved", () => {
+      const result = assertExpectation(expect_, [
+        "search_skills",
+        "search_nodes",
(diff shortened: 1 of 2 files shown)


malibio:
### Code Review Summary

Net positive, low-risk, well-scoped change. This is exactly the kind of test-only PR that should merge quickly: it adds direct unit coverage for `assertExpectation` — the pure function that is the entire enforcement mechanism for the aichat-matrix eval — without touching any runtime behavior of the CLI script itself. I ran the tests (`bun test scripts/aichat-matrix.test.ts`: 21 pass, 28 assertions) and `bun run quality:fix` (clean, no diff) to confirm both acceptance criteria beyond code review.

All 5 `Expectation` kinds have direct coverage, including the two edge cases the issue named explicitly (empty tool list for `toolSequence`/`noExtraTypes`/`toolOnce`, and `search_skills` interleaving for `noTools`/`toolOnce`/`toolSequence`/`noExtraTypes`). The `import.meta.main` refactor is a clean, behavior-preserving extraction — I diffed the CLI body line-by-line and it's a straight move, not a rewrite (usage message, exit codes, baseline-comparison logic, and the process.exit(1)/exit(0) paths are byte-identical, just re-indented). Threading `aichatPath` through `runTurn`/`newChat` instead of the closure-captured `AICHAT` const is a mechanical, low-risk change — same value, same computation, just passed explicitly since `AICHAT` is now scoped inside the `if` block.

No other module imports `scripts/aichat-matrix.ts` (grep confirms only prose references in `routing-eval.ts` comments and `mlflow-dev.md` docs, not code imports), so the export changes are safe.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- **`scripts/aichat-matrix.test.ts`, `noRetry` / search_skills test (lines ~106-112)**: The test name says "search_skills is filtered out before the run-length check, so it does not break up a repeat" and asserts `passed: false` — this is correct and the accompanying comment is exactly the kind of "why" documentation that pays for itself, since it's the one place in the suite where the intuitive expectation (routing tool interleaved → should break adjacency) is wrong. Good catch flagging this in the implementation notes and in the test itself; nothing to change, calling it out as a good example other tests in this file could imitate more consistently (e.g. the `toolSequence` "fails when out of order" test at line ~76 has no comment on *why* `["update_node", "search_nodes"]` fails the subsequence check for `["search_nodes", "update_node"]` — mildly asymmetric documentation, but not blocking).

- **`assertExpectation`'s `noTools` case is untested for the case where `actions` contains only routing tools that are NOT `search_skills`**: minor — `ROUTING_TOOLS` currently has exactly one entry, so this is a non-issue today, but if `ROUTING_TOOLS` ever grows, the tests would need a matching entry to keep catching regressions in `actionTools`'s filter logic specifically (vs. `assertExpectation`'s branch logic). Not a blocker; flagging as a latent coverage gap tied to a hardcoded array size of 1, which is fine for now (YAGNI).

#### Nitpicks

- **Nit**: `scripts/aichat-matrix.test.ts` line ~62, `toolSequence` "fails when a required tool is missing" test asserts `result.failure` contains `"search_nodes,update_node"` — this pins the exact `join(",")` format of the failure message. It's not unreasonable (failure messages are part of the debugging UX for a human reading eval output), but it's the kind of assertion that will need updating if the message format ever changes for readability (e.g. adding spaces after commas) even though the underlying assertion logic is unchanged. Consider `toMatch(/search_nodes/)` + `toMatch(/update_node/)` if brittleness starts to bite; not worth changing now.

- **Nit**: The file header comment in `scripts/aichat-matrix.test.ts` says "no Svelte/DOM dependency here" as the reason for using `bun:test` over Vitest — the more precise/primary reason (per the PR description and confirmed by inspecting `packages/desktop-app/package.json`'s `test` script) is that `scripts/` is simply outside the Vitest project's file glob, not a DOM-dependency issue. Worth a one-word tweak ("scripts/ isn't part of any Vitest project" vs. framing it around DOM) so a future reader doesn't go hunting for a DOM dependency that was never the actual reason. Optional.

- **Nit**: `quality:fix` (per its `package.json` composition) doesn't touch `scripts/` at all — it only runs ESLint/svelte-check on `packages/desktop-app`, `packages/skill`, and `cargo fmt`/`clippy` on Rust. So "passes `bun run quality:fix`" (an acceptance criterion) is trivially true for this file regardless of its content, since nothing in that command ever inspects `scripts/*.ts`. Not this PR's job to fix the workspace-wide lint gap, but worth noting for whoever eventually decides `scripts/` should get its own lint pass — the acceptance criterion as written doesn't actually validate the new file's lint cleanliness.

### Verification performed
- `bun test scripts/aichat-matrix.test.ts` → 21 pass, 0 fail, 28 expect() calls.
- `bun run quality:fix` → clean, `git status` shows no diff after.
- `grep -rn "aichat-matrix"` across the repo → only doc/comment references, no code imports of the module, confirming the export/`import.meta.main` changes are safe.
- Manual line-by-line diff of the `import.meta.main` block against the pre-refactor script body → confirmed behavior-preserving (usage message, exit codes, baseline comparison, and closure variable→parameter threading all unchanged in substance).

**Recommendation: Approve.** This satisfies all three acceptance criteria in issue #1630. The suggested improvements above are optional polish, not blockers.

malibio:
## Address-Review Summary

**Addressed** (1/1 blocking-adjacent item):
- ✅ Fixed the file-header doc comment in `scripts/aichat-matrix.test.ts` to correctly attribute why `bun:test` is used (scripts/ is outside every Vitest project's glob, not a DOM-dependency issue). Pushed in `a2f34327`.

**Skipped** (4 items — all explicitly labeled optional/non-blocking/YAGNI by the reviewer):
- Asymmetric doc comment on the `toolSequence` "fails when out of order" test — the test name plus inputs already make the failure self-evident; skipping per project convention against comments that restate what code does.
- Untested `noTools` case for a hypothetical second `ROUTING_TOOLS` entry — reviewer explicitly framed this as YAGNI since `ROUTING_TOOLS` has exactly one entry today.
- Brittle `join(",")` string-match assertion — reviewer said "not worth changing now"; assertion is meaningful and passing.
- `quality:fix` not covering `scripts/` lint — out of scope for this PR (repo-wide gap, not introduced or fixable here).

No critical or important findings were raised. Recommendation from review was **Approve**.

## Re-Review Decision

**Rationale**: The only change made was a 2-line doc-comment wording fix with zero logic or behavior impact. All 21 tests still pass; `quality:fix` remains clean. The skipped items are all suggestions/nitpicks the reviewer themselves marked as optional and non-blocking, not corrections to verify. The PR is ready for merge pending your decision.

PR #1633, opened 2026-07-11
Align CLI + SKILL.md with local-agent capability and guidance (#1529) — Phase 1: CLI capability gap

malibio:
## Summary

Phase 1 of a multi-phase implementation of #1529. This PR will stay open and accumulate incremental commits across phases (WIP → review → address-review, repeated per phase) rather than opening a new PR per phase.

**This phase closes #1529's "Close the CLI capability gap" scope section** — every row in the issue's capability parity matrix that was ❌/⚠️ is now ✅, except the explicitly-deferred `search_semantic` full parity (documented as an intentional CLI subset) and the intentionally-out-of-scope `search_skills` verb.

### What's in this phase

- New/extended proto RPCs on `NodeService`: `CreateSchema`/`UpdateSchema`, `ExecuteQuery`, `CreateRelationship`/`GetRelatedNodes`
- Daemon handlers delegating to existing `nodespace_core` ops/schema functions (no logic duplication — `query_ops::execute_query` was split into a `Vec<Node>`-returning core plus a thin typed-JSON wrapper so the gRPC path and the local-agent tool path share one validated implementation)
- New CLI verbs: `nodespace schema create/update`, `nodespace query`, `nodespace node update --property`, `nodespace node set-status`, `nodespace relationship create/get`
- `nodespace search` now wires `collection`/`collection_id`/`filters`/`threshold` through to the already-existing `SearchRequest` proto fields
- 10 new CLI integration tests + 2 pre-existing tests fixed for struct-shape changes

Full technical detail is in the WIP commit message (`git log`).

### Not in this phase (tracked in #1529)

- Section 2: Guidance alignment (SKILL.md rewrite)
- Section 3: Propagation tooling (seed-node → SKILL.md staleness check)
- Section 4: PTY chat wiring to SKILL.md

### Incidental findings (not fixed here, flagged for follow-up)

- Explicit schema fields passed to `create_schema`/`update_schema` reject `custom:`-namespaced names (colon character), even though CLAUDE.md documents that convention as required. The `custom:` prefix is only auto-applied on the natural-language description-parsing path, not for explicit `fields` arrays. Pre-existing `packages/core` behavior, not introduced here — worth a follow-up issue since it affects what the Phase 2 (guidance) work can honestly document.
- Filed #1632 for ~17 pre-existing failures in `packages/agent/tests/ollama_integration.rs`, confirmed reproducing identically on `main` before this branch — unrelated to this diff (zero files touched under `packages/agent/`).

## Test plan
- [x] `bun run test` — 159 files / 4031 tests passed (frontend baseline, unchanged)
- [x] `cargo test -p nodespace-cli --test cli_integration` — 24/24 passed
- [x] `cargo test -p nodespace-core --lib query_ops` — 10/10 passed
- [x] `cargo test -p nodespace-daemon --lib node_service` — 27/27 passed
- [x] `cargo test --workspace --exclude nodespace-app --exclude nodespace-nlp-engine --exclude nodespace-agent` — all passed
- [x] `cargo test -p nodespace-agent --lib` — 291/291 passed
- [x] `bun run quality:fix` — clean
- [x] Pre-push gate (`test:all` + `cargo build --bin nodespaced` + `test:e2e`) — passed

https://claude.ai/code/session_01DMPgbnyNk6bDFnpqtTPbKr


malibio:
### Code Review Summary

This is a well-scoped, atomic Phase 1 slice of #1529's "Close the CLI capability gap" checklist. It adds five new gRPC RPCs, thin daemon handlers that delegate to existing `packages/core` ops/schema functions, and matching CLI verbs, backed by 10 new integration tests exercising the full client→daemon→core stack. The diff does not touch SKILL.md, `skill_pipeline.rs`, or PTY wiring, correctly deferring those to later phases as the PR description states.

The core engineering judgment call — JSON-passthrough (`params_json`/`filters_json`/`edge_data_json`) instead of fully-modeled proto messages — is a reasonable, well-justified choice for this internal, single-repo protocol, and it's executed cleanly: errors from malformed JSON are mapped to `Status::invalid_argument` with the serde error message, not `.unwrap()`/`.expect()`/panics, at every one of the five new RPC handlers. This is a net positive change. A few real gaps remain, mostly around test coverage depth and one genuine validation hole in `ExecuteQuery`'s `limit` handling, detailed below.

### `params_json` / JSON-passthrough Pattern — Verdict

**Verdict: reasonable pragmatic tradeoff, correctly executed. Approve the pattern with two tightenings.**

Arguments in favor, confirmed by reading the actual code:
- Both ends of this protocol (CLI, daemon) live in the same repo, built and versioned together — there are no external API consumers to break, and no cross-language/cross-team contract to enforce via strict typing at the wire boundary. This is exactly the situation where a `oneof`-heavy or `google.protobuf.Struct`-based proto model buys safety nobody needs at the cost of duplicating structs (`SchemaField`, `SchemaRelationship`, `FieldRename`, `AgentFilterItem`, `AgentSortItem`) that already have a single source of truth in `packages/core`. Modeling them again in `.proto` would create exactly the kind of drift-prone duplication the PR is trying to avoid, and directly serves DRY.
- The daemon-side deserialization is correct Rust: every one of the five new handlers (`execute_query`, `create_relationship`, `get_related_nodes`, `create_schema`, `update_schema` in `packages/daemon/src/services/node_service.rs`) uses `serde_json::from_str(...).map_err(|e| Status::invalid_argument(...))?` — never `.unwrap()`/`.expect()`. `handle_create_schema`/`handle_update_schema` in `packages/core/src/schema/mod.rs` likewise propagate via `?` with typed `MarkdownError` variants mapped to gRPC status codes (`markdown_error_to_status`). A malformed JSON blob from the CLI reaches the user as a clean `INVALID_ARGUMENT` status with the underlying serde message, not a 500/panic.
- CLI construction of the JSON is safe in the common path: `nodespace query`/`relationship create`/`node update --property` build filter/property values via `serde_json::json!`/typed `serde_json::Value` construction (e.g. `parse_property` in `node.rs`, `json!({"status": args.status})` in `set_status`), not raw string concatenation. The one path where the *user* supplies raw JSON text directly (`schema create --params '<json>'`, `query --filters '<json>'`) is inherent to the pattern — it's the whole point of a passthrough — and is exactly analogous to how the local-agent tool call already accepts this same shape, so there's no ergonomics regression relative to the existing capability.

Real gaps, not hedges:
- **`ExecuteQueryRequest.limit` has no upper bound.** `packages/daemon/src/services/node_service.rs`'s `execute_query` handler passes `req.limit as usize` straight through when nonzero, with no cap. A CLI user (or, more importantly, a buggy/compromised local-agent caller reusing this path in a future phase) can request an unbounded result set. `search.rs`'s existing `SearchRequest.limit` at least defaults sanely; this one should have a documented server-side ceiling (mirroring whatever `execute_query`'s existing agent-tool caller already assumes) so `nodespace query --type task --limit 4000000000` (u32 max) can't be used to force a huge scan/serialization. **[Improvement]**
- **No proto-level or handler-level size bound on `params_json`/`filters_json`/`sorting_json`/`edge_data_json`.** Because these are opaque strings, gRPC's default message-size limits are the only backstop. Worth confirming (not necessarily fixing here) that tonic's server has a sane max-message-size configured, since a passthrough string field is an easy vector for an oversized payload compared to a structured message where individual fields would naturally cap growth. **[Nit — likely already covered by tonic defaults / existing SearchRequest.filters precedent, but flag for confirmation]**
- The pattern's downside — no compile-time contract for CLI/daemon field names — is real but low-risk here specifically because both ends are one repo, one CI, one test suite; a rename in `CreateSchemaParams` will be caught by the new integration tests failing, not by a silent runtime drift in production. That containment is what makes the tradeoff acceptable; it would not be acceptable if either side were a separately-deployed/versioned artifact.

Net: this is the correct pattern for an internal CLI-to-daemon protocol where both sides ship atomically. It should **not** be used as precedent for anything crossing a real API boundary (e.g. if NodeSpace ever exposes this RPC set externally), and that boundary condition is worth writing down (e.g. as a comment on the proto file, which is partially done already via the "avoid duplicating the nested Rust struct shapes" comments) so a future contributor doesn't reflexively apply params_json to a public-facing RPC.

### Findings

#### Critical Issues

None. No security vulnerabilities, no panics on untrusted input in production code paths, no architectural regressions.

#### Suggested Improvements

- **`packages/daemon/src/services/node_service.rs` — `execute_query` handler**: `req.limit` (a `u32`) is cast to `usize` and passed through with no upper bound when nonzero. Add a server-side max (e.g. clamp to a few thousand) so a large `--limit` value can't force an unbounded query/serialization. This is the one place in the diff where "the daemon validates" (the schema.rs doc comment's stated justification for JSON-passthrough) doesn't actually hold for a plain scalar field, not just the JSON blobs.

- **No daemon-layer unit tests for the five new RPCs.** `packages/daemon/src/services/node_service.rs`'s own `#[cfg(test)] mod tests` (27 tests, matching the PR's claimed count) has no additions — coverage for `execute_query`, `create_relationship`, `get_related_nodes`, `create_schema`, `update_schema` comes entirely from `packages/cli/tests/cli_integration.rs`, which drives them through a real spawned daemon + gRPC client. That's legitimate integration coverage and arguably higher-value than isolated unit tests, but it means there's no fast, daemon-only regression signal if a future change to `node_service.rs` breaks one of these handlers' error-mapping logic specifically (e.g. `markdown_error_to_status`'s four-way match) — you'd only find out via the slower CLI integration suite. Not a blocker given the integration tests are real and pass, but worth a follow-up if `node_service.rs`'s unit suite is meant to be the fast-feedback layer.

- **`schema_create_rejects_malformed_params` test asserts a weak/ambiguous condition.** In `cli_integration.rs`, the assertion is `err.contains("params_json") || err.contains("rpc failed")` — the second branch means the test would still pass even if the specific "invalid params_json" message were accidentally dropped or genericized, as long as the RPC merely failed for any reason. Tighten to assert the message actually contains the serde error content (or at minimum drop the OR-fallback) so this test would actually catch a regression in the error-message quality that the params_json design depends on for good CLI ergonomics.

- **`relationship create` has no test for a nonexistent target/source node or an undefined relationship name** — only the "invalid direction" error path is tested for `relationship get`, and `relationship create`'s happy path assumes both nodes and the relationship already exist via schema. Given `create_relationship` in `rel_ops.rs` maps the underlying error to a generic `OpsError::Internal(format!("Failed to create relationship: {}", e))`, it's worth confirming (via a test) that "relationship name not defined on source schema" and "target node doesn't exist" produce distinguishable, non-500-flavored errors rather than both collapsing into an opaque internal error — this matters directly for the find-then-edit / clarifying-question guidance the epic's Section 2 is going to need to write against this verb's actual failure modes.

#### Nitpicks

- **Nit:** `packages/cli/src/commands/node.rs::parse_property` silently falls back to treating a value as a plain string when `serde_json::from_str` fails (`unwrap_or_else`). That's the right UX default (so `--property name=Alice` doesn't require `--property name='"Alice"'`), but it means a typo'd JSON value like `--property tags=[1,2` (missing bracket) silently becomes the literal string `"[1,2"` instead of erroring. Worth a doc-comment caveat or a `--property-json` escape hatch later if this bites someone; not worth blocking on for Phase 1.

- **Nit:** `set_status`'s `VALID_STATUSES` list (`open, in_progress, done, cancelled`) is hardcoded identically in three places now across the codebase (CLI, `tools.rs` description string, `tools.rs` enum/validation) rather than being a single shared constant. Pre-existing duplication pattern (not introduced by this PR — the CLI mirrors what `tools.rs` already had), but three-way duplication is worth consolidating into one `const`/enum in `packages/core` if a status value is ever added, since the CLI list would otherwise silently drift out of sync with the tool the guidance work in Section 2 needs to describe consistently.

- **Nit:** `print_schema_result` in `schema.rs` takes `_json: bool` and ignores it, always pretty-printing — reasonable given the stated "structured data either way" rationale in the comment, but the unused parameter is a small smell; consider dropping it from the signature (breaking the call site's `json` pass-through) rather than keeping a permanently-ignored flag, since a future reader may reasonably expect `--json` to change output shape here as it does for every other command.

- **Nit:** `GetRelatedNodesRequest.direction` and `RelationshipAction::GetArgs.direction` are bare `String` rather than a proto `enum`/CLI `ValueEnum` restricted to `in`/`out`. Validation exists (`direction != "in" && direction != "out"` in the daemon handler, tested), but a typed enum would make the invalid state unrepresentable at the type level instead of relying on a runtime check plus a test asserting `Code::InvalidArgument`. Minor, given clap's `ValueEnum` would be a bigger surface change for one field.

### Scope / Atomicity Check

Confirmed: every changed file (`packages/proto/proto/node_service.proto`, `packages/daemon/src/services/node_service.rs`, `packages/core/src/ops/query_ops.rs`, `packages/cli/src/commands/{mod,node,query,relationship,schema,search}.rs`, `packages/cli/src/lib.rs`, `packages/cli/tests/cli_integration.rs`) is directly in service of the stated "close the CLI capability gap" scope. No unrelated refactors, no SKILL.md/skill_pipeline.rs/PTY changes bundled in. The `query_ops::execute_query` → `execute_query_nodes` + thin `execute_query` wrapper split is a justified, minimal refactor to let the new gRPC path reuse the same validated core logic as the existing agent-tool path, rather than duplicating filter/sort validation — good adherence to DRY and exactly the kind of shared-core-logic pattern the epic wants (per its "one canonical implementation, two vocabularies" framing). The incidental findings section of the PR description (the `custom:` namespace rejection on explicit schema fields, and the pre-existing `ollama_integration.rs` failures) is correctly flagged as out-of-scope-but-noted rather than silently absorbed or silently ignored.

### Recommendation: APPROVE

The architecture is sound, the params_json pattern is a defensible and correctly-implemented tradeoff for this internal boundary, security-sensitive paths (identifier validation via the existing whitelist-based `validate_identifier`, no raw string-built queries, no secrets) are unaffected or reused safely, and the change is atomic and well-tested at the integration level. The `execute_query` limit bound is the one item I'd want addressed before or shortly after merge (low effort, real if narrow risk); the rest are non-blocking improvements and nits suitable for fast-follow or the next phase.

malibio:
## Addressed review feedback

Pushed `39b58823` addressing the `/pragmatic-code-review` findings (recommendation was **APPROVE**; these were suggested improvements/nitpicks, not blockers):

**Addressed:**
- `execute_query`'s client-supplied `limit` is now capped at 500 server-side (was unbounded).
- `relationship get --direction` is now a `clap::ValueEnum` instead of a bare `String` — invalid values rejected at parse time.
- Dropped `print_schema_result`'s unused `json` parameter.
- Tightened `schema_create_rejects_malformed_params` to assert a precise `Status::InvalidArgument` + message content instead of a permissive OR-assertion.
- Renamed/reworked the invalid-direction test to drive the raw gRPC client directly (since clap now rejects invalid directions before the CLI handler runs, but the daemon-side validation is still real for non-CLI clients).
- Added `relationship_create_rejects_relationship_undefined_on_source_schema` — verifies the find-then-edit-relevant error surfaces correctly when a relationship isn't schema-defined.

**Skipped (with reasoning):**
- `parse_property`'s silent JSON-fallback-to-string: intentional, not a bug (`--property name=Alice` must work without quoting).
- `VALID_STATUSES` triple-duplication: real but pre-existing (already duplicated twice in `local_agent/tools.rs` before this PR); consolidating is a cross-package refactor out of scope here.
- Daemon-layer unit tests for the 5 new RPCs: matches this codebase's established pattern of CLI-integration-test coverage against the real gRPC handlers rather than mocked daemon-only tests.

All test suites re-verified green after these changes (25/25 CLI integration tests, full workspace `cargo test`, frontend baseline unchanged, `quality:fix` clean).

## Re-Review Decision

**Rationale**: All changes are low-risk, narrowly scoped to the review's own suggested-improvement/nitpick items (not critical findings), and verified by the full test suite. No new functionality or architectural changes were introduced — this was a straightforward hardening pass (bounds-checking, type-safety, test precision) on already-approved code.

malibio:
## Review: commit `2476e8a3` only (Section 3 — propagation tooling)

Scope: everything between `96b28004` and `HEAD` (`skill_rules.rs`, `gen_skill_md.rs`, `skill_pipeline.rs`, `SKILL.md`, `lib.rs`, `package.json`). Phases 1/2 are out of scope and not re-reviewed.

### Summary

The design (shared `SchemaRule`/`InteractionRule` constants, real generation for the 9 schema rules, presence-checking for the 7 interaction rules) is sound and well-documented — module docs in `skill_rules.rs` and `gen_skill_md.rs` clearly explain the asymmetry and why full generation isn't attempted for interaction rules. `package.json`'s `rust:test` change correctly starts running previously-untested `nodespace-agent` code (`--lib --bins` correctly excludes the `[[test]]`-declared `ollama_integration`/`search_skills_latency` integration binaries in `Cargo.toml` — verified). `cargo clippy -p nodespace-agent --lib --bins` is clean, and the checked-in `SKILL.md` generated block does currently match generator output.

However, there is one concrete, demonstrated bug that undermines the core guarantee this commit is supposed to provide, plus a couple of smaller robustness gaps.

### Findings

#### Critical

**`packages/agent/src/bin/gen_skill_md.rs`, `render_schema_rules_block()` (line ~38): the "Editing" paragraph is hardcoded, not interpolated from `EDIT_DONT_RECREATE.prose` — silently defeats drift detection for that rule.**

All 8 other `SchemaRule`s are correctly threaded through `format!()` args (`one_schema_per_request = ONE_SCHEMA_PER_REQUEST.prose`, etc.), but the "Editing" paragraph is a bare string literal baked directly into the format template:

```rust
"{one_schema_per_request}\n\n{schema_already_exists}\n\n\
 **Editing:** to add, remove, or rename a field, or change a relationship on an existing schema, use `schema update` with only the fields that need changing (`add_fields`/`remove_fields`/`rename_fields`, or an updated `description`/`title_template`). Don't re-create the whole schema for a small change.\n\n\
 **Schema fields:** {no_name_title_field} {name_placeholder_exception}\n\n\
 ..."
```

`EDIT_DONT_RECREATE.prose` exists in `skill_rules.rs` and happens to be byte-identical to this literal today, which is why `--check` currently passes. But it is never read by the generator. I verified this concretely: I edited `EDIT_DONT_RECREATE.prose` in `skill_rules.rs` (appended a marker string), left `SKILL.md` untouched, and ran `cargo run -p nodespace-agent --bin gen_skill_md -- --check` — it printed "up to date" and exited 0. This is exactly the drift scenario the whole mechanism exists to prevent, occurring inside the generator that's supposed to prevent it. (Reverted the probe edit afterward; working tree is clean.)

Note `skill_pipeline.rs`'s `schema_creation_guidance()` does correctly interpolate `EDIT_DONT_RECREATE.imperative` — the bug is isolated to the generator side only.

Fix: replace the hardcoded literal with `{edit_dont_recreate}` / `edit_dont_recreate = EDIT_DONT_RECREATE.prose` like its siblings.

#### Improvements

1. **`skill_pipeline.rs`, `graph_editing_guidance()` / `node_deletion_guidance()`: reusing the generic `FIND_THEN_ACT.imperative` drops verb-specific detail that used to live in the same sentence.** Before this refactor, "Graph Editing" said "...locate it. **Then call `update_node` with the ID and only the fields that need changing.**" and "Node Deletion" said "...locate it. Confirm the title matches..., then call `delete_node`." Now both instead splice in the shared imperative "...locate it. Then act on the resolved ID — do not guess IDs," which is generic by design (it's shared across 3 call sites) but reads slightly redundant/deflated at each specific call site — e.g. Node Deletion now reads "...to locate it. Then act on the resolved ID — do not guess IDs. Confirm the title matches what the user described, then call `delete_node`..." (two consecutive "act on it" beats). Not a correctness bug — the specific instruction is still present later in each block (e.g., "update_node FIELDS:" section) — but it's worth a read-through pass to confirm nothing reads as awkward filler now that it's machine-assembled. Low severity, but since this is precisely the kind of interpolation-boundary smoothness the review was asked to check, flagging it.

2. **`splice_generated_block` uses `.find()` (first occurrence) for both markers with no check for a second, duplicate marker pair.** Given this is a small, single-purpose CI tool with loud panics on missing markers, and today's `SKILL.md` has exactly one marker pair (verified), this is low risk in practice. Still, a one-line `debug_assert!`/check that the marker appears exactly once (or a comment noting the assumption) would make the "graceful failure over silent misbehavior" intent explicit rather than implicit.

3. **`skill_md_key_phrase` presence-checking is a legitimately weak (but not worthless) guard.** It catches wholesale removal of a rule's substance from `SKILL.md`, which is a real and plausible failure mode (someone rewrites a section and drops a clause). It does not catch semantic drift/rewording that keeps the exact phrase but changes the guidance around it, and — as a substring match — it could theoretically false-pass if the key phrase happens to appear in an unrelated sentence. Given the constraint (avoiding fragmenting hand-tuned prose), this is a reasonable proportional tradeoff, not a design flaw, but the module doc's honesty about the limitation ("catches outright removal, not rewording") is doing real work here and should stay.

#### Nitpicks

- Nit: `skill_rules.rs`'s `no_rule_text_is_empty` and the id-uniqueness tests are tautological in the sense that they test the constants against themselves rather than against either consumer, but they're still worthwhile cheap guards against copy-paste mistakes (e.g., duplicate `id`, an empty field from a bad merge) — no change needed, just noting they're a sanity net, not drift protection.
- Nit: `render_schema_rules_block()`'s inline comment ("NO_NAME_TITLE_FIELD and NAME_PLACEHOLDER_EXCEPTION share the **Schema fields:** paragraph") is helpful; consider adding a similar one-line note for why the Editing paragraph doesn't have a named interpolation binding, if the fix above isn't applied for some reason — but the fix is preferable to documenting the gap.

### Recommendation

One Critical fix needed (the hardcoded Editing paragraph in `gen_skill_md.rs`) before this is fully doing its job — it's a small, mechanical fix (swap the literal for `{edit_dont_recreate}` + add the format arg) and everything else in the commit is solid. Everything else here is Improvement/Nit level, not blocking.

malibio:
## Address-review: Section 3 (commit 2476e8a3), reviewed in [this comment](https://github.com/NodeSpaceAI/nodespace-core/pull/1633#issuecomment-4949281475)

Fixed in `52ae5f6e`:

- **🔴 Critical** — `gen_skill_md.rs`'s `render_schema_rules_block()` hardcoded the "Editing" paragraph instead of interpolating `EDIT_DONT_RECREATE.prose`, so `--check` couldn't detect drift in that rule. Fixed by importing and interpolating it. Reproduced the reviewer's own probe post-fix (edit the rule's prose → `--check` now exits 1 as expected; restored → passes again).
- **🟡 Improvement** — `graph_editing_guidance()`'s "FIND THEN UPDATE" line lost verb-specific detail ("Then call `update_node` with the ID...") when swapped for the generic `FIND_THEN_ACT` interpolation. Restored the tail after the interpolated rule.
- **🟡 Improvement** — `splice_generated_block()` had no guard against duplicate marker occurrences. Added two assertions so a duplicate marker fails loudly instead of splicing into the wrong occurrence.

Skipped (nits, no code change needed): the tautological `skill_rules.rs` self-tests (still useful copy-paste guards, not worth removing) and the "Schema fields" grouping comment note (moot once the Critical fix landed — `EDIT_DONT_RECREATE` is now a named, interpolated binding like its siblings).

Verified: `bun run rust:test` (979 nodespace-core + 295 nodespace-agent lib + 3 gen_skill_md bin tests, all green), `cargo clippy -p nodespace-agent --lib --bins` clean, `cargo fmt -p nodespace-agent` applied, `cargo check --workspace` clean, pre-push gate passed in full.

## Re-Review Decision

**Rationale**: All three fixes are small, mechanical, and directly verified against the reviewer's own repro steps (the Critical fix was confirmed using the exact drift scenario the reviewer used to find it). No new design surface, no architectural changes, no new files. Low risk.

malibio:
### Code Review Summary

Reviewed diff range `abdd1700..ae2fe297` (commit `ae2fe297`, "Wire PTY chat to SKILL.md — closes #1529 Section 4") in isolation, per scope. This is a clean, well-motivated fix: it replaces a confirmed-dead shim mechanism and unreachable-tool-vocabulary guidance with an unconditional `include_str!`-embedded copy of `packages/skill/SKILL.md` into every PTY session directory. Verified independently (not just taking the commit message's word for it):

- `include_str!("../../../skill/SKILL.md")` in `packages/agent/src/acp/context_assembly.rs` resolves correctly to `packages/skill/SKILL.md` (confirmed via `realpath` from the file's directory).
- `grep -rn` for every deleted symbol (`with_shim_dir`, `shim_source_dir`, `ShimTool`, `SHIM_TOOLS`, `write_shim_files`, `shim_files_for`, `shim_subdir_for`, `tools_section`) across `packages/` returns zero hits — the deletion is complete, no dangling references, no dead imports left behind.
- `agent_guidance.rs`'s `SCHEMA_CREATION_RULES`/`TOOL_STRATEGY_RULES`/`NODE_REFERENCE_FORMAT` are still consumed, and now *only* consumed, by `prompt_assembler.rs` — confirmed via `grep -rln`.
- `cargo build -p nodespace-agent` and `cargo test --lib acp::context_assembly` (22 tests) both pass clean.
- `ContextError::WriteFailed(#[from] std::io::Error)` makes `write_skill_md`'s bare `?` propagation correct and idiomatic — matches the existing pattern in `write_context_to_dir` in the same file.

This is a net-positive, atomic, well-scoped change. One real (non-blocking) substance gap and one test-depth gap below; everything else is nit-level.

### Findings

#### Critical Issues

None.

#### Suggested Improvements

- **`packages/agent/src/agent_guidance.rs` `NODE_REFERENCE_FORMAT` has no SKILL.md equivalent.** `NODE_REFERENCE_FORMAT` ("Reference nodes with bare `nodespace://` URI — no markdown links, no backticks") was dropped from the PTY context file along with the other two guidance constants, and I could not find any equivalent instruction in `packages/skill/SKILL.md` (grepped for `URI`, `nodespace://`, `markdown link`, `reference` — no hits addressing output-formatting conventions). Unlike `SCHEMA_CREATION_RULES`/`TOOL_STRATEGY_RULES` (which are genuinely daemon-tool-call vocabulary, correctly identified as inapplicable to a PTY agent that only has the `nodespace` CLI), this rule is about how the agent should *format node references in its own chat replies to the user* — a concern that applies just as much to a PTY-spawned CLI agent as to the local in-app agent, since presumably NodeSpace's chat UI gives bare `nodespace://` links special rendering (clickable/resolvable) regardless of which agent produced them. If that rendering exists, PTY agents will now emit un-rendered markdown links or raw IDs instead. Worth either adding a one-line equivalent to SKILL.md or confirming (and noting in the PR) that the chat UI doesn't special-case `nodespace://` links from PTY output, making this moot.

- **Test coverage stops at the file-write layer; no PTY-level assertion that SKILL.md is actually delivered via `launch()`.** The new `write_context_file_writes_both_context_file_and_skill_md` test is solid and exercises a real `GraphContextAssembler` over a real `SqliteStore`/`NodeService` (not tautological) — but it calls `write_context_file` directly, not `PtySession::launch()`. Separately, `pty::session::tests::launch_for_test_runs_command_and_emits_output` (and the other `launch_for_test`-based tests) spawn via `PtySession::launch_for_test`, which calls `spawn_in_pty` directly and never touches `GraphContextAssembler` at all (confirmed by reading `launch_for_test`'s body — it only does `create_dir_all` on the session dir, no assembler call). So there is currently no test — old or new — that exercises the actual `launch()` path end-to-end and confirms a real PTY session directory ends up with both `CLAUDE.md`/`AGENTS.md` and `SKILL.md` before the child process starts. Given `launch_for_test` was deliberately split out to avoid needing a full `GraphContextAssembler` fixture in PTY tests, closing this gap isn't a small addition — it's reasonable to defer, but flagging it since it's the one place a real regression (e.g. a future refactor of `launch()` that reorders or drops the `write_context_file` call) could land undetected until manual QA.

#### Nitpicks

- **Nit:** `packages/agent/src/acp/context_assembly.rs` module doc: "the same way it would in any directory a user `cd`s into with `SKILL.md` present" (in `write_context_file`'s doc comment) is a slightly indirect way to describe the mechanism — a one-clause note that this relies on each CLI harness's own project-local-`SKILL.md` discovery convention (which SKILL.md's own delivery docs presumably already establish) would help a future reader who isn't already familiar with that convention.
- **Nit:** The commit deletes `shim_tools_match_gemini_manifest`, a drift-detection test that guarded against `packages/skill/shims/gemini/nodespace-tools.json` diverging from the Rust-side tool registry. That's correct here (the Rust-side registry it protected is gone), but since `packages/skill/shims/*` is explicitly called out as still-alive (external-harness installer path), it's worth double-checking in a future PR that *something* still guards `nodespace-tools.json` against drifting from whatever it's meant to match now that this Rust-side check is gone — not a blocker for this PR since that's out of scope here.

### Verification Checklist (per review-focus questions in the task)

1. **`include_str!` path correctness** — Verified correct; resolves to `packages/skill/SKILL.md`.
2. **Unconditional `SKILL.md` copy for every `AgentType` including `Pi`** — Sound. `Pi` never had a working shim before (the mechanism was dead for all agent types), so this is strictly additive for every agent type, not a regression for any of them. No agent-type-specific reason found to make it conditional — SKILL.md is agent-agnostic CLI guidance.
3. **Substance gap from removing `SCHEMA_CREATION_RULES`/`TOOL_STRATEGY_RULES`/`NODE_REFERENCE_FORMAT`** — `SCHEMA_CREATION_RULES` and `TOOL_STRATEGY_RULES` are fully superseded by SKILL.md's CLI-vocabulary equivalents (schema-before-instance ordering, search-before-mutate, confirmation semantics, structured vs. semantic query selection — all present in SKILL.md, verified by reading its "Tool Decision Guide" and CLI reference sections). `NODE_REFERENCE_FORMAT` has no equivalent — see Suggested Improvements above.
4. **Error handling in `write_skill_md`** — Correct; uses existing `#[from] std::io::Error` conversion via `?`, consistent with sibling functions in the same file.
5. **Test quality** — New tests are genuine, not tautological (`write_context_file_writes_both_context_file_and_skill_md` spins up a real `SqliteStore`/`NodeService`). Deleting the shim/tools tests is appropriate — the underlying mechanism is gone, not merely adjusted, so keeping those tests would be testing dead code. See Suggested Improvements for the one coverage gap (no `launch()`-level PTY test).
6. **Stale doc comments / dead imports** — None found. Doc comments in all four touched files were updated consistently and accurately describe the new behavior. `PathBuf` import in `context_assembly.rs` is still used (return types), not dead.
7. **Satisfies #1529 Section 4 acceptance criteria** — Yes, substantially: the PTY chat path now delivers SKILL.md into every session directory, replacing dead/unreachable guidance. The one gap (item 2 above, node-reference-format parity) is minor and doesn't block the core wiring goal of "PTY chat path consumes SKILL.md."

### Recommendation

**APPROVE** (with the two Suggested Improvements above worth a follow-up — neither blocks merge; the `NODE_REFERENCE_FORMAT` gap is worth a quick confirm-or-fix before or shortly after merge since it's a one-line addition if needed, and the `launch()`-level test gap is reasonable to track separately rather than block this PR).

malibio:
Addressed review (commit ae2fe297, Section 4): both suggestions investigated, no code changes needed.

**1. `NODE_REFERENCE_FORMAT` dropped without a SKILL.md equivalent** — verified this is correctly dropped, not a gap. Checked the PTY terminal pipeline (`pty-terminal.svelte`): raw output bytes go straight into xterm.js's default renderer via `terminal.write(bytes)`, with no link addon loaded (no `@xterm/addon-web-links`, no custom `nodespace://` URI matcher) and no markdown parsing. Every place in the codebase that *does* render `nodespace://` specially (`chat-markdown.svelte`, `content-processor.ts`, `app-shell.svelte`, `node-card-inline.svelte`, `backlinks-panel.svelte`) belongs to the local in-app chat's markdown-rendering pipeline — a structurally separate UI surface the PTY terminal doesn't share. The rule only ever mattered for that rendering path, so its absence from the PTY context file / SKILL.md is a non-issue.

**2. No test exercises `PtySession::launch()` end-to-end to confirm SKILL.md lands in a real spawned session** — `launch_for_test` (the test-only constructor used by all existing real-PTY-spawn tests) deliberately bypasses `GraphContextAssembler` by design, so it can spawn plain shell utilities (`sh`, `cat`) instead of depending on a real agent binary being present in CI. The actual integration point is `GraphContextAssembler::write_context_file` writing into a session directory, which the new `write_context_file_writes_both_context_file_and_skill_md` test already covers against a real `NodeService`/`SqliteStore`. Extending `launch_for_test` to thread through an assembler, or requiring a real `claude`/`codex` binary on PATH for this test, would be a larger change out of proportion to Section 4's scope (delivery/wiring, not the PTY-spawn mechanics themselves).

No re-review needed — recommendation stands at APPROVE.

Issue #1057, opened 2026-04-07
Complete skill-to-tool library with intent→skill→tool pipeline tests


malibio:
## Overview

Build out the complete skill library covering all NodeSpace operations, with comprehensive tests for the intent→skill→tool pipeline.

## Problem

- Only 5 seed skills exist, not all tools are covered by skill guidance
- Zero tests for "given this user message, does the right skill get selected?"
- Intent extraction patterns are incomplete for many operations
- The `find_skills` pull-based path fails when model queries don't semantically match skill descriptions (e.g., "project management" vs "Schema Creation")

## Skill Coverage Matrix

| Skill | Tools | Example Prompts | Status |
|-------|-------|----------------|--------|
| Research & Search | search_semantic, search_nodes, get_node | "What do I know about X?", "Find all tasks" | ✅ Seed exists |
| Node Creation | create_node, get_node | "Create a new task: Do laundry" | ✅ Seed exists |
| Schema Creation | create_node, get_node | "Create a new type 'Project' with fields..." | ✅ Seed exists (with guidance) |
| Graph Editing | update_node, get_node | "Mark task X as done", "Change the title to Y" | ✅ Seed exists |
| Relationship Management | create_relationship, get_related_nodes, get_node | "Connect invoice to customer", "What's related to X?" | ✅ Seed exists |
| Node Deletion | delete_node, get_node | "Delete the old meeting notes", "Remove node X" | ❌ Missing |
| Bulk Import | create_nodes_from_markdown | "Import this document", "Create nodes from markdown" | ❌ Missing |
| Organization | create_relationship (member_of), get_node | "Add to collection X", "Organize by category" | ❌ Missing |

### 1. Intent Extraction Coverage (unit tests)
For each skill, test that representative user prompts extract the correct intent:
```
"What do I know about X?" → "search"
"Create a new type Project" → "create schema"  
"Mark task as done" → "update"
"Delete old notes" → "delete"
"Connect invoice to customer" → "relate"
```

### 2. Skill Matching (integration tests with embeddings)
Given an extracted intent, test that the correct skill is returned by the skill pipeline:
```
intent "search" → skill "Research & Search"
intent "create schema" → skill "Schema Creation"
intent "create" → skill "Node Creation"
```

### 3. Tool Argument Construction (mock engine tests)
Given a skill's guidance prompt, verify the model would receive the right context:
- Schema Creation guidance includes field structure example
- Graph Editing guidance includes enum value format
[...]

PR #1635, opened 2026-07-11
ollama_integration test suite has ~17 failing tests due to empty tool-call responses from mistral:7b

malibio:
Closes #1632


malibio:
### Code Review Summary

**Review type:** Initial review (no prior reviews/comments on PR #1635).

This is a well-executed triage of a flaky/failing integration test suite. The implementer did real root-cause work — verified via raw Ollama API calls bypassing NodeSpace code, cross-checked with model-size comparisons (3B/7B fail, 9B/12B succeed) — rather than reflexively slapping `#[ignore]` on red tests. All four of issue #1632's acceptance criteria are met. One correctness issue in the diff is worth flagging before merge (a stray `#[ignore]` reason string typo), plus one pre-existing convention violation this PR's own new doc comment perpetuates.

### Requirements Check (issue #1632)

- [x] **Root cause identified** — `mistral:7b` narrates tool calls as prose once ~6+ tools are offered in a single turn; verified independently of NodeSpace code and confirmed via model-size comparison.
- [x] **Failing tests triaged** — all 17 originally-listed failures are accounted for: 13 marked `#[ignore]` with specific reasons, 1 (`test_skill_pipeline_organization_real_model`) had its assertion corrected, others pass now that the model is pinned deterministically.
- [x] **Suite passes reliably** — confirmed by the main agent across two consecutive full runs (16 passed / 0 failed / 13 ignored).
- [x] **No regression elsewhere** — `cargo test -p nodespace-agent` (291 passed) and full `bun run test:all` both green; this is the only file touched.

### Findings

#### Critical Issues
None.

#### Important

- **`packages/agent/tests/ollama_integration.rs:721`** — `test_skill_pipeline_research_real_model`'s `#[ignore]` reason claims the failure occurs "even with only 3 tools offered," but `tools_for_skill("Research & Search", ...)` scopes to whatever `seed_skill_nodes()` whitelists for that skill, and the module doc comment's stated ceiling is "~6+ tools." If the Research skill's whitelist is genuinely 3 tools, this is a *second*, distinct model-capability failure mode (ambiguous-phrasing narration failure, not a tool-count ceiling) and deserves to be called out as such rather than folded into the "6+ tools" framing implied by the rest of the file — right now a maintainer skimming reasons will assume all 13 ignores share one root cause when this one doesn't. Worth a quick check that "3 tools" is accurate and, if so, a one-clause disambiguation (e.g., "distinct from the 6+-tool ceiling — fails even at low tool count on ambiguous phrasing").

- **Doc-comment issue-number references** (pre-existing, not newly introduced, but touching this territory) — `documentation.md`'s "no issue refs" rule explicitly states "the same rule applies to source-code comments." Lines 333 (`issue #1443`) and 463 (`Issue #1057`) already violated this before the PR and are untouched by the diff, so this isn't a blocking issue for this PR. However, the new module doc comment this PR adds sits directly above/beside that pattern and includes "see the linked follow-up issue for tracking" (line 40) — a live instance of exactly the anti-pattern the convention exists to prevent (a doc comment pointing at a ticket instead of describing the constraint directly). Since this PR is already touching this doc block extensively, it's a reasonable time to either inline the actual constraint (why `ornith:9b` can't reuse KV-cache across ReAct iterations) or drop the "see issue" pointer — leaving a dangling forward-reference to an issue number not present anywhere in this codebase is the kind of stale-link risk the convention is trying to avoid.

#### Suggested Improvements

- **PINNED_OLLAMA_MODEL vs. `first_ollama_model()` coexistence** — confirmed correct, not an oversight: `resolve_engine()` (used by all 13 ignored tests + the 6 non-ignored skill-pipeline tests, i.e. everything that depends on tool-calling behavior) always pins to `mistral:7b`. `first_ollama_model()` is retained only for 5 earlier tests (`test_ollama_inference_generate`, `test_ollama_inference_with_tools`, `test_ollama_inference_with_production_prompt`, `test_ollama_model_info_context_window_detection`, `test_ollama_model_info`) that assert on generic inference/metadata behavior, not tool-call structure — those genuinely don't care which model runs. No action needed, but a one-line note in the module doc comment ("first_ollama_model() remains in use for tests that don't depend on tool-calling structure") would save the next reader from re-deriving this distinction, since the PR description anticipated this exact question.

- **Coverage erosion via `#[ignore]`** — 15 of 29 tests in the file are now `#[ignore]`d (52%, using `grep -c '#\[ignore'`; note the issue description's own "~17 failing" estimate and any "45%" framing undercounts slightly). Since this suite isn't part of any CI gate, `#[ignore]`d tests here don't rot in the sense of silently failing in CI (they were never run there) — but they can rot in the sense of nobody manually re-running them, so an actual future model upgrade or tool-scoping fix that resolves the underlying ceiling would go undetected indefinitely. Given the issue's explicit non-goal ("not proposing to add this suite to any CI gate"), this is an accepted trade-off rather than a defect, but it's worth a lightweight follow-up habit (e.g., re-run `--ignored` when the default Ollama model changes) rather than leaving these as permanently-ignored dead weight.

#### Nitpicks

- **Nit:** The 13 `#[ignore]` reason strings use two templated variants verbatim (the "unscoped 10-tool StubToolExecutor set" ceiling and the "multi-field structured create_schema args" ceiling), each restating "Verified deterministic (not flaky) via raw Ollama API calls bypassing all NodeSpace code" nearly word-for-word 13 times. This is intentional redundancy for `cargo test -- --ignored --list` scannability (each reason is self-contained without needing the module doc comment open), which is a defensible choice for CI-adjacent tooling, but a maintainer reading the file top-to-bottom will find it repetitive. Not worth churning now — flagging only because it's the kind of thing that invites a "let's DRY this up" drive-by refactor later; a shared `const` string fragment interpolated per-test would preserve scannability while cutting duplication if that ever becomes annoying.

- **Nit:** `packages/agent/tests/ollama_integration.rs:32-44` — the "Evidence gathered" paragraph mixes investigation findings (model-size comparison) with unrelated production-model-picker rationale (`ornith:9b` KV-cache exclusion) and a memory-pressure OOM tangent. Consider splitting into "why mistral:7b specifically" vs. "why not the alternatives" for skimmability, though this is a judgment call and the current version is not hard to follow.

Net positive and well above the bar for a test-triage change: genuine root-cause investigation (not just silencing red tests), an assertion fix verified against the actual `tool_whitelist` in `seed_skill_nodes()` (confirmed correct by reading `packages/agent/src/skill_pipeline.rs:305` directly — the Organization skill's real whitelist is `["create_relationship", "get_node", "search_semantic", "search_nodes"]`, matching the widened assertion exactly), and no regressions. The two Important items above are not blockers — one is a pre-existing minor accuracy nit in a single ignore-reason string, the other is a pre-existing doc-comment convention gap this PR doesn't worsen in kind (only in surface area). Both are fine to leave as-is or address in this PR at the author's discretion.

malibio:
## Addressed Review Feedback

Both **Important** findings from the review addressed:

1. **`test_skill_pipeline_research_real_model` ignore-reason ambiguity** — confirmed the "3 tools offered" figure is accurate for what this test actually sends (the seed's real `tool_whitelist` for Research & Search is 4 tools including `execute_query`, but the test harness's hand-written `all_skill_tools()` stub doesn't define `execute_query`, so only 3 make it through `tools_for_skill`). Reworded the reason string to explicitly flag this as a **distinct** low-tool-count phrasing-ambiguity failure, separate from the 6+-tool-count ceiling the other 12 ignored tests share, so a reader skimming `--ignored --list` output doesn't conflate the two root causes.

2. **Dangling issue-ref pointer in the new doc comment** — removed "see the linked follow-up issue for tracking" and inlined the actual constraint (ornith:9b's recurrent/SSM layers can't reuse KV-cache across ReAct iterations, so multi-step turns re-decode the full prompt from scratch) directly, per this repo's no-issue-refs-in-doc-comments convention. Also split the doc comment into "why mistral:7b" / "why not the larger alternatives" subsections per the reviewer's skimmability suggestion, and added a one-line note explaining why `first_ollama_model()` still coexists with the new `PINNED_OLLAMA_MODEL` constant (used only by the 5 tests that assert on generic inference/metadata behavior where model identity doesn't matter).

**Skipped** (Suggestions/Nits, at my discretion per the reviewer's own framing as non-blocking):
- Deduplicating the 13 near-identical ignore-reason strings into a shared const fragment — intentional redundancy for `--ignored --list` scannability, reviewer explicitly called this defensible.
- Coverage-erosion tracking habit (re-run `--ignored` on future model changes) — process suggestion, not a code change; noted for future reference rather than actioned in this PR.

Re-ran the full suite after changes: still **16 passed, 0 failed, 13 ignored**. `bun run quality:fix` clean. Pushed as `d1843b10`.

## Re-Review Decision

**NO RE-REVIEW NEEDED.** Both changes are doc-comment/string wording only — no logic, assertions, or test behavior changed. Verified via a full re-run of `cargo test -p nodespace-agent --test ollama_integration --features testing` (same 16/0/13 result as before the edit) that nothing regressed.

malibio:
### Re-Review: commit 7e85ffb2 (search_nodes/execute_query guidance fix)

**Scope note:** This is a re-review of the single new commit `7e85ffb2` ("Fix search_nodes-vs-execute_query guidance for unfiltered listing requests"), added after the prior review (recommend: APPROVE, findings addressed) which covered the `packages/agent/tests/ollama_integration.rs` triage work. This commit touches only `packages/agent/src/agent_guidance.rs` (2 lines changed) and is production prompt-guidance code, not test code. The prior review's scope and conclusion stand unchanged; this covers only the delta.

### Summary

Clean, minimal, well-targeted change to two lines of `TOOL_STRATEGY_RULES`. Both edits are evidence-driven: change #1 (unfiltered-listing → `search_nodes`) is verified to fix a real routing bug via a real end-to-end eval harness (10/12 → 11/12, stable across two isolated runs). Change #2 (one-word-only query instruction) is verified present verbatim in the real production prompt via `NODESPACE_PROMPT_DUMP`, but honestly documented as not yet changing this model's behavior, with a real tracked follow-up issue (#1637) proposing the actual fix (query decomposition). This is a net positive change. No blocking issues found.

### Findings

#### Critical Issues
None.

#### Verification of the five requested checks

1. **Wording clarity/consistency** — Both additions match the file's established style: imperative, ALL-CAPS emphasis markers for hard constraints (`LISTING ALL OF A TYPE WITH NO FILTER`, `ONE WORD ONLY`), a worked example (consistent with e.g. the `DATABASE\" = SCHEMA` and `ENUM FIELDS` rules elsewhere in the same file), and inline parenthetical examples matching the surrounding bullets' density. Line 48 and line 52 are now visually long (~450 and ~640 chars) compared to neighboring bullets, but every bullet in this constant is already a single long line by convention (it's a `\n\`-continued string, not prose meant to be read as source) — consistent with existing style, not a new problem.

2. **Existing unit tests asserting on content** — Checked `agent_guidance.rs`'s own `#[cfg(test)]` module (lines 66-113). None of the four tests assert on the specific substrings that changed (they check `"ALWAYS search_nodes first"`, `"META QUESTIONS"`, `"SCHEMA CLAIMS WITHOUT VERIFICATION"`, `"IDENTICAL TOOL CALLS"` — all untouched by this commit). Ran `cargo test -p nodespace-agent agent_guidance` — 4/4 pass, no update needed.

3. **Other consumers of `TOOL_STRATEGY_RULES`** — Two: `prompt_assembler.rs` (local Ollama agent's seeded prompt nodes) and `acp/context_assembly.rs` (external-agent `CLAUDE.md`/`AGENTS.md` generation, gated by a 50,000-token `DEFAULT_TOKEN_BUDGET`). Checked both:
   - `prompt_assembler.rs::seed_prompt_bodies_match_expected_bytes` (line 415) does an exact-bytes comparison, but derives its expected value via `format!("{}\n\n{}", SCHEMA_CREATION_RULES, TOOL_STRATEGY_RULES)` from the live constants rather than a hardcoded literal — it self-updates and cannot drift. Ran it directly: passes.
   - `prompt_assembler.rs::assembled_prompt_contains_full_tool_strategy_body` (line 516) only asserts on unrelated needles (`"TOOL STRATEGY:"`, `"CLARIFICATION CONTRACT"`, etc.), none touched by this diff. Ran it: passes.
   - `acp/context_assembly.rs` (line 539) also builds guidance via live `format!()`, not a snapshot — no test there hardcodes the string either.
   - Token budget: the diff adds ~220 characters (~50-60 tokens) against a 50,000-token budget — negligible, and `guidance_tokens` is reserved off the top of the budget before other sections consume it (`context_assembly.rs:546-547`), so this can't cause guidance truncation.
   - `cargo fmt -p nodespace-agent --check` passes cleanly on the changed file.

4. **Is keeping change #2 defensible despite not measurably helping?** Yes. The commit message is explicit and honest about the outcome (verified present in the real prompt via `NODESPACE_PROMPT_DUMP`, but did not change behavior on the one remaining failing scenario), and files the actual fix as a separate tracked issue (#1637 — confirmed open, and its description matches: query decomposition for requests that bundle multiple implicit semantic decisions). This is the right call, not scope creep or wishful-thinking dead weight:
   - The instruction is objectively correct guidance regardless of whether *this* model complies with it — other/future models in the Ollama slot may well benefit.
   - It cost two sentences in an already-large prompt constant against a 50k-token budget — not a meaningful tax.
   - The alternative (reverting) would mean discarding verified-correct guidance because one specific quantized small model doesn't yet follow it — that's optimizing the source of truth for a single model's current limitation rather than fixing the actual gap (decomposition), which is exactly what #1637 targets.
   - The one risk worth naming: if #1637 lands and makes this line's guidance obsolete or redundant (e.g. decomposition removes the need for `search_nodes` to receive a raw user query at all), this line should be revisited then — not a blocker now, just a note for whoever picks up #1637.

5. **General diff quality** — Clean and minimal: 2 lines changed, no unrelated edits, no test files touched because none needed touching, commit message documents the verification methodology (real harness, real prompt dump, real isolated DB runs) rather than asserting success by fiat. This is exactly the kind of evidence trail that should accompany a change to a prompt string that has no compile-time contract — the actual behavior only shows up at inference time.

#### Suggested Improvements
None blocking. Optional for a future pass: consider whether `execute_query`'s own docstring/tool-schema description (not reviewed here — out of scope for this file) should also mention the "no filter → don't use me" guidance redundantly, in case a future model reads tool schemas more carefully than system-prompt bullets. Not required for this PR.

#### Nitpicks
- **Nit:** `packages/agent/src/agent_guidance.rs:48,52` — the two changed bullets are now among the longest in the file (~450/~640 chars). Not a real problem given the file's existing convention of single-line bullets, but if this constant keeps growing, it may be worth revisiting whether `TOOL_STRATEGY_RULES` should be decomposed into named sub-constants (search-tool-selection, mutation-flow, etc.) for easier targeted review in future diffs. Not blocking here.

### Recommendation

**APPROVE.** This commit is a well-evidenced, narrowly-scoped fix to production prompt guidance. Both changes are verified through real end-to-end mechanisms (harness pass-rate delta, prompt-dump inspection) rather than asserted by inspection alone, no existing tests needed updates (confirmed by running them), no other consumer of the changed constant is put at risk (confirmed via token-budget and exact-bytes-test analysis), and the one change that didn't measurably help is transparently documented with a real tracked follow-up rather than silently left in or silently dropped.

malibio (commented):
Re-review of commit 7e85ffb2 (agent_guidance.rs prompt fix) — recommendation: APPROVE (GitHub blocks self-approval since I'm the PR author's account; posting as a review comment instead). Full analysis in the PR comment above. Ran the relevant Rust tests directly (agent_guidance, seed_prompt_bodies_match_expected_bytes, assembled_prompt_contains_full_tool_strategy_body — all pass, no updates needed), verified no other consumer of TOOL_STRATEGY_RULES is at risk (both prompt_assembler.rs and context_assembly.rs derive expected content live via format!(), not hardcoded snapshots), confirmed the token-budget headroom is trivially sufficient, and confirmed issue #1637 (cited as the tracked follow-up for the guidance that didn't measurably change model behavior) is real, open, and on-topic. No blocking issues.

PR #1639, opened 2026-07-12
Adopt Gemma 4 E4B as the locked native model (supersedes ADR-046's Ministral 8B lock)

malibio:
Closes #1636


malibio:
### Code Review Summary

Net positive, mechanically clean, and well-justified by ADR-056 (which I read in full and treat as sound background per the review brief). The two-file diff correctly and minimally implements the "recommendation logic" half of the ADR-056 decision. However, verification of downstream call sites surfaced a real, user-facing scope gap: the actual frontend model picker and local-agent dropdown are gated by a separate curated constant that this PR does not touch, and that constant still excludes all Gemma 4 variants. As written, this PR changes what the daemon *recommends* internally but does not change what a user can actually *see or select* in the app. I'd request changes on that basis, not on the core logic itself.

### Findings

#### Critical Issues

- **`packages/desktop-app/src-tauri/src/commands/chat_models.rs:54`** — `EXPOSED_GGUF_MODEL_IDS: &[&str] = &["ministral-8b-q4km"]` is unchanged by this PR. Its own doc comment states: *"Gemma 4 was previously exposed but pulled after issues; the remaining lineup is the RAM-tier-appropriate set considered stable today."* This constant gates `chat_model_list` (`chat_models.rs:78`), which is the sole source for:
  - `model-store.svelte.ts`'s `recommendedModel` getter (`model-store.svelte.ts:153-163`) — a **client-side re-implementation** of "pick the best model," derived from `this.models` (populated only via `chatModelList()`, `model-store.svelte.ts:186-188`), filtered by RAM/size — entirely independent of the Rust `recommended_model_id()` this PR changes.
  - `agent-store.svelte.ts`'s real (non-mock) local-agent list (`agent-store.svelte.ts:75,145`), which is what actually populates the chat agent dropdown users pick from.

  Net effect: after this PR merges, `gemma-4-e4b-q4km` is not present in `EXPOSED_GGUF_MODEL_IDS`, so it cannot appear in the Settings model manager, cannot appear in the chat agent picker, and the frontend's own `recommendedModel` heuristic can never select it — it will keep recommending Ministral 8B (the only entry in the list) regardless of what the daemon's `recommended_model_id()` now returns. Separately, `chatModelRecommended()`/`chat_model_recommended` (the Tauri command wired straight through to the very RPC this PR's `recommended_model_id()` change affects) is registered (`lib.rs:530`) and implemented, but **is not called from any UI component** — only exercised by a type-existence assertion in `tauri-commands.test.ts`. So the code path this PR actually modifies is reachable only by the e2e test and by direct gRPC calls, not by any real user flow.

  This means the ADR-056 decision — "Gemma 4 E4B is now the locked native model" — is true at the daemon/CLI layer but not true in the shipped desktop app: a real user cannot download, select, or be recommended Gemma 4 E4B through the UI today. That's a materially incomplete implementation of the ADR's intent, not just a docs/test nit. Recommend either (a) adding `"gemma-4-e4b-q4km"` to `EXPOSED_GGUF_MODEL_IDS` as part of this PR (small, in-scope, mechanical — matches the size/shape of the rest of this diff), or (b) if there's a deliberate reason it's still withheld (e.g. the Scenario 5/6 gaps below are considered blocking for UI exposure specifically, even though not blocking for the native-path default), that reasoning needs to be stated explicitly somewhere — it is not currently in ADR-056, the PR description, or a code comment, so as it stands this reads as an oversight rather than an intentional phased rollout.

#### Suggested Improvements

- **`packages/agent/src/local_agent/model_manager.rs:462-464`** — Per the specific scrutiny requested: the doc comment above `recommended_model_id()` is good and does explain the RAM-tiering deviation clearly. But the actual `GEMMA_4_E4B.id` line itself carries no inline marker, and the function is four lines long with no runtime guard — a future maintainer skimming a diff (not reading doc comments, which is common in quick "simplify this" refactors) could plausibly "clean up" `fn recommended_model_id() -> &'static str { GEMMA_4_E4B.id }` into `Self::recommended_model_id_for(ModelFamily::Gemma4)` for consistency with how every other family works, silently reintroducing 12B/31B on high-RAM machines. Doc comments don't show up in a GitHub diff view of just the changed line, and this is exactly the kind of one-line "obvious" refactor that slips through review. Suggest a short inline comment directly on or immediately above the `GEMMA_4_E4B.id` return, e.g. `// NOT recommended_model_id_for(Gemma4) — that RAM-tiers into 12B/31B, which are parked per ADR-046/ADR-056.` Belt-and-suspenders given how easy the regression is to introduce and how non-obvious the consequence (silently re-enabling a defective tier) is from the line alone.

- **Testing gap, acknowledged not blocking**: there's no test asserting `recommended_model_id()` stays E4B under simulated high-RAM conditions — i.e., nothing would catch the regression described above if it happened. This is because `detect_system_ram()` (`model_manager.rs:1117-1121`) reads real system RAM via `sysinfo` and has no injection seam (no trait, no env-var override, no constructor parameter). Making it mockable purely to cover this one regression risk is probably not worth it on its own (YAGNI — the function is trivial today), but note it as an accepted gap rather than silently having no coverage. The existing tests (`recommended_model_returns_valid_id`, `recommended_spec_has_valid_fields`) correctly pin the *current* output but cannot exercise the RAM-tiering bypass logic at all, on any machine.

- **Missing follow-up issue reference (per ADR-056's own stated condition)**: ADR-056's Consequences section says *"A follow-up issue should attempt prompt/skill tuning"* for the Scenario 5 (query-tool mismatch) and Scenario 6 (incomplete update turn) gaps, but no issue number appears anywhere in the ADR or this PR. I checked `gh issue list` for anything referencing these scenarios or #1636 follow-up work and found nothing. I agree with the risk framing itself — Scenario 6 (search fires, `update_node` never called) is a truncated turn, not data corruption, and is meaningfully different in kind from Ministral 8B's write-side over-action (an unrequested `create_schema` mutation actually landed). That asymmetry (recoverable-omission vs. actual-unwanted-mutation) is a reasonable basis for accepting E4B's regression. But the ADR conditions merging this on a follow-up issue existing, and none does yet — that's a real, checkable gap, not a judgment call. Recommend filing the tracking issue before or immediately after merge (doesn't need to block merge itself, but shouldn't be allowed to silently fall through the cracks either).

#### Nitpicks

- Nit: `packages/desktop-app/src/lib/stores/agent-store.svelte.ts:8,42` — the module doc comment ("Local model agents (Ministral 3B/8B)") and `MOCK_LOCAL_AGENT` (id `local:ministral-3b-q4km`) are now stale relative to the new native default. Low stakes — this is dev-mode/non-Tauri mock data and an illustrative code comment, not real selection logic, so it doesn't affect production behavior. Worth a follow-up cleanup pass but not this PR's problem to solve, especially since real exposure is blocked by the `EXPOSED_GGUF_MODEL_IDS` issue above regardless.

- Nit: the PR is otherwise a clean, atomic, single-purpose change (recommendation logic + the one e2e test that exercises it) — good shape for review. The main issue is not scope creep but scope *shortfall*: it stops one layer short of making the ADR-056 decision observable to an actual user.

### Requirements Check

- Diff matches commit message/PR description: yes, exactly as described.
- `recommended_model_id()` returns `GEMMA_4_E4B.id` directly (not RAM-tiered): confirmed correct and intentional; doc comment explains why but could use a stronger inline guard (see above).
- Scope check on other Ministral-default call sites: the `agent_loop.rs:1513` `MockInferenceProvider` fixture is correctly out of scope (test-only, exercises agent-loop logic against a fake spec, not a real default). The `Ollama | OpenAiCompat` arm of `recommended_model_id_for` is correctly out of scope (sizing heuristic for non-GGUF-catalog models, a materially different concern). However, the scope check missed `chat_models.rs:54` (`EXPOSED_GGUF_MODEL_IDS`) and its two frontend consumers (`model-store.svelte.ts`, `agent-store.svelte.ts`) — these are real production code, not test fixtures, and they are the actual UI-facing path this ADR is meant to affect. This is the review's central finding.
- llama-cpp-2 crate correctly left untouched/pinned: confirmed via diff — not present in this PR.
- Test coverage: the two updated unit tests are correct and sufficient for what they cover (recommendation value, spec fields) but cannot and do not cover the RAM-tiering-bypass regression risk; noted as an accepted, non-blocking gap given `detect_system_ram()`'s lack of a test seam.
- Atomicity: diff itself is atomic and well-scoped for what it does; the problem is completeness relative to the ADR's intent, not bundling of unrelated changes.

### Recommendation

**Request changes.** The core logic change (`recommended_model_id()`, its doc comment, and the two unit tests) is correct, intentional, and well-reasoned — I'd approve that part standalone. But `EXPOSED_GGUF_MODEL_IDS` in `chat_models.rs` still excludes every Gemma 4 variant, which means this PR does not actually make Gemma 4 E4B selectable, visible, or recommended anywhere in the shipped desktop app — only in the daemon's internal default and in test code. Given the PR title's claim ("Adopt Gemma 4 E4B as the locked native model") and that this is the ADR-056 decision's entire practical purpose, I don't think this should merge without either exposing E4B in the frontend catalog or an explicit, stated reason (from the author, not inferred by me) for why UI exposure is intentionally deferred to a separate PR/issue.

malibio (commented):
REVIEW VERDICT: Request Changes (posted as --comment because GitHub disallows a formal request-changes review on your own PR). See the detailed findings comment above. Summary: core recommendation-logic change is correct and well-justified by ADR-056, but EXPOSED_GGUF_MODEL_IDS in packages/desktop-app/src-tauri/src/commands/chat_models.rs:54 still excludes all Gemma 4 variants, so the frontend model picker and local-agent dropdown cannot surface gemma-4-e4b-q4km at all post-merge — the ADR-056 decision is not yet observable to a real user. Recommend addressing before merge.

malibio:
## Review feedback addressed

**Critical — fixed:**
- `EXPOSED_GGUF_MODEL_IDS` in `chat_models.rs` now includes `gemma-4-e4b-q4km` instead of `ministral-8b-q4km`. This is the real gate behind `chat_model_list`, which feeds both the CLI and the Settings UI model picker/agent dropdown — without this, the daemon-level `recommended_model_id()` change was invisible to any real user. Confirmed the frontend's `recommendedModel` getter (`model-store.svelte.ts`) picks this up transitively since it derives from `chatModelList()`.
- Cleaned up stale "Ministral 3B/8B" doc-comment examples and the non-Tauri `MOCK_LOCAL_AGENT` fallback in `agent-store.svelte.ts` to match (flagged as a nit, addressed while in the area).

**Important — addressed:**
- Added an inline guard comment directly on the `GEMMA_4_E4B.id` return line in `recommended_model_id()`, warning against "simplifying" it to `recommended_model_id_for(ModelFamily::Gemma4)` (which would silently reintroduce the parked 12B/31B tiers on higher-RAM machines). Doc comments don't show in a single-line diff view, so this is a second line of defense per the review's reasoning.
- Filed #1640 to track the Scenario 5/6 prompt-tuning work ADR-056 already called for, and linked it from the ADR (previously just said "a follow-up issue should...", now points to #1640).

**Accepted as non-blocking gaps (per review's own framing):**
- `detect_system_ram()` has no test seam, so the RAM-tiering-bypass regression the guard comment defends against isn't directly unit-testable. Not fixing this now — the function is trivial and adding a mock seam purely for this one risk isn't proportionate.

**Separately, corrected ADR-056 itself:** the review didn't flag this, but a follow-up question from the user prompted me to independently verify (rather than take on faith) the #1634 tool-count-ceiling claim I'd cited as supporting evidence for the switch. Built a throwaway native-path probe (`LlamaChatInferenceEngine`, the same code path production uses) and tested `ministral-8b-q4km` directly at 2/3/5/7/10 offered tools — it did **not** degrade at any count, unlike `mistral:7b`/`ministral-3:3b` on Ollama (which I separately reproduced and confirmed does degrade, matching #1634's numbers). Retracted the tool-count-ceiling claim from ADR-056's Context, Consequences, Re-Introduction Path, and Alternatives Considered sections rather than leave a disproven rationale in an accepted ADR. This narrows the ADR's actual case to: the old ADR-046 baseline is unverifiable, and the re-measured Ministral 8B failure (an unrequested extra `create_schema` write) is a more expensive error class than either of E4B's read-side/incompleteness failures.

All changes pushed (`b8fac96d`, `f3b4c8e1`). `test:all` and the pre-push gate (including the real e2e `ai_chat_send_to_idle_test`) pass on both.

## Re-Review Decision

**RE-REVIEW NEEDED.** The critical finding involved a real, user-facing scope gap and a production code change (`EXPOSED_GGUF_MODEL_IDS`), not just a docs/style fix — worth a second pass to confirm the fix is complete and didn't miss another call site.

malibio:
### Code Review Summary (Re-Review)

Confirmed this is a re-review — the prior "Request Changes" comment (2026-07-12T01:35:11Z) is present on the PR, and its critical finding (`EXPOSED_GGUF_MODEL_IDS` still hardcoded to Ministral 8B) is correctly fixed by `b8fac96d`. However, tracing forward one more layer surfaced a **second, independent allowlist** with the exact same defect, in a file neither review has touched: `LOCAL_FAMILIES` in `model-manager.svelte`. Net: closer to done, but not yet fully closing the gap the PR title claims.

### Findings

#### Critical Issues

- **`packages/desktop-app/src/lib/components/settings/model-manager.svelte:34`** — `const LOCAL_FAMILIES: ModelFamily[] = ['ministral', 'qwen35'];` does not include `'gemma4'`. This constant gates two real UI surfaces:
  - `localModels` (line 35, rendered at lines 249/252) — the card grid in Settings → Models that lists downloadable/loadable local models. With `gemma4` excluded, `gemma-4-e4b-q4km` (now returned by `chatModelList()` thanks to the `EXPOSED_GGUF_MODEL_IDS` fix) is filtered out before it ever reaches the template. The Settings model manager will render its empty state ("no local models") even though the daemon is correctly serving Gemma 4 E4B as the one exposed GGUF model.
  - `buildDefaultOptions()` (line 121) — populates the "Default model" dropdown's `native:<model-id>` entries the same way. Gemma 4 E4B can never appear as a selectable default-model option here either, even once downloaded through some other path.

  The comment directly above the constant (line 32-33) is explicit about the dependency and was simply not updated when the sibling allowlist was fixed: *"Settings shows the curated set already filtered at the Tauri layer (see `EXPOSED_GGUF_MODEL_IDS` in `chat_models.rs`); families listed here must match whatever that allowlist currently exposes."* That invariant is now broken — `EXPOSED_GGUF_MODEL_IDS` exposes `gemma4`, `LOCAL_FAMILIES` does not.

  This is the same class of bug as the first review's finding (a second gate the intended-visible model must pass, silently left stale) and it's user-facing: a user going to Settings → Models specifically to find/download/select Gemma 4 E4B will see nothing there. Fix is one line: `const LOCAL_FAMILIES: ModelFamily[] = ['gemma4', 'qwen35'];` (drop `'ministral'` too, unless Ministral models should remain manageable from Settings even though delisted from the exposed catalog — worth an explicit call either way, since leaving `'ministral'` in is harmless today but silently permits a future re-expansion of `EXPOSED_GGUF_MODEL_IDS` to leak back in without a corresponding `LOCAL_FAMILIES` update, i.e. the same bug in reverse).

  Scope check confirms this is the only remaining gap of this kind — I re-traced every consumer of `chat_model_list`/`chatModelList()`: `model-store.svelte.ts` (no family allowlist, derives `family` from id prefix, correctly picks up `gemma4` automatically), `agent-store.svelte.ts`'s real `modelToAgent()` path (no family allowlist, chat agent dropdown is correctly fixed by the `EXPOSED_GGUF_MODEL_IDS` change alone), and `model-manager.svelte`'s `LOCAL_FAMILIES` (the one above). No other stale Ministral-default call sites found; all other `ministral` references in the tree are catalog entries for the (still valid, non-default) Ministral family, test fixtures, or historical regression-test comments — correctly out of scope.

#### Confirmed correct

- Full diff (`63d485bf`, `b8fac96d`, `f3b4c8e1`) matches all three commit messages exactly — no surprises, no scope creep.
- `EXPOSED_GGUF_MODEL_IDS` fix is correct and its doc comment was properly updated to reflect the new rationale (Gemma 4 E4B locked, 12B/31B parked) rather than left stale.
- Guard comment on `recommended_model_id()` (`model_manager.rs:461-463`) is reasonable, correctly placed directly on the `GEMMA_4_E4B.id` return line as promised, and introduces no issue of its own.
- `agent-store.svelte.ts` doc-comment/mock cleanups (module doc, `localAgentModelId` example, `MOCK_LOCAL_AGENT`) are correct and consistent.
- Issue #1640 exists, is open, has a clear problem statement/acceptance criteria matching ADR-056's Consequences section, and is correctly linked from ADR-056's Consequences section (`Tracked in issue #1640`).
- ADR-056 (nodespace-docs, committed to main as of `348a414`) reads honestly. The tool-count-ceiling retraction is not a quiet deletion — it has its own dedicated subsection ("The tool-count ceiling — ruled out as a rationale for this decision") explaining what was originally claimed, what was tested, and what was found, and every downstream section that referenced the original claim (Alternatives Considered, Re-Introduction Path) was consistently updated to match rather than left inconsistent. No overclaiming found; the ADR is appropriately hedged (e.g. "This is a closer call than the original draft implied").
- Test coverage for what changed in this PR (the two `model_manager.rs` unit tests, the e2e test's model swap) is adequate for the logic actually touched. The previously-noted `detect_system_ram()` test-seam gap is accepted as non-blocking, consistent with the first review's framing.

### Recommendation

**Request changes** (one more small round). The daemon-level and chat-agent-dropdown paths are now genuinely fixed end-to-end, but `model-manager.svelte`'s `LOCAL_FAMILIES` is a third, independent gate that still excludes Gemma 4 E4B from the Settings model-management UI and the default-model dropdown — the same "recommended but not actually visible/selectable" gap the first review found, recurring in a sibling file. This should be a fast fix (add `'gemma4'` to the array, decide on `'ministral'`'s fate), and I'd suggest a final grep for `LOCAL_FAMILIES`-style per-surface allowlists before the next re-review to make sure there isn't a fourth one.

malibio:
## Round 2 review addressed + scope expanded

**Round 2 finding fixed:** \`LOCAL_FAMILIES\` in \`model-manager.svelte\` (a second, independent allowlist gating the Settings model-grid and default-model dropdown) still had \`['ministral', 'qwen35']\` — same bug class as round 1's \`EXPOSED_GGUF_MODEL_IDS\` finding, just in a sibling file the first review didn't check. Fixed to \`['gemma4']\`. Swept the rest of the frontend for other per-surface allowlists per the review's own recommendation — none found.

**Scope expanded, at explicit user direction mid-review: full removal of the Qwen-family model catalog**, not just excluding it from exposure. This is a separate decision from the Gemma 4 E4B swap, landed in the same PR/session, documented in ADR-056 (nodespace-docs).

Removed entirely (catalog entries, not just hidden):
- \`qwen3-8b-q4km\`, \`qwen35-9b-q4km\`, \`qwen36-35b-a3b-q4km\`, and the Qwen3.5-based \`ornith-1-9b-q4km\`
- \`ModelFamily::Qwen3/Qwen35/Qwen36\` (Rust and TypeScript)
- The "Format D" Ornith/Qwen3.5 XML \`<tool_call>\` parser (\`parse_tool_calls_xml\`, \`parse_xml_tool_call_block\`) and its tests
- The \`<tool_call>\` XML interception branch in \`emit_oai_delta\`
- \`ThinkSplitter\` — a struct threaded through the streaming generation loop's control flow (carry-state routing alongside Gemma's \`ChannelSplitter\`) that existed solely for Ornith's \`<think>\` reasoning convention; required restructuring \`emit_oai_delta\`'s signature and the finalize path, not just deleting a block
- \`strip_ornith_tool_call_leak\` response-cleanup stripper and its tests
- The Qwen/Ornith UI notice and family references in Settings

Net diff: 706 lines removed, 31 added. Closed #1477 and #1464 (Ornith re-enable/evaluation) as moot — there's no catalog entry left to re-enable or evaluate.

**Verification:**
- Full workspace build clean (\`cargo build --workspace\`)
- \`bun run test:all\` (frontend + skill + Rust) passes, no new failures
- \`bun run quality:fix\` clean (0 lint errors, 0 warnings, clippy clean)
- Pre-push gate passed, including the real e2e \`ai_chat_send_to_idle_test\`

Pushed as \`35085d05\`.

## Re-Review Decision

**RE-REVIEW NEEDED.** This is a large, cross-cutting deletion (catalog, enum, streaming-pipeline internals, response processing, frontend types) touching working code paths, not just the model-swap scope the first two rounds reviewed. Worth a dedicated pass checking for any remaining dangling reference, orphaned test, or behavior change in the streaming/reasoning-routing restructuring specifically — that's the highest-risk piece (ChannelSplitter's carry-state logic changed shape when ThinkSplitter was removed from the finalize path).

malibio:
### Code Review Summary (Round 3 — Qwen-family removal, commit `35085d05`)

Reviewed the full PR (all 4 commits, `origin/main...HEAD`) with focus on the highest-risk piece: the `ThinkSplitter` removal and `emit_oai_delta`/finalize-path restructuring in `packages/nlp-engine/src/chat/mod.rs`. Independently rebuilt (`cargo build --workspace` — clean) and ran the affected test suites directly rather than trusting the PR's own report. Everything checked out correct. One blocking-but-mechanical item found: **the PR now has a real merge conflict with `origin/main`**, unrelated to this PR's own correctness.

### Findings

#### Critical Issues

- **Merge conflict with `origin/main` (`mergeable: CONFLICTING`).** `packages/desktop-app/src-tauri/tests/ai_chat_send_to_idle_test.rs` was independently modified by `ee3d5e06` ("Align CLI + SKILL.md with local-agent capability and guidance", merged to `main` the same day) — that commit is unrelated to this PR's model-lock decision but happens to touch the exact same lines (`MODEL_ID`, `MODEL_FILENAME`, and the doc comment above them), reverting/keeping them at `ministral-8b-q4km` with different stall-investigation prose than what predates this branch. `git merge-tree` confirms 3 real conflict hunks, all in this one file, all on the model-ID lines this PR intentionally changes to `gemma-4-e4b-q4km` / `gemma-4-E4B-it-Q4_K_M.gguf`. This isn't a defect in this PR's logic — it's a stale branch needing a rebase — but it must be resolved (keeping this PR's `gemma-4-e4b-q4km` values, matching ADR-056) before merge is possible at all. No other files conflict.

#### Confirmed correct (verified independently, not just re-reading the diff)

- **`ThinkSplitter` removal and `emit_oai_delta` finalize restructuring** — the highest-risk part of this round. Confirmed:
  - `partial_marker_suffix_len` is still used exactly once, correctly, inside `ChannelSplitter::push` — not orphaned, not accidentally removed alongside `ThinkSplitter`.
  - The finalize-path simplification (`channel_splitter.finalize()` now directly produces `final_answer`/`final_reasoning`, no `think_splitter.push()`/`.finalize()` routing step) is behaviorally correct: `think_splitter` was a pure pass-through no-op for any non-`<think>` text, so removing the routing step changes nothing for Gemma 4 or any surviving model family.
  - Ran `cargo test -p nodespace-nlp-engine --features chat-service chat::` directly: **78 passed, 0 failed**. All `channel_splitter_*` tests (marker-within-delta, split-across-deltas, opener-without-closer, multibyte, multiple-channels-concatenate, trailing-partial-marker) are untouched and green — `ChannelSplitter` itself was not damaged by the `ThinkSplitter` removal.
  - `emit_oai_delta_finalize_returns_channel_splitter_carry` (the rewrite of the old `..._routes_channel_carry_through_think_splitter` test) still meaningfully exercises `ChannelSplitter`'s carry-across-finalize behavior — it asserts a trailing `<|chan` partial marker survives `finalize()` as answer text with no reasoning emitted, which is a real, non-trivial assertion about `ChannelSplitter`'s own carry state, not a no-op. The original regression it reproduced (channel carry leaking past `think_splitter` unrouted) is structurally impossible to reproduce anymore since there's nothing left to route through, but the surviving assertions are still doing real work.
  - `regression_gemma4_12b_oai_delta_emits_tool_call_chunks` passes with the `think_splitter` argument dropped from its `emit_oai_delta` call — confirmed this Gemma-4-specific regression test is otherwise unmodified and still tests the same thing.
- **No dangling references anywhere in the tree** (Rust or TypeScript, outside appropriately historical doc comments) to `ThinkSplitter`, `parse_tool_calls_xml`, `parse_xml_tool_call_block`, `XML_TOOL_CALL_OPEN`/`CLOSE`, `strip_ornith_tool_call_leak`, `ornith_tool_call_re`, or `ModelFamily::Qwen3`/`Qwen35`/`Qwen36` (Rust `agent_types.rs` or TS `agent-types.ts`). The one remaining hit (`chat_models.rs:47-48`) is a doc comment correctly explaining historical context ("Qwen-family catalog entries ... have been removed from the catalog entirely").
- **`response_processing.rs`**: `stray_channel_marker_re` is confirmed still used by the surviving `strip_channel_blocks` (not orphaned by the `strip_ornith_tool_call_leak` removal). The `apply()` pipeline call list in `normalize_response_traced` has no gap — `strip_ornith_tool_call_leak`'s entry was cleanly excised, all 7 remaining strippers still chain correctly. Ran `cargo test -p nodespace-agent response_processing`: **28 passed, 0 failed**.
- **`model_manager.rs`**: `cargo test -p nodespace-agent model_manager`: **47 passed, 0 failed**. `list_returns_all_catalog_models`'s updated count (13 → 9) is arithmetically correct: 13 original minus the 4 removed Qwen-family entries (`QWEN3_8B`, `QWEN35_9B`, `QWEN36_35B_A3B`, `ORNITH_1_9B`) = 9, matching the current `CATALOG` array exactly (Ministral 3B/8B/14B, Mistral NeMo 12B, Mistral Small 3.2, Gemma 4 E4B/12B/12B-Unsloth/31B).
- **Frontend**: `LOCAL_FAMILIES` (`model-manager.svelte`) is now `['gemma4']`, correctly matching `EXPOSED_GGUF_MODEL_IDS` (`chat_models.rs`) = `['gemma-4-e4b-q4km']` — the round-2 finding's invariant is intact. `model-store.svelte.ts`'s family-inference function is correctly simplified to drop the `qwen35`/`ornith` branch. `ModelFamily` TS union correctly drops `'qwen35'`, matching the Rust enum. Linted the 4 changed frontend files directly (`bunx eslint`) — clean, 0 errors/warnings.
- **ADR-056** (read fresh from `nodespace-docs`): accurately documents the removal in its Decision and Consequences sections, names every removed symbol correctly (cross-checked each one against the actual diff — all present, none stale or missing), and is appropriately hedged rather than overclaiming (e.g., explicitly notes the tool-count-ceiling retraction in its own subsection rather than silently dropping it).
- **Issue closures #1477 and #1464**: both accurately describe what was removed and correctly state that re-adding Ornith/Qwen would require re-implementation from scratch, not "re-enabling" — consistent with the actual code state (no dead/disabled code left behind, full deletion confirmed by the dangling-reference sweep above).

#### Nitpicks

- Nit: none beyond the merge conflict above — this round's diff is clean.

### Recommendation

**Do not merge yet — rebase required.** The Qwen-removal work itself (`35085d05`) is correct, complete, well-tested, and accurately documented; I found no logic defect, no dangling reference, and no test-coverage regression in the highest-risk restructured code (`ChannelSplitter`/`ThinkSplitter`/finalize path). But GitHub reports `mergeable: CONFLICTING`, and I confirmed a real 3-hunk conflict in `ai_chat_send_to_idle_test.rs` against `origin/main`'s `ee3d5e06` (an unrelated same-day PR that touched the same lines). Rebase onto current `main`, keep this PR's `gemma-4-e4b-q4km`/`gemma-4-E4B-it-Q4_K_M.gguf` values (per ADR-056) when resolving, re-run `test:all` + the e2e test post-rebase, then this is ready to merge.

malibio:
## Rebase conflict resolved

Per round 3's finding: rebased onto \`origin/main\` (picked up \`ee3d5e06\`, the unrelated same-day \`#1529\` CLI-capability merge). One real conflict in \`ai_chat_send_to_idle_test.rs\` — both branches touched the same MODEL_ID/MODEL_FILENAME/doc-comment lines. Resolved by keeping this PR's \`gemma-4-e4b-q4km\` values (per ADR-056) while preserving \`main\`'s useful historical context (the \`ministral-3b-q4km\` → \`ministral-8b-q4km\` interim-fix backstory) in the merged doc comment, rather than discarding either side's content wholesale.

Verified after rebase:
- \`cargo build --workspace\` clean
- \`bun install\` (picked up #1529's dependency changes, no issues)
- Pre-push gate passed in full on force-push, including \`test:all\` and the real e2e \`ai_chat_send_to_idle_test\`
- \`gh pr view --json mergeable,mergeStateStatus\` now reports \`MERGEABLE\` / \`CLEAN\`

Force-pushed (\`--force-with-lease\`) as \`ae5a0841\`. Commit content is unchanged from what round 3 reviewed — only SHAs changed from being replayed onto the new base.

Ready for merge, pending final confirmation.

PR #1641, opened 2026-07-12
Replace hand-synced proto/domain conversions with a round-trip-tested boundary

malibio:
Closes #1502


malibio:
### Code Review Summary

Clean, well-scoped fix. The schema-primitive dedup is a straightforward re-export with no behavioral drift, and the two new proptests are genuinely load-bearing — I traced both by hand against hypothetical dropped-field scenarios and confirmed they fail (not trivially green). The `into_node` snake_case fix is a real bug fix, surfaced by the round-trip proptest exactly as the commit message claims, and I traced the full daemon → gRPC → Tauri → frontend path and found no consumer that reads the raw wire properties directly — every response goes through `node_to_typed_value`/`task_node_to_value`, which tolerantly reads both cases. This is a net-positive change that should ship.

### Requirements Check

- [x] **A round-trip proptest covers the conversion boundary** — `task_node_survives_node_round_trip` (packages/core/src/models/task_node.rs:911-950) covers the domain↔Node boundary; `task_promotes_all_stored_fields` / `ai_chat_promotes_all_stored_fields` / `schema_promotes_all_stored_fields` (packages/nodespace-types/src/convert.rs:321-419) cover the Node→wire-JSON promotion boundary. Both verified non-trivial (see Findings).
- [x] **Domain types with no behavior over proto are removed** — The 7 byte-for-byte-duplicated schema primitives (SchemaField, SchemaRelationship, EnumValue, EdgeField, RelationshipDirection, RelationshipCardinality, SchemaProtectionLevel) are deleted from core and re-exported from nodespace-types (packages/core/src/models/schema.rs now 16 lines, down from 912). The remaining TaskNode/AiChatNode/SchemaNode struct duplication is explicitly out of scope per the issue framing (dependency-graph-forced — desktop-app/src-tauri has zero nodespace-core dependency) and I did not flag it as unaddressed. See Suggested Improvements for a related but distinct naming issue this scoping decision surfaces.
- [x] **Code passes `bun run quality:fix`; tests pass** — Per baseline: nodespace-types 38 passed, core lib 955 passed, daemon 64 passed, desktop-app 42 passed + integration, frontend 4031 passed, clippy clean, pre-push gate passed. Not re-run; spot-checked the diff for obvious lint/format issues and found none.

### Code Review Findings

#### Critical Issues

None found.

#### Suggested Improvements

- **packages/core/src/models/node.rs:54 vs packages/nodespace-types/src/schema.rs:79** — Two distinct `RelationshipDirection` enums now coexist in the core crate under the same name: `nodespace_core::models::node::RelationshipDirection` (traversal direction for `NodeRelationship`/subtree queries, `Copy`-derived, pre-existing) and `nodespace_core::models::schema::RelationshipDirection` (schema relationship direction, re-exported from nodespace-types by this PR). `models/mod.rs` re-exports the former at the `models::` top level but deliberately omits the latter from the same re-export list to dodge the collision, so callers needing the schema variant must spell out `models::schema::RelationshipDirection`. This predates the PR (the schema module already existed at that unexported path) and isn't a regression, but this PR is exactly the point where it became visible/relevant — dedup work often surfaces latent naming collisions like this. Per DRY/single-source-of-truth, consider renaming one (e.g. `SchemaRelationshipDirection`) in a follow-up so `use nodespace_core::models::*` doesn't require path-qualification to disambiguate two unrelated "direction" concepts.
- **packages/nodespace-types/src/schema.rs:79-82 vs packages/core/src/models/node.rs:50-56** — Confirmed the serde attribute divergence flagged for scrutiny (`rename_all = "camelCase"` in nodespace-types vs. `rename_all = "lowercase"` in core's old — now deleted — copy) is a non-issue for the *current* variant set: `Out`/`In` are single-word, so both attributes serialize identically ("out"/"in"). No wire-format regression. Flagging only because it's a latent trap: if a future multi-word variant is ever added to `nodespace_types::RelationshipDirection` (e.g. `OutIn` or a friendlier `Bidirectional`), the camelCase attribute would silently diverge from what any lowercase-styled test or client expects. Not a blocker — just worth a code comment noting the current single-word coincidence so a future contributor doesn't add a multi-word variant without checking.

#### Nitpicks

- **Nit:** packages/core/src/models/task_node.rs:494 — The updated doc comment on `into_node` ("Uses snake_case property keys — the canonical on-disk storage form...") is a good improvement, but the struct's own top-of-file doc-comment JSON example at task_node.rs:19-30 still shows `"dueDate": null` in the "serialized (for Tauri/HTTP responses)" example. That's still accurate for the *wire* output (via `task_node_to_value`'s promotion), but could read as inconsistent with the new `into_node` comment to a reader who doesn't trace both paths — consider a one-line clarifying note that this is the post-promotion wire shape, distinct from `into_node`'s on-disk shape.
- **Nit:** packages/core/proptest-regressions/models/task_node.txt — the checked-in regression seed (`completed_at = Some("2000-01-01")`, empty content, all-else-default) is fine to keep per proptest convention, but worth confirming in CI that this seed doesn't silently mask a distinct future regression by always re-passing before novel cases run (standard proptest caveat, not specific to this PR).

### Verification Notes (scrutiny items from review brief)

1. **`into_node` snake_case change safety** — Traced end-to-end: daemon's `update_task_node` (packages/daemon/src/services/node_service.rs:861) calls `task.into_node()` and returns the raw proto `NodeResponse` via `node_to_proto` — no flattening happens daemon-side. On the Tauri side, every command that surfaces a task node response (packages/desktop-app/src-tauri/src/commands/nodes.rs:425/473/805, collections.rs:69/283/320/401) converts the proto back to a `Node` and immediately calls `node_to_typed_value`, which routes task nodes through `task_node_to_value` (packages/nodespace-types/src/convert.rs:96-154) — this reads `dueDate`.or_else(`due_date`), `startedAt`.or_else(`started_at`), `completedAt`.or_else(`completed_at`). No consumer reads raw wire properties directly; the frontend only ever sees the promoted, case-tolerant top-level fields. Confirmed safe.
2. **Proptest strength** — Both new proptests compare against independently-generated input values (not self-referentially derived from the code under test), so a silently dropped field in either direction of either conversion causes a real assertion failure, not a vacuous pass. Verified by manual trace of what happens if `into_node`/`from_node`/`task_node_to_value` each forgot a field — all three scenarios produce a mismatch against the originally-generated value.
3. **`RelationshipDirection` camelCase vs lowercase divergence** — Confirmed both variants are single-word (`Out`, `In`), so old core's `lowercase` and nodespace-types' `camelCase` serialize identically. No wire-value regression from the re-export. See Suggested Improvements for the latent-risk note.
4. **Stale camelCase-pinned tests** — Searched all `"dueDate"`/`"startedAt"`/`"completedAt"` string literals repo-wide. Found one pre-existing test (packages/core/src/models/task_node_test.rs:137, `test_due_date_clear`) asserting `task.as_node().properties.get("dueDate").is_none()` — but this fires when `due_date` is `None`, and neither old nor new `into_node` inserts the key at all in that case, so the assertion is casing-agnostic and not stale. No test found that asserts presence of a camelCase key with a `Some(..)` value that the new snake_case behavior would break.

### Recommendation

**APPROVE**

malibio (commented):
Initial pragmatic code review posted above. Recommendation: APPROVE (see full report in PR comments).

PR #1618, opened 2026-07-11
Remove ci.yml and restore local-only test gating per ADR-047

malibio:
Closes #1617

PR #1655, opened 2026-07-14
Resolve the registry path through NODESPACE_HOME so a redirected DB can't poison it (closes #1648)

mstomar125:
## Problem

`~/.nodespace/databases.toml` (the user-level database registry) could be seeded with a temp-dir path by test harnesses or temp-DB daemon runs, after which every boot served a doomed temp database as "Default". macOS purges `/var/folders`, so all data written there silently disappears — and the boot log printed a different path than the one actually served.

Root cause: the registry path was hard-wired to the real `$HOME/.nodespace` while the served DB path honored `NODESPACED_DB_PATH`. A harness that redirected the DB but inherited `HOME` wrote its temp path into the real registry.

## Fix

- **`NODESPACE_HOME`** — new `nodespace_home()`/`nodespace_dir()` resolver; both the registry path and `resolve_db_path`'s default route through it, so one override isolates *all* NodeSpace state in lockstep.
- **Boot repair** — `repair_doomed_default` re-points the default to the standard path when it lives under a system temp dir while the registry itself does not; a no-op for intentionally isolated (temp) registries, so test fixtures are never clobbered.
- **Truthful log** — boot now logs the registry-resolved served path (`serving default database`) instead of the requested boot path; the `Starting` line's field is renamed `requested_db_path`.
- **Harness isolation** — the real-binary spawn harnesses (test-support `SpawnedDaemon`, the e2e `daemon-harness.ts`, and the benchmark script) set `NODESPACE_HOME` under their temp dir, so test runs never touch `~/.nodespace`.

## Tests
- New daemon unit tests: `default_is_doomed` truth table, `repair_doomed_default` no-op for an isolated registry, `set_standard_default` replaces the previous default. Daemon lib suite green (67 tests).
- Full pre-push gate (`test:all` + `cargo build --bin nodespaced` + `test:e2e`) green locally.

## Acceptance
- [x] Running the test suites never mutates `~/.nodespace/databases.toml`.
- [x] A registry whose default path is gone/temp produces a visible warning + safe fallback, and the boot log reflects the DB actually served.

Closes #1648.

### DIFF
--- a/packages/daemon/src/lib.rs
+++ b/packages/daemon/src/lib.rs
@@ -14,24 +14,44 @@ use std::path::PathBuf;
 
 use anyhow::{Context, Result};
 
+/// Resolve NodeSpace's home directory — the parent of the `.nodespace/` state
+/// directory that holds the database registry (`databases.toml`) and the
+/// managed database files.
+///
+/// Honors `NODESPACE_HOME` so a test harness or alternate deployment can
+/// redirect *all* NodeSpace state — registry and databases together — with a
+/// single override. This is what keeps a redirected database from poisoning the
+/// real user's registry: without it, pointing only the database elsewhere (via
+/// `NODESPACED_DB_PATH`) while inheriting the real home dir would seed the real
+/// `~/.nodespace/databases.toml` with a throwaway path (ADR-053). Falls back to
+/// the user's home directory.
+pub fn nodespace_home() -> Result<PathBuf> {
+    if let Ok(custom) = std::env::var("NODESPACE_HOME") {
+        return Ok(PathBuf::from(custom));
+    }
+    dirs::home_dir().context(
+        "Cannot determine the NodeSpace home directory: home directory is unknown and NODESPACE_HOME not set",
+    )
+}
+
+/// The `.nodespace/` state directory under [`nodespace_home`].
+pub fn nodespace_dir() -> Result<PathBuf> {
+    Ok(nodespace_home()?.join(".nodespace"))
+}
+
 /// Resolve the on-disk database path the daemon (and any in-process clients
 /// such as the CLI's `diagnostics` subcommand) should consult.
 ///
 /// Honors `NODESPACED_DB_PATH` if set so integration tests and alternate
-/// deployments can redirect storage without recompiling; otherwise defaults
-/// to `$HOME/.nodespace/database/nodespace.db`.
+/// deployments can redirect a single database file without recompiling;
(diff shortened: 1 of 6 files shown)


mstomar125:
**APPROVE** — correct, conservative fix for the data-loss bug; the subtle repair-vs-fixture logic is right.

### Root cause closed
The bug was the split: registry path hard-wired to `$HOME/.nodespace` while the served DB honored `NODESPACED_DB_PATH`, so a harness that redirected only the DB but inherited `HOME` wrote its `/var/folders` temp path into the *real* registry → every boot served a doomed temp DB that macOS later purges. The fix routes **both** the registry path and `resolve_db_path`'s default through `nodespace_home()`/`nodespace_dir()` (honoring `NODESPACE_HOME`), so one override isolates all state in lockstep — and the real-binary harnesses (`test-support SpawnedDaemon`, e2e `daemon-harness.ts`, the benchmark script) now set `NODESPACE_HOME` under their temp dir, so test runs never touch `~/.nodespace`. That directly removes the poison vector.

### The repair predicate is precise (the part most likely to be wrong, and it's right)
`default_is_doomed(registry_path, default_path)` = `!is_under_system_temp(registry_path) && is_under_system_temp(default_path)`:
- Real registry poisoned by a temp default → registry not-temp + default temp → **repair** (`repair_doomed_default` re-points to the standard path). ✅
- Intentionally-isolated test/dev registry (registry itself under temp) → `!is_under_system_temp(registry)` is false → **no-op**, fixture preserved. ✅ (Covered by `repair_doomed_default_is_noop_for_isolated_registry`.)
- `is_under_system_temp` checks `std::env::temp_dir()` plus the well-known roots `/var/folders`, `/private/var/folders`, `/tmp` — correctly handling the macOS `/var` ↔ `/private/var` symlink twin, so the check holds regardless of the daemon's `$TMPDIR`.

### Also right
- **Truthful logging** — boot now logs `served_db_path` ("serving default database") = the registry-resolved path actually opened, and renames the `Starting` field to `requested_db_path`; fixes the "log showed a different path than served" half of the finding.
- Tests: `default_is_doomed` truth table + repair no-op + `set_standard_default`; daemon lib 67 green; full pre-push gate green locally (note: server-side CI is gone since #1618, so this rests on the local gate).
- Conservative + community-safe: on a normal install nothing is under temp → `repair` is a no-op; it only acts on a clearly-poisoned real registry.

Closes #1648. Approve.

PR #1656, opened 2026-07-14
Wire the dead Search sidebar item to a working search view (closes #1646)

mstomar125:
## Problem

The **Search** item in the left sidebar highlighted but did nothing — `handleNavItemClick` only branched on `daily-journal`, so Search (and Favorites) fell through to a cosmetic active-flag update with no action. Search was also the only path to a collection's page (which hosts the **Collaboration** tab), so this left the admin UI unreachable from the sidebar.

## Fix

Search is not a node, so it opens a singleton non-node tab the same way Settings does, rather than the daily-journal node-tab path:

- Add a `'search'` tab type; render it as a new `SearchPane` in `pane-content`.
- `SearchPane` queries the existing (previously unused) `search_roots` Tauri command and opens a node tab when a result is clicked, reusing an already-open tab. Debounced, with out-of-order response guarding, loading/empty/error states.
- Wire the Search nav item to open/focus that tab; allow the `'search'` type through tab persistence so it survives reload.

**Favorites** has no store or view yet and is a much larger feature; left out of scope for this bug.

No backend changes — `search_roots` was already registered and returns `Vec<Node>`.

## Tests
- `search-pane.test.ts` (new): renders, queries `search_roots`, lists results, opens a node tab on click, empty state.
- `navigation.test.ts`: search-tab singleton open/reuse, title, and persistence-validation survival.
- Frontend lint + svelte-check clean; full pre-push gate (`test:all` + build + `test:e2e`) green locally.

Closes #1646.

### DIFF
--- a/packages/desktop-app/src/lib/components/layout/navigation-sidebar.svelte
+++ b/packages/desktop-app/src/lib/components/layout/navigation-sidebar.svelte
@@ -216,6 +216,31 @@
     }
   }
 
+  /**
+   * Open (or focus) the Search view. Search is not a node, so it opens a
+   * singleton non-node tab — mirroring how Settings is opened from the menu —
+   * rather than the daily-journal node-tab path.
+   */
+  function openSearchTab() {
+    const currentState = navigationStore.state;
+    const existingTab = currentState.tabs.find((tab) => tab.type === 'search');
+
+    if (existingTab) {
+      setActiveTab(existingTab.id, existingTab.paneId);
+    } else {
+      addTab(
+        {
+          id: 'search',
+          title: 'Search',
+          type: 'search',
+          closeable: true,
+          paneId: getTargetPaneId()
+        },
+        true
+      );
+    }
+  }
+
   // Handle navigation item clicks
   function handleNavItemClick(itemId: string) {
     // Close sub-panels when clicking non-collection nav items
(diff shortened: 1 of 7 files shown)


mstomar125:
**APPROVE** — clean, correctly-guarded fix; restores the sidebar Search route (and with it the only path to a collection page / Collaboration tab).

### Verified
- **Out-of-order response guard is the correct pattern:** `const token = ++requestToken` captured before the `await invoke('search_roots')`, then `if (token !== requestToken) return` after it (and again before clearing `loading`), so a slow earlier query can never clobber a newer query's results. Debounced (`DEBOUNCE_MS`), Enter bypasses the debounce, `onDestroy` clears the timer. No stale-result or leaked-timer bug.
- **Right tab model:** Search isn't a node, so it opens a singleton non-node `'search'` tab (same shape as Settings) rather than the daily-journal node-tab path — open/focus reuses an already-open tab, and the `'search'` type is allowed through tab persistence so it survives reload. Result click opens a node tab.
- **No backend change** — `search_roots` was already a registered Tauri command returning `Vec<Node>`; this just wires the previously-dead UI to it. Additive, community-safe (Search is a core feature, not Pro-gated).
- Tests: `search-pane.test.ts` (render/query/list/open-on-click/empty) + `navigation.test.ts` (singleton open/reuse/persistence). Lint + svelte-check clean; pre-push gate green locally.

### Note (scope)
This unblocks the *route* to a collection page (which hosts the Collaboration/admin tab), but **#1645** (admin UI pinned at `enable-prompt`) has a separate root cause — nothing sets `DatabaseSettingsNode.sync_enabled`, so the Pro-UI variant machine still won't advance past `enable-prompt` even now that Search is reachable. So #1656 is necessary-but-not-sufficient for the admin UI; #1645 still needs its own `sync_enabled`-setter fix. Correctly scoped here (and Favorites explicitly deferred).

Closes #1646. Approve.

Issue #297, opened 2025-10-20
Implement nodespace:// Link Navigation with Dynamic Tab Creation


malibio:
## Problem Statement

Users can create `nodespace://{node-uuid}` links via the @mention system, but clicking these links currently does nothing. We need to implement a navigation handler that:

1. Intercepts clicks on `nodespace://` links
2. Resolves the node UUID to determine the appropriate viewer type
3. **Regular Click**: Navigate within current tab (browser-like behavior) ✨
4. **Cmd+Click**: Create new tab or switch to existing tab for side-by-side viewing
5. Supports Cmd+Click for opening in a new tab vs. regular click for current tab

**📝 Implementation Note**: The final implementation uses **in-tab navigation** for regular clicks instead of tab switching. This provides a more intuitive browser-like experience where clicking links navigates within the current context, while Cmd+Click allows opening in new tabs for comparison.

This is the **foundation** for the entire tab management system - without link navigation, users cannot leverage the multi-tab architecture.

## Current State

**What Already Exists:**
- ✅ `nodespace://{uuid}` links created by @mention system
- ✅ Basic tab system (`tab-system.svelte`) with keyboard navigation
- ✅ Tab store (`navigation.ts`) with `addTab()`, `setActiveTab()`, `closeTab()`
- ✅ Multiple viewer types: DateNodeViewer, TaskNodeViewer, TextNodeViewer, AIChatNodeViewer (all verified to exist)
- ✅ SharedNodeStore with **synchronous** `getNode(nodeId)` API
- ✅ Current routing in `/packages/desktop-app/src/routes/+page.svelte`

**What's Missing:**
- ❌ Click handler for `nodespace://` links
- ❌ Service to resolve node UUID → node type → appropriate viewer
- ❌ Logic to create/switch tabs dynamically
- ❌ Cmd+Click support for new tab behavior

[Rest of issue content remains the same...]

PR #1657, opened 2026-07-14
Add the enable-sync intent + first-Pro consent modal (part of #1645, #1654)

mstomar125:
## Problem

The registry-driven Pro UI gates its collaboration surfaces on the active database's `DatabaseSettingsNode` `sync_enabled` / `auth_status`, but **no production code ever wrote those fields** (#1645), so the admin/collaboration UI was permanently locked at the enable prompt. And there was no consent gate before local data could reach the shared public workspace (#1654).

This is the **core half** of a cross-repo feature. The Pro daemon `EnableSync` handler and the `auth_status → connected` mirror land in a follow-up nodespace-sync PR that pins the core rev this adds; the two `closes` when both land + the Monday two-user isolation check passes.

## What this adds

**Backend (core)**
- `NodeService::set_sync_enabled` / `set_auth_status` — merge single fields into the `database-settings` singleton, preserving siblings (shared merge helper). These are the authoritative writers the Pro UI's variant machine needs.

**Bridge (desktop-app src-tauri)**
- New `nodespace.pro.v1` `EnableSync` RPC (client copy) + `pro_enable_sync` Tauri command, modelled on `pro_signout` (no-op in community mode).

**Frontend consent flow (Pro-gated; community build unchanged)**
- First-Pro **consent modal**: *merge into the shared public workspace* — with a required, checkbox-gated **irreversible** warning, and the confirm button is never the default focus — vs *keep this database local-only*. Merge records consent via `pro_enable_sync` (flips `sync_enabled`), then starts sign-in; **nothing leaves the device before the opt-in**. Declining shares nothing.
- The enable-sync pill and the previously dead-end locked-Collaboration placeholder now open the modal; registered as the `enable-prompt` app-shell modal contribution.

## Consent sequencing

Consent-first: the modal is opened from the enable-sync affordance and the merge choice happens *before* sign-in / any cloud contact — strictly safer than prompting after sign-in, and it needs no "first-signed-in" detection. Re-clicking the pill re-opens it, so there's no auto-prompt to suppress.

## Tests
- NodeService roundtrip: `sync_enabled` / `auth_status` merge preserves the tenant binding.
- Modal gating + irreversible warning; slot merge → `pro_enable_sync` → sign-in; pill opens the modal; updated ui-extensions contribution counts.
- Full frontend suite (4037) + `clippy --all-targets -D warnings` + pre-push gate (`test:all` + build + `test:e2e`) green locally.

Part of #1645, #1654.

### DIFF
--- a/packages/core/src/services/node_service/mod.rs
+++ b/packages/core/src/services/node_service/mod.rs
@@ -1095,6 +1095,76 @@ impl NodeService {
         Ok(())
     }
 
+    /// Merge fields into the DatabaseSettingsNode singleton's `database-settings`
+    /// namespace, preserving every sibling field (tenant binding, sync state,
+    /// auth status). The singleton is seeded on database open, so a missing one
+    /// is an error rather than a silent no-op.
+    async fn merge_database_settings(
+        &self,
+        fields: &[(&str, serde_json::Value)],
+    ) -> Result<(), NodeServiceError> {
+        let node = self
+            .query_nodes_by_type("database-settings", None)
+            .await?
+            .into_iter()
+            .next()
+            .ok_or_else(|| {
+                NodeServiceError::InitializationError(
+                    "cannot update database settings: DatabaseSettingsNode singleton not found"
+                        .to_string(),
+                )
+            })?;
+
+        let mut properties = node.properties.clone();
+        let root = properties.as_object_mut().ok_or_else(|| {
+            NodeServiceError::InitializationError(
+                "DatabaseSettingsNode properties are not a JSON object".to_string(),
+            )
+        })?;
+        let settings = root
+            .entry("database-settings")
+            .or_insert_with(|| serde_json::json!({}))
+            .as_object_mut()
+            .ok_or_else(|| {
+                NodeServiceError::InitializationError(
+                    "DatabaseSettingsNode `database-settings` is not a JSON object".to_string(),
+                )
+            })?;
+        for (key, value) in fields {
+            settings.insert((*key).to_string(), value.clone());
+        }
+
+        self.update_node(
+            &node.id,
+            node.version,
+            NodeUpdate::new().with_properties(properties),
+        )
+        .await?;
+        Ok(())
+    }
+
+    /// Enable (or disable) per-database cloud sync by writing `sync_enabled` on
+    /// the DatabaseSettingsNode singleton (ADR-053). This is the field the
+    /// registry-driven Pro UI gates its collaboration surfaces on; nothing else
+    /// advances it, so this is the authoritative writer the first-Pro consent
+    /// flow calls once the user opts in.
+    pub async fn set_sync_enabled(&self, enabled: bool) -> Result<(), NodeServiceError> {
+        self.merge_database_settings(&[("sync_enabled", serde_json::Value::Bool(enabled))])
+            .await
+    }
+
+    /// Set the per-database cloud auth status (`local` or `connected`) on the
+    /// DatabaseSettingsNode singleton (ADR-053). The Pro daemon advances this to
+    /// `connected` once identity is bound and back to `local` on sign-out, which
+    /// drives the Pro UI from the sign-in variant to the connected variant.
+    pub async fn set_auth_status(&self, status: &str) -> Result<(), NodeServiceError> {
+        self.merge_database_settings(&[(
+            "auth_status",
+            serde_json::Value::String(status.to_string()),
+        )])
+        .await
+    }
+
     /// Seed core schema definitions if database is fresh
     ///
     /// Checks if schema nodes exist. If not, creates all core schemas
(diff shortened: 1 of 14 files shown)


mstomar125:
**REQUEST CHANGES**

Reviewed `07760d76` against the code. The mechanics are solid — my only blocker is the consent copy making a security promise that is **false under the live `tenant_public` model**, on the exact screen where the user gives irreversible consent.

**Blocker — the consent copy understates the risk (`first-pro-consent-modal.svelte`):**

> "…the shared **public workspace**, where other people using NodeSpace can **read** them. **Editing and deleting stay yours alone.**"

That last sentence is not true against the current `tenant_public` RLS. As surfaced in **sync#297** (CRITICAL, still open), `tenant_public` admits every authenticated user to a *shared* collection with open write — so other users can not only read but **edit and delete** your nodes. The modal is the informed-consent gate for an action it correctly labels irreversible; promising edit/delete exclusivity that the backend does not enforce makes the consent misinformed in exactly the dimension that matters (integrity/availability, not just confidentiality).

Note the copy describes a *third* model that matches neither reality: current state is public-read **+ public-write** (the sync#297 bug); the likely post-fix state (per-install restricted collection) is **no** third-party read at all. "Public read, owner-only write" is neither — so whichever way sync#297 resolves, this sentence needs to change.

Two acceptable resolutions (author/human's call):
1. **Fix the copy** to state that, until per-install write-restriction lands, other users can also **edit or delete** shared notes — i.e. don't promise "editing and deleting stay yours alone."
2. **Gate activation** so this consent flow cannot reach a live sign-in until sync#297's owner-only-write fix ships, at which point option-1's promise becomes true and the copy is accurate as written.

**Cross-layer deploy-order flag:** this is the core half — `pro_enable_sync` calls `EnableSync` on the `ProClient`, whose daemon handler + `auth_status→connected` mirror land in a follow-up nodespace-sync PR (per the PR body). That follow-up is what actually pushes the graph into `tenant_public`. So the ordering constraint is: **the sync-side `EnableSync` activation must not ship to external users before sync#297's restricted-collection fix** — otherwise consenting users' graphs land world-writable in `tenant_public/c000…0001`. This is consistent with sync#297's fix-direction #2 ("don't ship default-public sign-in outside the team until per-install collections are restricted").

**What's correct (no changes needed there):**
- `merge_database_settings` (`node_service/mod.rs`) reads the `database-settings` singleton fresh, clones properties, does a sibling-preserving insert into the sub-object, and writes with `update_node(id, node.version, …)` — proper OCC, same shape as `set_bound_tenant` (#1606). `set_sync_enabled`/`set_auth_status` are the authoritative writers #1645 was missing (nothing wrote `sync_enabled`, which pinned the UI at `enable-prompt`); roundtrip test confirms one field flips without clobbering the sibling. ✓
- Community-safety: `pro_enable_sync` returns `Ok(())` early with no `ProClient` → no-op in community, matching the other Pro commands. Community build unchanged. ✓
- The consent-first architecture itself is exactly right — checkbox-gated danger button, never default-focused, Escape/overlay = keep-local, nothing leaves the device before opt-in. The *only* fix needed is making the stated risk match what the backend actually allows.

Happy to flip to APPROVE once the copy tells the truth about third-party edit/delete (or activation is gated behind the sync#297 fix).

mstomar125:
**Follow-up to my REQUEST CHANGES above** — reframing now that **cloud#74** ("Owner-scope UPDATE/DELETE on open collections", closes cloud#73) has appeared and I've reviewed it (APPROVE).

cloud#74 makes the write/delete leg of the `tenant_public` leak owner-only: on an open collection only `created_by = auth.uid()` can UPDATE/DELETE a node (verified — its `has_restricted_ancestor` walk is character-identical to `user_can_write_node`, so the decomposition is exact). That is precisely what makes this modal's copy — *"Editing and deleting stay yours alone"* — **true by design**, with third-party **read** remaining the consented tradeoff the modal discloses.

So my blocker is no longer "the copy is false" — it's now a pure **deploy-order dependency**:

- The copy is accurate **iff cloud#74 is applied to the live `tenant_public` schema** before this consent flow can reach a real sign-in for external users.
- cloud#74's own fix-file notes the Management-API WAF/1010 gotcha, so applying it to `tenant_public` is a deliberate psql/SQL-editor step — it won't happen automatically on merge.

**Recommendation:** keep this PR gated behind that ordering — the core-half here + the follow-up sync `EnableSync` handler should not ship to external users until cloud#74 is **deployed to `tenant_public`** (not merely merged). Once that's confirmed, the copy is correct as written and I'll flip to APPROVE. If you'd rather decouple, the alternative is softening the copy to not promise edit/delete exclusivity until deployment is confirmed — but given cloud#74 exists and is correct, gating on its deployment is the cleaner path and keeps the honest copy.

Everything else in this PR remains good (the `merge_database_settings` OCC writers, community no-op, consent-first UX). This is purely about the three legs — cloud#74 (write/delete) + this modal (#1654 consent) + the #297 daemon gate — landing *and deploying* as a coherent set.

PR #1658, opened 2026-07-14
Re-hydrate the Pro signed-in state on webview reload (closes #1647)

mstomar125:
## Problem

On a webview reload a signed-in Pro user appears signed out. Two mechanisms conspire:

- `pro:tier-detected` is a one-shot event carrying the initial status — it does **not** re-fire on reload.
- `sync:status` only pushes on a *change*, so a steady signed-in session emits nothing after reload.

The daemon session is fully intact; only the frontend's view of it is lost until the next daemon transition (which may be minutes away, or never while idle).

## Fix

A pull-based snapshot to complement the two push channels:

- **`pro_client.rs`** — `ProClientInner` already caches `last_status` from the capability probe. Added `set_last_status()` so the `WatchSyncStatus` forwarding task keeps it current (the probe snapshot alone goes stale immediately).
- **`commands/pro_sync.rs`** — the stream-forwarding task now calls `pro_client.set_last_status(evt.clone())` per event; added a `pro_current_status` command returning a `SyncStatusSnapshot { state, detail, user_email }` (or `null` if the daemon hasn't emitted yet).
- **`pro-sync.svelte.ts`** — `start()` invokes `pro_current_status` after `pro_tier` and hydrates `state`/`detail`/`userEmail` from the snapshot. Deterministic: neither event needs to fire for the signed-in state to be restored.

Community mode is unchanged — no `ProClient` in managed state → the command returns `null` and the store stays signed-out.

## Tests

- `pro-sync.test.ts` — new `reload re-hydration (#1647)` describe: re-hydrates identity from the snapshot on `start()`; leaves the store signed-out when the daemon has no status yet (`null`).
- Full frontend suite green locally; `cargo build --bin nodespaced` + `test:e2e` pre-push gate passed.

### DIFF
--- a/packages/desktop-app/src-tauri/src/commands/pro_sync.rs
+++ b/packages/desktop-app/src-tauri/src/commands/pro_sync.rs
@@ -46,6 +46,32 @@ pub async fn pro_tier(app: AppHandle) -> Result<ProTier, String> {
     }
 }
 
+/// The daemon's most recent sync status (state, detail, signed-in email), cached
+/// from the `WatchSyncStatus` stream. Lets the frontend re-hydrate the signed-in
+/// state on demand — on a webview reload the one-shot `pro:tier-detected` event
+/// (which carries the initial status) does not re-fire, and `sync:status` only
+/// pushes on change, so without this a signed-in Pro user appears signed out.
+/// `None` in community mode or before any status is known.
+#[tauri::command]
+pub async fn pro_current_status(app: AppHandle) -> Result<Option<SyncStatusSnapshot>, String> {
+    let Some(pro) = app.try_state::<ProClient>() else {
+        return Ok(None);
+    };
+    Ok(pro.last_status().await.map(|s| SyncStatusSnapshot {
+        state: s.state,
+        detail: s.detail,
+        user_email: s.user_email,
+    }))
+}
+
+/// Serializable snapshot of `SyncStatusEvent` for [`pro_current_status`].
+#[derive(serde::Serialize)]
+pub struct SyncStatusSnapshot {
+    pub state: i32,
+    pub detail: String,
+    pub user_email: String,
+}
+
 /// Start a long-lived `WatchSyncStatus` subscription on the daemon
 /// and forward each event to the frontend as a Tauri event named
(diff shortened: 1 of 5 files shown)


mstomar125:
**APPROVE**

Reviewed `7b82abc6` against the code. Clean, correctly-scoped root-cause fix for the reload-logout (#1647). +106/-0, well tested.

**Root cause & fix are right.** On a webview reload the one-shot `pro:tier-detected` event (which carries the initial status) doesn't re-fire, and `sync:status` only pushes *on change* — so a signed-in Pro user renders signed-out until the next daemon transition. The fix caches each `WatchSyncStatus` event into `ProClient.last_status` via `set_last_status` (called from the forwarding task in `pro_subscribe_sync_status`) and adds `pro_current_status` so the store can re-hydrate deterministically on `start()`. This reflects the intact daemon session rather than re-establishing anything.

**Details verified:**
- `pro.inner().clone()` (pro_sync.rs) resolves to Tauri's `State::inner() -> &ProClient` → Arc-backed cheap clone, giving the spawned forwarding task an owned handle to keep the cache current. Idiomatic; the "Owned handle (Arc-backed)" comment is accurate. ✓
- `set_last_status` / `last_status` are short, non-nested write/read critical sections on the same `RwLock` — no lock-ordering or deadlock risk. ✓
- Community-safety: `pro_current_status` returns `Ok(None)` when there's no `ProClient` (community mode), and the store leaves state untouched on `null` — covered by the "leaves the snapshot alone … (null)" test. Community build unchanged. ✓
- No interaction with the open sync#297 leak: this is display-hydration only — it neither activates sync nor moves any data; it just re-reads the daemon's last-known status. Independent of #1657. ✓
- Tests cover the two branches that matter (re-hydrate signed-in identity from snapshot; no-op on null).

**One assumption worth a sentence (non-blocking):** the cache correctly reflects sign-*out* only if the daemon emits a `WatchSyncStatus` event on signout (so a reload after signout doesn't restore a stale `connected`). Since `sync:status` pushes on change and signout is a state change, `set_last_status` catches it — so this holds; flagging only so it's a conscious invariant.

No changes required.

PR #1659, opened 2026-07-14
Disable mermaid HTML labels so diagram labels survive SVG sanitization (closes #1650)

mstomar125:
## Problem

Every mermaid diagram in imported docs rendered as bare boxes with **no text** — the single biggest readability loss in the docs. DOM inspection showed 0 `<text>` and 0 `<foreignObject>` elements: the labels were gone, not hidden.

## Root cause

Mermaid's default `htmlLabels: true` emits node labels as HTML wrapped in `<foreignObject>`. `sanitizeSvg()` runs DOMPurify's `svg` profile, which **drops `<foreignObject>`** (it can host arbitrary HTML) — so every label is stripped. The existing sanitizer tests never asserted that label text survives, which is how this shipped.

## Fix

Initialize mermaid with `flowchart: { htmlLabels: false }` and `class: { htmlLabels: false }` so labels are emitted as native SVG `<text>`, which the sanitizer preserves. **The sanitizer is unchanged** — no relaxation of the XSS defense. sequence/state/er diagrams already emit `<text>` natively and need no flag.

## Tests

- `sanitizeSvg` strips `<foreignObject>` HTML labels but preserves native `<text>` labels (documents the root cause).
- `renderMermaid` initializes with `htmlLabels` disabled on flowchart + class.
- Full frontend suite + `cargo build --bin nodespaced` + `test:e2e` pre-push gate green locally.

### DIFF
--- a/packages/desktop-app/src/lib/services/mermaid-render.ts
+++ b/packages/desktop-app/src/lib/services/mermaid-render.ts
@@ -106,7 +106,9 @@ function getDOMPurify() {
   // executing in any modern webview — revisit if that CSP is ever relaxed.
   purify.addHook('uponSanitizeAttribute', (_node, data) => {
     if (data.attrName === 'style') {
-      data.attrValue = data.attrValue.replace(/(?:javascript|vbscript):/gi, '').replace(/expression\s*\(/gi, '');
+      data.attrValue = data.attrValue
+        .replace(/(?:javascript|vbscript):/gi, '')
+        .replace(/expression\s*\(/gi, '');
     }
   });
   return purify;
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Reviewed `3a42b59b` against the code. Correct root-cause fix for the wordless-diagram bug (#1650), and it closes the test gap that let it ship.

- **Root cause & fix are right.** Mermaid's default `htmlLabels:true` wraps node labels in `<foreignObject>` (arbitrary HTML), which `sanitizeSvg()`'s DOMPurify `svg` profile drops — so labels vanish. Setting `flowchart:{htmlLabels:false}` and `class:{htmlLabels:false}` makes mermaid emit native SVG `<text>`, which survives sanitization. The comment correctly scopes this to flowchart/class (the diagram types that use foreignObject labels); sequence/state/er emit `<text>` natively and need no flag.
- **No security regression.** `securityLevel:'strict'` is retained, and `htmlLabels:false` *reduces* the HTML surface rather than expanding it — labels are now plain SVG text.
- **Tests now assert the thing that broke.** The new `strips foreignObject … but preserves native <text> labels` case pins the sanitizer behavior, and `initializes mermaid with HTML labels disabled` asserts the init config carries `flowchart.htmlLabels===false` / `class.htmlLabels===false`. That's exactly the assertion whose absence let this regress (per the PR body). The theme-toggle across two renders to guarantee an `initialize()` call is a nice touch given the module-level theme cache.
- **Universal, no Pro coupling** — a rendering-quality fix in the shared desktop-app; community users get readable diagrams too, no behavioral regression (blank → labelled).

The `data.attrValue` change is a cosmetic multi-line reflow, unrelated to the fix — harmless. No changes required.

PR #1660, opened 2026-07-14
Surface bulk-import phase-2 failures instead of reporting false success (closes #1649)

mstomar125:
## Problem

Bulk dir import runs Phase 2 (collection resolution, node insert, membership + mention writes) in a fire-and-forget background task. Every sub-step failure was logged with `tracing::error` and swallowed, and the terminal event always reported `complete`. So a run could leave every node with **zero `member_of` edges** — searchable but invisible in the Collections UI (exactly the state the team DB is in) — while the CLI exited 0. A panic in the task ended the stream after the last progress message, also indistinguishable from success.

## Fix

- Collect the first phase-2 failure (collection resolution / bulk insert / collection assignment / mention creation) into a single error and **fold it into every per-file result**, so the terminal event reports the affected files as failed with the reason instead of a false `complete`.
- **Guard the background task**: an outer task awaits the phase-2 `JoinHandle` and, on a panic/cancellation (`JoinError`), closes the stream with a `Status::internal` error — so an abrupt death is never silent.

The existing `bulk_create_hierarchy_trusted` failure handling is unified into this single path (it previously marked results failed inline).

## Test

`apply_phase2_error` marks all file results failed with the reason on error, and leaves them untouched (successful) when there is no error. Full daemon lib tests + clippy green locally; `cargo build --bin nodespaced` + `test:e2e` pre-push gate passed.

## Note

This surfaces the failure; deterministic idempotent **re-import** (so a retry after a surfaced failure doesn't duplicate) is the separate #1651 work.

### DIFF
--- a/packages/daemon/src/services/import_service.rs
+++ b/packages/daemon/src/services/import_service.rs
@@ -481,8 +481,15 @@ async fn run_batch_import(
     let store = Arc::clone(node_service.store());
     let node_service_clone = (*node_service).clone();
     let tx_bg = tx.clone();
+    let tx_guard = tx.clone();
 
-    tokio::spawn(async move {
+    let phase2 = tokio::spawn(async move {
+        // First failure seen in phase 2. When set, the import did NOT fully
+        // succeed (nodes may exist but be unlinked from collections, or mentions
+        // may be missing), so every file result is marked failed rather than
+        // reporting a false "complete".
+        let mut phase2_error: Option<String> = None;
+
         send_progress(
             &tx_bg,
             5,
@@ -502,6 +509,7 @@ async fn run_batch_import(
             Ok(map) => map,
             Err(e) => {
                 tracing::error!("Failed to bulk resolve collections: {:?}", e);
+                phase2_error.get_or_insert_with(|| format!("Collection resolution failed: {e}"));
                 HashMap::new()
             }
         };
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE**

Reviewed `e06a7933` against the code. This fixes a real data-integrity-*reporting* bug — the exact "nodes land with zero `member_of` edges while the CLI reports success" state (#1649) — and does it carefully.

- **The disease is correctly identified and fully covered.** Previously only a bulk-insert failure marked results failed; collection-resolution, membership (`bulk_add_to_collections`), and mention (`bulk_create_mentions`) failures were `tracing::error`-swallowed while the terminal event still said `complete`. The new `phase2_error: Option<String>` accumulates the **first** failure across *all four* phase-2 steps (`get_or_insert_with` = first-failure-wins, most-upstream cause surfaced), and `apply_phase2_error` folds it into every per-file result → no false `complete`. The membership/mention paths were the actual culprit for the unlinked-nodes state, so widening coverage to them is the meaningful part.
- **The panic/cancel hole is closed.** Wrapping the phase-2 body in a `JoinHandle` and awaiting it in a guard task that sends `Err(Status::internal(…))` on `JoinError` means an aborted task can no longer end the stream after the last progress message and read as success. And since `send_progress(9,"complete",…)` is the final statement of the phase-2 closure, there's no panic-after-complete window → no double-terminal race (guard only fires when `phase2.await` is `Err`, i.e. the terminal was never sent).
- **Control flow stays correct on the continue-anyway paths.** Collection-resolution failure sets the error and proceeds with an empty map; downstream membership assignment is then empty/skipped, mentions still attempted, and `apply_phase2_error` marks everything failed at the end — best-effort work, accurate final reporting. `failed_results` (phase-1 failures) are already `success:false`, so folding only into `phase1_results` is right.
- **Tested.** `phase2_error_marks_all_results_failed` covers both branches (None → untouched; Some → all failed with the message). Good.
- **Community-safe** — core daemon import path, present in both builds; this is strictly-more-correct reporting with no behavioral regression for free users.

One non-blocking nit: message consistency — collection-resolution/assignment/mention errors interpolate the raw `{e}` into the user-facing message, while bulk-insert uses a generic "see daemon logs". Fine for a local single-user daemon, but you may want the same raw-vs-generic choice across all four for predictable CLI output. Not a blocker.

No changes required.

PR #1661, opened 2026-07-14
Degrade the database store gracefully under dev:browser (closes #1652)

mstomar125:
## Problem

In `dev:browser`, `DatabaseStore.load()` called the Tauri `list_databases` invoke unconditionally. The Tauri bridge is absent there and the dev-proxy exposes no DatabaseService, so it threw `Cannot read properties of undefined (reading 'invoke')` at boot and the switcher rendered its failure fallback — blocking multi-database debugging in the officially supported dev loop (and the WKWebView debug target).

## Fix

Gate `load()` on the Tauri bridge (`__TAURI__` / `__TAURI_INTERNALS__` in window — the same probe the model, agent, and external-link surfaces already use). When it's absent, present a single implicit **Default** database instead of erroring: the dev-proxy binds to one daemon/database, so one concrete, non-switchable entry is accurate. The Tauri app, where the bridge is present, is unchanged.

## Tests

`load()` with no bridge yields one implicit database, no error, and no `list_databases` invoke. Existing `load` tests set the bridge present so they still exercise the invoke path. Full frontend suite + pre-push gate green locally.

### DIFF
--- a/packages/desktop-app/src/lib/stores/database.svelte.ts
+++ b/packages/desktop-app/src/lib/stores/database.svelte.ts
@@ -37,6 +37,34 @@ interface DatabaseListing {
 }
 
 /**
+ * Whether the Tauri IPC bridge is present. Absent under `dev:browser`, where the
+ * app runs in a plain browser against the dev-proxy — which forwards NodeService
+ * but has no DatabaseService, so calling the `list_databases` invoke throws
+ * `Cannot read properties of undefined (reading 'invoke')`. Same probe the model,
+ * agent, and external-link surfaces use.
+ */
+function isTauriBridgePresent(): boolean {
+  return (
+    typeof window !== 'undefined' && ('__TAURI__' in window || '__TAURI_INTERNALS__' in window)
+  );
+}
+
+/**
+ * The single implicit database presented in browser dev mode: the dev-proxy is
+ * bound to one daemon/database and exposes no registry, so the switcher shows one
+ * concrete, non-switchable entry instead of erroring at boot.
+ */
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Reviewed `2b5f17a3` against the code. Clean, minimal fix for the `dev:browser` boot crash (#1652).

- **Right root cause & guard.** `DatabaseStore.load()` invoked `list_databases` unconditionally; under `dev:browser` there's no Tauri bridge (the dev-proxy forwards NodeService but exposes no DatabaseService), so the invoke threw `Cannot read properties of undefined (reading 'invoke')` at boot and the switcher showed its failure fallback. Gating on `isTauriBridgePresent()` (`__TAURI__` / `__TAURI_INTERNALS__` in `window`) — the same probe the model/agent/external-link surfaces already use — and presenting a single implicit **Default** database is the correct degrade. It matches reality: the dev-proxy is bound to one daemon/database with no registry.
- **Tauri app is provably unaffected.** In the shipped webview `__TAURI_INTERNALS__` is injected → guard is true → the original `list_databases` path runs unchanged. The `typeof window !== 'undefined'` prefix also makes it SSR-safe.
- **Tests are right.** The new browser-mode describe deletes `__TAURI_INTERNALS__` and asserts `list_databases` is never invoked, no error, a single implicit db, and a concrete active selection; the existing suite now sets `__TAURI_INTERNALS__ = {}` in `beforeEach` so it keeps exercising the invoke path. Good separation.
- **Community-safe** — this only changes a dev-loop surface; shipped behavior is identical.

The remaining hunks (one-lined `create`/`mutate`, the `activeDatabaseId` assignment) are cosmetic prettier reflow. No changes required.

PR #1662, opened 2026-07-14
Add sync_enabled + auth_status readers to the DatabaseSettingsNode accessors

mstomar125:
## What

The DatabaseSettingsNode singleton (ADR-053) already exposed `set_sync_enabled` and `set_auth_status` writers but had no matching readers, so any consumer wanting to gate on those fields had to `query_nodes_by_type("database-settings", …)` and reach into the properties JSON by hand.

Adds two readers that mirror the existing `get_bound_tenant` read pattern and its unbound-default semantics:

- `get_sync_enabled() -> bool` — returns `false` when the field is absent or the singleton has not been seeded yet.
- `get_auth_status() -> String` — returns `"local"` when the field is absent or the singleton has not been seeded yet.

## Why

`get_sync_enabled` is the authoritative gate the Pro daemon reads before pushing local changes to the cloud, so an un-opted-in database never leaves the device. `get_auth_status` completes the accessor pair alongside `set_auth_status`. Both are consumed by the Pro daemon in `nodespace-sync` (EnableSync handler test + the push opt-in gate).

## Tests

Extends `test_set_sync_enabled_and_auth_status_roundtrip` to exercise both getters through their fresh-install defaults and after each write. `cargo test --lib -p nodespace-core` — 48 node_service tests pass, 0 failures. Pre-push gate (test:all + build + e2e) green.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/core/src/services/node_service/mod.rs
+++ b/packages/core/src/services/node_service/mod.rs
@@ -1165,6 +1165,53 @@ impl NodeService {
         .await
     }
 
+    /// Read whether per-database cloud sync is enabled, from the DatabaseSettingsNode
+    /// singleton (ADR-053). Returns `false` when the field is absent or the singleton
+    /// has not been seeded yet — a fresh install defaults to sync disabled until the
+    /// first-Pro consent flow opts in. This is the authoritative gate the Pro daemon
+    /// reads before pushing local changes to the cloud, so an un-opted-in database
+    /// never leaves the device.
+    pub async fn get_sync_enabled(&self) -> Result<bool, NodeServiceError> {
+        let Some(node) = self
+            .query_nodes_by_type("database-settings", None)
+            .await?
+            .into_iter()
+            .next()
+        else {
+            return Ok(false);
+        };
+        Ok(node
+            .properties
+            .get("database-settings")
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE**

Adversarial re-review (author=reviewer, per the core re-review policy):

- `get_sync_enabled` / `get_auth_status` mirror `get_bound_tenant` exactly — same `query_nodes_by_type(...).into_iter().next()` singleton read and the same unseeded/absent → default branch. The singleton uniqueness these rely on is covered by `test_create_second_database_settings_is_noop`.
- Defaults are safe under malformed data: `.as_bool().unwrap_or(false)` and `.as_str().unwrap_or("local")` — a wrong-typed or missing field degrades to the seed default rather than panicking.
- `false`/`"local"` defaults match `seed_database_settings_if_needed`, so a seeded fresh install and an (defensive) unseeded read agree.
- Pure additive readers, no new deps, no behavior change to existing writers. Roundtrip test extended to assert both getters through defaults and after each write; `cargo test --lib -p nodespace-core` green (48/48), pre-push gate green.

No blockers.

PR #1663, opened 2026-07-14
Idempotent markdown re-import with --replace, plus stripped collection member titles (S3/S4)

mstomar125:
## Summary

Completes the remaining work on the readable-docs import pipeline: **S3 idempotent re-import** and **S4 title polish**. With S1 (phase-2 failure surfacing) and the S2 mermaid-label fix already merged, this makes refreshing the docs corpus a safe, repeatable, one-command operation.

## S3 — Idempotent re-import

Each imported document's root node id is now **deterministic** — a UUIDv5 of the file's base-directory-relative path — so a re-import addresses the *same* root instead of minting a new one. Phase 2 detects roots that already exist:

- **Plain re-import** leaves an already-imported document untouched (skipped) — no duplicate documents, no orphan "ghost" roots.
- **`--replace`** drops the stale child subtree and **keeps the root node** (so inbound `[[links]]`/mentions to the doc root survive), recreates children from the fresh parse, and re-marks the root's embedding stale.
- Collection membership is re-asserted idempotently on every run, so a doc that lost its `member_of` edge in a prior partial import self-heals.

The `replace` option is threaded through the proto, the CLI (`nodespace import dir --replace`), the Tauri import command (so the desktop/Pro path gets it too), and both the batch and single-file daemon paths. The `import dir` summary now reports the honest daemon result (new / refreshed / already-present) instead of always claiming to have re-imported everything.

## S4 — Title polish

Collection member lists rendered imported header roots as raw markdown (`# ACP Integration Architecture`) because those nodes carry the heading in `content` with no separate `title`. The member list now falls back to the content with its markdown stripped. (Tab titles already strip the marker via `formatTabTitle`.)

## S2 — Render smoke

The mermaid-label fix is already merged and is directly guarded by hermetic unit tests (`mermaid-render.test.ts`: `<foreignObject>` stripped, native `<text>`/label text preserved, `htmlLabels:false` on flowchart + class). A corpus-walking E2E test was intentionally **not** added — it would depend on a non-hermetic sibling docs checkout and duplicate coverage the unit tests already lock.

## Testing

- **New Rust tests** (`import_service.rs`): deterministic id is stable & path-sensitive; single-file re-import is a no-op and `--replace` refreshes in place; batch re-import never duplicates and `--replace` keeps every root id + collection membership.
- **New frontend test** (`collections.test.ts`): an untitled imported doc root renders with the heading marker stripped.
- **Live smoke on the real `nodespaced` + `nodespace` binaries** against the actual architecture docs (29 files): fresh import → 29 roots / 2976 nodes; plain re-import → **exactly** 29 / 2976 (zero duplication — the original defect created ~2805 ghost nodes); `--replace` → 29 / 2976 with membership intact.
- Local: frontend `4050/4050`, `cargo test` for core + daemon + cli green, full pre-push gate green.

Closes #1651.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6043,6 +6043,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "sha1_smol"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbfa15b3dddfee50a0fff136974b3e1bde555604ba463834a7eb7deb6417705d"
+
+[[package]]
 name = "sha2"
 version = "0.10.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -7629,6 +7635,7 @@ dependencies = [
  "getrandom 0.4.2",
  "js-sys",
  "serde_core",
+ "sha1_smol",
  "wasm-bindgen",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ serde_json = "1.0"
 tokio = { version = "1.0", features = ["full"] }
 anyhow = "1.0"
 chrono = { version = "0.4", features = ["serde"] }
-uuid = { version = "1.0", features = ["v4", "serde"] }
+uuid = { version = "1.0", features = ["v4", "v5", "serde"] }
 thiserror = "1.0"
 async-trait = "0.1"
 libsql = "0.6"
--- a/packages/cli/src/commands/import.rs
+++ b/packages/cli/src/commands/import.rs
@@ -34,6 +34,12 @@ pub struct ImportFileArgs {
     /// Route files to collections based on directory structure.
     #[arg(long)]
     pub auto_collection_routing: bool,
+
+    /// Refresh an already-imported document in place: replace its child subtree
+    /// from the fresh parse, keeping the root node so inbound links survive.
+    /// Without this, an already-imported document is left untouched.
+    #[arg(long)]
+    pub replace: bool,
 }
 
 #[derive(Args, Debug)]
@@ -56,6 +62,13 @@ pub struct ImportDirArgs {
     /// Directory names to exclude (repeatable, e.g. --exclude node_modules).
     #[arg(long = "exclude")]
     pub exclude_patterns: Vec<String>,
+
+    /// Refresh already-imported documents in place: replace each existing
+    /// document's child subtree from the fresh parse, keeping its root node so
+    /// inbound links survive. Without this, already-imported documents are
+    /// skipped (a plain re-import never duplicates).
+    #[arg(long)]
+    pub replace: bool,
 }
 
 pub async fn run(client: &mut ImportClient, action: ImportAction, json: bool) -> Result<()> {
@@ -91,6 +104,7 @@ async fn run_file(client: &mut ImportClient, args: ImportFileArgs, json: bool) -
         auto_collection_routing: args.auto_collection_routing,
         exclude_patterns: vec![],
         base_directory: String::new(),
+        replace: args.replace,
     };
 
     let mut stream = client
@@ -152,6 +166,7 @@ async fn run_dir(client: &mut ImportClient, args: ImportDirArgs, json: bool) -> 
         auto_collection_routing: args.auto_collection_routing,
         exclude_patterns: args.exclude_patterns,
         base_directory: args.directory,
+        replace: args.replace,
     };
 
     let mut stream = client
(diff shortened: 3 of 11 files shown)


mstomar125:
**APPROVE**

Reviewed `c215377f` against the code. Completes the readable-docs import pipeline (S3 idempotent re-import + S4 title polish, #1651). Correct mechanics, well-tested; my notes are all non-blocking.

**Core mechanic is sound.**
- Deterministic root id = `UUIDv5(IMPORT_ROOT_NAMESPACE, base-dir-relative-path)`, replacing the old `Uuid::new_v4()`. Fixed namespace with the right warning ("MUST stay fixed"). Re-import addresses the same root → idempotent. ✓
- `--replace` path: `delete_children_subtree_unchecked(root_id)` — I checked its impl (`sqlite_store/nodes.rs:630`): a `WITH RECURSIVE` CTE over `has_child` that gathers the **entire** descendant subtree (grandchildren included) and deletes them, so no orphans — then keeps + content-updates the root (inbound links/mentions to the doc root survive) and recreates children, and re-marks the root's embedding stale via `create_stale_embedding_markers_bulk`. Correct. ✓
- Skip path (plain re-import of an existing doc): creates nothing, but **re-asserts collection membership idempotently** — which self-heals the exact "#1648 zero `member_of` edges" state on a corpus that had a prior partial import. Nice. ✓
- S4: `node.title || stripMarkdown(node.content)` so an imported header root shows "ACP Integration Architecture", not "# ACP …". ✓

**Tests cover the behaviors that matter:** `deterministic_root_id_is_stable_and_path_sensitive` (same path→same id, different→different, valid UUID), `reimport_single_file_is_idempotent_and_replace_refreshes` (same root id on re-import, replace succeeds), `batch_reimport_does_not_duplicate_and_replace_keeps_roots`, and the collections-store strip test. ✓

**Community-safe:** `--replace` is additive (defaults false); the only behavior change to a plain community `import dir` is that re-import is now idempotent (skips instead of duplicating) — that's the fix, and free users benefit. ✓

Four non-blocking notes:
1. **Child-level inbound mentions don't survive `--replace`.** Only the root id is preserved; children are recreated with fresh ids, so any `[[link]]` targeting a *child* block/heading of a replaced doc dangles. Fine if links are only ever expected to target the doc root (which the PR body states) — just confirming that's the intended scope.
2. **Post-upgrade wrinkle:** docs imported before this change carry random v4 root ids, which the new deterministic scheme won't recognize as "existing" — so the *first* post-upgrade re-import creates a parallel deterministic-id copy rather than skipping/replacing the old one. One-time ghost-duplicate, not data loss; worth knowing when refreshing an existing corpus (a clean re-import into a fresh DB sidesteps it).
3. **Replace isn't atomic** — subtree-delete → root-update → children-insert are separate store calls; a mid-failure can leave a root with no fresh children. Not silent, though: it feeds the phase-2 error path (#1660), so the file is reported failed. Acceptable given that.
4. **Single-file vs batch key differ slightly** — single-file uses `strip_prefix(base_directory)`-or-full-path, batch uses `file_read.relative_path`; the same file imported through both entry points could hash to different roots. Edge-case only (a corpus goes through batch). 

No changes required.

mstomar125:
**Adversarial re-review run** (per the core merge policy). The reviewer confirmed the partition logic, child-id/mention/membership idempotency, single-file id honoring, and completion math are correct, and surfaced two real robustness defects in the new code, both now fixed in `76e00f68`:

1. **`get_nodes_by_ids` wasn't chunked** — one bound parameter per file overflows SQLite's ~999-parameter ceiling on a large directory import, which would error the existence check, default to an empty existing-set, and then fail with a duplicate-PK on re-import (or falsely fail a first import). Now chunked at 900 like the sibling bulk queries.

2. **`--replace` was destructive on failure** — it deleted a document's child subtree *before* the (non-transactional) bulk insert, so a mid-batch failure left already-processed docs with their children permanently deleted and not recreated. Restructured to insert-first-then-prune: the root is kept + updated, the fresh subtree is inserted, and only then are the captured old child ids deleted (new `delete_nodes_by_ids_unchecked`) — and only on a successful insert. A failed insert now leaves the previous content intact. Also de-duplicates repeated root ids within one batch.

New store-primitive tests (`reimport_store_primitives_test.rs`) cover the 900+ id chunk boundary and delete-by-ids exactness. Re-ran the live real-binary smoke against the 29 architecture docs after the refactor — still stable at 29 roots / 2976 nodes across import → re-import → `--replace`, membership intact.

mstomar125:
**APPROVE** — S3/S4 complete and correct. Adversarial re-review ran, both surfaced defects fixed and covered by new tests, live real-binary smoke green (import → re-import → --replace stable at 29 roots / 2976 nodes, membership intact), full pre-push gate green. Self-merging per merge policy.

PR #1666, opened 2026-07-15
Reorder the first-Pro consent flow to sign-in-first

mstomar125:
Closes #1665.

## Problem (found in live validation, 2026-07-15)
The Pro-sync surface demanded the public-workspace **publish** consent *before* the user had signed in, behind a pill labelled "Enable sync" that didn't read as "publish publicly" — and its "Keep local" decline could be a silent no-op (`disabled` mid-call) with no confirmation.

## What changed (frontend variant machine)
Reordered to **sign-in-first**, splitting the two-signal state machine into five variants:

| variant   | when                              | pill            | modal        | collab tab |
|-----------|-----------------------------------|-----------------|--------------|------------|
| teaser    | not Pro                           | teaser          | —            | —          |
| sign-in   | Pro, not authed, not enabled      | sync pill (OAuth) | —          | locked     |
| consent   | Pro, **authed**, not enabled      | turn-on-sync    | first-pro-consent | locked |
| relogin   | Pro, enabled, session lapsed      | sync pill       | pro-relogin  | live       |
| connected | Pro, authed, enabled              | sync pill       | pro-relogin  | live       |

- **Sign in first** (the sync pill drives OAuth); the publish consent is presented only **after** auth, in the `consent` variant.
- The consent modal **auto-opens once per fresh sign-in episode** and no longer starts OAuth on merge (the user is already signed in) — merge only records the consent (`pro_enable_sync`).
- **Pill relabeled** "Turn on sync" and both pill + modal name the public-workspace publish.
- **"Keep local" is never disabled into a no-op**, records the declined sign-in episode so the prompt doesn't immediately reopen, and shows a brief "Kept local — sync stays off" confirmation.
- `collaboration-locked` branches: reopen the consent modal when signed in, else point the user to the toolbar to sign in first.
- `isProSyncActive` stays "sync enabled" (`relogin | connected`) — the truth set (enabled databases) is unchanged, so membership/recovered-items consumers are unaffected.

## Guarantee preserved
The consent gate is intact: the daemon still only **pushes** after `sync_enabled`, which the merge flips. Signing in first connects and sets `auth_status=connected` but publishes nothing (the push sweep is server-side gated). The sync-side companion that makes enabling-after-connect actually upload the existing graph is nodespace-sync#305 (merged).

## Behavioral notes (transparency)
- The daemon connects on sign-in, so a user who declines ("Keep local") stays authenticated and the realtime **pull** remains active while **push** stays gated — nothing they own is published. Gating pull too (truly-idle-until-consent) is a possible follow-up; flagging for Anoop, not in scope here.
- The onboarding invitations inbox (in `pro-sync-pill`) now surfaces after **enabling** sync (connected) rather than in the consent window — consistent with sync being opt-in.

## Tests
- Variant-registry states + the sign-in → consent → connected transition, decline-registers, auto-open-per-episode, keep-local-actionable-while-pending.
- Full frontend suite green: **163 files / 4057 tests**. `eslint --fix` + `svelte-check` clean. Pre-push gate (test:all + `cargo build --bin nodespaced` + test:e2e) green.

### DIFF
--- a/packages/desktop-app/src/lib/components/collaboration/collaboration-locked.svelte
+++ b/packages/desktop-app/src/lib/components/collaboration/collaboration-locked.svelte
@@ -1,17 +1,26 @@
 <!--
   Collaboration locked placeholder.
 
-  Shown in the collection viewer's Collaboration tab for the `enable-prompt`
-  variant — a Pro daemon whose active database has `sync_enabled: false`. People &
-  roles live in the cloud tenant a database is bound to, so there is nothing to
-  show until sync is enabled. Purely static: no membership calls, no daemon
-  commands. Accepts the host collection's `nodeId` for a uniform viewer-extension
-  signature (surfaced as a data attribute for debugging).
+  Shown in the collection viewer's Collaboration tab for the `sign-in` and
+  `consent` variants — a Pro daemon whose active database has `sync_enabled:
+  false`. People & roles live in the cloud tenant a database is bound to, so there
+  is nothing to show until sync is enabled. Purely static: no membership calls, no
+  daemon commands. Accepts the host collection's `nodeId` for a uniform
+  viewer-extension signature (surfaced as a data attribute for debugging).
+
+  The action depends on where the user is in the sign-in-first flow: once signed
+  in (`consent`) the button reopens the publish-consent modal (mounted for that
+  variant); before sign-in (`sign-in`) there is no consent slot to open, so it
+  points the user at the toolbar sync button to sign in first.
 -->
 <script lang="ts">
   import { proSync } from '$lib/stores/pro-sync.svelte';
+  import { resolveProSyncVariant } from '$lib/plugins/ui-extensions.svelte';
 
   let { nodeId }: { nodeId: string } = $props();
+
+  // Signed in but not yet opted into sync ⇒ the consent modal is available here.
+  const canConsent = $derived(resolveProSyncVariant() === 'consent');
 
   function openConsent() {
     proSync.consentPromptOpen = true;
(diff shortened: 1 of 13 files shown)


mstomar125:
**Adversarial re-review: APPROVE** (independent pass, per core merge policy).

Traced all five variant combinations, the registry mappings, the auto-open derivation, timer lifecycle, and every consumer of the renamed variants. Verified: the consent guarantee holds (merge calls only `pro_enable_sync`, gated behind the acknowledged checkbox; auto-open never merges); `isProSyncActive` truth-set is unchanged (`relogin | connected` == "sync_enabled true"); Keep-local is never inert; timer `$effect` cleanup is leak-free; `collaboration-locked` has no dead button; no unmigrated old variant names.

Two items surfaced:
1. **Cross-repo verify (not a defect here):** the `sign-in → consent` edge needs the daemon to stamp `auth_status='connected'` while `sync_enabled` is still false. Confirmed on the sync side — the orchestrator sets it unconditionally on connect, independent of the push gate — so the flow is expressible. Re-confirm end-to-end in live validation with the rebuilt `nodespaced-pro` sidecar (from sync main, which now also kicks the post-enable push sweep).
2. **MINOR (fixed):** the consent modal auto-re-opened on every reload/restart even after "Keep local", because the decline lived only in-memory and session rehydration re-bumps `signedInEpisode`. Repeatedly popping an irreversible-publish dialog undermines informed consent (and violates the issue's "present … once"). Fixed in `99367536`: the decline is now persisted per database id in localStorage and gates the auto-open, so it's a one-time nudge that survives reload; a different database is unaffected, and the pill still reopens it deliberately. +3 tests.

PR #1671, opened 2026-07-15
Add shared-workspace (multi-user) guidance to the NodeSpace skill

mstomar125:
## Summary

Part of the **shared agent workspace** epic (NodeSpaceAI/nodespace-sync#303, workstream NodeSpaceAI/nodespace-sync#316): two engineers' Claude Code sessions share one synced NodeSpace collection as cross-session working memory.

The skill teaches an agent to use NodeSpace as **single-user** working memory. Several of those assumptions break when the Pro daemon is bound to a **shared, synced** collection (a teammate's agent reads/writes the same graph). This adds a `## Shared Workspaces (Multi-User)` section to `packages/skill/SKILL.md` covering:

- **Binding, not per-command target** — the daemon launch binding decides the shared collection; the agent creates nodes normally, and everything saved is visible to the teammate.
- **You're not the only writer** — prefer additive writes; pass the OCC `version` via `node batch-update` when editing a shared node (avoid `node update` without a version).
- **Attribute what you save** so a teammate can trace provenance.
- **Eventual consistency + semantic-search lag** — use structured `query`/`node query` for immediate cross-engineer recall; `search` over a peer's node only works after local embedding backfill (embeddings are local-only, ADR-045).
- **Don't file shared memory under date nodes** — they don't round-trip through sync yet.
- **Private + opt-in.**

## Safety / scope

- The addition sits **outside** the generated schema-rules block — `cargo run -p nodespace-agent --bin gen_skill_md -- --check` passes ("SKILL.md is up to date").
- `SKILL_MD` is embedded via `include_str!` in `acp/context_assembly.rs`; its tests compare the const to what's written (self-consistent) and `contains("NodeSpace")` — unaffected.
- Markdown-only; **community build unaffected**. Full pre-push gate green.

Implements NodeSpaceAI/nodespace-sync#316 (I will close it on merge — cross-repo).

### DIFF
--- a/packages/skill/SKILL.md
+++ b/packages/skill/SKILL.md
@@ -27,6 +27,26 @@ Use NodeSpace as a working memory across sessions:
 
 Date nodes make temporal retrieval reliable: if a finding is time-bound, attach it under today's date node so future searches can scope by day.
 
+## Shared Workspaces (Multi-User)
+
+A NodeSpace collection can be **synced and shared** with a teammate through NodeSpace Pro. When the daemon is bound to a shared collection, another engineer — or their agent — reads and writes the same graph, so you share context across sessions and across people. It is opt-in and private to its members. When you are working in a shared workspace, adjust how you use NodeSpace:
+
+**You share the whole workspace, not a per-command target.** You do not pick the shared collection per write — the daemon is launched already bound to it, so nodes you create sync into it automatically. Everything you save here is visible to your teammate. Keep private scratch notes in a separate, unshared database (`nodespace database create`/`--database`) if you need them.
+
+**You are not the only writer.** A node here may have been created or last edited by your teammate. Don't assume a node is yours or that its content is stable across your session. Search at session start to pull what your teammate has already saved.
+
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE** (adversarial re-review by author)

Markdown-only addition to `packages/skill/SKILL.md`; verified each claim against the current-state map before merging:

- **Daemon-binding targeting** — correct: the CLI `node create` hardcodes `collection: None`; push routes new nodes to the daemon's bound `config.collection` (`--collection`), so the agent creating normally is the right instruction.
- **OCC guidance** — matches the skill's own CLI reference: `node batch-update` is version-aware, `node update` without a version bypasses OCC. Telling agents to pass the read version when editing shared nodes is correct.
- **Semantic-search lag / embeddings local-only** — matches ADR-045 (embeddings regenerated per-machine, not synced); structured `query` as the immediate-recall path is accurate.
- **Date nodes don't sync** — matches the documented sync gotcha; the single-user SKILL elsewhere recommends date-node parenting, so carving out the shared exception is necessary, not contradictory.
- **Private + opt-in** — matches the consent gate (no push until sync enabled) and restricted-collection RLS.

Generation contract intact (`gen_skill_md --check` passes — edit is outside the generated block); `SKILL_MD` include_str tests are self-consistent and unaffected. Full gate green. Community build unaffected.

PR #1677, opened 2026-07-16
Harden pane hydration + poll daemon readiness until healthy (relates #1672)

mstomar125:
## What this is
Hardening of the pane-hydration path plus a persistent daemon-readiness poll. It **relates to #1672** (the fresh-install "Daily Journal Loading… forever" P0) and is the most likely fix, but the root cause is not yet confirmed on a live device, so **#1672 stays open** until the fresh-install validation leg runs (round-2, on the validation Mac). Do not auto-close #1672 on this merge.

## Why it's framed as hardening, not a confirmed fix
The round-1 report notes "reload does not fix it" and "no WatchNodes activity" across the hung session — both consistent with a failure at or below the app→daemon transport (e.g. a wedged lazy gRPC channel) rather than a swallowed one-shot fetch alone. A frontend re-invoke can only heal the latter. This PR fixes a real terminal-failure path (a hydration fetch that fails during the daemon boot window is never retried) and closes the signal hole that would defeat that retry — but confirming it resolves the user-visible hang needs a live fresh-install run, which can't be done headlessly.

## Changes
- `pane-content.svelte`: retry the active tab's hydration when the daemon becomes healthy (guarded by store-presence; not-found→close-tab semantics unchanged).
- `daemon-status.ts`: **poll `check_daemon_status` until the daemon is first observed healthy, then stop** and let the push channel carry later transitions. Previously the "healthy" signal came only from the backend's one-shot readiness emit (bounded by a fixed timeout) — a slower fresh-install boot would emit a single non-healthy status and never emit healthy, so `onDaemonReconnect` (this retry, and the sidebar stores) would wait forever. This makes the signal reliable for **every** consumer.

## Tests
- `daemon-status.test.ts` (+2): a boot where the push never carries healthy but a later poll observes it still fires `onDaemonReconnect`; the poll stops once healthy (no further pulls).
- `pane-content-hydration.test.ts` (+1): a hydration that fails then succeeds on the reconnect retry.
- Full frontend suite green (4063).

## Addresses the review
- **F1** — reframed to hardening; #1672 kept open for live validation (this comment + title).
- **F2** — the persistent readiness poll removes the one-shot-emit dependency. (The mid-session-daemon-restart-emits-nothing case is a broader daemon-status gap filed separately — see the follow-up issue.)
- **F3** — added tests that pin the actual daemon-status poll wiring (the load-bearing new behavior), not just the hydrate mirror.
- **F4** — the readiness poll logs each retry distinctly; the hydrate retry rides the existing error log.

### DIFF
--- a/packages/desktop-app/src/lib/components/layout/pane-content.svelte
+++ b/packages/desktop-app/src/lib/components/layout/pane-content.svelte
@@ -1,9 +1,10 @@
 <script lang="ts">
-  import { setContext, untrack } from 'svelte';
+  import { onMount, setContext, untrack } from 'svelte';
   import BaseNodeViewer from '$lib/design/components/base-node-viewer.svelte';
   import { navigationStore, updateTabContent, closeTab } from '$lib/stores/navigation.svelte';
   import { pluginRegistry } from '$lib/plugins/plugin-registry';
   import { sharedNodeStore } from '$lib/services/shared-node-store.svelte';
+  import { onDaemonReconnect } from '$lib/services/daemon-status';
   import type { Pane } from '$lib/stores/navigation.svelte';
   import { createLogger } from '$lib/utils/logger';
   import SettingsPane from '$lib/components/settings/settings-pane.svelte';
@@ -34,7 +35,11 @@
 
   // Load viewer when needed - moved to function called from onMount to avoid derived context issues
   async function loadViewerForNodeType(nodeType: string) {
-    if (viewerComponents.has(nodeType) || viewerLoadErrors.has(nodeType) || viewerLoading.has(nodeType)) {
+    if (
+      viewerComponents.has(nodeType) ||
+      viewerLoadErrors.has(nodeType) ||
+      viewerLoading.has(nodeType)
+    ) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


mstomar125:
**REQUEST CHANGES** (adversarial review of `215b898d`)

The 19-line retry is legitimate hardening of a real terminal-failure path and is correctly built (registration order, initial-fire semantics, cleanup, multi-pane behavior all check out). But:

- **F1 (blocker):** the PR claims "Fixes #1672" on an unvalidated root-cause theory that the round-1 evidence itself contradicts — the issue records "reload does not fix it" (a reload re-runs hydration against a then-healthy daemon and would heal without this fix), and the report records zero WatchNodes activity across the hung session (the watcher rides the same channel), both pointing at a failure below the webview (e.g. a wedged Rust-side lazy channel) that a frontend re-invoke cannot heal. Requires live fresh-install validation before it may close #1672; until then this is hardening.
- **F2 (should-fix):** the retry hinges on `onDaemonReconnect`, but "healthy" is emitted at most once, from a one-shot 30s-bounded startup task (`src-tauri/src/lib.rs:277` is the only emit). A fresh-install boot >30s emits "not_running" and no healthy ever follows — the retry never fires; mid-session daemon restarts emit nothing at all (which also invalidates the PR's #1644 claim). The healthy signal needs to be persistent (poll until healthy + keep observing after).
- **F3 (should-fix):** the new test is a hand-rolled mirror that imports neither the component nor daemon-status — deleting the entire `onMount` registration leaves the suite green.
- **F4 (nit):** a failed retry is as silent as the original failure; log attempt/outcome distinctly so device logs can distinguish "never retried" from "retried and failed".

Fix direction: persistent daemon-status polling so the reconnect signal is reliable for every consumer; retitle to hardening with "Relates #1672"; pin the actual wiring in tests; distinct retry logs. #1672 stays open pending the live fresh-install leg.

mstomar125:
Addressed the review; merging as **hardening** with **#1672 kept open** for the live fresh-install validation leg (round-2):
- **F1** — retitled/reframed to hardening; not closing #1672 on the unverified theory. The live repro will confirm whether this resolves the user-visible hang or whether the cause is below the webview.
- **F2** — added a persistent readiness poll in `daemon-status.ts`: it pulls `check_daemon_status` until the daemon is first observed healthy, then stops. This removes the dependency on the backend's one-shot readiness emit that a slow boot could miss. The mid-session-restart variant (no push event at all) is filed as its own systemic gap: #1678.
- **F3** — added two `daemon-status.test.ts` cases that pin the actual poll wiring (boot where push never carries healthy but a later poll observes it → reconnect fires; poll stops once healthy).
- **F4** — the readiness poll logs each retry distinctly.

Issue #321, opened 2025-10-21
Expose semantic search via MCP (search_containers method)


malibio:
## Overview

Expose the existing semantic search functionality to MCP clients like Claude Code. This enables AI agents to search NodeSpace content using natural language queries instead of just basic filtering, completing the AI agent workflow: **search → read → understand → modify**.

### Q1: EmbeddingService Location

**A: EmbeddingService already exists in core package!** ✅

**Location:** `packages/core/src/services/embedding_service.rs`  
**Name:** `NodeEmbeddingService` (not `EmbeddingService`)

**No refactoring needed** - just use the existing service. It's already accessible from core package.

```rust
// Already exists in core!
use nodespace_core::services::NodeEmbeddingService;

pub struct NodeEmbeddingService {
    nlp_engine: Arc<EmbeddingService>,
    db: Arc<DatabaseService>,
}

impl NodeEmbeddingService {
    pub async fn search_containers(&self, query: &str, threshold: f32, limit: usize) -> Result<Vec<Node>, NodeServiceError>
    pub async fn exact_search_containers(&self, query: &str, threshold: f32, limit: usize) -> Result<Vec<Node>, NodeServiceError>
}
```

### Q2: MCP Server Service Pattern

**A: Use Option B - Create a combined struct.**

**Why:**
- Cleaner API (single parameter vs multiple)
- Easier to extend (add more services later without changing signature)
- More Rust-idiomatic

**Implementation:**
```rust
// packages/core/src/mcp/server.rs

pub struct McpServices {
    pub node_service: Arc<NodeService>,
    pub embedding_service: Arc<NodeEmbeddingService>,
}

pub async fn run_mcp_server_with_callback(
    services: McpServices,  // Single combined struct
    callback: Option<ResponseCallback>,
) -> anyhow::Result<()> {
    // ...
    let response = handle_request(&services, state, request).await;
}

async fn handle_request(
    services: &McpServices,
    state: &ServerState,
    request: MCPRequest,
) -> MCPResponse {
    let result = match request.method.as_str() {
        "initialize" => { /* ... */ }
        "create_node" => nodes::handle_create_node(&services.node_service, request.params).await,
        "search_containers" => search::handle_search_containers(&services.embedding_service, request.params).await,
        // ...
    };
}
```

### Q3: Testing Scope

**A: Focus on unit tests and manual echo testing. Skip Claude Code integration for now.**

**What to implement:**
[...]

PR #1679, opened 2026-07-16
Render the consent variant reliably; stop the pill claiming "Synced" while sync is off (#1674)

mstomar125:
## Summary
Fixes #1674 — the first-Pro consent flow (#1665 "Part B") never rendered. On a genuinely fresh DB after sign-in the daemon reported exactly the consent-variant state (`auth_status=connected`, `sync_enabled=false`) but no consent modal appeared and the pill showed "Synced".

## Root cause
The variant machine's second axis — the per-database `database-settings-singleton` node — was read **once** at boot (fire-and-forget) and then relied on a single `node:updated` watch event to learn that sign-in flipped `auth_status`. That event has unrecoverable loss modes (watcher reconnect backoff, broadcast lag drops, a failed coalescer refetch), so the variant stayed stuck at `sign-in` (no modal), while the pill labelled purely off the realtime axis and showed "Synced" — realtime connects on sign-in by design even though push stays gated (#321).

## Changes
- **`database.svelte.ts`** — `refreshDatabaseSettings()` force-refetches the settings node (epoch-guarded against a database switch); a reconnect hook retries `load()` when nothing is loaded yet.
- **`pro-sync.svelte.ts`** — `setState` re-hydrates the settings node on every sync-state **transition edge**, so axis 2 follows axis 1 without depending on the lossy watch event.
- **`first-pro-consent-slot.svelte`** — refresh after EnableSync so consent→connected doesn't depend on the watch event either.
- **`pro-sync-pill.svelte`** — fail-safe mapping: never render "Synced" unless the settings node confirms `sync_enabled`; otherwise "Signed in — sync off" (label, tone, aria-label, tooltip). `isProSyncActive()` fails safe to false when the settings node is unhydrated.

## Tests
+18: a `sync:status` edge refetches the settings node and flips `resolveProSyncVariant()` sign-in→consent (real proSync + real store + real ui-extensions); pill label/tone pairs for connected/syncing × sync_enabled true/false + unhydrated fail-safe; refresh bypasses cache + epoch-drop; reconnect reload; merge triggers refresh. Full suite **4078** green; svelte-check 0/0; eslint/prettier clean; `free-user-guardrail` 11/11.

⚠️ Pairs with the daemon consent-gate (nodespace-sync#321, merged): UI + enforcement together. The "Signed in — sync off" copy is new; live validation on a real Pro sidecar (the #1665 repro) is still worthwhile in round-2.

### DIFF
--- a/packages/desktop-app/src/lib/components/first-pro-consent-slot.svelte
+++ b/packages/desktop-app/src/lib/components/first-pro-consent-slot.svelte
@@ -85,7 +85,10 @@
     } finally {
       pending = false;
       proSync.consentPromptOpen = false;
-      // Enabling flips the variant to `connected`, unmounting this slot.
+      // Enabling flips the variant to `connected`, unmounting this slot. Re-pull
+      // the settings node directly so that flip doesn't depend on the
+      // `node:updated` watch event, which can be lost for good (#1674).
+      databaseStore.refreshDatabaseSettings();
     }
   }
 
--- a/packages/desktop-app/src/lib/components/pro-sync-pill.svelte
+++ b/packages/desktop-app/src/lib/components/pro-sync-pill.svelte
@@ -11,6 +11,11 @@
     green      — CONNECTED
     red        — ERROR
 
+  Exception (#1674): CONNECTED/SYNCING only prove the daemon's realtime session
+  is live — i.e. signed in. When the active database has not opted into sync
+  (`sync_enabled: false` on its DatabaseSettingsNode, via isProSyncActive), the
+  pill shows a neutral "Signed in — sync off" instead, never "Synced".
+
   Click: when the daemon is signed out (DISCONNECTED, AUTH_REQUIRED,
   UNSPECIFIED, ERROR), clicking triggers the PKCE flow via
   `pro_initiate_oauth`. The daemon opens the browser; this UI just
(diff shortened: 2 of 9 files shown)


mstomar125:

Traced the full reactive chain, import closure, epoch semantics, and pill logic; ran all 6 changed test files (52 tests green). Every attack surface resolves in the PR's favor:
- The refresh→setNode→resolveProSyncVariant chain is unbroken (same SharedNodeStore singleton, SvelteMap per-key reactivity; the skip-while-editing guard doesn't drop the settings node since it's read-only) — empirically confirmed: the test drives a real `sync:status` edge and asserts the variant flips sign-in→consent (fails if refresh were a no-op).
- No import cycle (`database.svelte` imports the plain `ui-extensions.ts` constant, not the `.svelte.ts` that pulls pro-sync; 96-module BFS clean).
- Epoch guard is correctly scoped to DB switches only (bumped solely by `clearAll`←`switchTo`, which re-refreshes after).
- Pill fail-safe: unhydrated settings → `isProSyncActive()=false` → never "Synced"; connected+syncing users not overridden. No false "Synced" flash.
- No refetch storm (edge-gated), no feedback loop, free build provably unaffected (guardrail 11/11).

One low-severity should-fix (non-blocking, self-healing, **not a regression**): `refreshDatabaseSettings` applies the fetched node without a version guard, so two out-of-order local reads could momentarily clobber — same version-unguarded behavior the pre-existing `node:updated` path already has for this node; heals on the next transition. Captured for follow-up. Merging.

PR #1680, opened 2026-07-16
Coalesce relationship sync events + O(1) parent index to unblock the UI during a large pull (#1673)

mstomar125:
## Summary
Fixes #1673 — the round-1 P0 where the app is unusable during the initial cloud-sync pull of a populated tenant (24,197 nodes / 39,183 relationships): the daemon floods the WatchNodes stream (69× "subscriber lagged; events dropped") and the webview applied each event synchronously. Every `has_child` relationship event ran an **O(N) parent scan** and copied the entire children map (a global reactive invalidation), so a 39k-edge burst was **O(N²)** on the main thread — the app sat on "Loading…".

## Changes
- **`reactive-structure-tree.svelte.ts`** — a reverse child→parent index makes `getParent` **O(1)** (verified against the children map, self-healing on a miss so a stale entry can't return a wrong parent). New `runBatch()` applies a burst of ops under a **single** `notifyChange` instead of one per op.
- **`tauri-sync-listener.ts`** (Pro path) — buffer `has_child` relationship events over the existing 16 ms coalescing window and flush them through one `runBatch` in **arrival order** (so create/delete of the same edge in a window resolves exactly as un-coalesced). Chunk the node-fetch flush (200/chunk) with a main-thread yield and a **per-chunk** epoch re-check, plus a reentrancy guard so overlapping flushes can't share tombstone state.

## Community build unchanged
Non-Pro relationship events still apply synchronously per event (a dedicated guardrail test asserts this); `free-user-guardrail.test.ts` 11/11.

## Tests
+19: single `notifyChange` per burst; O(1) `getParent` asserted via a counting map (zero full-map iterations over 500 calls); create→delete and delete→create ordering both ways; per-chunk epoch drop on a mid-flush database switch; reverse-index consistency across add/remove/reparent/clear/restore. Full frontend suite **4079** green; svelte-check 0 errors; eslint/prettier clean.

Companion: the daemon-side double-spawn + startup-progress work is nodespace-sync#324 (progress `detail` renders here). Round-1 evidence: `validation-round1/reports/val-P0-ui-blocked-during-sync.md`.

### DIFF
--- a/packages/desktop-app/src/lib/services/tauri-sync-listener.ts
+++ b/packages/desktop-app/src/lib/services/tauri-sync-listener.ts
@@ -55,6 +55,11 @@ const log = createLogger('TauriSync');
  *  collect a daemon-flushed batch without adding perceptible latency. */
 const REPLAY_COALESCE_WINDOW_MS = 16;
 
+/** Max node fetches dispatched and applied per chunk of a coalesced flush.
+ *  Chunks are separated by a macrotask yield so an initial-pull burst of tens
+ *  of thousands of nodes cannot monopolize the webview main thread. */
+const NODE_FETCH_CHUNK_SIZE = 200;
+
 const pendingNodeIds = new Set<string>();
 let coalesceTimer: ReturnType<typeof setTimeout> | null = null;
 // Ids deleted while a flush is mid-fetch (between the snapshot and the apply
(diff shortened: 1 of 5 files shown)


mstomar125:

Tried hard to break it across all 7 surfaces; no correctness defect found. All 19 new tests pin the claimed behaviors and pass (123 tests green in the affected suites).
- **Reverse index** stays correct on every mutation path incl. the adversarial add A→B / reparent A→C / delete-stale-A→B sequence (the stale-delete guard is doubly protected: drops the index entry only when the edge existed under that parent AND the index still points there). The self-heal **cannot** return a wrong non-null parent — `getParent` verifies the index hit against the children map before returning.
- **Reactivity preserved**: `getParent` reads `this.children` unconditionally before the index lookup (including the miss path), so `$derived`/`$effect` consumers still invalidate — matches the old O(N) tracking exactly.
- **Ordering** is arrival-order preserved (create→delete absent, delete→create present, both tested); **epoch guard** re-checks per chunk and a mid-flush DB switch drops remaining chunks with no half-populated store (switch does clear()+clearAll() synchronously); **reentrancy guard** defers, never drops, deferred ids.
- **Community build** byte-unchanged (isPro-gated; guardrail 11/11 + a dedicated immediate-apply test).
- Tests are causal (CountingMap asserts zero full-map iterations over 500 getParent calls; runBatch asserts one map-reassignment per burst) — they'd fail on regression.

Nits only (non-blocking): runBatch assumes a synchronous mutate (sole caller is sync); reparent-prune now depends on index accuracy (safe — children is never externally mutated); +16ms latency on Pro has_child events (intended, symmetric with the node coalescer). Ship it.

PR #1681, opened 2026-07-16
Create the DB file on `database create`; error on wrong-database reads instead of serving the active DB

mstomar125:
## Summary

Two round-1 daemon findings on the multi-database (ADR-053) path:

1. **`database create` never created the DB file.** `DatabaseManager::create()` registered the database in the registry but never created/opened the SQLite file, so a freshly-"created" database 500'd on first use. `create()` now creates and opens the file *before* registering, and rolls the registry entry back if the open fails. `register()` now errors when the target file is absent instead of silently registering a phantom.

2. **A wrong-`--database` read silently served the active DB.** When a request carried an `x-ns-database-id` header naming a database the daemon has no `DatabaseManager` for, routing fell through to the default services and returned the *active* database's data — a silent wrong-answer. `routed_database_services()` now returns `UNIMPLEMENTED` for a named-DB request when no `DatabaseManager` is present, while a header-less request keeps the default-fallback behavior. Verified safe for normal flows: the CLI and desktop app only stamp the header on an explicit database-switcher selection, never on the default single-DB path.

## Testing
- `bun run test:all` + `cargo build --bin nodespaced` + `test:e2e` (full pre-push gate) green.
- Unit coverage: create-then-open rollback, register-absent-file error, routed-services header cases.

Closes #1675
Closes #1676

### DIFF
--- a/packages/cli/tests/cli_integration.rs
+++ b/packages/cli/tests/cli_integration.rs
@@ -1291,11 +1291,15 @@ async fn database_registry_round_trip() {
         .expect("second registered")
         .id
         .clone();
+    // Create makes the database file before registering it — it exists on disk
+    // as soon as the command returns, without any request having routed to it.
+    assert!(
+        tempdir.path().join("second.db").exists(),
+        "create must create the database file"
+    );
 
-    // Open the second database by routing a write to it, so its file exists on
-    // disk — this lets the remove assertion below prove the file is preserved
-    // rather than merely never created (a freshly created DB is "missing" until
-    // first opened).
+    // Route a write to the second database, proving the freshly created
+    // database serves requests immediately.
     let mut node_second = node_client_for(&sock, &second_id).await;
     node_second
         .create_node(CreateNodeRequest {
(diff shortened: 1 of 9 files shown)


mstomar125:

Verified the one intended behavior change is isolated and no normal flow breaks:
- All four `Server::builder()` sites in `packages/daemon/src/main.rs` install `DbManagerLayer`, so the core daemon always has a manager → the new `UNIMPLEMENTED` branch is unreachable in-core.
- Both client interceptors stamp the `x-ns-database-id` header only on explicit selection (`packages/cli/src/lib.rs`, `packages/desktop-app/src-tauri/src/services/grpc_client.rs` insert nothing when the id is `None`), so the default single-DB path is unaffected.
- `create()` creates+opens before registering and rolls the registry entry back atomically on save failure; `failed_create_rolls_back_and_leaves_the_daemon_serviceable` proves the daemon stays serviceable. e2e/CLI/database_service tests were strengthened, not weakened.

Two low-severity nits accepted as best-effort on exceptional paths (documented behavior, match existing `register`/`get_or_open` patterns): an orphan `-wal`/`-shm` window if a registry-save fails while a model is loaded, and a concurrent same-path `create` TOCTOU.

The Pro-build validation item (external Pro daemon may omit `DbManagerLayer`; confirm the desktop switcher passes `None` for the default DB rather than its concrete id) is tracked in nodespace-sync#327 — a live check for Wave 5, not a blocker here.

Merging.

PR #1682, opened 2026-07-17
Pin the pane's daemon-reconnect rehydration by mounting the real component

mstomar125:
## Summary

Round-1 fix validation turned up a real coverage gap on the journal-hydration fix.

The existing `pane-content-hydration` suite drives a **local copy** of `hydrateNode()`'s logic rather than the component, and says so in its own header. That means it cannot catch the wiring regressing: if the pane stopped subscribing to `onDaemonReconnect`, or subscribed with a callback that no longer re-hydrates, every one of those tests would still pass. This is the gap the #1677 adversarial review flagged ("the new test pins nothing — it doesn't import the component/daemon-status"). It matters because that subscription is the only thing that rescues a first hydration fetch which raced the daemon's boot window — without it the pane sits on "Loading..." forever.

`pane-content-reconnect-hydration.test.ts` mounts the real component and asserts the observable contract:
- it subscribes to daemon reconnect while mounted;
- a boot-window fetch failure is **retried** when the daemon reports healthy, and the tab survives (a failed fetch is not a missing node, so it must not be closed);
- reconnect is a **no-op** once the node is already present (no redundant refetch on a mid-session daemon restart);
- an unmounted pane stops retrying (subscription released on destroy).

**Mutation-tested:** with the `onDaemonReconnect` subscription removed from the component, the first two tests fail (`expected +0 to be 1`, `[1 call] != [2 calls]`) — so those genuinely pin the fix. To be precise about what each test buys: the last two (no-op-when-hydrated, stop-after-unmount) still *pass* under that mutation, since they only assert `ensureCalls` did not grow, which is trivially true when nothing is subscribed. They are kept as guards against an over-eager refetch, not as anti-regression cover for this fix.

The panes pane-content routes to are stubbed to keep the import graph shallow; pulling the real viewer tree in (`bits-ui` et al) is exactly what pushed the sibling suite into mirroring the logic.

One assertion is deliberately omitted and commented: that "Loading..." disappears after the retry. That would only be testing the real store's reactivity (`isNodeHydrated` is `$derived` off the store cell) — the mock store is a plain Map and this suite stubs runes non-reactively. The store's own suite covers that. Residual gap worth naming: nothing here proves the reconnect clears the UI end-to-end; that stays live-validation territory.

### Drive-by: unblock `quality:fix`

`import_service.rs` had a double-negative (`!(exists && !replace)`) that clippy rejects under `-D warnings`, which was failing `bun run quality:fix` for the whole repo (pre-existing, from `3489a323d` — not from this branch). Simplified to `!exists || replace`: identical by De Morgan, and it is the exact negation of the `if exists && !replace { skipped_roots… }` branch above it, so children are contributed for precisely the new + replaced roots — matching the comment. No semantic change to `--replace`.

Note the Rust diff is **not** a one-liner: running `quality:fix` also applied `cargo fmt`, which reflowed ~6 unrelated pre-existing lines in the same file (pure line-wrapping, no semantic change).

## Testing
- Full FE suite: **4102 passed / 4102** (166 files).
- `quality:fix`: 1767 files, **0 errors, 0 warnings** (was failing before this).
- `cargo test -p nodespace-daemon import`: green.
- Pre-push gate (test:all + `cargo build --bin nodespaced` + e2e): green.

Relates to #1672 — which stays **open**: this pins the hardening, but the root cause is still unconfirmed and needs a live fresh-install repro in the real Tauri app (browser mode can't exercise it — `daemon-status` is Tauri-only).

### DIFF
--- a/packages/daemon/src/services/import_service.rs
+++ b/packages/daemon/src/services/import_service.rs
@@ -587,8 +587,8 @@ async fn run_batch_import(
                 // Refresh in place. Update the root now (non-destructive) so its
                 // id and inbound links/mentions survive, and capture its current
                 // children to prune only after the fresh subtree is inserted.
-                let update = nodespace_core::NodeUpdate::new()
-                    .with_content(prepared.root_content.clone());
+                let update =
+                    nodespace_core::NodeUpdate::new().with_content(prepared.root_content.clone());
                 if let Err(e) = store
                     .update_node(&prepared.root_id, update, Some("import".to_string()))
                     .await
@@ -603,8 +603,7 @@ async fn run_batch_import(
                             prepared.root_id,
                             e
                         );
-                        phase2_error
-                            .get_or_insert_with(|| format!("Subtree replace failed: {e}"));
+                        phase2_error.get_or_insert_with(|| format!("Subtree replace failed: {e}"));
                     }
                 }
                 replaced_roots.push(prepared.root_id.clone());
@@ -627,7 +626,7 @@ async fn run_batch_import(
 
             // Children are (re)created for new and replaced roots. Skipped roots
             // keep their existing subtree, so contribute no children here.
-            if !(exists && !replace) {
+            if !exists || replace {
                 for child in &prepared.children {
                     let parent = child
                         .parent_id
@@ -748,7 +747,10 @@ async fn run_batch_import(
         // insert) means a failed bulk insert leaves the previous content intact
         // instead of truncating it — re-import is never destructive on error.
         if !bulk_insert_failed && !prune_after_insert.is_empty() {
-            if let Err(e) = store.delete_nodes_by_ids_unchecked(&prune_after_insert).await {
+            if let Err(e) = store
+                .delete_nodes_by_ids_unchecked(&prune_after_insert)
+                .await
+            {
                 tracing::error!(
                     "Failed to prune {} stale node(s) after replace: {}",
                     prune_after_insert.len(),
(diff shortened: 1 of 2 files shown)


mstomar125:

- **Boolean rewrite verified correct.** Truth table over all four combinations: `!(exists && !replace)` ≡ `!exists || replace` (plain De Morgan). It is also the exact negation of the `if exists && !replace { skipped_roots… }` branch above, so children are contributed for precisely the new + replaced roots and skipped roots contribute none — matching the comment. No semantic change to `--replace`.
- **Mutation claim independently reproduced.** Removing the `onMount(() => onDaemonReconnect(...))` block yields exactly 2 failed / 2 passed; `src/tests/components/` is green at 38 files / 549 tests with the file restored. This is a real improvement over the sibling suite, which mirrors `hydrateNode()` and would stay green through that deletion.
- **Noted honestly rather than papered over:** tests 3 and 4 are vacuous under the mutation (they only assert `ensureCalls` did not grow, trivially true with nothing subscribed) — kept as over-eager-refetch guards, not as cover for this fix. The omitted "Loading… disappears" assertion is legitimate (it would test the mock, not the component), but leaves a real residual gap: nothing proves the reconnect clears the UI end-to-end. That is live-validation territory and is tracked against #1672, which stays open.
- `settle()` is deterministic (all mock async resolves in microtasks, no timers) and `beforeEach`/`afterEach(cleanup)` reset every hoisted handle — no cross-test bleed.

PR description corrected: the Rust diff is not a one-liner — `quality:fix` also applied `cargo fmt`, reflowing ~6 unrelated pre-existing lines in that file (formatting only). A stray fmt-only change to `sqlite_store/nodes.rs` from the same `quality:fix` run was reverted to keep this PR focused.

Merging.

PR #1683, opened 2026-07-17
Poll daemon-status steady-state so a mid-session restart re-fires onDaemonReconnect (#1678)

mstomar125:
## Summary
Fixes #1678. The shared daemon-status service fans a reconnect signal to every daemon-dependent store (schemas, collections, children-tree, pane hydration) via `onDaemonReconnect`, fired on a not-healthy → healthy transition. That signal was reliable for the **initial** boot, but a **mid-session daemon restart** emitted nothing: all `daemon-status` emit sites live in the backend's one-shot startup task (`src-tauri/src/lib.rs`), so a daemon that dies and is relaunched while the app is running pushes no `healthy` event and no consumer retries — a wedged pane/store stays wedged until an app restart.

## Fix
Add a low-frequency steady-state poll of `check_daemon_status` (which probes the socket live each call) in `daemon-status.ts`, started after the listener wires up. A `healthy → not_running → healthy` cycle is now observed and `applyStatus` re-fires `onDaemonReconnect`. Reconnect callbacks fire only on the not-healthy → healthy edge, so steady healthy polls are no-ops. Also adds `stopDaemonStatusListener` (test teardown; production never calls it) and `.unref()`s the interval so it can't keep a process alive on its own.

## Why it matters for #1672
This closes one of the two plausible root causes of the journal "Loading…" hang: the case where the daemon's `healthy` transition is never *observed* by the frontend (the push fires before the webview registers its listener, and the single pull-on-subscribe caught `starting`), so the pane's `onDaemonReconnect` retry never runs. With the steady-state poll the daemon is re-observed and the pane self-heals.

## Tests
- `re-probes daemon health on a steady-state interval after start`
- `re-fires onDaemonReconnect when a steady-state poll observes a healthy → not_running → healthy restart` (the #1678 scenario, driven purely by the poll — no push)
- Full daemon-status suite 16/16; pre-push gate green (test:all + cargo build + e2e).

Relates: #1672, #1644, #1677.

### DIFF
--- a/packages/desktop-app/src/lib/services/daemon-status.ts
+++ b/packages/desktop-app/src/lib/services/daemon-status.ts
@@ -35,6 +35,19 @@ const log = createLogger('DaemonStatus');
 /** How long to show a "connecting" banner before any status is known. */
 const CONNECTING_GRACE_PERIOD_MS = 1500;
 
+/**
+ * How often to re-probe daemon health once the listener is running.
+ *
+ * The backend pushes `daemon-status` only from its one-shot startup task, so a
+ * daemon that dies and is relaunched mid-session emits nothing — every
+ * reconnect consumer would stay wedged until an app restart. A low-frequency
+ * pull observes the `healthy → not_running → healthy` cycle so `applyStatus`
+ * re-fires `onDaemonReconnect`. Kept slow because the only work on a steady
+ * healthy poll is a cheap socket probe; the reconnect callbacks fire only on
+ * the not-healthy → healthy edge, never on repeated healthy polls.
+ */
+const STEADY_STATE_POLL_MS = 15000;
+
 export interface DaemonStatusState {
   /** True until the first status is known or the grace period elapses. */
   connecting: boolean;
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE** (adversarial self-review)

Went looking for ways this breaks:
- **Reset on error:** `pollInFlight` is cleared in `.finally()`, so a thrown `getCurrent` can't wedge the poll permanently. ✓
- **Reconnect storm:** `applyStatus` fires `onDaemonReconnect` only on the not-healthy→healthy edge, so steady healthy polls are no-ops — the expensive store reloads don't re-run every 15s. Only the `_status.set` repeats, and downstream `$derived` memoizes identical values → no visible churn. ✓
- **No cross-thread race:** JS single-threaded; `applyStatus` is synchronous; `pollInFlight` only guards overlapping `getCurrent`. ✓
- **unref on browser-style handle:** optional-chained (`?.unref?.()`), so a numeric `setInterval` id is a safe no-op; node's Timeout gets `.unref()` so it can't keep a process/test-worker alive. ✓
- **Test isolation:** `afterEach` stops the singleton (same module instance the case used, since `resetModules` runs in `beforeEach`), so intervals don't leak across cases — 16/16 green. ✓

Pre-push gate green (test:all + cargo build + e2e), twice. Minimal, correct, tested. Merging.

PR #1687, opened 2026-07-17
Collapse execute_query into search_nodes as one agent query tool (closes #1640)

malibio:
## What & why

E4B scored 10/12 on `aichat-matrix.ts`. Scenario 5 ("List all my invoices") called `execute_query` instead of the expected `search_nodes`. The root cause was an **ambiguous tool boundary, not a bad description**: `execute_query` was a near-superset of `search_nodes` — it listed by type, filtered by title/content, filtered by property, **and** returned properties — so no amount of "prefer `search_nodes`" prose gave the model a clean rule for "list all X".

## Change

Collapse the two agent query tools into **one** — `search_nodes`:

- Covers title/keyword lookup, list-all-of-type, and typed property filtering (`filters` + `sorting`), and always returns node `properties`.
- `execute_query` / `Tool::ExecuteQuery` are deleted (13 tools now, was 14).
- Internally routes by capability: title/type listing → title index (`NodeService::query_nodes`, the only path that filters by title); property filters → `QueryService` SQL `json_extract` (`query_ops::execute_query`), so property queries keep the correct SQL path (not the slow in-memory one).
- **Unchanged (separate layer):** backend `query_ops::execute_query`, the daemon gRPC `ExecuteQueryRequest`, and the CLI `query` command.

Skills, tool-strategy guidance, and test fixtures updated to the single tool. Docs updated in nodespace-docs (data-layer.md Query System, prompt-skill-system.md whitelist, ADR-056 consequence).

## Verification — `aichat-matrix.ts` vs `gemma-4-e4b-q4km`: **11/12** (was 10/12)

| | Scenario | Tools called |
|---|---|---|
| ✅ | 1. Greeting | `(none)` |
| ✅ | 2. Capability | `(none)` |
| ✅ | 3. Schema creation | `create_schema` |
| ✅ | 4. Instance creation | `search_skills,search_nodes,create_node` |
| ✅ | **5. List/query** | **`search_nodes`** (was `execute_query` — the fix) |
| ❌ | 6. Update | `search_nodes` (stops before `update_node`) |
| ✅ | 7. Empty-result query | `search_nodes` |
| ✅ | 8a. Create type: book | `create_schema` |
| ✅ | 8b. Create type: contact | `create_schema` |
| ✅ | 8c. Instance: book | `search_skills,create_node` |
| ✅ | 8d. Instance: contact | `search_skills,create_node` |
| ✅ | 8e. Query across types | `search_nodes` |

Scenario 5 is fixed; no regression on the other passing scenarios.

**Scenario 6 is the sole remaining failure — by design, moved to #1637.** `Expected sequence [search_nodes,update_node]... got: search_nodes`: the model finds the node but stops before the write. This is not prompt-fixable (proven in #1637 via prompt-dump analysis) and needs query decomposition + read-then-write turn completion, both of which #1637 now owns. E4B reaches 12/12 once #1637 lands.

> Matrix-run note: on the RAM-constrained test machine, the shipped #1638 KV-headroom logic floored E4B's context to 4096 (prompt overflow). The run above was captured with a throwaway local `KV_HEADROOM_BYTES` reduction to size the context to 11520; that change was reverted and is **not** in this PR (only the 5 agent/core source files are).

Unit tests: agent 285, core-ops 40, nlp-chat 84, frontend 4104, aichat-matrix 21 — all pass. `quality:fix` clean.

Closes #1640.


malibio:
## Pragmatic Code Review — PR #1687

**Verdict: REQUEST CHANGES** (one small but real functional bug; everything else is a nit). The architecture is sound and the tool-collapse is the right call — a single change away from mergeable.

### Summary

Collapsing the two overlapping query tools into one `search_nodes` is a genuine net-positive: it removes the ambiguous tool *boundary* that was the documented root cause of Scenario 5, and does it with the correct architectural seam — the agent tool routes by capability while the backend `query_ops`/gRPC/CLI layer stays untouched. The registry-driven design (single `Tool` enum → definitions, dispatch, seeds, whitelists all derived) means the removal is enforced by the compiler and the whitelist-validation test, so there is very little room for silent drift. Verification is thorough (11/12, with the 12th failure legitimately owned by #1637). The one blocker is a default-limit mismatch that silently truncates the exact "list all X" flow this PR is about.

---

### Findings

#### [Critical/Blocker] Default-limit mismatch silently truncates "list all X"

- `packages/agent/src/local_agent/tools.rs:277` — schema advertises `"Max results to return (default 50)"`
- `packages/core/src/agent_params.rs:35` — doc-comment says `Default: 50`
- `packages/agent/src/local_agent/tools.rs:887` — code uses `params.limit.unwrap_or(DEFAULT_SEARCH_LIMIT)` where `DEFAULT_SEARCH_LIMIT = 10` (tools.rs:124)

When the model omits `limit` (the common case for "list all my invoices" / "show all tasks"), it receives **10** rows, not the **50** the schema promises. For the headline use case this PR fixes, a user with >10 invoices gets a silently truncated list with no `truncated`/`has_more` signal, and the model will confidently report a partial set as complete. This is a correctness regression against the tool's own advertised contract.

Pick one and make all three agree:
- If 50 is intended: `let limit = params.limit.unwrap_or(50);` (or bump `DEFAULT_SEARCH_LIMIT` — but that constant is shared, so verify no other caller relies on 10), **or**
- If 10 is intended: change the schema string and the `agent_params.rs` doc back to "default 10".

Given this is a *listing* tool now, 50 is the better default — but either way the contract and the code must match. (Principle: least astonishment / self-consistent API contract.)

#### [Improvement] Two stale in-repo references to the removed agent tool

The PR says docs were updated in `nodespace-docs`, but two in-repo references to "the local agent's `execute_query` tool" were missed and now describe a tool that no longer exists:

- `packages/cli/src/commands/query.rs:6` — "...plus sorting, matching the local agent's `execute_query` tool."
- `packages/skill/SKILL.md:225` — "This is the CLI counterpart of the local agent's `execute_query` tool."

Both are in hand-written sections (not the `<!-- BEGIN GENERATED -->` block). The CLI/gRPC `execute_query` still exists as a backend op, so the fix is to reword to reference the backend query op / `nodespace query` command rather than "the local agent's tool" — e.g. "matching the backend structured-query path (`query_ops::execute_query`)". Low blast radius, but it's exactly the kind of drift that misleads the next reader.

#### [Nit] `query` + `filters` routes the keyword to `content`, not `title`

`tools.rs:922-933`: when both a keyword and property filters are present, the keyword is appended as a `content contains` filter (QueryService has no title path). This is a reasonable, well-commented degradation and I agree with the trade-off — flagging only so it's a conscious decision. If "list invoices named X with status=open" ever becomes a real path, title-substring semantics would silently differ from the no-filter path (which matches `title`). Acceptable for now.

---

### Focus-area checklist (as requested)

1. **Hybrid routing correctness** — Correct. No-filter path → `node_ops::query_nodes` with a `title contains` filter (or `None` when query empty, correctly listing all of a type); filter path → `query_ops::execute_query` with `target_type = node_type or "*"`. Verified `"*"` is an explicitly-supported wildcard in `validate_identifier` (query_ops.rs:167), so the omitted-node_type case is safe. Keyword-as-content on the filter path is acceptable (nit above). The **only** correctness issue is the limit default (blocker above).
2. **Dangling references** — Code references: clean (`cargo check -p nodespace-agent -p nodespace-core` passes; the only surviving `execute_query` mentions in the agent crate are the intended internal routing calls). Whitelists: `Research & Search` correctly drops `execute_query` and the `all_skill_whitelist_tools_are_registered` test guards against re-introduction. Two doc strings missed (improvement above).
3. **Test-fixture retargeting** — Faithful. The narrated-tool-call detector keys off `Tool::ALL` membership; `execute_query` is no longer a member and `search_nodes` is, so retargeting the fixtures to `search_nodes(...)` preserves the original test intent (both the "detects a real narrated call" and "does not trip on prose mention" cases still exercise a registered-vs-registered distinction). The `summarize_executions` fixture correctly updated its expected humanized label to `node search`. Args payload updated from `{"target_type":"task"}` to `{"query":"","node_type":"task"}`, matching the new schema.
4. **Requirements validation (#1640, updated scope)** — Legitimate. Deferring Scenario 6 to #1637 is well-documented: #1637 exists, is scoped to both query decomposition and read-then-write completion, and contains verified prompt-dump evidence that the failure is not prompt-fixable. The updated #1640 acceptance criteria explicitly allow "remaining gap documented with reasoning," which is satisfied.
5. **Security/maintainability** — No security surface change (same ops, same `validate_identifier` SQLi guard on the property path). Maintainability improves: one query tool, registry-enforced, drift-detected by tests.

### Requirements checklist

| # | Criterion (updated #1640) | Status |
|---|---|---|
| 1 | Root cause identified for Scenario 5 | ✅ Ambiguous tool boundary, documented |
| 2 | Root cause identified for Scenario 6 → moved to #1637 | ✅ Documented, #1637 owns it |
| 3 | Tool-collapse + skills/guidance/descriptions updated | ✅ (2 stale doc strings missed — improvement) |
| 4 | aichat-matrix vs E4B: Sc.5 passes, no regression, gap documented | ✅ 11/12 reported |
| 5 | quality:fix clean; tests pass | ✅ Reported; `cargo check` confirmed clean |

**Recommendation: REQUEST CHANGES** — fix the limit default (blocker), ideally sweep the two stale doc references while in there. Both are one-line changes; no re-architecture needed.

malibio (commented):
Review posted as a summary comment. NET: REQUEST CHANGES in spirit — one blocker (default search limit is 10 in code via DEFAULT_SEARCH_LIMIT but the schema/param doc advertise 50, silently truncating the 'list all X' flow this PR targets) plus two stale in-repo doc references to the removed 'execute_query' agent tool (cli/src/commands/query.rs:6, skill/SKILL.md:225). Architecture and the tool-collapse are sound; both fixes are one-liners. (Cannot use --request-changes on your own PR, hence a comment review.)

malibio:
## Review addressed (commit 30c0de64)

**✅ 🔴 Default-limit mismatch** — bumped `DEFAULT_SEARCH_LIMIT` 10 → 50. The tool now honors its own schema contract ("default 50") and matches the server/CLI default; a bare "list all X" is no longer silently capped at 10. Verified the const has no other caller.

**✅ 🟡 Stale `execute_query` doc refs** — reworded both in-repo strings that called it "the local agent's tool" (it's backend-only now):
- `cli/src/commands/query.rs` → "backed by the same QueryService query op the local agent's `search_nodes` tool routes property filters through"
- `skill/SKILL.md` (hand-written CLI section) → "CLI counterpart of the property-filtering path of the local agent's `search_nodes` tool". `gen_skill_md --check` still passes (generated block untouched).

**⏭️ 🟢 Nit (query+filters → content, not title)** — skipped intentionally. QueryService has no title-filter path; the degradation is deliberate and commented. Adding a title filter to QueryService is #1638/QueryService scope, not this PR's.

Tests: agent lib (incl. tools) pass; SKILL.md sync check passes; pre-push gate (test:all + build + e2e) green.

### Re-Review Decision
**NO RE-REVIEW NEEDED** — all changes are one-line, low-risk (a constant value + two doc-comment rewords). No logic paths changed beyond the default value the schema already advertised; the skipped nit is documented.

PR #1691, opened 2026-07-18
Daemon does not suppress same-origin WatchNodes echo — frontend guesses via content-comparison heuristic instead of authoritative client identity

malibio:
Closes #1689

### DIFF
--- a/packages/cli/src/lib.rs
+++ b/packages/cli/src/lib.rs
@@ -26,6 +26,17 @@ use tonic::transport::Channel;
 /// nothing, letting the daemon fall back to its default database; that variant
 /// is applied uniformly so a client's type is the same whether or not a
 /// database was selected.
+///
+/// Deliberately does NOT stamp `x-ns-client-id` (ADR-026 C5 extension,
+/// implemented in the desktop app's own `DatabaseIdInterceptor` at
+/// `packages/desktop-app/src-tauri/src/services/grpc_client.rs`) — a separate,
+/// same-named struct in a different crate. The CLI is a one-shot process per
+/// invocation that never opens `WatchNodes`, so it has no same-origin echo to
+/// suppress; leaving its writes untagged means they carry no
+/// `source_client_id` and are therefore always visible as foreign writes to
+/// any other subscriber (e.g. a desktop window), which is the correct
+/// behavior. If the CLI ever needs its own stable client id, add it here to
+/// this struct — the desktop-app copy is independent and not shared.
 #[derive(Clone)]
 pub struct DatabaseIdInterceptor {
     // Some(id) → stamp header on every request; None → stamp nothing (daemon
--- a/packages/daemon/src/services/node_service.rs
+++ b/packages/daemon/src/services/node_service.rs
@@ -65,6 +65,17 @@ use crate::nodespace::{
 ///
 /// `embedding_state` is `None` while the model is loading or when the NLP
 /// engine is absent. Semantic search returns `UNAVAILABLE` in both cases.
+///
+/// **Adding a new unary RPC handler**: call `self.route(&request).await?`
+/// first, exactly like every existing handler — this both resolves ADR-053
+/// database routing and scopes the returned `node_service` for same-origin
+/// write tagging (ADR-026 C5 extension) via the `x-ns-client-id` header.
+/// **Adding a new streaming (subscribe-style) handler**: if it should also
+/// suppress the subscriber's own writes, read `x-ns-client-id` via the
+/// standalone `client_id_header()` helper *before* calling `route()` — see
+/// `watch_nodes` for the reference implementation. Do not rely on `route()`
+/// alone for a subscribing handler: `route()`'s use of the header scopes
+/// *writes*, which a read-only streaming handler never performs.
 #[derive(Clone)]
 pub struct NodeServiceImpl {
     node_service: Arc<CoreNodeService>,
--- a/packages/daemon/src/services/node_service.rs
+++ b/packages/daemon/src/services/node_service.rs
@@ -108,12 +108,44 @@ impl NodeServiceImpl {
     /// routing middleware installed a header-less request falls back to `self`
     /// while a header-carrying one is rejected rather than silently served from
     /// the active database.
+    ///
+    /// Also applies same-origin write tagging (ADR-026 C5 extension): a
+    /// request carrying `x-ns-client-id` gets its `node_service` scoped via
+    /// `with_client(id)`, so any event this request's write emits is stamped
+    /// with that id as `source_client_id` — the signal `watch_nodes` uses to
+    /// drop the writer's own echo. Applied here, once, so every RPC handler
+    /// (which all start with `let this = self.route(&request).await?`) picks
+    /// it up without individually touching the client-id header.
     async fn route<T>(&self, request: &Request<T>) -> Result<NodeServiceImpl, Status> {
-        match crate::db_routing::routed_database_services(request).await? {
-            Some(services) => Ok(services.node_service_grpc.clone()),
-            None => Ok(self.clone()),
+        let mut this = match crate::db_routing::routed_database_services(request).await? {
+            Some(services) => services.node_service_grpc.clone(),
+            None => self.clone(),
+        };
+        if let Some(client_id) = client_id_header(request).map_err(|e| *e)? {
+            this.node_service = Arc::new(this.node_service.with_client(client_id));
         }
+        Ok(this)
     }
+}
+
+/// Read the `x-ns-client-id` header off a request, if present. Absent →
+/// `Ok(None)`, meaning the caller's writes emit no `source_client_id` and are
+/// never suppressed on any `WatchNodes` stream — behavior identical to before
+/// this header existed. `Status` is boxed on the error path so this small
+/// `Option<String>` success type doesn't blow up `Result`'s stack size next
+/// to it (`clippy::result_large_err`).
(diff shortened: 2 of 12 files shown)


malibio:
### Code Review Summary

This is a clean, well-scoped architectural fix that removes a client-side guessing heuristic in favor of daemon-side authoritative adjudication, exactly per ADR-026 C5's original intent. The implementation correctly threads client identity through the write path (`route()` → `with_client()`) and separately through the subscribe path (`watch_nodes` reading the header directly, pre-`route()`), which is the subtle part of this change and it's handled correctly. All four acceptance-criteria-adjacent concerns I was asked to dig into check out. Verified: `cargo test -p nodespace-daemon` (31/31 unit tests incl. 3 new echo-suppression tests, plus the real-transport integration test) and `bun run test` (166 files / 4095 tests) both pass; `cargo clippy` on daemon/proto is clean.

This is a net positive and I see no blockers.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- **`packages/daemon/src/services/node_service.rs` (`route()`, `watch_nodes`)**: The write-scoping vs. subscribe-filtering split is correct but relies on every future RPC handler continuing to call `route()` first and every future streaming handler continuing to read the header *before* calling `route()`. This is well-documented in comments but not structurally enforced — a future handler that reads the client-id header only via `route()`'s side effect (rather than the standalone `client_id_header()` helper) would silently break. Not a blocker; worth a one-line note in `NodeServiceImpl`'s doc comment steering future RPC additions toward the existing pattern, since this is the kind of thing that's easy to get subtly wrong by copy-pasting the wrong handler as a template.

- **`packages/desktop-app/src-tauri/src/services/grpc_client.rs`**: `client_id` is threaded as a required `MetadataValue<Ascii>` parameter through `DatabaseIdInterceptor::none()`/`for_id()`, generated once in `from_channel()` via `generate_client_id()`. This is correct and the "one id per process/window lifetime" invariant is well-documented, but it's asserted only by convention (comments) — nothing stops a future call site from constructing a second `DatabaseIdInterceptor` with a freshly-generated id and accidentally splitting one window's writes/subscription across two identities, silently defeating suppression for that window. Given the existing `GrpcClientInner.client_id` field already centralizes this, the risk is low; flagging only because the failure mode (stale client-id ⇒ own writes stop being suppressed ⇒ old bug quietly returns) is exactly the class of bug this PR is designed to eliminate.

- **CLI foreign-write correctness (point c in the review brief)**: `packages/cli/src/lib.rs` has its own, separate `DatabaseIdInterceptor` (not the one modified in `grpc_client.rs`) that never stamps `x-ns-client-id`. That's correct today — CLI writes carry no `source_client_id` and are therefore never suppressed on any subscriber's stream, so a future second desktop window will correctly see them as foreign. Worth calling out explicitly since it means the CLI and desktop-app now have two independently-maintained interceptor structs with the same name in different crates; a future contributor adding client-id support to one is likely to assume it covers both. Consider a shared helper or at least a cross-reference comment between the two, though this is a pre-existing duplication (not introduced by this PR) and not a blocker.

#### Nitpicks

- Nit: `packages/daemon/src/services/node_service.rs::client_id_header` boxes the `Status` error to satisfy `clippy::result_large_err` — reasonable and well-commented, matches the pattern in `db_routing`.
- Nit: The new integration test's helper naming (`window_a`/`window_b` as clones of the same `client`) reads slightly oddly since they're literally the same tonic client value with different per-request headers, not independent connections — a one-line comment clarifying that a single tonic client is being reused to simulate two logical windows (since the header, not the transport connection, is what the daemon keys on) would help a future reader who expects two separate channels.

### Verification of the four specific concerns raised

**(a) write-scoping vs. subscribe-filtering split in `route()`/`watch_nodes`**: Correct. `watch_nodes` reads `x-ns-client-id` via the standalone `client_id_header()` helper *before* calling `self.route(&request)`, specifically because `route()`'s use of the same header is for scoping the returned `node_service` for writes, which `watch_nodes` never performs. Confirmed via the doc comments on both call sites and the passing `watch_nodes_untagged_write_is_delivered_to_tagged_subscriber` test, which specifically isolates this from the write-tagging path.

**(b) removing `serverConfirmedVersions`/OCC-stash — no reintroduced OCC-defeat**: Confirmed safe. The old stash was a client-side "avoid a conflict I predict is spurious" optimization that was only safe when the own-echo classification was correct — and that classification was the exact recurring bug (#1437/#1586/#1684). `computeOccVersionForUpdate` now always reads the local node's own version; the next `UpdateNode` RPC either succeeds normally or the daemon's real OCC check (`isVersionConflict`/`x-version-conflict`) surfaces a genuine conflict. This is strictly more correct than the old stash-based shortcut, not a regression — the daemon's version check was always the actual source of truth; the stash only ever changed whether a real conflict got preempted by (potentially wrong) client guesswork.

**(c) CLI writes still treated as foreign by a future second desktop window**: Confirmed. See "Suggested Improvements" above — the CLI's interceptor is untouched and never stamps the header, so its writes carry no `source_client_id` and are never suppressed.

**(d) local-agent writes correctly NOT tagged with the frontend's own client id**: Confirmed. `packages/daemon/src/services/local_agent_service.rs` calls `self.inner.node_service` directly — it never goes through the gRPC `NodeServiceImpl::route()` path or reads `x-ns-client-id`, so agent writes never carry a `source_client_id` and will always appear as foreign writes on any `WatchNodes` subscription, including the desktop window that spawned the agent session. This is the correct behavior per the issue's design (the frontend must still see agent-originated writes).

### Acceptance Criteria (issue #1689)

All checked:
- [x] Daemon assigns a stable identity per client connection, threaded through `NodeService::with_client()` for writes
- [x] `watch_nodes` suppresses same-origin events so the frontend no longer content-compares
- [x] `isPlausibleOwnEcho`/`lastPersistedContent` deleted (not just reduced to a fallback)
- [x] No regression to genuine foreign-write conflict detection — verified via passing tests and the real-OCC-check argument above
- [x] Debounced-write echo race eliminated by construction (no content comparison exists anymore)
- [x] Companion ADR-026 amendment (nodespace-docs commit 12acef5, already pushed)
- [x] `quality:fix` clean (clippy clean, `git diff --stat` shows only the intended files)
- [x] `test:all`-equivalent passing (4095/4095 frontend, 31/31 + 1 integration daemon tests)

malibio:
## Addressed Review Feedback

**Addressed (4):**
- Added a steering note in `NodeServiceImpl`'s doc comment for future RPC handlers on the write-scoping (`route()`) vs. subscribe-filtering (`client_id_header()` pre-`route()`) split.
- Cross-referenced the CLI's and desktop-app's independent, same-named `DatabaseIdInterceptor` structs, documenting why the CLI's copy intentionally never stamps `x-ns-client-id`.
- Clarified the real-transport test's `window_a`/`window_b` naming (same tonic client, different per-request headers — the daemon keys suppression on the header, not the transport connection).
- Fixed grammar in the `grpc_client.rs` doc comment.

**Skipped (2, both explicitly non-blocking per the review):**
- Making the "one client id per `GrpcClient` lifetime" invariant type-safe instead of convention-enforced — the field is already centralized in `GrpcClientInner`, the review called the risk low, and a structural fix is disproportionate for a single-process desktop app with exactly one `GrpcClient` instance.
- Merging the CLI's and desktop-app's `DatabaseIdInterceptor` structs into a shared helper — pre-existing duplication not introduced by this PR, out of scope for this fix.

**Also found and fixed during verification** (not from the review, but surfaced by the pre-push gate's real-transport test suite): `optimistic_echo_race_test.rs` drove writes and watched for their echo through the *same* `GrpcClient` — a scenario the daemon-side suppression in this PR makes architecturally impossible (a connection's own writes are now never echoed back to it). Rewrote both tests in that file to use two independent `GrpcClient`s (matching the existing pattern in `cross_window_hierarchy_sync_test.rs`), so the watching window observes a genuinely foreign writer's broadcasts instead — the scenario that remains real per the issue's multi-client carve-out.

## Re-Review Decision

**Rationale**: All changes in this round were either (a) documentation-only steering comments, (b) a straightforward two-client rewrite of an existing test file following an established pattern already used elsewhere in the same test directory, or (c) grammar fixes. No production logic changed. Full `test:all` (166 files / 4095 tests) and the full `nodespace-app` ADR-048 Tauri-seam integration suite (incl. the rewritten `optimistic_echo_race_test.rs`) pass, and the pre-push gate's complete gRPC round-trip + Tauri-seam suite passed on push.

malibio:
## Merge conflict resolved

Rebasing on \`main\` surfaced a substantive conflict with #1693 (merged after this PR was reviewed): #1693 independently added a version-comparison guard (\`incomingIsNewer\`) to \`decideRemoteUpdate\` to fix the same visible symptom ("conflict toast on every keystroke"), but via a **different mechanism** targeting a **different echo source** — stale replays from \`nodespace-sync\`'s in-process writes (\`NodeService::with_client("sync-service")\`, ADR-027), which this PR's \`x-ns-client-id\` gRPC header does not cover (the sync service never opens a gRPC connection).

**Resolution**: kept #1693's \`incomingIsNewer\` version guard (still required — it's the only remaining protection against stale sync-service replays) but dropped the content-based \`isPlausibleOwnEcho\`/\`stashVersion\` machinery it was paired with, since that's now unreachable — same-gRPC-connection echoes never arrive at all after this PR's daemon-side fix. Updated \`remote-update-policy.ts\`'s doc comment and tests to describe both mechanisms and why each is independently still necessary.

Full \`test:all\` (166 files / 4098 tests) and the complete pre-push gate (gRPC round-trip + Tauri-seam ADR-048 suites) pass post-merge.

PR #1704, opened 2026-07-20
Deterministic collection ids to stop --auto-collection-routing creating duplicates (#1703)

mstomar125:
Addresses the duplicate-collections facet of #1703 (98 collection nodes for 18 names in `tenant_demo`).

## Root cause
`CollectionService::create_collection` minted a fresh **random UUID** (`Node::new`) for each collection node. `resolve_path`/`bulk_resolve_collections` dedup *within a single store* by name lookup, but two **independent** imports — a different device, a re-import on a fresh local db, or a concurrent double-create — each generate a different id for the same-named collection. After sync those distinct ids accumulate as duplicate collection nodes.

## Fix
Collection names are globally unique (documented invariant), so derive the collection node id **deterministically** from the normalized name (UUIDv5):

```rust
pub fn deterministic_collection_id(name: &str) -> String {
    Uuid::new_v5(&COLLECTION_ID_NAMESPACE, normalize_collection_name(name).as_bytes()).to_string()
}
```

`create_collection` becomes an idempotent **get-or-create by that id**: reuse an existing node (a prior create, a synced copy, or a race winner) instead of erroring on the duplicate id. Independent stores now converge on one id per collection, so sync upserts them into a single node; a concurrent double-create resolves to the same node.

- Non-breaking for existing data: `resolve_path` still finds pre-existing (random-id) collections by name and reuses them — only brand-new collections get deterministic ids.
- The namespace UUID is fixed and must not change (it anchors collection identity).

## Scope
This PR fixes the **duplicates** facet. The second facet noted on the issue — imported sub-collections having no *person* `member_of` edge, so they don't surface in the sidebar — is a separate concern (the import path has no importer-identity context; it needs identity plumbing or a sidebar/reachability change). It's being ground-truthed against a fresh re-import + membership pass and will be split into a focused follow-up.

## Tests
- `deterministic_collection_id` stable + case/whitespace-insensitive.
- Two independent stores converge on one collection id (the cross-device dedup that sync relies on).
- Re-resolving a path and paths sharing a segment create no duplicate nodes.
- Full `nodespace-core` lib suite (960) + daemon `import_round_trip` green; clippy + fmt clean; pre-push gate (frontend + Rust + e2e) passed.

### DIFF
--- a/packages/core/src/ops/collection_ops.rs
+++ b/packages/core/src/ops/collection_ops.rs
@@ -5,6 +5,7 @@
 
 use crate::models::{Node, NodeUpdate};
 use crate::ops::OpsError;
+use crate::services::collection_service::deterministic_collection_id;
 use crate::services::{
     CollectionService, CreateNodeParams, InsertPositionOwned, NodeService, NodeServiceError,
 };
@@ -308,9 +309,14 @@ pub async fn create_collection(
         serde_json::json!({ "description": input.description })
     };
 
+    // Deterministic id from the (globally-unique) name so a UI-created collection
+    // converges with the same-named collection created on another device or by import
+    // (`CollectionService::create_collection` derives the same id) instead of minting a
+    // random UUID that syncs up as a duplicate. The name-existence check above already
+    // rejects a local duplicate.
     let collection_id = node_service
         .create_node_with_parent(CreateNodeParams {
-            id: None,
+            id: Some(deterministic_collection_id(&input.name)),
             node_type: "collection".to_string(),
             content: input.name,
             parent_id: None,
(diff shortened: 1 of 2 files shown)


mstomar125:
## Adversarial review + resolution

Independent adversarial review verified the deterministic-id mechanism as safe and correct on the import path (all six checks CONFIRMED):
- Dedup is a pure function of the normalized name; the name-lookup (`LOWER(title)`) and id-derivation agree, and the deterministic `get_node(id)` is a self-consistent backstop (it even closes a latent Unicode `LOWER()` vs Rust `to_lowercase()` gap).
- Get-or-create race: the loser re-fetches the winner; a real (non-duplicate) error still propagates (no node at `id` → `get_node` None → `Err` propagates).
- Root/default collection (`c0000000-…-001`, version nibble `0`) can never be produced or collided with by a v5 UUID; not name-resolved through this path — no shadowing.
- Mixed legacy+new: a pre-existing random-id collection is found by title and reused (no second local node).
- UUIDv5 usage / const namespace / 128-bit literal all valid; no cross-node-type id collision (dedicated v5 namespace).

**Gap found and fixed (commit 0c369a6c):** the fix originally covered only the import path — the interactive `ops::create_collection` (Tauri/gRPC "New collection") still minted a random UUID (`create_node_with_parent { id: None }`), so a UI-created collection wouldn't converge cross-device (the change's own goal). Now it passes `id: Some(deterministic_collection_id(&name))` too, so all create paths converge. Added a cross-path test (UI-create on store A + import resolve on store B → same id).

Known bounded limitation (documented): pre-existing legacy random-id collections aren't retro-actively re-id'd (forward-only), and the sidebar-membership facet of #1703 is tracked separately.

`nodespace-core` lib suite 961 green; clippy + fmt clean; pre-push gate (frontend + Rust + e2e) passed. **APPROVE.**

Issue #338, opened 2025-10-23
Add index-based child operations and fix sibling chain integrity


malibio:
# Add Index-Based Child Operations and Fix Sibling Chain Integrity

## Overview

This issue addresses critical gaps in the MCP API for hierarchical node manipulation and fixes a data integrity bug in node deletion. We need to add intuitive index-based operations for child node management and ensure sibling chains remain intact during deletions.

### Current Limitations

1. **No hierarchy visualization** - Agents can't see parent-child relationships or sibling ordering
2. **No intuitive insertion** - Agents must understand linked-list pointers (`before_sibling_id`)
3. **No deletion preview** - Agents don't know what children will be cascade-deleted
4. **Broken sibling chains** - Deleting a node leaves dangling `before_sibling_id` references (DATA INTEGRITY BUG 🔴)

### Current Data Structure

NodeSpace uses a **linked-list structure** for siblings:
```
Node 1 (before_sibling_id = None)        ← First child
Node 2 (before_sibling_id = "node-1")    ← Points to previous sibling
Node 3 (before_sibling_id = "node-2")    ← Points to previous sibling
```

This is efficient for the database but unintuitive for AI agents who think in array indexes.

### Problem 1: Sibling Chain Integrity Bug 🔴 CRITICAL

**Current behavior:**
```
Before deletion:
- uuid-1 (before_sibling_id=None)
- uuid-2 (before_sibling_id="uuid-1")  ← Deleting this
- uuid-3 (before_sibling_id="uuid-2")

After deletion:
- uuid-1 (before_sibling_id=None)
- uuid-3 (before_sibling_id="uuid-2")  ← BROKEN! Points to deleted node
```

**Impact:**
- Orphaned references to deleted nodes
- Broken tree traversal
- Incorrect sibling ordering in UI
- Potential crashes when dereferencing deleted IDs
- Data corruption

**Root cause:**
`packages/core/src/services/node_service.rs:552-575` - `delete_node()` only does:
```rust
conn.execute("DELETE FROM nodes WHERE id = ?", [id]).await
```

It doesn't update the next sibling's `before_sibling_id` to maintain chain integrity.

### Problem 2: No Index-Based Operations

AI agents naturally think: "Insert at position 3" or "Move to first position"

**Current workaround (cumbersome):**
```javascript
// Agent wants: "Insert at position 3"
// Must do:
1. get_markdown_from_node_id(parent) → Get content
2. Infer structure from markdown
[...]

PR #1707, opened 2026-07-20
Add bulk_create_has_child store primitive for the batched edge sweep (sync #345)

mstomar125:
Foundation for nodespace-sync#345 (batch the reconnect edge sweep). The pull's node apply is already batched (sync #338); the `has_child` edge apply is still per-row (`create_parent_edge` → `move_node`, one write each) and is now the dominant cost of a large cold pull (~133 edges/s in a 33k-edge tenant).

## Change
- `SqliteStore::bulk_create_has_child(&[(parent, child, order)])` — insert many `has_child` edges in ONE transaction. Direction matches `move_node`/`get_children` (`in_node = parent, out_node = child`), order stored at `properties.$.order` (which `get_children` sorts by ASC), so a fresh parent's children reproduce the sender's sibling order exactly. **Idempotent**: a child that already has a parent is skipped (a node has at most one parent) — it never re-parents.
- `NodeService::bulk_create_has_child_edges(&[(parent, child, order)])` — thin wrapper that emits one `RelationshipCreated` per created edge (matching the per-row path), coalesced by the caller's `begin_batch_emit` guard.

Intended for the from-scratch (cold) sweep where every parent is fresh, so the sender's `order` values are the final sibling order. Repositioning / non-fresh parents stay on the per-row `create_parent_edge`.

## Tests
`bulk_create_has_child_edges_reproduces_order_and_is_idempotent`: attach children out of order (b=1, a=2, c=3) → `get_children` returns [b,a,c]; a re-run skips already-parented children (0 created, order unchanged). Full `nodespace-core` lib suite (962) green; clippy + fmt clean; pre-push gate passed.

The sync-side integration (batch the cold rel sweep, per-row fallback) follows in nodespace-sync#345.

### DIFF
--- a/packages/core/src/db/sqlite_store/relationships.rs
+++ b/packages/core/src/db/sqlite_store/relationships.rs
@@ -499,6 +499,65 @@ impl SqliteStore {
         .await
     }
 
+    /// Bulk-create `has_child` edges (parent → child) in ONE transaction, for the
+    /// batched reconnect edge sweep (issue #345). Each tuple is `(parent, child,
+    /// order)` where `order` is the sender's sibling order; `get_children` sorts by
+    /// `json_extract(properties, '$.order')` ASC, so a fresh parent's children
+    /// reproduce that order exactly.
+    ///
+    /// Idempotent and safe for a from-scratch sweep: a child that ALREADY has a
+    /// parent `has_child` edge is skipped (a node has at most one parent), so this
+    /// only ever attaches genuinely-unparented children — it never re-parents.
+    /// Direction matches `move_node`/`get_children`: `in_node = parent, out_node =
+    /// child`. Returns the edges actually inserted, so the caller can emit one
+    /// `RelationshipCreated` per edge.
+    pub async fn bulk_create_has_child(
+        &self,
+        edges: &[(String, String, f64)],
+    ) -> Result<Vec<(String, String, f64)>> {
+        if edges.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        let now = Utc::now().to_rfc3339();
+        let tx = self
+            .db
+            .transaction()
+            .await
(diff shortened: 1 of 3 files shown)


mstomar125:
## Adversarial review + resolution

Verified all seven correctness points against `move_node`/`get_children`/`get_parent`, the schema, and the event machinery:
1. **Direction** — `in_node=parent, out_node=child` matches `move_node`'s insert, `get_children` (`JOIN ON out_node=n.id WHERE in_node=?`), and `get_parent` (`WHERE out_node=child`). No flip.
2. **Idempotency / no second parent** — the `WHERE out_node=child AND type='has_child'` skip-check is correctly BROADER than the `UNIQUE(in_node,out_node,type)` index (which alone would NOT catch a same-child-different-parent duplicate). Read-your-writes inside the tx makes it hold within a single batch too.
3. **Order fidelity** — `{"order": f64}` sorted by `json_extract(...'$.order')` ASC; integer-valued f64 compares fine.
4. **Atomicity** — one tx, `created` built in memory and returned only after commit; any error rolls back the whole batch and emits nothing → clean per-row fallback.
5. **FK fallback** — `in_node/out_node` are `NOT NULL REFERENCES node(id)` with `foreign_keys=ON`, so a missing endpoint rolls back the batch → `Err` → per-row fallback (the intended trigger).
6. **Event parity** — same `RelationshipEvent::new` constructor/id-shape/from-to as `create_parent_edge`; byte-identical events.
7. **Insert-only** — skips already-parented children, never deletes/repositions → can't corrupt existing structure.

Addressed the two non-blocking notes (commit): corrected the doc comment (relationship events broadcast immediately, NOT coalesced by `begin_batch_emit` — matches the per-row path), and added a within-batch idempotency test (same child under two parents → attaches to the first only). **APPROVE.**

PR #1709, opened 2026-07-21
Skill installer: prepend Claude Code frontmatter + honor $CLAUDE_CONFIG_DIR (#1708)

mstomar125:
Closes #1708.

## Problem
The shipped `packages/skill/SKILL.md` has no Claude Code frontmatter, so the skill installer's copy isn't surfaced by native skill discovery — a session doesn't find NodeSpace on its own, which is the point of the installer. And `agents.ts` hardcodes `~/.claude`, so a separate profile (e.g. `CLAUDE_CONFIG_DIR=~/.claude-ns`) never gets the skill.

## Change
- **`installer.ts`** — when writing `SKILL.md` for an agent that defines `skillFrontmatter`, prepend it (`frontmatter + body`); everything else is copied verbatim. The shared `SKILL.md` source is unchanged (agent-agnostic; the generated schema-rules block is untouched), so other agents still get it as-is.
- **`agents.ts`** — Claude Code `detectionDir`/`installDir` derive from `$CLAUDE_CONFIG_DIR` when set, else `~/.claude`; adds the Claude Code frontmatter constant (`name` + `description` + `allowed-tools: Bash(nodespace:*)`).
- **`types.ts`** — `AgentConfig.skillFrontmatter?: string`.
- **tests** — assert claude-code SKILL.md gets frontmatter while other agents stay verbatim; assert `$CLAUDE_CONFIG_DIR` is honored (and falls back). Isolates the top-level `AGENTS` import from an inherited `$CLAUDE_CONFIG_DIR` so the suite is deterministic. 20/20 pass.

## Validation
Built + ran the installer against a throwaway `CLAUDE_CONFIG_DIR`: SKILL.md starts with the frontmatter, body preserved, installed into the target profile. End-to-end: a headless Claude Code session, given only a natural request (no CLI in the prompt), discovered the installed skill and drove the `nodespace` CLI unprompted.

Follow-up (out of scope): the `nodespace-hook.ts` shim's `hook()` runtime appears unused by current Claude Code; if confirmed dead it can be removed separately.

### DIFF
--- a/packages/skill/src/agents.ts
+++ b/packages/skill/src/agents.ts
@@ -4,12 +4,36 @@ import type { AgentConfig } from './types.js';
 
 const home = homedir();
 
+// Claude Code reads its config from $CLAUDE_CONFIG_DIR when set (e.g. a separate
+// `claude-ns` profile), falling back to ~/.claude. Detect + install into the
+// active profile rather than always ~/.claude so the skill lands where the
+// running Claude Code will actually look for it.
+const claudeConfigDir = process.env.CLAUDE_CONFIG_DIR?.trim() || join(home, '.claude');
+
+// Claude Code surfaces a skill via its YAML frontmatter (name + description).
+// The shared SKILL.md body is agent-agnostic and carries none, so the installer
+// prepends this Claude-Code-specific frontmatter when writing the skill.
+const CLAUDE_CODE_SKILL_FRONTMATTER = `---
+name: nodespace
+description: >
+  Read and write the NodeSpace knowledge graph — a local-first, synced working
+  memory that persists across sessions and can be shared with a teammate's agent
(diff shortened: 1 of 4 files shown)


mstomar125:
**APPROVE**

Reviewed the diff: frontmatter is prepended only for `SKILL.md` and only for agents that define `skillFrontmatter` (claude-code) — other agents stay verbatim (test asserts both). `$CLAUDE_CONFIG_DIR` is honored with a correct empty/unset fallback to `~/.claude`. Tests isolate the top-level `AGENTS` import from an inherited env value (avoids the false failure + real-profile mutation the gate hit) and cover frontmatter + config-dir behavior; 20/20 pass. The generated schema-rules block and shared `SKILL.md` source are untouched. Pre-push gate (test:all + Rust) green. Merging.

PR #1710, opened 2026-07-21
daemon: expose NodeServiceImpl::node_service() accessor (unblocks Pro multi-DB routing)

mstomar125:
## What
Adds `NodeServiceImpl::node_service(&self) -> Arc<NodeService>` — a public accessor for the underlying `NodeService` the gRPC impl serves.

## Why
Foundation for the Pro daemon multi-database routing work (nodespace-sync#327, ADR-053). When `nodespaced-pro` opens its default database **through** the `DatabaseManager` (so `x-ns-database-id` routing works and `database list` reports it open), cloud-sync must bind to the *same* `NodeService` the router's handlers serve. The manager's bundle (`DatabaseServices`) only exposed `node_service_grpc: NodeServiceImpl`, and its inner `node_service` was private — so cloud-sync couldn't reuse it and would otherwise construct a second `NodeService` on the same file (single-writer contention). This accessor closes that gap.

## Test
Added `node_service_accessor_returns_the_same_underlying_service` — asserts `Arc::ptr_eq` between the accessor's result and the service passed to `new`. Pre-push gate (test:all + Rust) green.

No behavior change for existing callers; purely additive.

### DIFF
--- a/packages/daemon/src/services/node_service.rs
+++ b/packages/daemon/src/services/node_service.rs
@@ -112,6 +112,15 @@ impl NodeServiceImpl {
         self
     }
 
+    /// The underlying `NodeService` this impl serves. Lets a caller that opens a
+    /// database through the registry (e.g. the Pro daemon binding cloud-sync to
+    /// the manager's default database) reuse the *same* `NodeService` the gRPC
+    /// handlers serve, rather than constructing a second one on the same file
+    /// (which would contend on the single-writer lock).
+    pub fn node_service(&self) -> Arc<CoreNodeService> {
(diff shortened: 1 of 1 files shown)


mstomar125:
**APPROVE** — purely additive accessor: returns `self.node_service.clone()` (an `Arc` clone, no behavior change for existing callers). Test asserts `Arc::ptr_eq` so it genuinely hands back the same allocation — which is the whole point (Pro daemon reuses the manager's default NodeService instead of double-opening the file). Pre-push gate green. Merging.

Issue #350, opened 2025-10-24
Fix create_nodes_from_markdown hierarchy: Same-level headers incorrectly nested


malibio:
## Overview

The `create_nodes_from_markdown` MCP method incorrectly nests same-level headers under each other instead of making them siblings under their parent. For example, multiple H2 headers should all be children of the H1, but currently H2[2] is a child of H2[1], H2[3] is a child of H2[2], etc.

## Current (Incorrect) Structure

From database after MCP import:

```
Container: # Product Launch Campaign Strategy (H1)
  id: 482fbcf2-916b-48ec-b8db-37cc8a383196
  parent_id: NULL
  container_node_id: NULL

├─ text: "This comprehensive product launch..."
│  parent_id: 482fbcf2... (CORRECT - child of container)
│
├─ ## Launch Overview (H2)
│  id: 7839aca9...
│  parent_id: 482fbcf2... (CORRECT - child of H1)
│
└─ ## Executive Summary (H2) ❌ WRONG
   id: 25be8813...
   parent_id: 7839aca9... (WRONG - should be 482fbcf2, not previous H2!)
   
   └─ ## Target Audience Analysis (H2) ❌ WRONG  
      id: 08f0b97c...
      parent_id: 25be8813... (WRONG - should be 482fbcf2!)
      
      └─ ### Primary Target Segment (H3) ✅ CORRECT
         parent_id: 08f0b97c... (CORRECT - child of H2)
```

## Problem

**Same-level headers are being nested sequentially instead of as siblings:**
- H2[1] → child of H1 ✅
- H2[2] → child of H2[1] ❌ (should be child of H1)
- H2[3] → child of H2[2] ❌ (should be child of H1)

**Expected behavior:**
- All H2 headers should be siblings, all children of the H1 parent
- H3 headers should be children of their preceding H2
- H4 headers should be children of their preceding H3

## Expected (Correct) Structure

```
Container: # Product Launch Campaign Strategy (H1)
├─ text: "This comprehensive..."
├─ ## Launch Overview (H2)
│  ├─ text: (any content under Launch Overview)
│  └─ ### Subsection (H3) - if exists
├─ ## Executive Summary (H2)  
│  ├─ text: (any content under Executive Summary)
│  └─ ### Subsection (H3) - if exists
└─ ## Target Audience Analysis (H2)
   ├─ text: (any content)
   └─ ### Primary Target Segment (H3)
      └─ text: demographic info
```

## Investigation Required

[...]

PR #1714, opened 2026-07-22
Re-target Pro cloud-sync when switching the active database (ADR-053)

mstomar125:
Closes #1711. Part of the per-database cloud-sync epic (#1535). The daemon-side single-active sync + `ActivateDatabase` RPC are done in nodespace-sync (#350–#358, merged).

> ⚠️ **HOLD — do not merge yet.** Per the author's request this stays open for **on-device validation** first (the switch → sync-retarget → pill behavior needs the running Tauri app). The change is unit-tested + reviewed; only the interactive end-to-end check remains.

## What
`databaseStore.switchTo(id)` re-points the routed gRPC clients via `set_active_database` (the data plane) but never told the Pro daemon to move its cloud-sync session — so switching database in the app changed which database is read/written while the daemon kept syncing the boot tenant. This wires the switch to the daemon's single-active `ActivateDatabase` RPC.

- Mirror the `ActivateDatabase` RPC + messages into the desktop's pro proto copy.
- Add `pro_activate_database(database_id)` Tauri command → `CloudSyncService.ActivateDatabase`; **no-ops in community mode** (no `ProClient`), matching the other `pro_*` commands, so the community DB switcher is unaffected.
- `switchTo` calls it after `set_active_database`, **best-effort**: a re-target failure is logged and does not abort the already-committed routing switch.

Desktop-only.

## Test plan
- `bun run test` — the database-store tests assert the switch re-targets sync (`pro_activate_database` invoked with `{ databaseId }`) **and** still completes when the re-target fails. `nodespace-app` compiles (proto regenerates). Pre-push gate (`test:all` + `cargo build` + `test:e2e`) green.
- **On-device (pending, the reason this is held):** with a Pro daemon + ≥2 bound databases, switching database in the app re-targets sync to the new tenant (Realtime/pill follows). The backend RPC is already headless-validated in nodespace-sync.

## Review
Independently reviewed (adversarial): proto parity byte-for-byte vs the daemon; command no-ops in community; `switchTo` wiring is genuinely best-effort; snake/camel arg convention correct; frontend tests cover both paths. No defects found.

## Scope
The tenant/database **picker UI** + per-DB sync pill (builds on #1536/#1537) are separate — this is only the switch→sync wiring.

### DIFF
--- a/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
+++ b/packages/desktop-app/src-tauri/proto/nodespace_pro.proto
@@ -45,6 +45,13 @@ service CloudSyncService {
   // can advance from the enable prompt to sign-in. Idempotent.
   rpc EnableSync (EnableSyncRequest) returns (EnableSyncResponse);
 
+  // Switch the cloud-sync session to follow a different local database (ADR-053
+  // single-active sync). The desktop calls this from its switch-database action,
+  // right after re-pointing routing, so switching database also switches which
+  // tenant syncs — not just which one is read/written. An empty database_id
+  // deactivates (local-only). Idempotent for the already-active database.
+  rpc ActivateDatabase (ActivateDatabaseRequest) returns (ActivateDatabaseResponse);
+
   // --- Team membership (M5, #147). Thin adapters over the cloud
   // membership RPCs; the daemon forwards the signed-in user's JWT, so
   // the same admin / last-admin / open-vs-restricted gates apply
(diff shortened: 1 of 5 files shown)


mstomar125:
**APPROVE**

Independent adversarial review — all 5 verification points pass, nothing merge-blocking:
- Community-mode safe (`try_state::<ProClient>()` None → `Ok(())`, no panic/error).
- Best-effort in `switchTo`: try/catch after the supersede guard and `activeDatabaseId = id`, so a superseded switch never re-targets and a failure doesn't abort the committed routing switch.
- Proto is byte-for-byte wire-compatible with the daemon proto (field numbers/names match); empty `database_id` = deactivate documented both sides.
- Command registered in the Tauri handler list.
- Test arg-name `{ databaseId: 'b' }` matches Tauri v2 camelCase→snake_case (established convention).

One low-severity, non-blocking note: the new `await invoke('pro_activate_database')` adds a suspension point to a previously-synchronous switch tail with no `seq !== switchSeq` re-check after it — two rapid switches could interleave their activate RPCs briefly. Self-heals on the next switch (daemon ActivateDatabase is single-active + idempotent); worth a one-line re-check if tightened later. Device-validated both directions.

PR #1724, opened 2026-07-23
Collections appear empty: content-aware counts + reliable multi-collection membership

mstomar125:
## What
Fixes "collections appear empty" and makes multi-collection membership replicable (core + daemon half). Closes #1723.

## Changes
- **Content-aware `member_count`** (`get_all_collections_with_member_counts`): counts only content members (excludes person/schema/database-settings/collection/horizontal-line), so the sidebar's count and empty-collection pruning match what the viewer lists — a roster-only collection is no longer a visible click-to-empty.
- **`bulk_add_to_collections`** now returns the created edges; new **`NodeService::bulk_add_to_collections_notify`** emits a `RelationshipCreated` per edge (same push path as `create_relationship`, throttled so a bulk import's burst doesn't overflow the event broadcast). The batch importer routes through it, so imported secondary memberships now reach cloud.
- **`DOMAIN_EVENT_CHANNEL_CAPACITY` 128 → 4096** so a bulk import's edge burst isn't dropped.
- **`SqliteStore::get_multi_membership_edges()`** enumerates secondary (multi-collection) memberships for the cloud-sync membership sweep (excludes `person` — RBAC state is server-managed).

## Testing
`cargo test -p nodespace-core` + `-p nodespace-daemon` green, incl. new: content-count exclusion, event-emission, `get_multi_membership_edges`, and an import self-heal (a present-but-orphaned doc regains its topic membership on a plain re-import). Pre-push gate (test:all + build + e2e) passed.

## Related
Pairs with nodespace-sync#365 (member_of push, merged) and the membership sweep (nodespace-sync#363, which consumes `get_multi_membership_edges`). Follow-ups: nodespace-sync#362, nodespace-cloud#87.

### DIFF
--- a/packages/core/src/db/sqlite_store/relationships.rs
+++ b/packages/core/src/db/sqlite_store/relationships.rs
@@ -306,6 +306,46 @@ impl SqliteStore {
         Ok(ids)
     }
 
+    /// Every `member_of` edge belonging to a node that is a member of MORE THAN
+    /// ONE collection, as `(member_id, collection_id)` pairs. A node's first
+    /// membership rides its atomic cloud node insert, so only these SECONDARY
+    /// memberships need a separate push; the cloud-sync membership sweep uses this
+    /// to converge them. Bounding to multi-membership nodes keeps the sweep cheap
+    /// — the single-membership majority (already covered atomically) is never
+    /// enumerated. Idempotent re-push of the one atomic-covered edge among a
+    /// multi-membership node's edges is a benign no-op on the cloud side.
+    ///
+    /// `person` members are excluded: their collection memberships are RBAC state
+    /// managed server-side by the membership RPCs (invite / set_member) and are
+    /// not the sweep's concern — re-asserting them would bypass those gates and
+    /// waste round-trips. Content and nested-collection memberships are kept.
+    pub async fn get_multi_membership_edges(&self) -> Result<Vec<(String, String)>> {
+        let mut rows = self
+            .db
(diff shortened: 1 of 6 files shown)


mstomar125:
**APPROVE** (after fixing the one blocker an adversarial review caught)

Independent review found a real blocker and it's now fixed:
- **Blocker (fixed):** the capacity bump landed on the wrong constant. There are two `DOMAIN_EVENT_CHANNEL_CAPACITY` — the `SqliteStore` one (inert: the store notifies via a `notifier` closure, nothing ever `event_tx.send`s or subscribes to it) and the `NodeService` one (the live channel `emit_event` writes and every consumer, incl. the cloud-sync push consumer, subscribes to). I'd bumped the dead one. Now: `NodeService` constant → 4096 (live), `SqliteStore` reverted to 128 (dropped from the diff). Added a guard test `bulk_membership_burst_does_not_overflow_the_event_channel` that emits a 300-edge burst and asserts a subscriber receives all with no `Lagged` — it fails at 128, passes at 4096.

Everything else verified correct by the review: content-count SQL exactly matches the frontend `NON_CONTENT_NODE_TYPES`; the `bulk_add_to_collections` return-type change has no stale callers; event direction matches `create_relationship`; `get_multi_membership_edges` correctly bounds to multi-membership non-person nodes; no `member_count` consumer relied on all-members semantics.

Pre-push gate (test:all + build + e2e) green on the amended commit.

Issue #1720, opened 2026-07-23
[design][ADR-053] Multiple databases per tenant (future — do not implement yet)


mstomar125:
**Status: design note only — not scheduled.** Flag to @Samartha-J-alt / Anoop before implementing (data-model + sync-engine change).

## Idea
Allow one cloud **tenant** to hold **multiple local databases**, instead of the current 1:1. This reverses the one-tenant-per-database invariant (#354, ADR-053 §4), which exists to stop two local databases from double-pushing the same tenant and defeating echo prevention. So this is a sync-engine change, not a UI tweak: each database under a tenant would need to sync a **scoped subset** (e.g. a single collection) so they don't collide.

## Model to design (per product direction)
- **Local databases prevail.** A user can always create a purely **local-only** database — bound to no tenant — that **never syncs**, for private/unsynced work. This mode is first-class and must stay simple.
- **Sync is opt-in per database.** When creating a database that could sync a tenant, the user chooses at creation: **keep it local**, or **sync with the tenant**. If left local, **no node under that database ever syncs, in any case** — local always wins.
- Under a tenant, multiple *synced* databases must be non-overlapping (per-collection or per-subset scoping) to preserve echo prevention / avoid double-push.

## Open questions for the design
- Sync scoping unit: whole-tenant today → per-collection? per-subtree? How is a database's scope chosen + enforced server-side (RLS)?
- Billing/identity implications (ADR-053 §5/§6 are per-database today).
- Migration from the current 1:1 registry model.

## Note
The **local-only database** half of this (create an unbound local database that never syncs; opt-in bind-to-tenant later) is already supported by the backend today (#350 unbound = local-only, #352 BindTenant) and is being surfaced in the Settings UI now. Only the *multiple-synced-DBs-per-tenant* part is deferred here.

PR #1725, opened 2026-07-23
DB/tenant Settings UI: switcher→Settings, per-row bound tenant, reload-persist

mstomar125:
## What
The Database/Tenant management UI (ADR-053), plus fixes so switching is clear and sticky. Closes #1722.

## Changes
- **Settings entry** below Favorites in the sidebar (`layout.svelte.ts` nav item + `navigation-sidebar.svelte` handler → shared `open-settings.ts`).
- **Database ≠ Tenant, made explicit.** The top-left database switcher is now a read-only indicator that opens Settings → Database (no inline tenant switching); the Database settings section explains a database is a local store that syncs to one cloud tenant, and shows **per row** whether it "Syncs to tenant · X" or is "Local only". Switching databases happens here, via an explicit **Open** button.
- **Per-row bound tenant** is full-stack: `DatabaseInfo.bound_tenant_schema` added to the proto, populated by the daemon's `database_service.rs`, surfaced through the Tauri `commands/database.rs` DTO and the `database.svelte.ts` store, and rendered in the settings row — so a database and its tenant are never confused.
- **Reload-persist fix**: the active database is remembered in `localStorage` and restored on load, so a refresh no longer snaps back to the default database.

## Testing
`bun run test` (frontend) green, incl. updated `layout.test.ts` (nav now has the Settings item). Daemon/proto compile; per-row tenant verified live (registry rows show their bound tenant).

## Notes / follow-ups
Multiple-databases-per-tenant (design) is tracked in #1720; the local-storage invitations popup gating in #1721. Builds on Slice C (#1714, merged).


mstomar125:
**APPROVE** (after fixing 2 type-gate blockers an independent review caught)

- **Blocker (fixed):** adding `boundTenantSchema` as a required `DatabaseInfo` field left it missing from `IMPLICIT_BROWSER_DATABASE` (database.svelte.ts) and the test factory `db()` (database.test.ts) → 2 `svelte-check` type errors (the pre-push gate runs vitest, which doesn't type-check, so it missed them). Added `boundTenantSchema: null` to both. `bunx svelte-check` now: **0 errors, 0 warnings**. (database.test.ts joins the PR as the necessary type fix.)

Everything else verified correct by the review:
- `bound_tenant_schema` end-to-end: proto field 8 unique; daemon `listing_to_info` reads the registry `Option<String>` (set on bind); Tauri DTO `#[serde(rename_all = "camelCase")]` → `boundTenantSchema`; frontend type + rendering match.
- Reload-persist restores the remembered id only if still registered (else default→first→null); `switchTo` persists; localStorage helpers try/catch-guarded.
- database-switcher rewrite has no dangling refs; `open-settings.ts` dedups the Settings tab; layout.test nav = 4 items ending in `settings`. No `any`/`{@html}`/`console.*`.

Non-blocking notes: `openSettings(category)` doesn't re-navigate an already-open Settings tab to the requested section (pre-existing pattern, works when opened fresh); the open-settings docstring's "File menu shares this" is stale (app-shell keeps its own copy) — cosmetic.

PR #1730, opened 2026-07-24
Verify model artifact SHA-256 at download and load time (closes #1715)

mstomar125:
## Summary

Closes the last unmet leg of the T9 supply-chain bar (ADR-058): the **build-time** model download did `curl -L` with **no SHA-256 verification**, so a substituted `nomic-embed-text-v1.5.Q8_0.gguf` (compromised upstream, redirect hijack, MITM on a CI runner, malicious mirror) would be code-signed and notarized as authentic, then parsed by native llama.cpp inside the daemon with full authority. This adds digest verification at **both** the download and the load site.

The runtime chat-model download (`model_manager.rs`) was already hardened (#1543); this is the separate build/bundle path it never touched, plus the companion load-time check.

## Changes

**Build-time — `scripts/download-models.ts`**
- Pin the HuggingFace commit in the URL (`resolve/0188c9bf…/`) so `main` can't move under us, and pin `MODEL_SHA256`.
- `curl -fL --proto =https` (fail-hard on HTTP errors, HTTPS-only). The digest is the authoritative gate regardless of redirect target.
- Streaming `sha256File()` (Bun.CryptoHasher) verifies the artifact **after** download; on mismatch it **deletes the file and throws** — never bundles an unverified model.
- Also verifies an already-present file; a stale/mismatched one is re-downloaded rather than trusted.

**Load-time — `packages/nlp-engine/`**
- `config.rs`: pinned `EMBEDDING_MODEL_SHA256` (single source of truth, kept in lockstep with the script), a streaming `compute_file_sha256()` (64 KB chunks — a 146 MB GGUF never buffers whole), and `verify_model_integrity()`.
- `embedding.rs`: right after the model path resolves and **before** `LlamaModel::load_from_file`, verify the on-disk bytes; refuse to load on mismatch. This closes the post-install tamper window — a model swapped on disk after install is rejected.
- `error.rs`: new `EmbeddingError::IntegrityError` so the failure is legible (not folded into a generic load error).
- `Cargo.toml`: `sha2 = "0.10"`.

**Digest rotation** is a single-change contract: rotating the model updates the commit + digest in the script AND the `EMBEDDING_MODEL_SHA256` const together (called out in both comments).

## Acceptance criteria

- [x] Pin an expected SHA-256 next to `MODEL_URL` + a matching constant at the Rust load site
- [x] After download, compute the digest and fail hard + delete the file on mismatch
- [x] Verify the on-disk model against the pinned digest at load time; refuse to load on mismatch
- [x] Download over HTTPS with normal cert validation (digest remains authoritative)
- [x] Rotating the model updates the pinned digest in the same change

## Verification

- Authoritative HF LFS digest for the pinned commit == the pinned constant (`3e24342164b3d94991ba9692fdc0dd08e3fd7362e0aacc396a9a5c54a544c3b7`); the shipped local model matches it.
- Ran the script live: verifies the real 146 MB model in ~0.2 s → "already downloaded and verified".
- New Rust unit tests (all green): NIST `sha256("abc")` vector, mismatch rejection, unreadable-file error, digest-format invariant.
- `cargo clippy -p nodespace-nlp-engine` clean; `cargo test --lib` green. No fake-model test exists, and integration tests skip when the model is absent / use the real default config, so the load gate introduces no test breakage.

Closes #1715.

### DIFF
--- a/packages/nlp-engine/Cargo.toml
+++ b/packages/nlp-engine/Cargo.toml
@@ -26,6 +26,9 @@ sysinfo = { version = "0.33", optional = true }
 # Caching layer
 lru = "0.12"
 
+# SHA-256 for load-time model-integrity verification (ADR-058)
+sha2 = "0.10"
+
 # Workspace dependencies
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/packages/nlp-engine/src/config.rs
+++ b/packages/nlp-engine/src/config.rs
@@ -1,10 +1,63 @@
 /// Configuration for the embedding service using llama.cpp
 use serde::{Deserialize, Serialize};
-use std::path::PathBuf;
+use sha2::{Digest, Sha256};
+use std::path::{Path, PathBuf};
 
 /// Offload all model layers to GPU. This is the llama.cpp convention
 /// where any value >= total layers offloads everything.
 pub const GPU_OFFLOAD_ALL_LAYERS: u32 = 99;
+
+/// SHA-256 of the pinned embedding model `nomic-embed-text-v1.5.Q8_0.gguf`
+/// (HuggingFace `nomic-ai/nomic-embed-text-v1.5-GGUF`, commit
+/// `0188c9bf409793f810680a5a431e7b899c46104c`), lowercase hex.
+///
+/// This is the load-time half of the model-integrity gate (ADR-058); the
+/// build-time download in `scripts/download-models.ts` pins the SAME digest.
+/// Rotating the model MUST update both constants in the same change.
+pub const EMBEDDING_MODEL_SHA256: &str =
+    "3e24342164b3d94991ba9692fdc0dd08e3fd7362e0aacc396a9a5c54a544c3b7";
+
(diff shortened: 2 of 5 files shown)


mstomar125:
**APPROVE**

Independent adversarial review (all 7 scrutiny points) returned CONFIRMED-GOOD: the load-time gate runs before `load_from_file` and is unbypassable (only two `load_from_file` sites in nlp-engine; the chat one is a separate model, tracked in #1729); Bun streaming + `.digest("hex")` matches Rust `format!("{:x}", …)` lowercase; fail-hard + delete is correct; 64 KB stack buffer + sha2 0.10 usage safe; non-optional `sha2` dep is correct. Pinned digest verified against the authoritative HF LFS metadata; script verified the real 146 MB model live. Tests green, clippy clean, no test breakage. The one behavior-change note (only pinned Q8_0 bytes load; `.f16`/override paths resolve but are rejected) is deliberate and now documented on `resolve_model_path`.

Issue #70, opened 2025-08-15
Phase 1.1: Implement ContentProcessor Service


malibio:
## Parent Issue
Part of Epic #69: Text Editor Architecture Refactor - Phase 1

## Summary
Extract content processing logic into a dedicated service to separate concerns and improve maintainability.

## Current Problem
Content processing is scattered across TextNode and MinimalBaseNode components, mixing UI logic with business logic.

## Proposed Solution
Create a dedicated ContentProcessor service that handles:

### Core Functionality
- Markdown to HTML conversion for display
- HTML to Markdown conversion for storage  
- Header level parsing and syntax stripping
- Content validation and sanitization

### Service Interface
```typescript
interface ContentProcessor {
  // Markdown processing
  markdownToDisplay(markdown: string): string;
  displayToMarkdown(html: string): string;
  
  // Content validation
  validateContent(content: string): ValidationResult;
  sanitizeContent(content: string): string;
  
  // Header detection
  parseHeaderLevel(content: string): number;
  stripHeaderSyntax(content: string): string;
}
```

## Implementation Tasks
- [ ] Create ContentProcessor service interface
- [ ] Implement markdown to HTML conversion
- [ ] Implement HTML to markdown conversion  
- [ ] Add header level parsing logic
- [ ] Add content validation and sanitization
- [ ] Create comprehensive unit tests
- [ ] Update TextNode to use ContentProcessor
- [ ] Remove content processing logic from components

## Acceptance Criteria
- [ ] ContentProcessor service passes all unit tests
- [ ] TextNode uses ContentProcessor instead of inline logic
- [ ] No content processing logic remains in UI components
- [ ] Markdown conversion maintains formatting fidelity
- [ ] Header detection works for H1-H6 levels
- [ ] Content validation prevents XSS and malformed content

## Dependencies
None - can be implemented independently

## Estimated Effort
3-4 days

## Definition of Done
- [ ] Service implemented with full test coverage
- [ ] TextNode refactored to use service
- [ ] All existing functionality preserved
- [ ] Documentation updated
- [ ] Code review completed

PR #1731, opened 2026-07-24
Protect ai-chat nodes from Backspace deletion/merge (closes #1697)

mstomar125:
## Summary

Fixes a data-loss bug: pressing Backspace at the start of an ai-chat node deleted or merged it away exactly like a plain text node, silently discarding the entire conversation. An ai-chat's messages live in `properties.messages` (ADR-028), separate from `.content`, so `combineNodes()` — which only concatenates `.content` strings and deletes the current node — destroyed the whole conversation with no protection or confirmation.

The existing `acceptsContentMerge` flag only guards the **previous** (target) node from *receiving* a merge; nothing guarded a node from being deleted/merged away as the **current** node.

## Change

Adds a new plugin flag `deletableViaBackspace`, mirroring `acceptsContentMerge` exactly:

- **`types.ts`** — optional `deletableViaBackspace?: boolean` on `PluginDefinition`, default `true` (existing types unaffected).
- **`plugin-registry.ts`** — `deletableViaBackspace(nodeType)` returning `true` for unknown/disabled/unspecified, `false` only when a registered+enabled plugin opts out.
- **`core-plugins.ts`** — `deletableViaBackspace: false` on the ai-chat plugin. Also sets `acceptsContentMerge: false` (like code-block/quote-block) so a node *below* an ai-chat can't Backspace-merge its text into ai-chat's `.content` and consume itself — the symmetric guard surfaced in review.
- **`base-node-viewer.svelte`** — both `handleCombineWithPrevious` and `handleDeleteNode` now check the flag against the **current/event** node's own type (`nodeManager.nodes.get(eventNodeId).nodeType`) right after the existence check, and no-op when it's `false`. (Previously the only type check was against the *previous* node.)

## Acceptance criteria

- [x] Backspace at the start of an empty ai-chat node does not delete it
- [x] Backspace at the start of a non-empty ai-chat node does not merge/delete it
- [x] Other node types' Backspace behavior unchanged (default-true; full suite green)
- [x] Flag documented consistently with `acceptsContentMerge` (JSDoc in-code + nodespace-docs PR #70)
- [x] `bun run test` passes
- [x] Passes lint/format

## Verification

- New unit tests mirror the existing `acceptsContentMerge` registry block: `plugin-registry.test.ts` (+5 — false/true/default/unknown/disabled) and `core-plugins.test.ts` (+1 asserting the real ai-chat plugin sets the flag). All green.
- Full frontend suite: **167 files, 4115 tests passing**; `svelte-check`: **0 errors / 1769 files**.
- Docs: `nodespace-docs` PR #70 adds the flag to `creating-node-types.md` alongside `acceptsContentMerge`.
- Independent adversarial review: CONFIRMED-GOOD. Verified the two guarded handlers are the *only* Backspace-at-start delete/merge paths (`MergeNodesCommand` → `deleteNode`/`combineWithPrevious`, no Cmd+Backspace or select-all bypass); `nodeType` is the correct field on the `Map<string, Node>`; default-true preserves all other types; and ai-chat reaches these handlers via node-row's fallback `BaseNode` branch (it has a viewer but no inline node component). The companion `acceptsContentMerge: false` was added per the review's secondary finding.

Closes #1697.

### DIFF
--- a/packages/desktop-app/src/lib/design/components/base-node-viewer.svelte
+++ b/packages/desktop-app/src/lib/design/components/base-node-viewer.svelte
@@ -453,7 +453,6 @@
       if (allNodes.length === 0) {
         // No persisted children - create initial placeholder if needed
         // Note: We already checked cache/DB above, so if allNodes is empty, no persisted children exist
-
         // No children at all - placeholder will be created automatically by viewerPlaceholder derived
         // Removed manual ID creation - getOrCreatePlaceholderId() handles it lazily
         // Focus is handled by BaseNode's onMount when autoFocus=true
@@ -819,6 +818,14 @@
         return;
       }
 
+      // Protect node types whose data lives outside `.content` (e.g. ai-chat,
+      // whose conversation is in properties.messages) from being merged away by
+      // a start-of-node Backspace. Guards the current/event node's own type.
+      const eventNode = nodeManager.nodes.get(eventNodeId);
+      if (eventNode && !pluginRegistry.deletableViaBackspace(eventNode.nodeType)) {
+        return;
+      }
+
       const currentVisibleNodes = visibleNodesFromStores;
       const currentIndex = currentVisibleNodes.findIndex((n) => n.id === eventNodeId);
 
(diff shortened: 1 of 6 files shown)


mstomar125:
**APPROVE**

Independent adversarial review returned CONFIRMED-GOOD across all 6 scrutiny points: the two guarded handlers are the *only* Backspace-at-start delete/merge paths (`MergeNodesCommand` → `deleteNode`/`combineWithPrevious`; no Cmd+Backspace or select-all bypass — `canExecute` is false with a selection); `nodeType` is the correct field on `Map<string, Node>`; the guard no-ops safely because `execute()` already called `preventDefault()` (cursor stays at 0, command still returns handled); default-true leaves all other node types unchanged; and ai-chat reaches these handlers via node-row's fallback `BaseNode` branch (it has a viewer but no inline node component). The review's secondary finding — a node *below* ai-chat could merge its text into ai-chat's `.content` — is closed by the companion `acceptsContentMerge: false`. Full frontend suite 4115/4115, svelte-check 0 errors.

PR #1732, opened 2026-07-24
Flatten typed properties in optimistic updateNode merge (closes #1695)

### LINKED ISSUE #1695: Optimistic updateNode merge doesn't flatten properties.* into top-level fields, causing ai-chat UI to lag behind RPC round-trip
## Overview

In an ai-chat node, two visible symptoms share one root cause: the optimistic (pre-round-trip) merge in `SharedNodeStore.updateNode()` never flattens namespaced `properties.*` writes into the top-level fields components actually read. The UI only becomes correct once the backend RPC round-trip completes and the response is spread onto local state — so anything gated on those top-level fields appears to "hang" for the duration of one network/IPC round trip.

## Problem Statement

**Symptom 1 — model-selection textarea delay:**
After starting a new ai-chat node and selecting a model, the textarea for typing a message takes a long time to appear.

**Symptom 2 — sent message appears late:**
After sending a message, it doesn't show up in the conversation until after the AI agent's response comes back — both appear to land together.

**Mechanism (both symptoms):**

`SharedNodeStore.updateNode()` (`packages/desktop-app/src/lib/services/shared-node-store.svelte.ts`, optimistic merge ~lines 990-996) applies the synchronous "optimistic" local update as a shallow `{ ...existingNode, ...changes }`. When `changes` is a nested/namespaced shape like `{ properties: { model: ... } }` or `{ properties: { messages: [...] } }`, the shallow merge only updates `node.properties.*` — it does not flatten those values into the top-level fields (`node.model`, `node.provider`, `node.messages`) that `AiChatNode` (`packages/desktop-app/src/lib/types/ai-chat-node.ts`) and the viewer read.

The top-level fields are only corrected later, once `backendAdapter.updateNode()`'s RPC resolves and the backend's already-flattened response (produced server-side via `node_to_typed_value`/`flatten_properties_for_api`) is spread onto the local node (~lines 1192-1211 and ~1601-1640 of the same file).

Concretely:
- `handleModelSelect` (`ai-chat-node-viewer.svelte:144-200`) writes `properties.provider`/`properties.model`. The textarea's visibility gate (`ai-chat-node-viewer.svelte:432`, `isMessageProvider && !model`) reads the top-level `node.model`, which stays stale until the RPC resolves.
- `handleSend` (`ai-chat-node-viewer.svelte:259-268`) writes `properties.messages`. The rendered message list (`persistedMessages`, `ai-chat-node-viewer.svelte:103-116`) reads the top-level `node.messages`, which also stays stale until the same round trip completes — so the user's own message and the agent's reply both become visible in the same post-round-trip window, making it look like the message "waited" for the response.

## Relation to Prior Work

This is directly adjacent to (but not fixed by) PR #1694 / issue #1690, "Content-only edits forced immediate persistence via redundant nodeType, and property-only updates never synced typed fields back from the backend." That PR fixed this exact class of bug — `properties.*` writes not syncing to top-level typed fields — in the **post-response reconciliation** path (the `Object.assign` applied after the backend RPC returns). It did not touch the earlier, purely local **optimistic merge** step, which has the identical shallow-merge gap one layer earlier. #1694 turned "field never becomes correct" into "field becomes correct only after a round trip" — this issue is about closing that remaining round-trip-gated gap.

ADR-049 (`decisions/049-canonical-reactivity-store-pattern.md`) independently documents that `SharedNodeStore` currently "fakes reactivity" and explicitly defers "separating the three hand-synced concerns (data, version/OCC, persistence coordination)" as future work this ADR makes possible but does not itself perform. This issue is a concrete instance of that deferred separation, not a reversal of ADR-049 — no companion `nodespace-docs` change is required, since ADR-049 already anticipates this class of follow-up work.

## Proposed Solution

Make the optimistic merge in `SharedNodeStore.updateNode()` flatten known typed properties into their top-level fields synchronously, at write time, instead of waiting for the backend's flattened response. This likely means factoring out (or reusing, if one already exists client-side) a small flattening step equivalent to what `node_to_typed_value`/`flatten_properties_for_api` do server-side, and applying it to `changes.properties` during the optimistic merge before it's assigned into the local node.

## Acceptance Criteria

- [ ] Selecting a model on an ai-chat node makes the message textarea appear immediately (no visible delay pending an RPC round trip)
- [ ] Sending a message makes it appear immediately in the conversation view, before the agent's response arrives
- [ ] `SharedNodeStore.updateNode()`'s optimistic merge flattens `properties.*` changes into corresponding top-level fields for all typed node kinds that rely on top-level reads (not just ai-chat)
- [ ] No regression in the post-response reconciliation path fixed by #1694/#1690 (backend-authoritative values still win once the RPC resolves)

### Reference Files

- **shared-node-store.svelte.ts**: `packages/desktop-app/src/lib/services/shared-node-store.svelte.ts` — optimistic merge (~lines 990-996), post-response reconciliation (~lines 1192-1211, ~1601-1640)
- **ai-chat-node.ts**: `packages/desktop-app/src/lib/types/ai-chat-node.ts` — top-level typed fields (`model`, `provider`, `messages`) read by the viewer
- **ai-chat-node-viewer.svelte**: `packages/desktop-app/src/lib/components/viewers/ai-chat-node-viewer.svelte` — `handleModelSelect` (144-200), textarea visibility gate (432), `handleSend` (259-268), `persistedMessages` (103-116)

### Implementation Details

Investigate whether the server-side flattening logic (`node_to_typed_value`/`flatten_properties_for_api`) has a client-side equivalent already, or whether one needs to be introduced. Whatever flattening the optimistic path applies must stay consistent with what the backend produces, so the post-response reconciliation doesn't fight the optimistic value (avoid introducing a second, diverging flattening implementation).

### Testing Requirements

Add coverage for `SharedNodeStore.updateNode()` asserting that a `properties.*`-shaped change is reflected in the corresponding top-level field synchronously, before any backend response is awaited (i.e., testable without mocking network timing).

## Related Issues

- #1690 (prior fix to the post-response reconciliation path for this same class of bug)

mstomar125:
## Summary

Fixes an ai-chat UI lag: the optimistic (pre-round-trip) merge in `SharedNodeStore.updateNode()` shallow-spread `changes`, so a namespaced `{ properties: {...} }` write (a) replaced `node.properties` wholesale and (b) never flattened `properties.*` into the top-level fields components read. As a result, selecting a model (textarea gated on top-level `node.model`) and sending a message (list reads top-level `node.messages`) both appeared to hang for one network/IPC round trip, until the backend's already-flattened response arrived.

The backend `node_to_typed_value` (`nodespace-types/src/convert.rs`) remains the single flattening authority; this adds a **minimal, optimistic-only** mirror so the pre-round-trip state reflects the same top-level fields the backend would promote. The backend response always overwrites these, so any drift degrades optimistic latency only — never correctness.

## Changes

- **`node-normalize.ts`** — new `OPTIMISTIC_TYPED_FIELDS` map (the exact per-type fields convert.rs promotes: ai-chat `status/provider/model/messages`, task `status/priority/dueDate/assignee/startedAt/completedAt`), plus `deepMergeProperties` (one-level merge + one level into the `[nodeType]` namespace) and `promoteTypedFields` (promotes a field **only if present in this write** — flat `properties.x` or nested `properties.<type>.x` — so an omitted field never clobbers a top-level value with `undefined`).
- **`shared-node-store.svelte.ts`** — the optimistic merge deep-merges `properties` and spreads the promoted top-level fields into the new node.
- **`update-protocol.ts`** — new `UpdateOptions.replaceProperties` (default deep-merge; `true` = use the incoming bag as-is).
- **`reactive-node-service.svelte.ts`** — `updateNodeProperties()` passes `replaceProperties: true` (it already computes the full merged/replaced bag), preserving its `merge: false` = wholesale-replace contract.
- **`schema-field-update.ts` / `schema-property-form.svelte`** — pass `replaceProperties: true` in their old-format migration paths (which deliberately drop flat keys and build the full bag), so the deep-merge default doesn't re-add the dropped keys to the optimistic node (surfaced by review).

## Anti-divergence

The typed-field list is one tiny table colocated in `node-normalize.ts` with a doc comment pointing at `convert.rs`, and is always overwritten by the authoritative backend echo — so it can never cause a *correctness* divergence, only an optimistic-latency one.

## Verification

- New tests: `node-normalize.test.ts` (+8 — deep-merge sibling/namespace/missing-bag; promote present-only/nested/null/no-map/map-alignment) and `shared-node-store.test.ts` (+5 — flat ai-chat promotion, message-send preserves provider/model via the omitted-field guard, deep-merge keeps sibling keys, nested task promotion, non-typed node untouched). All green.
- Caught + fixed a regression: `updateNodeProperties(merge=false)` expects wholesale replace; now flagged via `replaceProperties: true`.
- **Full frontend suite: 166 files, 4091 tests passing**; `svelte-check`: **0 errors / 1768 files**.
- Independent adversarial review: **CONFIRMED-GOOD**. Key finding: the deep-merge changes only the *local optimistic* node — the persisted wire payload is still built from the raw `changes.properties` (`shared-node-store.svelte.ts` capture path), so the backend stays the flattening authority and no persistence/corruption path is introduced; the backend echo reconciles. All `updateNode({ properties })` call sites audited; the two schema-migration sites above were the only ones needing `replaceProperties`. Content typing (`{ content, nodeType }`, no `properties`) short-circuits the new block entirely, so the real hot path is unaffected.

Closes #1695.

### DIFF
--- a/packages/desktop-app/src/lib/components/property-forms/schema-property-form.svelte
+++ b/packages/desktop-app/src/lib/components/property-forms/schema-property-form.svelte
@@ -26,7 +26,12 @@
   import { Input } from '$lib/components/ui/input';
   import { backendAdapter } from '$lib/services/backend-adapter';
   import { sharedNodeStore } from '$lib/services/shared-node-store.svelte';
-  import { type SchemaNode, type SchemaField, type EnumValue, isSchemaNode } from '$lib/types/schema-node';
+  import {
+    type SchemaNode,
+    type SchemaField,
+    type EnumValue,
+    isSchemaNode
+  } from '$lib/types/schema-node';
   import type { Node } from '$lib/types';
   import { parseDate, type DateValue } from '@internationalized/date';
   import { createLogger } from '$lib/utils/logger';
@@ -123,7 +128,10 @@
 
     // For strongly-typed nodes (TaskNode, etc.), check top-level fields first
     // Type-specific fields like status, priority, dueDate are at the top level
-    if (fieldName in node && (node as unknown as Record<string, unknown>)[fieldName] !== undefined) {
+    if (
+      fieldName in node &&
+      (node as unknown as Record<string, unknown>)[fieldName] !== undefined
+    ) {
       return (node as unknown as Record<string, unknown>)[fieldName];
     }
 
@@ -184,7 +192,7 @@
       // Look up label from enum values
       if (statusField && status) {
         const enumValues = getEnumValues(statusField);
-        const enumValue = enumValues.find(ev => ev.value === status);
+        const enumValue = enumValues.find((ev) => ev.value === status);
         statusLabel = enumValue?.label || formatEnumLabel(status);
       }
     }
@@ -249,10 +257,13 @@
       ? { title: evaluateTitleTemplate(schema.titleTemplate, migratedNamespace) || null }
       : {};
 
+    // updatedProperties is the full intended bag (the old-format branch above
+    // deliberately drops flat keys), so replace rather than deep-merge.
     sharedNodeStore.updateNode(
       nodeId,
       { properties: updatedProperties, ...titleUpdate },
-      { type: 'viewer', viewerId: 'schema-property-form' }
+      { type: 'viewer', viewerId: 'schema-property-form' },
+      { replaceProperties: true }
     );
   }
 
@@ -392,7 +403,9 @@
                 <!-- Enum Field → shadcn Select Component (Fixed with bind:value) -->
                 {@const enumValues = getEnumValues(field)}
                 {@const currentEnumValue = getEnumValue(field)}
-                {@const currentLabel = enumValues.find(ev => ev.value === currentEnumValue)?.label || formatEnumLabel(currentEnumValue)}
+                {@const currentLabel =
+                  enumValues.find((ev) => ev.value === currentEnumValue)?.label ||
+                  formatEnumLabel(currentEnumValue)}
                 <Select.Root
                   type="single"
                   value={currentEnumValue}
--- a/packages/desktop-app/src/lib/design/components/schema-field-update.ts
+++ b/packages/desktop-app/src/lib/design/components/schema-field-update.ts
@@ -93,10 +93,14 @@ export function updateSchemaField(
         [targetNode.nodeType]: updatedNamespace
       };
 
-  // Persist via sharedNodeStore
+  // Persist via sharedNodeStore. This builds the full intended properties bag
+  // itself (and the old-format branch deliberately drops flat keys), so replace
+  // rather than deep-merge — otherwise the dropped flat keys would reappear on
+  // the local optimistic node.
   sharedNodeStore.updateNode(
     targetNodeId,
     { properties: updatedProperties },
-    { type: 'viewer', viewerId }
+    { type: 'viewer', viewerId },
+    { replaceProperties: true }
   );
 }
--- a/packages/desktop-app/src/lib/services/node-normalize.ts
+++ b/packages/desktop-app/src/lib/services/node-normalize.ts
@@ -18,3 +18,86 @@ export function normalizeNodeData(nodeData: Node): Node {
   }
   return nodeData;
 }
+
+/**
+ * Optimistic-only mirror of the backend's typed-field promotion
+ * (`node_to_typed_value` / `flatten_properties_for_api` in
+ * `packages/nodespace-types/src/convert.rs`). For each node type, lists the
+ * type-specific fields the backend lifts from the stored `properties` bag up to
+ * the TOP LEVEL of the node (the fields viewers actually read).
+ *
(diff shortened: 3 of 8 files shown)


mstomar125:
**APPROVE**

Independent adversarial review returned CONFIRMED-GOOD. The de-risking finding: the deep-merge changes only the *local optimistic* node — the persisted wire payload is still built from the raw `changes.properties`, so the backend remains the flattening authority and no persistence/corruption path is introduced; the backend echo reconciles. Every `updateNode({ properties })` caller was audited: ai-chat handlers get the intended fix, full-bag builders are no-ops, and the only sites needing `replaceProperties: true` (updateNodeProperties + the two schema old-format migration branches that deliberately drop flat keys) now set it. Undefined-clobber guard, nested-vs-flat handling, array identity, hot-path short-circuit (content typing has no `properties`), and rollback all verified. Full frontend suite 4091/4091, svelte-check 0 errors.

PR #1733, opened 2026-07-24
Exclude ai-chat nodes from embedding/vectorization (closes #1696)

mstomar125:
## Summary

AI-chat conversation nodes were embedded and made semantically searchable. This stops that: ai-chat nodes are never vectorized. Conversations are not general knowledge and must not surface in semantic search.

This reverses the per-type decision in **ADR-029** (embeddable-but-hidden) to **not embeddable at all**. The behavior-owned embedding mechanism itself is unchanged — only ai-chat's `get_embeddable_content()` result changes.

## Changes

- **`behaviors/mod.rs`** — `AiChatNodeBehavior::get_embeddable_content()` now returns `None` unconditionally (was: extract user/assistant text from `properties.messages[]`). Updated the method + struct doc comments. Consolidated the two now-obsolete extraction tests into one `test_ai_chat_node_never_embeddable` (asserts `None` for rich, empty, tool-only, and no-messages shapes); flipped the `get_embeddable_content` matrix assertion to `is_none()`.
- **`embedding_service.rs`** — flipped the ai-chat assertion in `test_behavior_driven_embedding_decision` to `is_none()`; updated its doc comment; added a clarifying comment at `matches_scope` noting `SearchScope::Conversations` (`node_type == "ai-chat"`) is now a no-op (no ai-chat rows exist in the vector index). The enum variant is retained for API stability.
- **`agent/local_agent/tools.rs` + `agent/skill_pipeline.rs`** — removed `conversations` from the `search_nodes` tool's advertised `scope` options and dropped the skill guidance that steered the model to `scope="conversations"` for past chats. Without this, the model would keep calling a scope that now returns nothing and report "searched conversations, found nothing." The core `SearchScope::Conversations` enum variant is retained (documented no-op) for API stability.
- **`nodespace-docs` ADR-029** (committed directly to docs `main` per that repo's workflow) — added a Revision note, updated the per-type behavior table (ai-chat → `None`) and the SearchScope section (Conversations/Everything surface no ai-chat content).
- **`Cargo.lock`** — adds the `sha2` entry for `nodespace-nlp-engine` that was missing (its `Cargo.toml` gained the dep in #1715 but the root lockfile was never committed alongside it, leaving `--locked` builds inconsistent). Drive-by correction.

## Notes

- Per CLAUDE.md's decision tree, embeddability for built-in types is behavior-trait-owned (no schema flag), so this is a Core hardcoded-behavior change.
- Existing ai-chat embeddings are cleaned up by the normal pipeline: `get_embeddable_content()` returning `None` triggers the embedding service's "don't embed me → clean up existing embeddings" path on reprocess (greenfield: no migration needed).
- `NON_EMBEDDABLE_CONTAINER_TYPES` unchanged (ai-chat isn't a container type).

## Verification

- `cargo test -p nodespace-core --lib`: **966 passed, 0 failed**; `cargo clippy -p nodespace-core`: clean.
- Independent adversarial review (crate-wide + other packages): confirmed no test breaks and container-parity holds. It surfaced one real behavioral gap — the agent still steered the LLM to the now-dead `conversations` scope — which this PR fixes (tools.rs + skill_pipeline.rs above).
- `cargo test -p nodespace-agent --lib` (scope + skill tests): green; `cargo clippy -p nodespace-agent`: clean.

Closes #1696.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3862,6 +3862,7 @@ dependencies = [
  "lru",
  "serde",
  "serde_json",
+ "sha2",
  "sysinfo",
  "tempfile",
  "thiserror 1.0.69",
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -407,8 +407,8 @@ fn def_search_semantic() -> ToolDefinition {
                 },
                 "scope": {
                     "type": "string",
-                    "enum": ["knowledge", "conversations", "everything"],
-                    "description": "Search scope: 'knowledge' (default, searches text/header/code/schema nodes), 'conversations' (searches conversation messages), 'everything' (all node types). Use 'conversations' when the user asks about past chats or conversation history."
+                    "enum": ["knowledge", "everything"],
+                    "description": "Search scope: 'knowledge' (default, searches text/header/code/schema nodes) or 'everything' (all embedded node types). Note: ai-chat conversations are not embedded, so they are never returned by search regardless of scope."
                 },
                 "include_archived": {
                     "type": "boolean",
--- a/packages/agent/src/skill_pipeline.rs
+++ b/packages/agent/src/skill_pipeline.rs
@@ -257,7 +257,6 @@ FETCH ADDITIONAL CONTENT: Only call get_node with format=markdown if you need fu
 PARAMETER GUIDANCE:
 - Use 'collection' to narrow search to a namespace/folder (e.g. collection="Architecture").
 - Use 'node_types' to filter by type (e.g. node_types=["task"]) — prefer over 'collection' for type-based filtering.
-- Use 'scope'="conversations" when the user asks about past chats or conversation history.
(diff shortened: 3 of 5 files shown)


mstomar125:
**APPROVE**

Independent adversarial review (crate-wide + agent/daemon/desktop-app): no tests break, container-parity holds (ai-chat's `can_have_children()` is false so it's excluded from `NON_EMBEDDABLE_CONTAINER_TYPES` parity), and the consolidated test compiles + asserts `None` across all message shapes. The review's one real finding — the agent still steered the LLM to the now-dead `conversations` scope (tools.rs / skill_pipeline.rs) — is fixed here by de-advertising it while keeping the core enum as a documented no-op. Existing ai-chat embeddings are cleaned via the normal `get_embeddable_content() == None → delete` path. `cargo test -p nodespace-core --lib` 966/966, agent scope+skill tests green, clippy clean.

PR #1747, opened 2026-07-25
Serialize concurrent sibling reorders to fix fractional-order race (closes #1561)

mstomar125:
## Summary

Fixes a data-corruption race: two concurrent `reorder_node` operations on distinct siblings under the same parent could produce a corrupted final sibling order. `move_node`'s order logic reads the current sibling order values, computes a new fractional-order key in Rust, and writes it back as **separate, non-transactional `await`ed steps** on the shared DB connection. Two concurrent moves interleave at those `await` points, both reading the same stale snapshot and computing overlapping/inconsistent order keys — so the final order reflects only one of the two moves. Version-based OCC doesn't catch it: it guards the moved node's own row, not the read of its siblings' order values.

## Fix

Serialize the read → compute → write-back with an application-level async mutex (`SqliteStore::reorder_lock`), held across the whole `move_node` order section.

**Why not a DB transaction (as the issue proposed):** the store uses a single shared `Arc<libsql::Connection>` for all tasks. A `BEGIN IMMEDIATE` on that connection can't serialize concurrent moves — the second task would hit *"cannot start a transaction within a transaction"* (nested transaction on the same connection), not wait. An app-level `tokio::sync::Mutex` serializes them deterministically regardless of the connection model, and also protects the cross-parent branch's read-compute-write (its `new_order` is likewise computed from an out-of-transaction read). Reorders are user-driven and infrequent, so the added contention is negligible.

## Changes

- `packages/core/src/db/sqlite_store/mod.rs` — new `reorder_lock: Arc<tokio::sync::Mutex<()>>` field + doc comment; initialized in `new()`.
- `packages/core/src/db/sqlite_store/nodes.rs` — `move_node` acquires `reorder_lock` immediately before the sibling-order read and holds it (through any rebalance) until the write-back completes. `rebalance_children_for_parent` does not take the lock, so there is no re-entrant deadlock.

## Verification

- The existing repro test `concurrent_reorders_of_distinct_nodes_all_succeed` (previously flaky ~1-in-15 under full-suite load) — **30/30 consecutive passes** against a `nodespaced` built with this fix.
- New `concurrent_reorders_keep_all_children_with_distinct_order` (multi-threaded, 16 simultaneous reorders): all moves succeed, all children retained, distinct order keys — 30/30 standalone + 3/3 full-suite passes.
- `cargo test -p nodespace-core --lib`: 967 passed; clippy clean.
- Independent adversarial review: **CONFIRMED-GOOD** — verified the guard covers the full read→compute→write across all three write branches (same-parent, cross-parent tx, make-root) and the rebalance; no deadlock (nothing in the guarded region re-takes the lock); `reorder_lock` is the store's only mutex (no lock-ordering hazard); and the test's distinctness check can't false-fail (serialized beginning-moves stay ~1.0 apart).

**Scope (follow-up #1746):** `reorder_lock` serializes `move_node` against `move_node` — the exact #1561 scenario. Review noted other sibling-order write paths (`create_child_node_atomic` append, `get_next_order_for_relationship`, `bulk_create_has_child`, `move_children_to_parent`) still bypass the lock and can race; those are pre-existing and tracked in #1746.

**Honest note on reproduction:** the race is timing/load-dependent and does not reproduce deterministically in isolation (single shared libsql connection tends to complete each task's statements together; the 1-in-15 flake needs full-suite pressure). Verification therefore rests on the fix's construction (the mutex makes the interleave impossible), the existing repro test passing against the fixed daemon, and the concurrency smoke test — not on a failing-then-passing unit reproduction.

Closes #1561.

### DIFF
--- a/packages/core/src/db/sqlite_store/mod.rs
+++ b/packages/core/src/db/sqlite_store/mod.rs
@@ -127,6 +127,16 @@ pub struct SqliteStore {
     event_tx: broadcast::Sender<crate::db::events::EventEnvelope>,
     valid_node_types: HashSet<String>,
     notifier: Option<StoreNotifier>,
+    /// Serializes `move_node`'s sibling-order read → fractional-key compute →
+    /// write-back. That sequence spans several `await`s on the single shared
+    /// `db` connection, so two concurrent moves would otherwise interleave and
+    /// compute new order keys against the same stale sibling snapshot, corrupting
+    /// the final order. A DB `BEGIN IMMEDIATE` can't serialize them here — every
+    /// task shares one connection, so the second would hit "cannot start a
+    /// transaction within a transaction" rather than waiting — so we serialize at
+    /// the application level. Reorders are infrequent (user-driven), so the
+    /// contention cost is negligible.
+    reorder_lock: Arc<tokio::sync::Mutex<()>>,
 }
 
 impl SqliteStore {
@@ -154,6 +164,7 @@ impl SqliteStore {
             event_tx,
             valid_node_types,
             notifier: None,
+            reorder_lock: Arc::new(tokio::sync::Mutex::new(())),
         })
     }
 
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Independent adversarial review returned CONFIRMED-GOOD across all 6 primary checks: the `_reorder_guard` is acquired before the sibling-order SELECT and held (as a local binding) to function return, covering all three write branches (same-parent UPDATE, cross-parent tx, make-root) plus the rebalance; no deadlock (nothing in the guarded region re-takes `reorder_lock` — `rebalance_children_for_parent`, the raw db calls, and `calculate_order` take no lock; the existence/cycle checks run before the guard); `reorder_lock` is a per-store field shared via `Arc<SqliteStore>` so all production callers serialize on it; the "BEGIN IMMEDIATE can't serialize on a shared connection" reasoning is sound; the tokio Mutex across DB awaits is correct and is the store's only mutex (no lock-ordering hazard); and the new test can't false-fail (serialized beginning-moves stay ~1.0 apart). Scope note (other order-write paths bypass the lock) filed as #1746. core lib 967 green, clippy clean, gate incl. the integration repro test against a fixed daemon green.

PR #1749, opened 2026-07-27
Add load-time SHA-256 integrity gate for the chat model (#1729)

mstomar125:
## Summary

The embedding model verifies its GGUF against a pinned SHA-256 before native llama.cpp loads it (ADR-058); the chat/LLM model had no equivalent. Chat models *are* verified at **download** time, but a model correctly downloaded and then **swapped/tampered on disk** was loaded without re-verification — the same post-install tamper window the embedding load gate closes, into C++ with full authority (integrity + memory-safety surface).

## Change

- **`nlp-engine/chat/mod.rs`** — `load_model(model_path, expected_sha256: Option<&str>)`. After the `exists()` check and **before** `backend()` / `load_from_file`, when a digest is present it verifies the on-disk file and returns `ChatError::IntegrityError` on mismatch. `None` is the documented escape hatch for a user-supplied / non-catalog model (loads with a warning — nothing to verify against).
- **`nlp-engine/chat/error.rs`** — new `ChatError::IntegrityError`.
- **`nlp-engine/config.rs`** — extracted a generic `verify_file_sha256(path, expected)` from `verify_model_integrity` (which now delegates to it); comparison is **case-insensitive** so a correctly-pinned digest can't be refused over casing.
- **`agent/model_manager.rs`** — `expected_sha256_for_filename(filename)` resolves the pinned digest from the catalog by GGUF basename.
- **`agent/inference.rs`** — `LlamaChatInferenceEngine::load` resolves the digest from the model file's basename and passes it down, preserving the `agent → nlp-engine` dependency direction (nlp-engine can't import the catalog).

The digest lookup keys off the same `entry.filename` that `model_path()` builds the load path from, so a catalog model's basename always resolves to `Some(digest)` — a wiring-invariant test guards that. The daemon load surface takes a `model_id` (never an arbitrary path), so the `None` branch is unreachable today (future-proofing).

## Acceptance criteria

- [x] Verify the on-disk file against the selected catalog entry's pinned SHA-256 before `load_from_file`, reusing the digests in `model_manager.rs`
- [x] Refuse on mismatch with a clear integrity error (`ChatError::IntegrityError`, mirrors `EmbeddingError::IntegrityError`)
- [x] Handle no-catalog / user-supplied paths explicitly (documented `None` escape hatch: load with a warning)

## Tests

- `verify_file_sha256` accepts match / rejects mismatch
- `expected_sha256_for_filename` resolves a catalog filename / `None` for unknown
- **every catalog model basename resolves to its pinned digest** (guards the load-path wiring)
- `load_model` returns `IntegrityError` on a mismatched digest **before** native load

nlp-engine 99 / agent 301 lib tests pass; clippy clean; full pre-push gate (test:all + `cargo build --bin nodespaced` + test:e2e + `cargo test -p nodespace-app`) green. Independent adversarial review: no bypass / correctness bug / regression.

Closes #1729.

Context: Threat T9, ADR-058 (Model Artifact Integrity Verification). Companion to the embedding-model load gate (#1715).

### DIFF
--- a/packages/agent/src/local_agent/inference.rs
+++ b/packages/agent/src/local_agent/inference.rs
@@ -42,8 +42,17 @@ impl LlamaChatInferenceEngine {
         let engine = ChatEngine::new(config)
             .map_err(|e| InferenceError::Engine(format!("Failed to create ChatEngine: {e}")))?;
 
+        // Look up the pinned digest for this model file so the load-time integrity
+        // gate refuses a swapped-on-disk GGUF before native llama.cpp parses it. A
+        // path outside the catalog (a user-supplied model) resolves to None and
+        // loads without the gate — there is nothing to verify against.
+        let expected_sha256 = std::path::Path::new(model_path)
+            .file_name()
+            .and_then(|f| f.to_str())
+            .and_then(super::model_manager::expected_sha256_for_filename);
+
         engine
-            .load_model(model_path)
+            .load_model(model_path, expected_sha256)
             .map_err(|e| InferenceError::Engine(format!("Failed to load model: {e}")))?;
 
         Ok(Self {
--- a/packages/agent/src/local_agent/model_manager.rs
+++ b/packages/agent/src/local_agent/model_manager.rs
@@ -1010,6 +1010,18 @@ fn find_catalog_entry(model_id: &str) -> Result<&'static CatalogEntry, ModelErro
         .ok_or_else(|| ModelError::NotFound(model_id.to_string()))
 }
 
+/// The pinned SHA-256 for a catalog chat model, looked up by its GGUF file name
+/// (basename). Returns `None` for a file that is not in the catalog — a
+/// user-supplied / custom model with no pinned digest, which the load path then
+/// loads without an integrity gate. Used at load time to hand the expected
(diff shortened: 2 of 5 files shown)


mstomar125:
**APPROVE**

Independent adversarial review — no bypass, correctness bug, or regression:
- The only production chat-GGUF load path (`local_agent_service → LlamaChatInferenceEngine::load → ChatEngine::load_model`) always builds the path from `models_dir.join(entry.filename)`, so the basename is always the catalog filename → digest resolves to `Some` → verified. A swapped catalog model keeps its filename → still refused. Filenames + digests are unique across the catalog (no wrong-digest collision).
- The "rename to a non-catalog name → `None`" path is unreachable: the daemon surface takes a `model_id`, never an arbitrary path.
- Comparison is lowercase-consistent and now case-insensitive; the gate runs before `load_from_file`; the `cfg(not(chat-service))` branch consumes the new arg.
- Signature change has exactly two callers, both updated; embedding gate behavior preserved (delegation).

Folded in the review's two LOW suggestions: case-insensitive digest compare, and a wiring-invariant test (`every_catalog_model_basename_resolves_to_its_pinned_digest`) so a regression that dropped the lookup can't slip through.

PR #1751, opened 2026-07-27
Fix move_children_to_parent order collision and serialize order writes (#1746)

mstomar125:
## Summary

Follow-up to #1561. Two fixes to sibling-order writes:

### 1. `move_children_to_parent` — single-threaded order collision (correctness)
It assigned the moved children fresh orders `1.0, 2.0, 3.0…` **without reading the new parent's existing children**, so moving into a **non-empty** parent collided their `has_child` order keys even without concurrency. It now reads the current max order under `new_parent_id` and **appends** the moved children after it (empty parent → unchanged behavior).

### 2. Bring `create_child_node_atomic` + `move_children_to_parent` under `reorder_lock`
#1561 added `reorder_lock` (a non-reentrant `tokio::Mutex`) to `move_node` only. These two other self-contained store paths do a sibling-order read → compute → write on the shared connection and could interleave with a concurrent reorder. They now hold the lock across that sequence.

**No re-entrancy / deadlock:** `reorder_lock` is acquired in exactly three store methods (`move_node`, `create_child_node_atomic`, `move_children_to_parent`) and none calls another (`rebalance_children_for_parent` deliberately doesn't take it); no caller holds the lock (it's private to `SqliteStore`).

## Scope

Deferred (needs a **layered** atomic store method, not a one-line lock) and filed as **#1750**: the ops-layer "add existing node as child" (`get_next_child_order` read → `create_generic_relationship` write, two separate store calls) and `bulk_create_has_child` (inserts pre-computed orders) still write `has_child` orders without the lock.

## Tests

- **New:** `move_children_into_non_empty_parent_appends_without_collision` — seeds the new parent with two children at 1.0/2.0 via `create_child_node_atomic`, moves two more in, asserts all four order keys are distinct and the moved children hold the two largest (appended after). The old code produced `1.0, 1.0, 2.0, 2.0`.
- Existing `test_move_children_to_parent_store_*`, `concurrent_reorders_keep_all_children_with_distinct_order`, and the ops-layer move tests still pass.

968 core lib tests pass; clippy clean; full pre-push gate green (one transient daemon-readiness e2e flake — a startup-timing timeout unrelated to this change — cleared on a longer `E2E_DAEMON_TIMEOUT`; passes cleanly standalone). Independent adversarial review: no deadlock / race / correctness bug / regression.

Closes #1746.

### DIFF
--- a/packages/core/src/db/sqlite_store/mod.rs
+++ b/packages/core/src/db/sqlite_store/mod.rs
@@ -1221,6 +1221,92 @@ mod tests {
         Ok(())
     }
 
+    /// Moving children into a NON-EMPTY parent must APPEND after its existing
+    /// children, not restart at 1.0, 2.0… The existing children are seeded via
+    /// `create_child_node_atomic` (which appends at 1.0, 2.0), so the old code —
+    /// which assigned the moved children a fresh 1.0, 2.0 ignoring existing
+    /// siblings — collided their order keys. Deterministic, single-threaded.
+    #[tokio::test]
+    async fn move_children_into_non_empty_parent_appends_without_collision() -> Result<()> {
+        let (store, _temp) = create_test_store().await?;
+
+        // new_parent already holds two children at orders 1.0 and 2.0.
+        let new_parent = Node::new("text".to_string(), "New Parent".to_string(), json!({}));
+        let new_parent_id = new_parent.id.clone();
+        store.create_node(new_parent, None, None).await?;
+        for i in 0..2 {
+            store
+                .create_child_node_atomic(
+                    &new_parent_id,
+                    "text",
+                    &format!("existing {i}"),
+                    json!({}),
+                    None,
+                )
+                .await?;
+        }
+
+        // Two children under a source parent, to be moved across.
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Independent adversarial review — no deadlock, race, correctness bug, or regression:
- **Re-entrancy clear:** `reorder_lock` is acquired in exactly three store methods (`move_node`, `create_child_node_atomic`, `move_children_to_parent`); none calls another, `rebalance_children_for_parent` deliberately doesn't take it, and no caller holds it (private field). `notify` is synchronous so it can't re-enter an async mutex. Lock-across-await-tx matches the pattern #1561 established.
- **Correctness:** `existing_max` read (DESC LIMIT 1, NULL→0.0, None on empty) + `prev = orders.last().or(existing_max)` correctly appends after existing; the "child already under new_parent" edge case is version-gated (old edge deleted, re-inserted at a strictly larger key).
- **No regression:** empty parent → identical sequence; all move tests + the concurrency test pass.

Scope (ops-layer `get_next_child_order`→write and `bulk_create_has_child`) deferred to **#1750** — they need a layered atomic store method, not a one-line lock.

PR #1755, opened 2026-07-27
Consolidate subtree-render + drop dead code + expose skill in nav (part of #1699)

mstomar125:
## Summary

Ships the **non-conflicting subset** of #1699 (ADR-057). Items 2 (skill guidance → markdown children) and 3 (retire the `prompt` type) are **deferred** — see below and the [issue comment](https://github.com/NodeSpaceAI/nodespace-core/issues/1699).

- **Consolidate duplicate subtree-render logic:** `skill_ops::render_skill_instructions` and `PromptAssembler::fetch_prompt_body` carried two independent copies of the same DFS subtree walk. Extracted one shared `flatten_subtree_content` util in `packages/core` (`node_service::hierarchy`, re-exported at `services::`), used by both. `fetch_prompt_body` now also skips empty-content nodes (previously emitted blank `\n\n` sections) — a harmless unification; nested children are still preserved.
- **Delete dead code:** `PromptAssembler::assemble_with_skill()` (zero callers, leftover from the push-based discovery design ADR-038 replaced).
- **Nav:** `'skill'` added to `SIDENAV_CORE_TYPES` — skill nodes browse via the generic path (no new UI; `skill` is already a core schema).

## Why 2 & 3 are deferred

The issue's premise that the `prompt` type has "zero remaining consumers" is **incorrect**: the daemon seeds prompt-typed nodes at startup (`daemon/services/assembly.rs:229`) and `PromptAssembler::fetch_prompt_overrides` queries `node_type == "prompt"` **every agent turn** (`agent_loop.rs:298` → `assemble()`, a live path). Retiring the type would make those internal system-prompt override nodes fall back to the embeddable `CustomNodeBehavior` → search pollution; item 2 has the same embeddability implication. Retiring `prompt` needs a design decision (migrate the override mechanism off the type — the `agent-config` schema the issue gestures at) that's out of scope here. Full detail on the issue.

## Verification

`flatten_subtree_content` is behavior-equivalent to both originals (same pre-order; the only change is the harmless empty-skip). core 973 / agent 301 lib tests pass; clippy clean; the `fetch_prompt_body` body-drop regression test still passes; pre-push gate green. Independent review: sound.

**Part of #1699** (issue stays open for items 2 & 3).

### DIFF
--- a/packages/agent/src/prompt_assembler.rs
+++ b/packages/agent/src/prompt_assembler.rs
@@ -11,7 +11,7 @@ use std::sync::Arc;
 
 use nodespace_core::markdown::NodeTemplate;
 use nodespace_core::models::Node;
-use nodespace_core::services::NodeService;
+use nodespace_core::services::{flatten_subtree_content, NodeService};
 
 use crate::agent_guidance::{NODE_REFERENCE_FORMAT, SCHEMA_CREATION_RULES, TOOL_STRATEGY_RULES};
 use crate::agent_types::ToolDefinition;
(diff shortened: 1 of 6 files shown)


mstomar125:
**APPROVE**

Independent review — sound, safe to ship:
- `flatten_subtree_content` is behavior-equivalent to both originals (identical pre-order stack walk); the only intended change is `fetch_prompt_body` now skipping empty-content nodes (cosmetic — collapses redundant blank paragraphs; children still walked). The missing-id-in-node_map divergence is a can't-happen store-invariant path that warns identically.
- Export path clean (`nodespace_core::services::flatten_subtree_content`); no dangling refs; import cleanup correct.
- 5 ported skill_ops walk tests + the `fetch_prompt_body` body-drop regression test cover it. Nav filter (`isCore && SIDENAV_CORE_TYPES.has`) surfaces `skill` with no new UI.
- Items 2/3 fully reverted (prompt schema/behavior intact, skill_pipeline still `Some("prompt")`).

Merging as part of #1699; issue stays open for the deferred prompt-retirement (needs the override-node re-spec).

Issue #1734, opened 2026-07-25
PersonNode duplicate prevention: schema-level uniqueness validation class, creation-time adoption, convergence policy


malibio:
## Overview

The same human can end up as two PersonNodes: a contact node created as content, plus the node minted/adopted when they join the tenant. The invite-time claim (nodespace-cloud#88: invites can carry a `person_node_id`, and the bind adopts that node) solves the common case. This issue is the prevention layer for everything else — pointing the schema system at the root cause instead of building merge machinery first.

## Problem Statement

Duplicate person nodes arise from ordinary data entry: a user creates a contact for someone who already has a node (bound or unbound). Today nothing discourages this — `validate()` in the behavior system is strictly per-node and cannot consult the store, so no uniqueness rule is expressible.

A constraint to respect: in a local-first system a global uniqueness rule cannot be *enforced* at creation — two offline devices can each validly create the same person, and the conflict only becomes visible at sync convergence, where hard rejection would turn a data-entry duplicate into a sync failure. Hard uniqueness exists only where it is identity: the cloud's `auth_identities` UNIQUE constraints already guarantee one bound node per account. Everything else must be suggest-don't-block plus a convergence policy.

### 1. Schema-system uniqueness validation class (the refactoring)

Extend the validation system with a store-aware rule kind: `unique` on a schema property (first consumer: `person.email`, case-insensitive, non-empty values only). This is a new validation class — unlike existing per-node rules it requires a store query, so the seam changes: validation gains an async store-backed phase (or the rule is evaluated by NodeService at create/update rather than inside `NodeBehavior::validate`). Scope is per-database. Design the capability generically (schema-driven types get it too), consistent with the hybrid core/extension behavior architecture.

### 2. Creation-time adoption UX (suggest, never block)

When a uniqueness rule matches at creation: surface "A person with this email already exists — use them instead?" offering adopt-existing / create-anyway. Same affordance on edit when an email collides.

### 3. Convergence policy (the residue)

For duplicates that slip through (offline creation, create-anyway): detect on sync apply (uniqueness scan over incoming person nodes), surface as a non-blocking "possible duplicate" indicator. Full admin-driven merge (re-pointing mentions/member_of edges, tombstoning the loser, sync-echo handling) is deliberately deferred until the prevention layers prove insufficient — record that decision here.

## Architecture Impact

[...]

PR #1756, opened 2026-07-27
Enforce root-only content membership for member_of edges (#1742)

mstomar125:
## Summary

ADR-059 §2: a **content node may hold a `member_of` edge only if it is a root** (no `has_child` parent). Membership carries access classification (ADR-037), so letting an interior node carry independent classification would make access a property of outline position — the conflation ADR-059 removes. Until now the rule held only by the absence of a UI affordance.

## Fix

A single shared `NodeService::assert_root_only_membership(&source)` — returns Ok for **collection** (nesting) and **person** (grantee membership, ADR-037 §4) sources; for a content source it requires `is_root_node`. Called from **both** `member_of` write paths:

- **`create_relationship`** (uses the source it already fetches) — covers single adds (CLI, `add_to_collection`, playbook `add_relationship`) **and the sync-apply of a remote edge**, which routes through `create_relationship` (`sync/nodespaced-pro/.../cloud_sync/apply.rs:305`) — so sync-apply parity holds with **no sync-repo change**.
- **`bulk_add_to_collections_notify`** (import) — validates the whole batch before writing.

One validation reused by every path, so the invariant can't be silently broken by whichever surface wrote the edge.

## Acceptance criteria
- [x] `member_of` from a content node with a `has_child` parent → rejected
- [x] `member_of` from a root content node → succeeds
- [x] Person-node `member_of` (grantee) unaffected
- [x] Collection→collection `member_of` (nesting) permitted
- [x] Declared as a shared validation, not hardcoded in a single write path

## Note on criterion 5

The rule is a single shared service-level validation reused by every write path (not a per-path check), which satisfies "not hardcoded in one write path." A fuller *schema-system-declared* validation framework (the issue's ideal phrasing, sibling to the seam #1734 gestures at) would be a separate refinement — flagged, not built here, to keep this focused on the enforcement.

## Verification

New test `root_only_content_membership_is_enforced` covers all four criteria + the bulk path. **Live-conflict check clean:** the full core (974) / agent (301) / daemon suites pass — nothing existing (seeds, skill/tool seeding, import) files an interior content node. Clippy clean; pre-push gate green.

Closes #1742.

### DIFF
--- a/packages/core/src/services/node_service/mod.rs
+++ b/packages/core/src/services/node_service/mod.rs
@@ -3356,6 +3356,88 @@ mod tests {
         assert!(node_ids.contains(&child2_id), "child2 event missing");
     }
 
+    /// ADR-059 §2: a content node may hold a `member_of` edge only if it is a
+    /// root. Enforced on every membership-write path; person (grantee) and
+    /// collection (nesting) sources are exempt.
+    #[tokio::test]
+    async fn root_only_content_membership_is_enforced() {
+        let (service, _temp) = create_test_service().await;
+        let coll = service
+            .create_node(Node::new(
+                "collection".to_string(),
+                "Docs".to_string(),
+                json!({}),
+            ))
+            .await
+            .unwrap();
+
+        // A ROOT content node may be a member.
+        let root_doc = service
+            .create_node(Node::new("text".to_string(), "Root doc".to_string(), json!({})))
+            .await
+            .unwrap();
+        service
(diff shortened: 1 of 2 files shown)


mstomar125:
Closing unmerged — adversarial review found the service-level enforcement is the wrong layer (misses the sync cold-sweep path, the reparent transition, and risks wedging incremental sync). Needs a store-level guard + cross-repo work; findings captured on #1742.

PR #1768, opened 2026-07-28
Enforce root-only content membership for member_of edges (store-level, ADR-059 §2)

mstomar125:
## What

Enforces **root-only content membership** for `member_of` edges (ADR-059 §2): a content node may be a member of a collection only when it is a **root** node (no `has_child` parent). `collection` nodes (nesting) and `person` nodes (grantee membership, ADR-037 §4) are exempt.

The rule is enforced at the **store** level so *every* write path is covered by construction — CLI, graph import, playbook `add_relationship`, and the sync-apply cold-sweep — rather than by a per-path check that a future path could miss.

## How

A new `SqliteStore::assert_root_only_membership(member_ids)` runs before all **three** `member_of` INSERT sites:

- `add_to_collection` (single, auto-order)
- `bulk_add_to_collections` (the path the sync-apply cold-sweep and daemon import call directly)
- `create_generic_relationship` when its `rel_type` is `member_of` — reached when a caller supplies an explicit `order` (`NodeService::create_relationship`'s non-auto-order fork, playbook `add_relationship`, CLI `relationship create --edge-data`)

The bulk check is a single chunked `IN (...)` query (node type + parent presence in one round trip), so the performance-sensitive cold-sweep path stays one query, not 2×N point lookups. Interior nodes are rejected with an actionable, node-naming error (`member_of_not_root: content node '<id>' …`).

## Acceptance criteria (issue #1742)

- ✅ interior content `member_of` write rejected; ✅ root content write succeeds
- ✅ person-node edges unaffected; ✅ collection-to-collection nesting permitted
- ✅ enforced on the sync-apply path (cold-sweep → `bulk_add_to_collections`), not only local writes
- ✅ rejection names the node and why; ✅ CLI path gated (routes through the guarded store methods)
- ✅ regression: interior node cannot be filed into a collection via the generic/explicit-order path
- ✅ regression: a restricted task inside an open project still works (interior task under a filed root holds no membership of its own)
- Criterion 5 (declared in the schema system as a store-aware relationship-validation *kind*) is left to the schema-seam work tracked with #1734; this PR is the enforcement.
- Criterion 8 (reparent transition) split to **#1765** (reject-the-indent); sync permanent-vs-transient apply classification split to **nodespace-sync#378**.

## Tests

Two store-level tests: `test_root_only_content_membership_is_enforced` (root ✓, interior ✗ via both `add_to_collection` and `create_generic_relationship`, person/collection exemptions, restricted-task-in-open-project end-to-end) and `test_bulk_root_only_membership_is_enforced` (all-root batch applies; a batch with one interior node is refused — cold-sweep parity). Full core lib suite: **975 passed / 0 failed**.

## Review

Two independent adversarial reviews. The first caught a real High-severity bypass — `create_generic_relationship` was an unguarded third `member_of` writer reachable by adding one `order` key — which is now fixed and covered by a test. A focused second pass confirmed the fix is complete (exactly three `member_of` writers, all guarded) and introduces no false positives.

Closes #1742

### DIFF
--- a/packages/core/src/db/sqlite_store/mod.rs
+++ b/packages/core/src/db/sqlite_store/mod.rs
@@ -519,6 +519,158 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_root_only_content_membership_is_enforced() -> Result<()> {
+        // ADR-059 §2: a content node may be `member_of` a collection only when it
+        // is a root node (no `has_child` parent). Enforced in `add_to_collection`,
+        // so every single-add path is gated — including the CLI (which routes
+        // through `ops::collection_ops` → `add_to_collection`).
+        let (store, _t) = create_test_store().await?;
+
+        let coll = Node::new("collection".to_string(), "Coll".to_string(), json!({}));
+        let coll_id = coll.id.clone();
+        store.create_node(coll, None, None).await?;
+
+        // A ROOT content node can be filed. (criterion: root content succeeds)
+        let root_text = Node::new("text".to_string(), "root doc".to_string(), json!({}));
+        let root_id = root_text.id.clone();
+        store.create_node(root_text, None, None).await?;
+        assert!(
+            store.add_to_collection(&root_id, &coll_id).await?.is_some(),
+            "a root content node must be fileable into a collection"
+        );
+
+        // An INTERIOR content node (has a `has_child` parent) is REJECTED. This is
+        // the CLI-path regression: filing an interior node into a (restricted or
+        // any) collection is refused with an actionable, node-naming error.
+        let interior = store
+            .create_child_node_atomic(&root_id, "text", "an interior child", json!({}), None)
+            .await?;
+        let err = store
+            .add_to_collection(&interior.id, &coll_id)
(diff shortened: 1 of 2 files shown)


mstomar125:
**APPROVE**

Store-level enforcement verified through two independent adversarial reviews.

- **Review 1** caught a real High-severity bypass: `create_generic_relationship` was a third, param-bound (`?4`) `member_of` writer reachable via `NodeService::create_relationship`'s explicit-`order` fork (and playbook `add_relationship` / CLI `relationship create --edge-data`). Fixed at the store choke point + covered by a test that drives exactly that path.
- **Review 2** (focused) confirmed CONFIRMED SOUND: exhaustive sweep shows exactly three `member_of` writers, all now guarded; the guard checks the member (in_node) with the correct root definition; no false positives; the exemptions (collection nesting, person grantee membership) hold; tests are non-tautological.

Full core lib suite 975/975; clippy clean; pre-push gate (test:all + daemon build + e2e) green. Author=reviewer=mstomar125 self-approval per merge policy.

PR #1772, opened 2026-07-28
Reject reparenting a collection member into an interior position (ADR-059 §2)

mstomar125:
## What

Adds the **reparent side** of the root-only content-membership rule (ADR-059 §2). #1742 enforces it at `member_of` *write* time; this closes the reverse transition — a node that already holds a `member_of` edge (a root member) being moved into an interior position, which would make it a forbidden interior member with no `member_of` write for the forward guard to catch.

## How

The guard is enforced at the **store's two `has_child` attach sites**, symmetric with how #1742 enforced the forward direction at the store's `member_of` INSERT sites. Adversarial review confirmed these are the only two store paths that attach an *existing* node to a parent (the rest create fresh nodes or move already-interior ones), so gating both covers every reparent path by construction:

- **`SqliteStore::move_node`** — the service `NodeService::move_node` (outline drag/indent) and `NodeService::upsert_node_with_parent` (the gRPC upsert / `save_node_with_parent` Tauri command; a *distinct* reparent path the first review caught bypassing a service-only guard).
- **`SqliteStore::bulk_create_has_child`** — the sync-apply cold sweep (`apply_cold_rel_sweep_batched`), mirroring #1742's guard on its sibling `bulk_add_to_collections`.

When a node's type is neither `collection` nor `person` and it holds one or more `member_of` edges, a move that would give it a `has_child` parent is **rejected** with an error naming the collections (a single chunked query on the bulk path).

- **Reject, not drop** — a node can hold several `member_of` grants, each an independent access path; silently dropping them on an indent would revoke access with no record (Michael's decision on the issue's criterion 8).
- **Move to root stays allowed** (`new_parent = None`) — the guard only fires when the node would gain a parent.
- **`collection` (nesting) and `person` (grantee, ADR-037 §4) exempt** — matching the forward guard.

Adversarial review of the first (service-level) attempt found that `upsert_node_with_parent` bypassed a service-only guard; moving the check to the store chokepoint closes that and any future reparent surface.

## Tests

- **Store** `test_reparent_guard_rejects_moving_a_member_under_a_parent`: `store.move_node` rejects reparenting a member; non-member moves freely; move-to-root allowed; person exemption; **and `bulk_create_has_child` rejects a member in a batch** (cold-sweep parity).
- **Service** `test_reparenting_a_collection_member_is_rejected`: rejection via **both** `move_node` and `upsert_node_with_parent`, proof the membership is not dropped (second attempt still rejects), move-to-root still allowed.

All core test targets green — **980 lib + integration + doctests, 0 failed**. Reviewed adversarially three times: the passes caught the `upsert_node_with_parent` bypass and the `bulk_create_has_child` cold-sweep gap in turn, both now closed and tested. The sync side of the cold-sweep rejection (permanent-vs-transient classification, so a bad incoming edge can't wedge the cursor) is tracked in nodespace-sync#378.

Closes #1765

### DIFF
--- a/packages/core/src/db/sqlite_store/mod.rs
+++ b/packages/core/src/db/sqlite_store/mod.rs
@@ -581,7 +581,12 @@ mod tests {
         // `add_to_collection` does. (Regression for the bypass where adding one
         // `order` JSON key skipped the guard.)
         let err_generic = store
-            .create_generic_relationship(&interior.id, &coll_id, "member_of", &json!({"order": 5.0}))
+            .create_generic_relationship(
+                &interior.id,
+                &coll_id,
+                "member_of",
+                &json!({"order": 5.0}),
+            )
             .await
             .expect_err("member_of via the generic path must reject an interior node");
         assert!(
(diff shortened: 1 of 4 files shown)


mstomar125:
**APPROVE**

Store-level reparent guard verified through three independent adversarial reviews, each of which caught a real completeness gap and drove the enforcement to the correct layer:

1. First attempt guarded the service `move_node` only → review found **`upsert_node_with_parent`** (gRPC / `save_node_with_parent`) reparents via `store.move_node` directly, bypassing it.
2. Moved to the store `SqliteStore::move_node` chokepoint → review found **`bulk_create_has_child`** (the sync reconnect cold sweep) attaches existing roots without going through `move_node`, asymmetric with #1742's guard on its sibling `bulk_add_to_collections`.
3. Gated BOTH store attach sites via a batched `assert_may_gain_parent`. The third review's exhaustive enumeration confirms these are the ONLY two store paths that attach an existing node (the rest create fresh nodes or move already-interior ones), so coverage is complete.

980 lib + all integration + doctest targets green; clippy clean; pre-push gate (test:all + daemon build + e2e) green. Cold-sweep permanent-vs-transient handling tracked in nodespace-sync#378. Author=reviewer=mstomar125 self-approval per merge policy.

PR #1781, opened 2026-07-28
Agent emits malformed tool calls: title_template referencing undefined fields, and calls with empty arguments

malibio:
Closes #1775


malibio:
Heads-up for the #1777 branch (`worktree-issue-1777-schema-namespace-validation`), which is in progress concurrently.

Both branches append test blocks to the end of `packages/core/src/schema/schema_test.rs`, so whichever merges second will hit a textual conflict there. It is benign — two independent blocks appended at the same place, no logic overlap. **Resolution is to keep both sides.**

One thing worth knowing rather than discovering during the resolve: my `test_create_schema_without_fields_key_is_unaffected` deliberately does NOT assert that description-only `create_schema` succeeds, and documents why — at the time it was written the `custom:` prefix made that path fail, which is precisely the bug #1777 fixes. Once #1777 lands, that test still passes (it only asserts the field-locating pass stays out of the way), and #1777's own test covers the success case. No change needed to either.

PR #1753, opened 2026-07-27
Implement ProjectNodeBehavior and the project core schema (#134)

mstomar125:
## Summary

Adds `project` as a core node type (Universal Graph model): a `project` `SchemaNode` in `core_schemas.rs` + a `ProjectNodeBehavior` registered in `NodeBehaviorRegistry::new()`.

- **Schema:** `status` (enum, Core, default `planning`, indexed, required), `priority` (enum, User, indexed, optional), `start_date`/`end_date` (date, User, optional). `is_core: true`. Mirrors the `task` schema.
- **Behavior:** name is the node `content`; typed data under `properties.project.*`; children are `has_child` edges; ownership/membership are graph edges — **no side table, no owner_id/team_members blob**.
- **Validation split (as `TaskNodeBehavior`):** the behavior TYPE-checks fields and enforces the cross-field `start_date <= end_date` rule (only when both parse as ISO `YYYY-MM-DD`); the schema system validates enum membership/values (not duplicated here). Empty name → `MissingField`; non-string `status`/`priority`/date → `InvalidProperties`.
- `can_have_children` → true, `supports_markdown` → true, `default_metadata` seeds `project.status = "planning"`. Projects are **embeddable roots** (name/description) — `get_embeddable_content`/`get_parent_contribution` return the content, unlike Task which returns `None` (intentional divergence, stated explicitly).

## Tests

5 behavior tests + a doctest: valid full project, minimal (name only), empty-name rejection, non-string field rejection, inverted date range rejection (+ non-ISO dates skip the range check, equal dates allowed), and `default_metadata` shape. Updated the enumeration tests (`test_registry_get_all_types` 18→19, `test_get_core_schemas_returns_all` 17→18) and renamed a `schema_test` example from "Project"→"Campaign", since `project` is now a reserved core type a user schema can't shadow.

973 core lib tests pass; clippy clean; doctest passes; pre-push gate green.

## Scope
Type only — no project UI/viewers, no status-transition workflows (#334), no `on_delete` (the trait has none). Per ADR corrections in the issue (Universal Graph; no `projects` table).

Closes #134.

PR #1782, opened 2026-07-28
Description-only create_schema always fails: namespaced field names contain ':' which the field-name validator rejects

malibio:
Closes #1777

### DIFF
--- a/packages/core/src/behaviors/mod.rs
+++ b/packages/core/src/behaviors/mod.rs
@@ -1250,11 +1250,17 @@ pub struct SchemaNodeBehavior;
 /// the bare name must be non-empty and contain only alphanumerics and underscores.
 ///
 /// Namespace prefixes are part of the stored field name, not a separate layer
-/// applied afterwards: user-defined fields are namespaced (`custom:`, `org:`,
-/// `plugin:`) before the schema node is validated and persisted, so validation
-/// must accept the prefixed form. Both schema-creation paths agree on this —
-/// the description path applies `custom:` during field inference, and the
-/// explicit-fields path passes through names the caller already namespaced.
+/// applied afterwards: fields are namespaced (`custom:`, `org:`, `plugin:`)
+/// before the schema node is validated and persisted, so validation must accept
+/// the prefixed form. The same stored-name convention is assumed downstream —
+/// CEL strips the prefix, the graph resolver matches on it, and identifier
+/// validation in the query path permits it.
+///
+/// This function decides only whether a name is *well-formed*, not whether it is
+/// namespaced. Both are accepted, because the two schema-creation paths differ:
+/// the description path applies `custom:` during field inference, while the
+/// explicit-fields path stores caller-supplied names verbatim and does not
+/// namespace them. Requiring a prefix here would reject the latter.
 fn validate_schema_field_name(name: &str) -> Result<(), NodeValidationError> {
     let invalid = |reason: &str| {
         Err(NodeValidationError::InvalidProperties(format!(
@@ -3771,6 +3777,142 @@ mod tests {
         assert!(
             behavior.validate(&namespaced_node).is_ok(),
             "Namespaced field names should pass validation"
+        );
+    }
+
+    #[test]
+    fn test_schema_node_validates_namespaced_names_at_nesting_depth() {
+        let behavior = SchemaNodeBehavior;
+
+        // Name validation recurses through `fields` and `item_fields`, so the
+        // namespace rule must hold at depth, not just on top-level fields.
+        let nested = Node::new(
+            "schema".to_string(),
+            "venue".to_string(),
+            json!({
+                "isCore": false,
+                "version": 1,
+                "fields": [
+                    {
+                        "name": "custom:address",
+                        "type": "object",
+                        "protection": "user",
+                        "indexed": false,
+                        "fields": [
+                            {
+                                "name": "custom:postal_code",
+                                "type": "string",
+                                "protection": "user",
+                                "indexed": false
+                            }
+                        ]
+                    },
+                    {
+                        "name": "custom:sessions",
+                        "type": "array",
+                        "protection": "user",
+                        "indexed": false,
+                        "item_fields": [
+                            {
+                                "name": "custom:room",
+                                "type": "string",
+                                "protection": "user",
+                                "indexed": false
+                            }
+                        ]
+                    }
+                ]
+            }),
+        );
+
+        assert!(
+            behavior.validate(&nested).is_ok(),
+            "Namespaced nested and item field names should pass validation"
+        );
+
+        // A malformed nested name must still be rejected at depth.
+        let bad_nested = Node::new(
+            "schema".to_string(),
+            "venue".to_string(),
+            json!({
+                "isCore": false,
+                "version": 1,
+                "fields": [
+                    {
+                        "name": "custom:address",
+                        "type": "object",
+                        "protection": "user",
+                        "indexed": false,
+                        "fields": [
+                            {
+                                "name": "custom:a:b",
+                                "type": "string",
+                                "protection": "user",
+                                "indexed": false
+                            }
+                        ]
+                    }
+                ]
+            }),
+        );
+
+        assert!(
+            behavior.validate(&bad_nested).is_err(),
+            "Malformed nested field names should still be rejected"
+        );
+    }
+
+    #[test]
+    fn test_schema_node_title_template_resolves_namespaced_field() {
+        let behavior = SchemaNodeBehavior;
+
+        // Templates are checked against the STORED field names, so a schema
+        // created from a description must reference the namespaced form.
+        let resolved = Node::new(
+            "schema".to_string(),
+            "venue".to_string(),
+            json!({
+                "isCore": false,
+                "version": 1,
+                "titleTemplate": "{custom:contact_email}",
+                "fields": [
+                    {
+                        "name": "custom:contact_email",
+                        "type": "string",
+                        "protection": "user",
+                        "indexed": false
+                    }
+                ]
+            }),
+        );
+
+        assert!(
+            behavior.validate(&resolved).is_ok(),
+            "titleTemplate referencing a namespaced field should validate"
+        );
+
+        // The bare name is a different key and must not resolve.
+        let unresolved = Node::new(
+            "schema".to_string(),
+            "venue".to_string(),
+            json!({
+                "isCore": false,
+                "version": 1,
+                "titleTemplate": "{contact_email}",
+                "fields": [
+                    {
+                        "name": "custom:contact_email",
+                        "type": "string",
+                        "protection": "user",
+                        "indexed": false
+                    }
+                ]
+            }),
+        );
+
+        assert!(
+            behavior.validate(&unresolved).is_err(),
+            "Bare name must not resolve against a namespaced field"
         );
     }
 
--- a/packages/core/src/behaviors/mod.rs
+++ b/packages/core/src/behaviors/mod.rs
@@ -1232,7 +1232,8 @@ impl NodeBehavior for DateNodeBehavior {
 /// Validation includes:
 /// - Non-empty content (schema name)
 /// - Properties must be valid JSON object
-/// - Field names must be unique (alphanumeric and underscores only)
+/// - Field names must be unique (alphanumeric and underscores, with an optional
+///   `<namespace>:` prefix — see [`validate_schema_field_name`])
 /// - Enum fields must have at least one value defined (in core_values or user_values)
 ///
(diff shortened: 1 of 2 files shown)


malibio:
# Pragmatic Code Review — PR #1782

**Recommendation: APPROVE** (with one follow-up worth filing, and one nit)

## Summary

This is a tight, well-scoped, net-positive fix. One commit, one root cause, minimal blast radius. The diagnosis in the commit message is correct and — importantly — the justification for *which* side to change is not hand-waved. I independently verified the claim that `custom:field` is already an accepted **stored** field name elsewhere in the system, and it holds:

- 📁 `packages/core/src/ops/query_ops.rs:178` — `validate_identifier` explicitly permits `':'`, and 📁 `packages/core/src/ops/query_ops.rs:358` asserts `validate_identifier("custom:field", "t").is_ok()`.
- 📁 `packages/core/src/playbook/cel.rs:205` — strips the colon namespace (`custom:amount` → `node.amount`).
- 📁 `packages/core/src/playbook/graph_resolver.rs:278` — matches bare keys against the `custom:` prefixed form.

So the schema field-name validator genuinely was the outlier, and loosening it (rather than deferring namespacing until after validation) is the correct call: it makes the validator agree with the storage contract the rest of the system already assumes, instead of introducing a second, conflicting notion of "field name" at the schema layer. Choosing the smaller-surface fix that removes an inconsistency rather than adding a compensating layer is the right instinct here.

Verification performed: all 3 new unit tests pass; `cargo test -p nodespace-core --lib schema` → 113 passed, `--lib behaviors` → 125 passed; `cargo clippy -p nodespace-core --lib --all-features` clean, no new warnings.

---

## Findings

### 🟢 Correctness of `validate_schema_field_name` — verified sound

📁 `packages/core/src/behaviors/mod.rs:1256-1298`

I extracted the parser logic and ran the edge cases against it directly rather than reasoning about the tuple match on paper. Results:

| input | result |
|---|---|
| `""` | rejected (empty bare) |
| `"a"` | accepted |
| `"a:b"` | accepted |
| `"a:b:c"` | rejected (multi) |
| `":"` | rejected (empty ns) |
| `"::"` | rejected (multi) |
| `"a:"` | rejected (empty bare) |
| `":a"` | rejected (empty ns) |
| `"cust_om:a_b"` | accepted |

All correct, and all match the documented rule. Two details worth calling out as *good*:

- The `(Some(bare), None, _)` arm correctly uses `_` for the third element — reachable only as `None`, and not pretending otherwise.
- `!s.is_empty()` inside `is_valid_segment` is load-bearing, not redundant: `chars().all(..)` is vacuously true on `""`, so without it `""`, `"a:"`, and `":a"` would all pass. The tests cover exactly these three, which suggests the author knew that rather than got lucky.

The `// split always yields at least one element, so this cannot panic` comment is the right kind of comment — it explains why the `_` fallback arm is not a silent correctness hole. Comments explaining *why*, not *what*.

### 🟢 `is_alphanumeric` vs `is_ascii_alphanumeric` — inconsistency is inherited, not introduced

The validator uses `is_alphanumeric()`, so `café` and `custom:café` (and full-width `１２３`) are accepted, whereas 📁 `packages/core/src/ops/query_ops.rs:178` uses `is_ascii_alphanumeric()` and would reject them. This looks like a discrepancy but is **not a regression from this PR** and is **not security-relevant**:

- The pre-existing validator this replaces already used `is_alphanumeric()` — the PR preserves the character class verbatim and only adds the prefix rule. Tightening to ASCII here would be an unrelated behavior change smuggled into a bug fix, correctly avoided (atomic-change discipline).
- 📁 `packages/core/src/schema/mod.rs:1032` (`normalize_field_name`) also filters on `is_alphanumeric()`, so the schema-inference path and this validator agree. Switching only the validator to ASCII would actually *create* a new inconsistency and break description-only creation for non-English descriptions.
- Not a security boundary: the SQL-facing gate is `validate_identifier`, which is stricter. A unicode field name that this validator admits would be rejected downstream at the query op, not injected.

Mentioning this only so it is on record as considered-and-deliberate rather than overlooked. No change requested.

### 🟢 Security — no widened injection surface

Field names do reach string-interpolated SQL at 📁 `packages/core/src/services/query_service/mod.rs:343-351` and `:365-369` (`json_extract(properties, '$.{}')`), which is a genuine pre-existing interpolation. But this PR does not widen it:

- The gate for that path is `validate_identifier` (`query_ops.rs:166`), which is **stricter** than the schema validator and already allowed `:`. Nothing newly admitted by `validate_schema_field_name` reaches SQL that could not already.
- The only new character admitted is `:`, which has no syntactic meaning in a SQLite JSON path or in string literal context, and cannot terminate the surrounding `'...'` quoting. Quote, backslash, `.`, `$`, `[`, whitespace all remain rejected.
- CEL: `cel.rs` treats `:` as a namespace separator it strips, so a namespaced name resolves to the bare identifier — the intended behavior, not an escape.

No injection concern. The `describe_malformed_fields` gate from #1781 (📁 `packages/core/src/schema/mod.rs:43`) runs earlier and only checks structural shape, so there is no ordering conflict with this change.

### 🟢 Interaction with #1781 (`validate_template_tokens`) — consistent, and the test suite confirms it

📁 `packages/core/src/behaviors/mod.rs:1335-1370` + `:1423-1436`

`validate_template_tokens` matches `{token}` against `defined_fields`, which is built from the **stored** (already-namespaced) names. That is self-consistent with this change: a description-created schema stores `custom:contact_email`, so a template must say `{custom:contact_email}`, and both the validator and the Rust renderer (📁 `packages/core/src/utils/markdown.rs:132-145`, which does a raw `properties.get(&field_name)`) agree on that. All 113 schema tests including the #1781 title-template suite pass unchanged. No conflict.

### 🟡 Acceptance criterion 3 is *documented*, not fully *achieved* — the explicit-fields path still does not namespace

This is the one substantive gap, and I want to be honest about it rather than tick the box.

The doc comment at 📁 `packages/core/src/behaviors/mod.rs:1245-1255` states the two paths "agree on this — the description path applies `custom:` during field inference, and the explicit-fields path passes through names the caller already namespaced." That second clause is the load-bearing one, and it is an *assumption about caller behavior that nothing enforces*. At 📁 `packages/core/src/schema/mod.rs:227-229` the explicit path is literally `(explicit_fields, Vec::new())` — a pass-through with no namespacing and no requirement of one.

So the real state is: the validator now accepts **both** namespaced and bare names, and the two paths produce **different** conventions for the same user intent — description-only yields `custom:capacity`, explicit-fields yields `capacity`. They no longer *conflict* (which is the bug this PR fixes, and that is genuinely fixed), but they do not *agree* either. The system tolerates both rather than converging on one.

Concretely, this matters because it is user-visible: two schemas created for the same domain via different call paths get different property keys, which means different `titleTemplate` tokens, different CEL/playbook selectors, and different query filter strings. `CLAUDE.md` states the rule as "No user properties without namespace prefix — conflicts with future core properties," and the explicit-fields path silently permits exactly that.

**I am not asking for it in this PR.** Namespacing the explicit path is a behavior change with real blast radius (existing explicit-field callers, the agent tool contract, the eval suite), and bundling it here would violate the atomic-change principle that this PR otherwise follows well. The right move is a follow-up issue: "explicit `create_schema` fields bypass `custom:` namespacing." Ship this fix now.

Suggest softening the doc comment so it describes what the code guarantees rather than what callers are trusted to do — e.g. note that the explicit path passes names through unchanged and does not enforce namespacing, so bare names remain valid. As written, a future reader could reasonably take "the two paths agree" as an invariant they can rely on, and it is not one.

### 🟢 Test quality — good, with two gaps worth noting

The three new tests are meaningful, not coverage theater:

- `test_schema_field_name_accepts_bare_and_namespaced` and `test_schema_field_name_rejects_malformed_names` (📁 `packages/core/src/behaviors/mod.rs:3710-3746`) hit the genuinely tricky boundaries — `"custom:"`, `":capacity"`, `"custom:a:b"` — i.e. exactly the cases where a naive `splitn`/`contains(':')` implementation would silently pass. The failure messages interpolate the offending name, so a regression names itself.
- `test_create_schema_from_description_only` (📁 `packages/core/src/schema/schema_test.rs:1081-1131`) is the real regression test for AC 1 and 2 and is well-built: it asserts the specific `custom:contact_email` name rather than just `is_ok()`, and — the part I'd have flagged if missing — it reads the schema back with `get_schema_node("venue")` rather than trusting the returned value. Asserting persistence, not just the return, is what makes this an end-to-end test instead of a unit test wearing a costume.

Gaps (neither blocking):

- **Nested/item field recursion is untested for namespaced names.** `validate_schema_field` recurses into `field.fields` and `field.item_fields` (📁 `packages/core/src/behaviors/mod.rs:1314-1325`), so `validate_schema_field_name` now runs at arbitrary depth. A nested field named `custom:x` is newly accepted and nothing covers that path. Low risk (same function, same rule), but it is a distinct call site.
- **No test for the namespaced-name/title-template combination** — a schema with field `custom:contact_email` and `titleTemplate: "{custom:contact_email}"`. This is precisely the seam between this PR and #1781, it is the case most likely to surprise someone later, and it is cheap to add. Worth including if you touch this again.

### 🟢 Project standards — compliant

Checked against `CLAUDE.md` explicitly:

- ✅ No lint suppression, no `#[allow(dead_code)]`, no `#[deprecated]`.
- ✅ No backward-compat cruft or transition-period language — old validation is deleted outright, not kept alongside.
- ✅ **No issue numbers in the new doc comment or code comments.** Verified directly; the new doc block at `:1245-1255` describes the constraint behaviorally with no `#1777` reference. The issue number appears only in the commit message, which is the permitted GitHub-mechanics case.
- ✅ Change is atomic: one root cause, one fix.

### 🟢 Nit: unrelated rustfmt churn in the diff

📁 `packages/core/src/behaviors/mod.rs:2917-2940` — the reflowing of `Node::new("project", ...)` and the `assert_eq!(behavior.get_embeddable_content(...))` in `test_project_node_behavior` is pre-existing formatting drift from #1753/#1781, not this change. It is 3 lines of noise in an otherwise surgical diff. Harmless and arguably good hygiene to leave fixed — just noting it so a future `git blame` on those lines doesn't send someone hunting for a connection to namespace validation that isn't there.

---

## Verdict

**APPROVE.** AC 1 and AC 2 are fully met and properly tested. AC 3 is met in the sense that matters — the rule is clearly stated in the code and the paths no longer contradict each other — but the doc comment slightly overstates the agreement, and the explicit-fields path remains un-namespaced. That is a real inconsistency, correctly left out of scope, and worth a follow-up issue rather than a blocker.

Suggested before merge (both optional, neither gating): soften the "the two paths agree" sentence to describe the actual guarantee, and file the explicit-fields-namespacing follow-up.

malibio (commented):
Pragmatic Quality review posted as a PR comment. Recommendation: APPROVE — no blockers. Verified independently: the claim that `custom:field` is already a valid stored field name holds (query_ops::validate_identifier accepts ':', cel.rs strips the prefix, graph_resolver matches on it), so loosening the schema validator is the correct side to change. Parser edge cases checked empirically (``, `a:`, `:a`, `a:b:c`, `::` all correctly rejected). Tests pass (schema 113, behaviors 125), clippy clean. One honest caveat on AC 3: the explicit-fields path still does not namespace, so the two paths no longer conflict but do not truly agree — suggest a follow-up issue and softening the doc comment. Non-blocking.

malibio:
Addressed the review in d108b94a.

**Doc comment (the 🟡 finding).** The review was right that "the two paths agree… the explicit-fields path passes through names the caller already namespaced" described trusted caller behavior as if it were an enforced invariant — nothing enforces it. Reworded to state what the function actually decides: whether a name is *well-formed*, not whether it is namespaced. Both forms are accepted precisely because the paths differ, and requiring a prefix here would reject the explicit path. The downstream stored-name convention (CEL, graph resolver, query identifier validation) is now named as the reason the prefixed form must be accepted.

**Test gaps.** Both added:
- `test_schema_node_validates_namespaced_names_at_nesting_depth` — name validation recurses through `fields` and `item_fields`, previously uncovered at depth. Includes a malformed nested name to pin that rejection also reaches depth.
- `test_schema_node_title_template_resolves_namespaced_field` — the seam with the title-template work: `{custom:contact_email}` must validate against a namespaced field, and bare `{contact_email}` must not.

Both negative assertions were mutation-checked (inverting each input fails the test), so neither passes vacuously.

**Explicit-fields path.** Filed as #1785 rather than fixed here, per the review's recommendation. Worth noting the issue does *not* presuppose the fix direction: making the explicit path namespace matches the CLAUDE.md rule but forces `custom:` into every agent-authored template and query, while dropping namespacing from the description path is less churn and keeps the agent contract stable. The issue asks for that convention to be decided and recorded as an ADR before implementation.

core 1019 passed, clippy clean, pre-push gate green.

PR #1784, opened 2026-07-28
Follow-up to #1775: title_template regression tests do not fail on the defect they claim to guard

malibio:
Closes #1783

### DIFF
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -995,11 +995,12 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
             // Otherwise loop back for another inference round
         }
 
-        // Reached only via the duplicate-call guard's `break` above. The
-        // max-iteration path (iteration == effective_max_iterations - 1) always
-        // returns early and never falls through here. Run one final text-only
-        // inference so the session always produces a response from the tool
-        // results already in history.
+        // Reached via either `break` above: the duplicate-call guard, or the
+        // consecutive-parse-failure guard. The max-iteration path
+        // (iteration == effective_max_iterations - 1) always returns early and
+        // never falls through here. Run one final text-only inference so the
+        // session always produces a response from the tool results already in
+        // history.
         on_status(LocalAgentStatus::Thinking);
 
         let mut messages = vec![ChatMessage::text(Role::System, system_content.clone())];
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -1784,7 +1784,9 @@ impl GraphToolExecutor {
         let result = handle_create_nodes_from_markdown(&ns, handler_args)
             .await
             .map_err(|e| {
-                ToolError::ExecutionFailed(format!("create_nodes_from_markdown failed: {:?}", e))
+                // Display, not Debug — MarkdownError's Display text is written to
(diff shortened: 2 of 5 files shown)


malibio:
## Code review: APPROVE (with one finding, now addressed)

Reviewed by `/pragmatic-code-review`. The reviewer was explicitly asked not to trust the authors claim that the ratchet works, since the author had already been wrong once about these exact assertions.

### Independently verified

The reviewer reconstructed the diverging pre-check from its original commit, re-inserted it at the original call site, and ran the suite:

```
test ..._empty_placeholder_rejected ... FAILED
  ... a different message here means a second validator spoke first:
  invalid params: title_template "{} {status}" references {status}, which is not
  defined in "fields". Either add {"name":"status","type":"text"} ...
test ..._unclosed_placeholder_rejected ... ok
```

**The ratchet fires, for exactly the claimed reason.** The file was then restored; worktree confirmed clean.

### Three things it tried to break and could not

- **Asserted literals** match `validate_template_tokens` byte-for-byte, including the `'{'` quoting. A typo would have made the "tightened" assertion weaker than the loose one it replaced — this was the highest-risk item and it is clean.
- **`fields: []` is not arbitrary.** It deserializes to `Some(vec![])` and takes the explicit-fields branch, whereas an absent key is `None` and falls to the description branch. Different paths; the chosen fixture is the one that mirrors what the pre-check actually read.
- **The "un-pinnable" claim on the unclosed case is a correct structural observation**, not a cop-out. Both validators return the same error variant, so message text is the only signal; anything stronger would need production-visible test machinery.

### 🟡 Finding — addressed in d3ca89be

The sweep was incomplete: three `MarkdownError` sites in `markdown/mod.rs` still produced `invalid params: Invalid parameters: ...`, one of them in `handle_create_nodes_from_markdown` — the very function this branch already touches from the `tools.rs` side. That is the same "one function away" argument this branch accepted for `schema/mod.rs`, so leaving them was inconsistent.

Fixed. Deliberately left, having checked each:
- `node_batch.rs:64/114` — plain `String` with no Display prefix, so the text is load-bearing.
- `ops/mod.rs:40` — thiserror attribute on a different error type.
- `markdown/mod.rs:1194` — "Invalid collection path", a distinct message, not a doubled preamble.

No `MarkdownError` construction now doubles the preamble.

### Re-review decision

**NO RE-REVIEW NEEDED.** `d3ca89be` is a mechanical string-prefix change with no behavioural or logic change, which is squarely the "trivial, low-risk" case. The reviewed commit `d43b7a24` is unchanged.

core 1013 passed · agent 309 passed · frontend 4104 passed · clippy clean · full push gate passed.

malibio:
Rebased onto `main` after #1777 landed (`dc3a2888`). Clean rebase, no conflicts — but it did change one test's meaning, so one commit added.

`test_create_schema_without_fields_key_is_unaffected` was written while description-only `create_schema` was broken by the `custom:` prefix bug. It deliberately asserted only that the field-locating pass stayed out of the way, guarded behind `if let Err`, so it would not couple to that bug's fix.

#1777 fixes that path — which makes the guard **vacuous**: the `Err` branch never runs, so the test passes without asserting anything. That is exactly the failure mode this PR exists to correct, so leaving it would be self-defeating. It now asserts the call succeeds and returns `schemaId: "venue"`, and the stale "separately broken on main" comment is gone.

Reviewed commit `d43b7a24` (now `751d149c`) is unchanged by the rebase apart from its new base.

core **1019** passed (up from 1013 — #1777's tests now included) · agent 309 · clippy clean · full push gate passed.

PR #1794, opened 2026-07-29
create_schema description route folds the type keyword into the inferred field name

malibio:
Closes #1789

### DIFF
--- a/packages/core/src/schema/mod.rs
+++ b/packages/core/src/schema/mod.rs
@@ -1020,7 +1020,8 @@ fn parse_single_field_description(
     // Extract field name (first word or phrase before parentheses/keywords)
     let field_name = extract_field_name(field_desc)?;
 
-    // Infer field type
+    // Infer field type from the full description, before the name is shortened, so a phrase
+    // like "capacity number" still yields `number` once the name becomes `capacity`.
     let field_type = infer_field_type(field_desc, &field_name);
 
     // Extract enum values if present
@@ -1031,6 +1032,20 @@ fn parse_single_field_description(
 
     // Collect warnings
     let mut warnings = Vec::new();
+
+    // A trailing type keyword is an annotation, not part of the name the caller wants stored.
+    let field_name = match strip_trailing_type_keyword(&field_name) {
+        Some((stripped, keyword)) => {
+            warnings.push(format!(
+                "Inferred field name '{}' (type {}) from '{}' — the '{}' keyword was read as the type. \
+                 Use '{}' in title templates, filters and property lookups.",
+                stripped, field_type, field_desc, keyword, stripped
+            ));
+            stripped
+        }
+        None => field_name,
+    };
+
     if field_type == "string" && contains_any(field_desc, &["amount", "price", "cost"]) {
         warnings.push(format!(
             "Field '{}' mentions monetary amount but inferred as string. Consider using number type.",
@@ -1093,6 +1108,77 @@ fn extract_before_keyword(text: &str, keywords: &[&str]) -> Option<String> {
         }
     }
     None
+}
+
+/// Bare type keywords that may be dropped from the tail of an inferred field name.
+///
+/// Deliberately narrower than the vocabulary [`infer_field_type`] recognises: words like
+/// "amount", "price", "count" and "total" also signal a type, but they are legitimate field
+/// names in their own right, so removing them would leave the caller with a worse name than
+/// the one they described.
+const STRIPPABLE_TYPE_KEYWORDS: &[&str] = &[
+    "number", "date", "text", "string", "boolean", "list", "array",
+];
+
+/// Words that make a trailing type keyword part of the field name rather than a type annotation.
+///
+/// "capacity number" describes a capacity held as a number; "invoice number" describes an
+/// identifier that is called a number. The phrases are structurally identical, so the intent
+/// cannot be derived — this table records the cases where the keyword is intrinsic.
+///
+/// "due" is listed for "date" because `due_date` is a core task property: stripping "due date"
+/// to "due" would make it disagree with the "due_date" spelling and would also slip past the
+/// reserved-core-property warning that a collision is supposed to raise.
+const INTRINSIC_KEYWORD_PREFIXES: &[(&str, &[&str])] = &[
+    (
+        "number",
+        &[
+            "invoice",
+            "phone",
+            "account",
+            "serial",
+            "order",
+            "reference",
+            "tracking",
+            "part",
+            "version",
+            "page",
+        ],
+    ),
+    ("date", &["birth", "due"]),
+];
+
+/// Drop a trailing bare type keyword from an inferred field name.
+///
+/// Returns the shortened name and the keyword that was removed, or `None` when the name should
+/// be left alone. Operates on an already-normalized snake_case name, so splitting on '_' is
+/// enough to find word boundaries.
+///
+/// The keyword is only removed when it is the final word, a non-empty name remains, and the
+/// preceding word does not make it intrinsic (see [`INTRINSIC_KEYWORD_PREFIXES`]).
+fn strip_trailing_type_keyword(name: &str) -> Option<(String, String)> {
+    let words: Vec<&str> = name.split('_').filter(|w| !w.is_empty()).collect();
+
+    // A bare keyword is the whole name ("number"): there is nothing left to call the field.
+    if words.len() < 2 {
+        return None;
+    }
+
+    let (keyword, head) = words.split_last()?;
+    if !STRIPPABLE_TYPE_KEYWORDS.contains(keyword) {
+        return None;
+    }
+
+    let preceding = head.last()?;
+    let is_intrinsic = INTRINSIC_KEYWORD_PREFIXES
+        .iter()
+        .find(|(kw, _)| kw == keyword)
+        .is_some_and(|(_, prefixes)| prefixes.contains(preceding));
+    if is_intrinsic {
+        return None;
+    }
+
+    Some((head.join("_"), (*keyword).to_string()))
 }
 
 /// Normalize field name to snake_case
@@ -1440,6 +1526,110 @@ mod tests {
     }
 
     #[test]
+    fn test_strip_trailing_type_keyword() {
+        // A trailing keyword is a type annotation: the caller wants the noun.
+        for (input, expected_name, expected_keyword) in [
+            ("capacity_number", "capacity", "number"),
+            ("score_number", "score", "number"),
+            ("tags_list", "tags", "list"),
+            ("bio_text", "bio", "text"),
+            ("start_date", "start", "date"),
+            ("labels_array", "labels", "array"),
+            // Only the final word is considered, so a multi-word head survives intact.
+            ("max_room_capacity_number", "max_room_capacity", "number"),
+        ] {
+            assert_eq!(
+                strip_trailing_type_keyword(input),
+                Some((expected_name.to_string(), expected_keyword.to_string())),
+                "'{input}' should strip its trailing type keyword"
+            );
+        }
+
+        // Names where the keyword is intrinsic, not an annotation.
+        for input in [
+            "invoice_number",
+            "phone_number",
+            "account_number",
+            "serial_number",
+            "order_number",
+            "tracking_number",
+            "part_number",
+            "page_number",
+            "date_of_birth",
+            // `due_date` is a core task property; stripping it would both diverge from the
+            // "due_date" spelling and dodge the reserved-core-property warning.
+            "due_date",
+        ] {
+            assert_eq!(
+                strip_trailing_type_keyword(input),
+                None,
+                "'{input}' keeps its keyword — it is part of the intended name"
+            );
+        }
+
+        // Nothing to strip: the keyword is the entire name, or there is no keyword.
+        for input in ["number", "date", "capacity", "customer_name", ""] {
+            assert_eq!(
+                strip_trailing_type_keyword(input),
+                None,
+                "'{input}' has no strippable trailing keyword"
+            );
+        }
+
+        // A keyword away from the tail is an ordinary word.
+        assert_eq!(strip_trailing_type_keyword("number_of_guests"), None);
+    }
+
+    /// The description route used to fold the type keyword into the stored name, so
+    /// "capacity number" produced a field called `capacity_number` while the explicit-fields
+    /// route produced `capacity` for the same intent. Stored names are user-visible keys, so
+    /// the keyword is now read as the type only.
+    #[test]
+    fn test_parse_field_descriptions_strips_type_keyword() {
+        let fields = parse_field_descriptions("capacity number");
+
+        assert_eq!(fields.len(), 1);
+        assert_eq!(fields[0].name, "capacity");
+        // The type still comes from the full phrase, even though the name no longer carries it.
+        assert_eq!(fields[0].field_type, "number");
+
+        // The caller is told which key was stored, so later title templates and filters can
+        // reference it without guessing.
+        assert!(
+            fields[0]
+                .warnings
+                .iter()
+                .any(|w| w.contains("capacity") && w.contains("capacity number")),
+            "Expected a warning naming the stored field and its source phrase, got {:?}",
+            fields[0].warnings
+        );
+    }
+
+    /// Naive stripping would turn "invoice number" into `invoice` — a different wrong answer.
+    #[test]
+    fn test_parse_field_descriptions_keeps_intrinsic_type_keyword() {
+        let fields = parse_field_descriptions("invoice number, phone number, due date");
+
+        assert_eq!(fields.len(), 3);
+        assert_eq!(fields[0].name, "invoice_number");
+        assert_eq!(fields[1].name, "phone_number");
+        assert_eq!(fields[2].name, "due_date");
+
+        // Nothing was rewritten, so nothing is reported as rewritten.
+        for field in &fields {
+            assert!(
+                !field
+                    .warnings
+                    .iter()
+                    .any(|w| w.contains("was read as the type")),
+                "'{}' should not report a stripped keyword, got {:?}",
+                field.name,
+                field.warnings
+            );
+        }
+    }
+
+    #[test]
     fn test_parse_field_descriptions_simple() {
         let desc = "invoice number, amount, status";
         let fields = parse_field_descriptions(desc);
@@ -1612,12 +1802,13 @@ mod tests {
         // Stored names are bare — no namespace prefix is applied during inference
         assert!(normalized.iter().all(|f| !f.name.contains(':')));
 
-        // Verify field names are present
+        // Exact names, not substrings: these are the keys title templates and filters use,
+        // and a `contains` assertion would pass whether or not a type keyword was folded in.
         assert_eq!(normalized.len(), 4);
-        assert!(normalized[0].name.contains("invoice"));
-        assert!(normalized[1].name.contains("amount"));
-        assert!(normalized[2].name.contains("status"));
-        assert!(normalized[3].name.contains("due"));
+        assert_eq!(normalized[0].name, "invoice_number");
+        assert_eq!(normalized[1].name, "amount");
+        assert_eq!(normalized[2].name, "status");
+        assert_eq!(normalized[3].name, "due_date");
 
         // Verify types are inferred correctly (following priority order)
         assert_eq!(normalized[0].field_type, "number"); // "number" keyword in "invoice number"
--- a/packages/core/src/schema/mod.rs
+++ b/packages/core/src/schema/mod.rs
@@ -1038,7 +1038,8 @@ fn parse_single_field_description(
         Some((stripped, keyword)) => {
             warnings.push(format!(
                 "Inferred field name '{}' (type {}) from '{}' — the '{}' keyword was read as the type. \
-                 Use '{}' in title templates, filters and property lookups.",
+                 Use '{}' in title templates, filters and property lookups. To store a different \
+                 name, pass \"fields\" explicitly instead of a description.",
                 stripped, field_type, field_desc, keyword, stripped
             ));
             stripped
@@ -1126,9 +1127,10 @@ const STRIPPABLE_TYPE_KEYWORDS: &[&str] = &[
 /// identifier that is called a number. The phrases are structurally identical, so the intent
 /// cannot be derived — this table records the cases where the keyword is intrinsic.
 ///
-/// "due" is listed for "date" because `due_date` is a core task property: stripping "due date"
-/// to "due" would make it disagree with the "due_date" spelling and would also slip past the
-/// reserved-core-property warning that a collision is supposed to raise.
+/// Recoverability is asymmetric, so this errs towards keeping: a missed strip leaves a slightly
+/// clumsy name, while a wrong strip changes a user-visible key that title templates, CEL
+/// selectors and query filters all resolve against. The "number" group is the larger one because
+/// those nouns *have* a number rather than *being* one.
 const INTRINSIC_KEYWORD_PREFIXES: &[(&str, &[&str])] = &[
     (
         "number",
@@ -1143,6 +1145,18 @@ const INTRINSIC_KEYWORD_PREFIXES: &[(&str, &[&str])] = &[
             "part",
             "version",
             "page",
+            "model",
+            "ticket",
+            "flight",
+            "id",
+            "security",
+            "registration",
+            "license",
+            "batch",
+            "lot",
+            "room",
+            "seat",
+            "case",
         ],
     ),
     ("date", &["birth", "due"]),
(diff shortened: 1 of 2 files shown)


malibio:
## Code Review: PR #1794 — Read trailing type keywords as types, not part of the field name

**Scope:** commit `271d8b31` (2 files, +295/-10). `origin/main` has advanced to `7758badd` (PR #1791) — that commit is unrelated and excluded.
**Verification performed:** `cargo test -p nodespace-core --lib schema` → **126 passed, 0 failed** (baseline 121). I also extracted `strip_trailing_type_keyword` / `extract_field_name` / `infer_field_type` into a standalone harness and ran ~45 realistic descriptions through the full pipeline to probe the edges; results are quoted below.

---

### Requirements Check

| # | Acceptance criterion | Status |
|---|---|---|
| 1 | Decide the inference rule, incl. names where the keyword is intrinsic | ✅ Rule is decided and documented at the point of definition (`STRIPPABLE_TYPE_KEYWORDS` + `INTRINSIC_KEYWORD_PREFIXES`), with the rationale for the narrower strip vocabulary written down |
| 2 | `"capacity number"` infers name `capacity`, type `number` | ✅ Verified end-to-end through `handle_create_schema` in `test_create_schema_paths_agree_on_stored_field_names_with_type_keyword`, and directly in my harness |
| 3 | Intrinsic keywords not mangled — `"invoice number"` stays `invoice_number` | ⚠️ Partially. Holds for the 12 listed prefixes; fails open for everything else (`model number` → `model`, `ticket number` → `ticket`, `flight number` → `flight`, `social security number` → `social_security`). See 🟡-1 |
| 4 | A test asserts both routes agree on the stored name using a description containing the type keyword | ✅ `schema_test.rs:1210` — asserts on the **stored node** via `get_schema_node`, not the return value, and additionally pins the type to prove the keyword was *read* rather than merely discarded |
| 5 | Existing `parse_field_descriptions` / `normalize_field_name` tests updated to the decided rule | ✅ And genuinely **strengthened**, not weakened — see the Testing note below |

---

### Assessment

This is a **net positive** and I'd merge it in its current shape. The core judgement is right and unusually well argued: the commit message and the doc comments both state plainly that `"capacity number"` and `"invoice number"` are *structurally identical*, so no amount of parsing sophistication recovers the intent — a lookup table is the honest mechanism, not a lazy one. The guards are the right three (final word only, non-empty remainder, keep-list), the strip vocabulary is deliberately narrower than `infer_field_type`'s (correctly excluding `amount`/`price`/`count`/`total`, which are legitimate names), and type inference deliberately runs on the full description *before* the name is shortened so the keyword informs the type and then disappears from the name. `strip_trailing_type_keyword` is a pure, total, separately-testable function with no allocation surprises — exactly the right seam.

I verified the two claims the description makes that a reader would have to take on trust, and both hold:

- **`due` keep-list entry:** confirmed correct. `RESERVED_CORE_PROPERTIES` (`mod.rs:17-29`) contains **both** `"due_date"` and `"due"`, and `due_date` is a real hardcoded task property (`packages/core/src/behaviors/mod.rs:724`). So stripping `"due date"` → `due` would have produced a name that (a) disagrees with the `due_date` spelling the UI reads and (b) *still* collides with a reserved property — the entry earns its place. Good catch by the author.
- **Warnings actually reach the caller:** confirmed. `parse_single_field_description` pushes onto `InferredField.warnings`, `normalize_inferred_fields:1407` extends the outer list with them, and `handle_create_schema:345` emits them in the response. Not a dead channel.

My findings are about the *blast radius* of the rule, not its correctness. Nothing below blocks merge.

---

### Findings

#### 🟡 Important

**1. `📁 packages/core/src/schema/mod.rs:1163-1180` — Stripping can now *manufacture* a reserved-core-property collision, and the keep-list only defends one of the five cases.**

The `due` entry exists precisely because stripping would produce a reserved name. But that hazard is not unique to `due` — it's structural, and the other instances are unguarded. Verified through the real pipeline:

```
"status text"   -> status     (RESERVED_CORE_PROPERTIES)
"priority text" -> priority    (RESERVED_CORE_PROPERTIES)
"content text"  -> content     (RESERVED_CORE_PROPERTIES)
"id number"     -> id          (RESERVED_CORE_PROPERTIES)
```

`"status text"` is an extremely plausible thing for an LLM agent to write. Before this change it stored `status_text` (no collision); now it stores `status` (collides with a core task property).

The mitigating fact — and the reason this is 🟡 not 🔴 — is that the reserved check at `mod.rs:1410` runs on `stored_name` *after* the strip, so the caller does get the "matches a reserved core property" warning. The system degrades to "warns loudly" rather than "silently shadows". That is acceptable behaviour, and arguably stripping-then-warning is *more* correct than the old fold-and-hide.

The design smell is that `due` is handled by a hand-written keep-list entry while four structurally identical cases are handled by a different mechanism (the downstream warning). Consider expressing the invariant once instead of enumerating instances — i.e. in `strip_trailing_type_keyword`, decline the strip when the shortened name would land in `RESERVED_CORE_PROPERTIES`, which subsumes the `due` entry and covers `status`/`priority`/`content`/`id` for free:

```rust
let candidate = head.join("_");
if RESERVED_CORE_PROPERTIES.contains(&candidate.as_str()) {
    return None; // shortening would collide with a core property
}
```

*Principle: DRY / single source of truth for an invariant.* One rule ("never shorten into a core property") replaces one table entry plus four unhandled cases, and it stays correct automatically when someone adds a core property later — the current keep-list would silently rot.

**2. `📁 packages/core/src/schema/mod.rs:1136-1152` — The keep-list is the right mechanism, but its omissions are load-bearing and the failure mode isn't cheaply recoverable.**

To be clear: I think a keep-list *is* correct here, for the reason the doc comment gives — the intent genuinely isn't in the phrase. My concern is calibration, not approach. Verified misses:

```
"model number"           -> model
"ticket number"          -> ticket
"flight number"          -> flight
"social security number" -> social_security
"id number"              -> id
"release date"/"start date"/"end date"/"expiry date"/"hire date"
                         -> release / start / end / expiry / hire
```

The `*_date` group is worth separating from the `*_number` group. For dates the strip is arguably **right** — `start`, `end`, `expiry` are perfectly good field names for date-typed fields, and `start_date` vs `start` is a naming-taste question. For `*_number` the strip is **wrong** whenever the noun names an identifier: `model` and `ticket` and `flight` are not numbers, they're things that *have* numbers. `model_number` → `model` on a Product schema is a genuinely bad key.

Worth noting the asymmetry in recoverability, because it's what makes this 🟡 rather than 🟢: a *missed* strip (`invoice_number` kept) is cosmetic and the caller can live with it; a *wrong* strip (`model_number` → `model`) changes a user-visible key that title templates, CEL selectors and query filters resolve against, so fixing it later is a breaking change to every call site (per CLAUDE.md's own rule on stored names). The rule's failure mode is therefore biased toward the more expensive direction. At minimum `model`, `ticket`, `flight`, `id`, `badge`, `license`, `registration` look like they belong on the list.

Maintainability of a hardcoded list: acceptable *provided* it stays a small curated table with the "why" comment it already has. It will drift, but the cost of drift is bounded and the alternative (asking the LLM to disambiguate, or requiring explicit `fields`) is more machinery than this deserves right now. **YAGNI applies — don't build a classifier.** Just widen the table.

**3. `📁 packages/core/src/schema/mod.rs:1037-1044` — Warning fires on every strip, including the unambiguous ones; and it's phrased as an explanation rather than an instruction.**

The message is consumed by an LLM agent, and it's emitted unconditionally on every successful strip. For a description like `"capacity number, score number, guest list"` the caller gets three near-identical warnings for three cases where the inference was simply *correct*. Warning channels that fire on correct behaviour get ignored, which erodes the channel for the reserved-property warning that shares it and actually needs attention.

*Principle: signal-to-noise — a warning that fires on success is not a warning.*

The content is good — naming the stored key and telling the caller to use it in templates/filters is exactly the actionable part. Two options, either is fine:
- Drop the warning for strips with no plausible ambiguity and keep it only where the head noun is identifier-shaped, or
- Keep it but lead with the imperative: `"Stored field 'capacity' (number) — use 'capacity' in title templates, filters and property lookups. Read 'number' in 'capacity number' as the type; if you meant a field literally named 'capacity_number', pass it in `fields` explicitly."`

The second half of that suggestion matters: right now the warning tells the caller *what happened* but not *how to override it*. Naming the escape hatch (explicit `fields`) makes it actionable.

#### 🟢 Suggestions

**4. `📁 packages/core/src/schema/mod.rs:1163` — `words.len() < 2` and the later `head.last()?` overlap.**
The `len() < 2` guard already ensures `head` is non-empty, so `head.last()?` can never return `None`. Harmless and arguably good defensive style, but the `?` reads as if a second failure mode exists. `Nit:` if you touch this again, the intent is slightly clearer as an `expect`-free destructure or a comment noting the guard makes it infallible.

**5. `📁 packages/core/src/schema/mod.rs:1171` — `.find(...).is_some_and(...)` could be `any`.**
`INTRINSIC_KEYWORD_PREFIXES.iter().any(|(kw, prefixes)| kw == keyword && prefixes.contains(preceding))` is one combinator instead of two and reads closer to the sentence it implements. `Nit:` purely stylistic.

**6. Repeated / all-keyword names behave sanely — no action needed, recording it as verified.**
I probed the degenerate cases from the review brief:
```
"date date"     -> date   (type date)      "number number" -> number (type number)
"text list"     -> text   (type array)     "list"          -> list   (unchanged)
"capacity  number" (double space) -> capacity   (normalizer collapses it)
"序号 number" / "größe number"      -> 序号 / größe  (unicode head preserved)
```
No panics, no empty names, no unicode mangling — `normalize_field_name` uses `char::is_alphanumeric` which is unicode-aware, and splitting on `_` is safe on UTF-8 because `_` is ASCII and can't occur inside a multi-byte sequence. `"text list"` → `text` typed `array` is odd but it's a nonsense input and the old behaviour (`text_list`) wasn't better.

**7. Interaction with `extract_before_paren` / `extract_before_keyword` is correct — verified.**
```
"capacity number (required)" -> capacity, type number
"amount number in USD"       -> amount,   type number
```
Both extraction paths hand `strip_trailing_type_keyword` an already-normalized snake_case name, so the ordering holds. One genuinely odd case, pre-existing and not caused by this PR: `"capacity number of seats"` → `capacity_number_of` (the 3-word `take(3)` fallback in `extract_field_name` truncates mid-phrase, leaving a trailing `of` that blocks the strip). Ugly, but it was ugly before and fixing `take(3)` is a separate concern.

**8. Re: the known pre-existing `contains_any` substring bug — it does interact, but not harmfully.**
`"notes text"` now yields name `notes` with type `boolean` (because `"notes"` contains `"no"`). Confirmed this misbehaves identically on `main`; this PR doesn't cause or worsen it. Worth noting only because the strip makes the *name* correct while the *type* stays wrong, which could make the bug read as newly-introduced when someone hits it. Not in scope here — the existing follow-up should stand.

---

### Testing

Tests are the strongest part of this PR. Specifically worth calling out:

- **No existing test was weakened.** `test_integration_full_schema_creation` (`mod.rs:1804-1810`) was *tightened* — four `assert!(name.contains(...))` substring checks became exact `assert_eq!`s, with a comment explaining that a `contains` assertion would have passed whether or not the keyword was folded in. That's the assertion that let the original bug hide, and replacing it is the right fix.
- The cross-route agreement test asserts on the **persisted node** (`get_schema_node`) rather than the response payload, which is what actually matters for user-visible keys.
- `test_create_schema_description_keeps_intrinsic_type_keyword` pins both halves in one schema (`"invoice number, capacity number"` → `invoice_number` + `capacity`), so a future over-eager strip can't pass by regressing only one side.
- The negative test asserts the *absence* of a strip warning for intrinsic names — a nice touch, since it catches "stripped then re-added" implementations.
- The main agent's mutation check (disabling the strip fails exactly the 3 new tests while keep-list tests still pass) is the right way to prove non-vacuity, and matches what I'd expect from reading them.

Gap: no test covers the strip-into-reserved-name interaction from 🟡-1. If you act on that finding, `"status text"` is the case to pin.

### Standards Compliance

✅ No lint suppression, no `#[allow(dead_code)]`, no `#[deprecated]`, no backward-compat shims or dual-mode paths — the old behaviour is simply gone, per the greenfield rule. ✅ No issue numbers in code comments (the `#1789` reference is confined to the commit message, where it belongs; the doc comments describe the constraint directly). ✅ No new dependencies. ✅ Rust-only change, no logging concerns. ✅ Change is atomic — one cohesive purpose, no unrelated edits. ✅ No security surface: pure string manipulation on already-validated input, no injection/authz/secrets dimension, bounded allocation.

### Performance

Negligible. `strip_trailing_type_keyword` allocates one `Vec<&str>` and at most two `String`s per field, on a schema-creation path that already does DB I/O. Not worth optimizing.

---

### Recommendation: **COMMENT** — approve-with-suggestions

The change is correct, well-reasoned, well-tested, and strictly better than what it replaces. None of the findings are blockers: 🟡-1's bad case still produces a loud warning rather than silent shadowing, 🟡-2 is a table that wants a few more rows, and 🟡-3 is polish on a message. My one recommendation if you address anything before merge is **🟡-1** — replacing the `due` special case with a general "don't shorten into a reserved core property" check is a small edit that removes a class of bugs rather than an instance, and it makes the `due` entry unnecessary. **🟡-2** (adding `model`/`ticket`/`flight`/`id` to the number keep-list) is cheap insurance against the expensive failure direction. Both are equally fine as follow-ups.

*(Posted as a comment rather than an approval because GitHub rejects self-approval — the author and reviewer are the same account.)*

malibio (commented):
Initial pragmatic code review posted above: **COMMENT** (approve-with-suggestions). Net positive, no blockers. Verified 126/126 schema tests pass and independently probed ~45 edge cases through an extracted copy of the inference pipeline.

Three 🟡 findings, all non-blocking:
1. Stripping can manufacture a reserved-core-property collision (`"status text"` → `status`, also `priority`/`content`/`id`). The `due` keep-list entry solves exactly this for one word; a general "don't shorten into RESERVED_CORE_PROPERTIES" check would subsume it and cover the rest.
2. Keep-list omissions bias toward the expensive failure direction (`model number` → `model`, `ticket`/`flight`/`id`). A wrong strip breaks user-visible keys; a missed strip is cosmetic.
3. Warning fires on every strip, including unambiguous ones, and doesn't name the override (explicit `fields`).

Confirmed the two load-bearing claims: `due_date` is a real core task property (behaviors/mod.rs:724) and both `due`/`due_date` are in RESERVED_CORE_PROPERTIES; and the warnings channel does reach the create_schema response. No existing test was weakened — `test_integration_full_schema_creation` was tightened from substring to exact assertions, which is precisely the assertion that let the original bug hide.

malibio:
## Code Review: PR #1794 — Re-Review of `6802414a`

**Review Type:** Re-Review
**Scope:** commit `6802414a` only (1 file, +63/-14). `271d8b31` was covered in the [previous review](https://github.com/NodeSpaceAI/nodespace-core/pull/1794#issuecomment-5111308031); unrelated `origin/main` commits excluded.
**Verification performed:** re-ran `cargo test -p nodespace-core --lib schema` → **127 passed, 0 failed** (confirms the claim; 126 after `271d8b31`, 121 at baseline). Re-extracted the post-commit `strip_trailing_type_keyword` / `extract_field_name` / `infer_field_type` / `normalize_field_name` / `RESERVED_CORE_PROPERTIES` into a standalone harness and re-ran the full parse pipeline over ~55 inputs, including every case I flagged last time. Independently ran **two mutation checks** — one on the new guard, one on the widened keep-list. Working tree restored and re-verified clean (`git status` empty, 127 passing) after both.

---

### Previous Review Summary

Five actionable recommendations were made (🟡-1, 🟡-2, 🟡-3, 🟢-4, 🟢-5). Findings 6, 7 and 8 were recorded-as-verified observations requiring no action.

| # | Recommendation | Status |
|---|---|---|
| 🟡-1 | Express the "don't shorten into a reserved core property" invariant once, instead of enumerating `due` by hand | ✅ **Addressed, exactly as suggested** — verified load-bearing by mutation |
| 🟡-2 | Widen the `number` keep-list (`model`/`ticket`/`flight`/`id`/…) | ✅ **Addressed** (12 entries added) — but ⚠️ **entirely untested**, see 🟡-A |
| 🟡-3 | Warning should name the escape hatch; consider not firing on unambiguous strips | ◐ **Half addressed** — escape hatch named (good); every-strip firing deliberately retained. Rebuttal assessed below, I **accept it** |
| 🟢-4 | Collapse redundant `len() < 2` / `head.last()?` guards | ✅ **Addressed** |
| 🟢-5 | `.find().is_some_and()` → `.any()` | ✅ **Addressed** |

**New issues introduced by this commit:** 1 (🟡-A — test coverage gap on the new keep-list entries). No behavioural regressions found.

---

### Requirements Check (issue #1789)

| # | Acceptance criterion | Status |
|---|---|---|
| 1 | Decide the inference rule, incl. names where the keyword is intrinsic | ✅ Rule now stated as three composable guards (final-word, keep-list, reserved), each documented at its definition |
| 2 | `"capacity number"` infers name `capacity`, type `number` | ✅ Re-verified end-to-end through the pipeline harness and via `test_create_schema_paths_agree_on_stored_field_names_with_type_keyword` |
| 3 | Intrinsic keywords not mangled — `"invoice number"` stays `invoice_number` | ✅ **Upgraded from ⚠️ to ✅.** All four cases I reported failing now hold: `model number` → `model_number`, `ticket number` → `ticket_number`, `flight number` → `flight_number`, `social security number` → `social_security_number` |
| 4 | A test asserts both routes agree on the stored name using a description containing the type keyword | ✅ Unchanged and still asserts on the persisted node |
| 5 | Existing tests updated to the decided rule | ✅ Extended, not weakened |

All five criteria now met. Criterion 3 was the one I could only pass partially last time.

---

### Assessment

**This is the right fix, and it's the version of 🟡-1 I'd hoped for rather than the cheap one.** The author took the general invariant over the enumeration, and the commit message argues the choice on its merits rather than citing the review. Verified through the real pipeline that all four previously-unguarded cases are closed and that the strip is declined rather than performed:

```
"status text"    -> status_text     (was: status)
"priority text"  -> priority_text   (was: priority)
"content text"   -> content_text    (was: content)
"id number"      -> id_number       (was: id)
```

I also checked the cases I had *not* probed, and the general guard covers them for free — which is precisely the argument for preferring it: `node type text` → `node_type_text`, `parent id text` → `parent_id_text`, `root id text` → `root_id_text`, `created at date` → `created_at_date`, `modified at date` → `modified_at_date`. The hand-enumerated alternative would have missed all five. **The DRY argument paid off immediately, not hypothetically.**

**Mutation check on the guard (independent of the main agent's):** replacing `if RESERVED_CORE_PROPERTIES.contains(...)` with `if false` fails exactly two tests — `test_parse_field_descriptions_does_not_strip_onto_reserved_property` (mod.rs:1652) and `test_strip_trailing_type_keyword` (mod.rs:1599). Confirms the guard is load-bearing and the new tests are **not vacuous**. The pipeline-level test is the one that matters: it exercises `parse_field_descriptions` rather than the private helper, so it pins the *behaviour* and would survive a refactor that moved the check elsewhere.

---

### Answering the three questions put to this re-review

**1. Is declining the strip the right response, versus stripping and letting the reserved warning fire?**

**Declining is correct**, and for a sharper reason than "avoids a collision". The two situations are not symmetric:

- `create_schema({description: "status"})` — the caller *described* a reserved name. Warning is right: report what they asked for, leave the choice with them (`normalize_inferred_fields`'s own doc comment commits to exactly this).
- `create_schema({description: "status text"})` — the caller described `status_text`, which is **not** reserved. A strip-then-warn design would have the tool *create* the problem and then bill the caller for it. The warning would be truthful but the collision would be an artifact of our inference, not of their input.

Declining keeps the invariant "we never make a caller's name worse than they wrote it", which is the same asymmetry-of-recoverability argument the keep-list rests on. The two mechanisms now share one rationale instead of two. I'm satisfied this is the better of the two designs.

**2. Does it interact correctly with the reserved warning at `normalize_inferred_fields:1429`? Is there a path where the caller gets no signal?**

Interaction verified correct — the guard and the warning are **disjoint by construction**. The guard fires only when the *shortened* name is reserved; the warning fires only when the *stored* name is reserved. When the guard declines, the stored name is the unshortened one, which by definition is longer than the reserved candidate and therefore not itself reserved. So there is no double-warn, no suppressed warn, and no ordering hazard. Confirmed empirically — `"status text"` produces **zero** warnings, `"status"` produces exactly one, `"due date"` produces exactly one (unchanged from `271d8b31`; the `due` keep-list entry already declined that strip, so this is not a regression).

**On "is there now a path where the caller gets no signal about an unexpected stored name?" — no, because the stored name is no longer unexpected.** When the guard declines, the stored name is what the caller literally wrote, normalized (`status text` → `status_text`). Silence is the correct signal: nothing surprising happened. The signal is owed when we *change* the name, and that path still warns.

**3. Is the `due` entry in `INTRINSIC_KEYWORD_PREFIXES` now dead code?**

**No — it is genuinely load-bearing, and the author's instinct to keep it was right even though the stated reason is the weaker one.** I tested this directly rather than reasoning about it. The guard checks `head.join("_")`, so it only subsumes the `due` entry when the head is *exactly* `due`. Multi-word heads ending in `due` escape it. Simulating removal of the entry:

```
payment_due_date  -> Some("payment_due")   // guard does NOT catch: "payment_due" not reserved
invoice_due_date  -> Some("invoice_due")   // likewise
rent_due_date     -> Some("rent_due")      // likewise
due_date          -> None                  // guard DOES catch this one
```

With the entry present, all four correctly return `None`. So `due` covers a strictly larger set than the reserved guard does. `payment_due_date` → `payment_due` would be a real bad key on an Invoice schema. Keeping it is correct; the doc comment's framing ("documents intent independently") undersells it — it's not redundant documentation, it's the only thing handling the multi-word-head case. `Nit:` worth a word in the comment so a future reader doesn't delete it as dead code after making the same first-order analysis I did. `birth` is likewise live (`date_of_birth` is handled by word-order, `birth_date` by this entry).

**Corollary worth noting:** the same multi-word-head limitation applies generally — `customer_id_number`, `order_id_number` and `user_id_number` all correctly return `None`, but via the `id` keep-list entry, **not** via the reserved guard (`customer_id` is not reserved). Adding `id` to the keep-list was therefore not redundant with the guard either. The two mechanisms genuinely compose.

---

### Findings

#### 🟡 Important

**A. `📁 packages/core/src/schema/mod.rs:1148-1159` — The 12 new keep-list entries are entirely untested. Removing all of them fails zero tests.**

This is the one new issue. I ran the same mutation check on the widened list that I ran on the guard: deleting `model`, `ticket`, `flight`, `security`, `registration`, `license`, `batch`, `lot`, `room`, `seat`, `case` from `INTRINSIC_KEYWORD_PREFIXES` and re-running the suite gives **127 passed, 0 failed** — an unchanged green suite. (`id` is separately covered by the new reserved-property test, so I excluded it from the deletion set.)

`test_strip_trailing_type_keyword`'s intrinsic block (mod.rs:1569-1581) still lists only the *original* entries — `invoice`, `phone`, `account`, `serial`, `order`, `tracking`, `part`, `page` — and was not extended alongside the table. So the fix for 🟡-2 landed in the code but not in the test, and nothing prevents a future edit from silently reverting it.

*Principle: a change worth making is worth pinning — untested data tables are the ones that rot.* The irony is sharp here: 🟡-2 was raised **because** the original list's omissions were load-bearing and expensive to fix later, and the new entries now carry that same exposure. This is also the cheapest possible fix — the test already loops over a slice, so it's adding strings to an existing array, not writing a new test:

```rust
"model_number", "ticket_number", "flight_number", "registration_number",
"license_number", "batch_number", "room_number", "seat_number", "case_number",
```

Not a merge blocker (the behaviour is correct today, and I verified all 12 work through the real pipeline), but it's the highest-value follow-up remaining on this PR and it costs one line.

#### 🟢 Suggestions

**B. `📁 packages/core/src/schema/mod.rs:1155-1157` — `room`, `seat`, `case`, `lot` block the `number` strip only; the same nouns still strip freely against other keywords.**

Checking the over-blocking risk raised in the brief: the keep-list is keyed on `(keyword, prefix)` pairs, so widening the `number` group has **no effect** on any other keyword. Verified: `room text` → `room`, `case text` → `case`, `case date` → `case`, `batch date` → `batch`, `lot text` → `lot` all still strip correctly. A Venue's `capacity` is untouched — `capacity number` → `capacity` still works, since the pair is `(number, capacity)` and `capacity` is not in the list. So the specific concern about `"room number"` blocking a legitimate `capacity` strip does not materialise; the tuple keying contains the blast radius by design.

The residual judgement call is `case` and `lot`, which are the two genuinely ambiguous additions — a legal-tech schema's `case number` is an identifier (correctly kept), but `lot number` in an auction context could plausibly be an integer the user wants to sort on. Both fail in the recoverable direction (clumsy name, not a wrong key), which is the direction the doc comment explicitly chose. I'd leave them. `Nit:` no action.

**C. `📁 packages/core/src/schema/mod.rs:1039-1043` — 🟡-3 rebuttal: sound, accept as-is.**

The author kept the warning firing on every strip, arguing the stored key differs from what the caller wrote and the agent needs it for subsequent title-template and filter calls. **I withdraw the finding.** My original framing ("a warning that fires on success is not a warning") assumed the reader is a human triaging a log. The consumer here is an LLM composing a *follow-up* call, and for that reader the message isn't a warning about a problem — it's the return value telling it which key to use. `"capacity number"` → `capacity` is a correct inference *and* a fact the agent cannot derive from its own input. Suppressing it on the unambiguous cases would remove information precisely where the agent is most likely to assume nothing happened. The noise cost to the reserved-property warning is real but second-order, and the two messages are textually distinct enough not to blur.

The escape-hatch sentence is a genuine improvement and closes the actionable half of the original finding — the message now says what happened *and* how to override it. `Nit:` the phrasing "instead of a description" slightly overstates the requirement (`fields` and `description` can coexist; only the named field needs to move), but no agent is going to be misled into a broken call by it.

**D. `📁 packages/core/src/schema/mod.rs:1176-1180` — 🟢-4 collapse is correct, and strictly better than what I suggested.**

Folding `words.len() < 2` into `head.is_empty()` on the post-`split_last` value removes the redundancy *and* the double source of truth about what "too short" means, rather than just annotating it. `split_last()?` on an empty vec still handles the `""` input (verified: returns `None`). The surviving `head.last()?` at line 1181 is now genuinely infallible given `head.is_empty()` was checked three lines up — same overlap I flagged, one layer down — but it's inside a `-> Option` function where `?` costs nothing and reads fine. Not worth another pass.

---

### Testing

The pipeline-level test the previous review asked for was added and is the right shape — `test_parse_field_descriptions_does_not_strip_onto_reserved_property` (mod.rs:1649-1656) exercises `parse_field_descriptions`, not the private helper, and covers three distinct reserved names in one call so a partial regression can't pass. The doc comment above it explains *why* the case matters ("manufactures a collision rather than reporting one") rather than restating the assertion. The unit-level cases added to `test_strip_trailing_type_keyword` include `due_date` in **both** the intrinsic block and the new reserved block — deliberate overlap that pins the behaviour from both mechanisms, so removing either one alone is caught.

Gap: 🟡-A above. The guard is well pinned; the keep-list widening is not pinned at all.

### Standards Compliance

✅ No lint suppression, no `#[allow(dead_code)]`, no `#[deprecated]`, no compat shims. ✅ No issue numbers in code comments — the `#1789` reference stays in commit messages where it belongs. ✅ No new dependencies. ✅ Atomic: one cohesive purpose (review follow-up), single file, no drive-by edits. ✅ Comments explain *why* (the "manufactures a collision" comment at 1188-1191 and the asymmetric-recoverability rationale at 1130-1133 are both intent, not mechanics). ✅ No security surface — pure string manipulation, bounded allocation, no injection/authz/secrets dimension. ✅ Commit message argues the design rather than narrating the diff.

### Performance

Unchanged and negligible. The new guard is one `contains` over an 11-element `&[&str]`, on a path that already does DB I/O. The widened keep-list adds 12 entries to a linear scan that runs at most once per field.

---

### Recommendation: **COMMENT** — approve-with-suggestions. Ready to merge.

`6802414a` addresses the substance of every finding from the previous review. 🟡-1 was fixed with the general invariant rather than the enumeration, and I verified it closes five cases beyond the four I originally reported — the DRY argument paid off immediately. 🟡-2 is fixed in behaviour. 🟡-3's rebuttal is sound and I withdraw the finding; the escape-hatch addition closes its actionable half. 🟢-4 and 🟢-5 are done. The `due` entry is not dead code — I tested it, and it covers multi-word heads (`payment_due_date`) that the reserved guard structurally cannot reach.

One new finding, non-blocking: **🟡-A**, the 12 widened keep-list entries have zero test coverage — deleting them leaves the suite green. That's a one-line fix to an existing test loop and is the only thing I'd still do on this PR, but it doesn't hold up the merge given the behaviour is verified correct today. Issue #1789's acceptance criteria are now fully met, including criterion 3 which I could only pass partially before.

*(Posted as a comment rather than an approval because GitHub rejects self-approval — the author and reviewer are the same account.)*

malibio (commented):
Re-review of `6802414a` posted above: all five prior findings addressed (🟡-1 fixed with the general invariant, 🟡-2 fixed in behaviour, 🟡-3 rebuttal accepted and withdrawn, 🟢-4/🟢-5 done). One new non-blocking finding — 🟡-A, the 12 widened keep-list entries have zero test coverage. Ready to merge.

PR #1795, opened 2026-07-29
Rust test gate covers 2 of 9 crates: nodespace-daemon's 92 tests never run automatically

malibio:
Closes #1792

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "review:full": "bun run scripts/review-manager.ts --mode full",
     "review:delta": "bun run scripts/review-manager.ts --mode delta",
     "review:help": "bun run scripts/review-manager.ts --help",
-    "rust:test": "cargo test --lib -p nodespace-core -- --test-threads=2 && cargo test --lib --bins -p nodespace-agent -- --test-threads=2",
+    "rust:test": "cargo test --lib --bins --workspace --exclude nodespace-app --exclude nodespace-app-test-support -- --test-threads=2",
     "rust:test:lib": "cargo test --lib -p nodespace-core -- --test-threads=2",
     "rust:test:long": "RUN_LONG_TESTS=1 cargo test --lib -p nodespace-core -- --test-threads=2",
     "rust:test:doc": "cargo test --doc",
--- a/packages/daemon/src/services/local_agent_service.rs
+++ b/packages/daemon/src/services/local_agent_service.rs
@@ -1971,7 +1971,7 @@ mod tests {
     /// Bounds the swap on its own task rather than awaiting it directly: if the
     /// snapshot timeout regresses away, `replace_engine` never returns, and an
     /// unbounded await here would hang the suite instead of failing it.
-    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[tokio::test]
     async fn engine_swap_completes_when_model_info_hangs() {
         let (mut svc, _node_service, _tempdir) = test_service().await;
         // Drive the timeout path without paying the production bound in
@@ -1981,22 +1981,36 @@ mod tests {
             .expect("sole owner before any clone")
             .model_spec_snapshot_timeout = short;
 
+        // Both calls run on a runtime of their own, on a dedicated OS thread,
+        // and the test waits on a channel. Neither may share a runtime with
+        // this test: if the swap timeout regresses, `replace_engine` never
+        // returns; and if `get_status` regresses to a live engine query, it
+        // blocks its thread outright. Either one, awaited on a shared runtime,
+        // hangs the whole suite rather than failing this test — which is what
+        // it did before, but only under concurrent load, making it a latent
+        // flake. Waiting on a channel cannot be starved by tokio scheduling.
         let swap_svc = svc.clone();
-        tokio::time::timeout(
-            short + std::time::Duration::from_secs(5),
-            tokio::spawn(async move {
+        let (done_tx, done_rx) = std::sync::mpsc::channel();
+        std::thread::spawn(move || {
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("swap runtime");
+            let result = rt.block_on(async move {
                 swap_svc
                     .replace_engine_if_changed("hanging-model", Arc::new(HangingModelInfoEngine))
+                    .await;
+                swap_svc
+                    .get_status(Request::new(GetLocalStatusRequest { session_id: None }))
                     .await
-            }),
-        )
-        .await
-        .expect("engine swap must not block on a stalled model_info")
-        .expect("swap task joins");
+            });
+            // A send failure means the test already timed out and moved on.
+            let _ = done_tx.send(result);
+        });
 
-        let status = svc
-            .get_status(Request::new(GetLocalStatusRequest { session_id: None }))
-            .await
+        let status = done_rx
+            .recv_timeout(short + std::time::Duration::from_secs(5))
+            .expect("engine swap must not block on a stalled model_info")
             .expect("get_status")
             .into_inner();
 
@@ -2042,24 +2056,36 @@ mod tests {
         });
         started_rx.await.expect("generation should take the lock");
 
-        // Run the status call on its own task and bound the join, not the call.
-        // A pre-fix `get_status` blocks its worker thread outright, so a
-        // timeout wrapped directly around the future could be starved of a
-        // thread to fire on; timing out the join keeps the failure clean.
+        // Run the status call on a runtime of its own, on a dedicated OS
+        // thread, and wait on a channel rather than on the task.
+        //
+        // A pre-fix `get_status` blocks the thread it runs on outright. Any
+        // construction that shares a runtime with it — a spawned task, a
+        // timeout wrapped around the call — can have every worker consumed
+        // before the timer is scheduled, which hangs the whole suite instead
+        // of failing this test. Whether that happens depends on what else the
+        // harness is running concurrently, so it is a latent flake in any
+        // shared-runtime form. An isolated runtime cannot be starved by the
+        // blocked call, so the failure stays clean and deterministic.
         let status_svc = svc.clone();
-        let status = tokio::time::timeout(
-            std::time::Duration::from_secs(5),
-            tokio::spawn(async move {
-                status_svc
-                    .get_status(Request::new(GetLocalStatusRequest { session_id: None }))
-                    .await
-            }),
-        )
-        .await
-        .expect("get_status must return while a generation holds the engine lock")
-        .expect("status task joins")
-        .expect("get_status")
-        .into_inner();
+        let (done_tx, done_rx) = std::sync::mpsc::channel();
+        std::thread::spawn(move || {
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("status runtime");
+            let result = rt.block_on(
+                status_svc.get_status(Request::new(GetLocalStatusRequest { session_id: None })),
+            );
+            // A send failure means the test already timed out and moved on.
+            let _ = done_tx.send(result);
+        });
+
+        let status = done_rx
+            .recv_timeout(std::time::Duration::from_secs(5))
+            .expect("get_status must return while a generation holds the engine lock")
+            .expect("get_status")
+            .into_inner();
 
         // Served from the cache, so the real geometry is still reported.
         assert_eq!(status.model_id, "held-model");
--- a/packages/daemon/src/services/local_agent_service.rs
+++ b/packages/daemon/src/services/local_agent_service.rs
@@ -1967,10 +1967,6 @@ mod tests {
     /// A remote engine that stalls on `model_info` must not hold up the swap.
     /// Losing the geometry degrades the status report; blocking here would
     /// block the model-load RPC that awaits `replace_engine`.
-    ///
-    /// Bounds the swap on its own task rather than awaiting it directly: if the
-    /// snapshot timeout regresses away, `replace_engine` never returns, and an
-    /// unbounded await here would hang the suite instead of failing it.
     #[tokio::test]
     async fn engine_swap_completes_when_model_info_hangs() {
         let (mut svc, _node_service, _tempdir) = test_service().await;


malibio:
## Review Type: Initial Review

Scope: one commit (`133ec7ee`), two files, +56/-30. Verified by independent re-execution, including two mutation tests and a load-stress run.

---

## Requirements Check (#1792)

| # | Requirement | Status |
|---|---|---|
| 1 | `rust:test` covers all workspace crates, not 2 of 9 | ✅ Verified — 10 test binaries across 7 crates now build and run |
| 2 | `nodespace-daemon`'s 92 tests actually run in the gate | ✅ Verified — visible in `bun run rust:test` output, 0.94s |
| 3 | `nodespace-nlp-engine` (108), `nodespace-types` (38), `nodespace-cli` (6) covered | ✅ Verified — exact counts reproduced |
| 4 | New crates covered by default (opt-out, not opt-in) | ✅ `--workspace` + `--exclude` inverts the failure mode correctly |
| 5 | Excluding `nodespace-app` loses no coverage | ✅ Verified independently — see below |
| 6 | Total test count 1,335 → 1,583 | ✅ Reproduced exactly: 309+3+6+0+1026+92+1+108+0+38 = **1,583** |

**Independent verification of #5.** I ran `cargo test -p nodespace-app --no-run` and confirmed the ADR-048 pre-push step (`scripts/test-gate.ts:42`) builds and runs **13 targets** — `src/lib.rs` (22 unit tests), `src/main.rs`, and all 11 `tests/*.rs` integration binaries. The exclusion is strictly correct: it is not merely justified by the `--test-threads=1` + `NODESPACED_TEST_BIN` requirement, it also loses nothing, because that step is broader than what `--lib --bins` would have run anyway. Good call.

**`--exclude nodespace-app-test-support` is also correct.** I checked `packages/desktop-app/src-tauri/test-support/src/lib.rs`: zero `#[test]`/`#[tokio::test]` functions and no `mod tests`. My first grep produced false positives from doc comments *about* `#[tokio::test]` semantics; a precise count returns 0. Nothing is skipped. That said, the two exclusions are asymmetric in nature — see 🟢 finding 4.

---

## Code Review Findings

### 🟢 The isolated-runtime + mpsc construction is correct, and deterministic — not merely "less likely" to hang

This was the question most worth scrutinizing, and the construction holds up structurally, not just empirically. The reason it is deterministic rather than probabilistic:

`std::sync::mpsc::Receiver::recv_timeout` blocks the **OS thread** on a condvar with an absolute deadline. It requires no tokio timer, no reactor turn, and no scheduler slot to expire. The previous form (`tokio::time::timeout` around a `tokio::spawn` join) depended on the shared runtime having a free worker to poll the timer — precisely the resource a thread-blocking `std::sync::Mutex` acquisition consumes. The rework removes that dependency edge entirely rather than widening it (which is what `worker_threads = 4` was doing: raising the bar for starvation, not eliminating it).

Because the code under test now runs on its own `new_current_thread` runtime pinned to its own `std::thread`, there is no shared resource between the blocked call and the timeout mechanism. That is a structural guarantee, not a race the scheduler usually wins.

I probed for a residual starvation path and found none. `MutexHeldDuringGenerationEngine::generate` uses `tokio::task::spawn_blocking` (`local_agent_service.rs:1864`), which consumes a thread from the *test's* runtime blocking pool — but the test never awaits anything served by that pool, so it cannot be starved by it either.

**Empirical confirmation.** Beyond re-running the author's numbers, I stress-tested the exact condition that produced the original latent flake: 8 saturating `yes` processes plus `--test-threads=8` (4x the gate's concurrency). Three consecutive runs: 92 passed in 0.78–0.81s, no hang, no flake. Five sequential clean runs at `--test-threads=2` were also stable at 0.91–0.94s. This is the right fix.

### 🟢 Both tests still genuinely guard their defects — mutation-verified, and each guards a *distinct* defect

I did not take the PR's claim on trust; I ran two separate mutations.

**Mutation 1** — reverted `get_status` to a live engine query (`local_agent_service.rs:789`), reintroducing #1791's defect:

```
test result: FAILED. 90 passed; 2 failed; finished in 5.94s
  engine_swap_completes_when_model_info_hangs
  get_status_does_not_block_while_a_generation_holds_the_engine_lock
```

Reproduces the claimed 5.94s and both test names exactly. Clean failure, **no hang**.

**Mutation 2** — removed the swap snapshot timeout (`replace_engine`, `local_agent_service.rs:227`), the defect the *swap* test exists for:

```
test result: FAILED. 91 passed; 1 failed; finished in 5.31s
  engine_swap_completes_when_model_info_hangs
```

This second mutation is the one that matters for review confidence, and it is worth stating why. After the rework, `engine_swap_completes_when_model_info_hangs` calls `get_status` inside its spawned thread, so it now fails under mutation 1 as well. Taken alone, that could look like the swap test has degenerated into a duplicate `get_status` guard. Mutation 2 disproves that: the swap test fails on a defect that leaves `get_status` entirely intact, so it retains independent discriminating power. Both guards are real. Worth noting that the swap test is now sensitive to two defects rather than one — mildly less surgical for triage, but strictly more coverage, and the assertions (`model_id == "hanging-model"`, `granted_n_ctx == 0`) still pin the degradation semantics the test was written for.

### 🟢 Dropping `multi_thread` breaks no assumption

The `worker_threads = 2`/`4` annotations existed solely to keep a worker free for the timeout timer. Once the timeout moved to `recv_timeout` on the test's own OS thread, that requirement disappears. Nothing else in either test body needs true parallelism: in `get_status_does_not_block…` the only concurrent work is `turn` (a `tokio::spawn`) and the test awaits `started_rx` before proceeding, so the current-thread runtime interleaves them at await points correctly. Verified empirically by the runs above. Dropping the annotation is a genuine simplification, not an incidental change.

### 🟢 `Arc::get_mut(&mut svc.inner).expect("sole owner before any clone")` remains sound

I re-derived this rather than assuming the rework preserved it. `LocalAgentServiceImpl::new` (`local_agent_service.rs:129-160`) constructs `inner: Arc::new(...)` and — importantly — spawns nothing and retains no clone of that `Arc` anywhere. `test_service()` returns it directly. The `Arc::get_mut` call sits before the first `svc.clone()` on the following line, so the refcount is provably 1. The rework moved the clone but did not reorder it ahead of the `get_mut`. Still sound.

Structurally this is the fragile kind of assertion — it is order-dependent and would break silently if a future edit hoisted a clone above it — but the `expect` message names the invariant precisely, so a break fails loudly with a self-explaining message. Acceptable as written.

### 🟢 `let _ = done_tx.send(result)` is the right call, and does not mask a real error

`send` on an `mpsc::Sender` fails only when the receiver has been dropped. The receiver lives on the test thread and is dropped only after `recv_timeout` returned `Err(Timeout)` and the subsequent `.expect(...)` panicked — i.e. the test has already failed loudly with a message naming the actual defect. Discarding the send error there is correct: the alternative (`.expect()`) would panic on a detached thread, printing a second, less informative panic that races the real one and could obscure it. The inline comment (`"A send failure means the test already timed out and moved on"`) states this precisely. Note also that `result` here is the `Result<Response, Status>` from the call — a *real* error from `get_status` is not swallowed; it travels through the channel and is unwrapped by `.expect("get_status")` on the receiving side. Only the *transport* failure is discarded. Correct distinction, correctly made.

### 🟡 `local_agent_service.rs:1974` — stale doc comment survives the rework

The doc comment above `engine_swap_completes_when_model_info_hangs` still reads:

```rust
/// Bounds the swap on its own task rather than awaiting it directly: if the
/// snapshot timeout regresses away, `replace_engine` never returns, and an
/// unbounded await here would hang the suite instead of failing it.
```

The test no longer bounds anything "on its own task" — `tokio::spawn` and `tokio::time::timeout` are both gone, replaced by a dedicated OS thread and `recv_timeout`. The paragraph describes the *previous* construction, and it now sits directly above an excellent in-body comment that explains the actual (and contradictory) mechanism. A future reader hitting the stale paragraph first has to work out which one is true.

The underlying *rationale* in that paragraph is still valid and worth keeping — it is only the mechanism wording that rotted. Suggest tightening to something like "Bounds the swap off the test's runtime rather than awaiting it directly: …", or folding it into the in-body comment which already covers this ground well.

This is the one thing I would fix before merge. Not a blocker — it is a comment, the code is correct, and CLAUDE.md's net-positive standard is clearly met — but stale comments about concurrency mechanics are exactly the kind that mislead the next person debugging a hang, and this PR exists *because* a subtle concurrency assumption went unexamined. Cheap to fix while the context is fresh.

### 🟢 Thread/runtime leak on failure is real but benign — correctly traded off

Worth reasoning through explicitly since the task flagged it. On the failure path the spawned `std::thread` is never joined and stays blocked forever, holding a current-thread runtime and (in the `get_status` case) a `std::sync::MutexGuard` held by a `spawn_blocking` task. Assessment:

- **Process exit is unaffected.** Rust's test harness does not join detached threads at exit; `main` returning terminates the process regardless of blocked non-daemon threads. I confirmed this empirically — both mutation runs left threads blocked forever (mutation 2's thread is parked in `std::future::pending()`, which by construction never completes) and the harness still exited promptly and reported results, in 5.94s and 5.31s respectively. No wedging.
- **Cross-test contamination is structurally impossible.** The leaked guard is on a `std::sync::Mutex` created inside `MutexHeldDuringGenerationEngine`, itself constructed fresh per test invocation and reachable only through that test's `svc`. No other test shares it. Likewise each leaked runtime is private to its thread.
- **It only happens when the suite is already failing.** In the passing path both threads complete and exit normally (`replace_engine_if_changed` returns after its 50ms timeout; `get_status` returns from cache). I verified no thread accumulation across 5 sequential and 3 loaded runs.

Joining the thread would require the very unbounded wait the design exists to avoid, so leaking is not a shortcut here — it is the correct trade. Fine as-is; no change requested.

### 🟢 `--bins` pulls in nothing hostile to a gate

I enumerated every binary the widened invocation newly builds. Three bin targets: `nodespaced` (`src/main.rs`, 1 test), `nodespace` CLI (`src/main.rs`, 0 tests), and `gen_skill_md` (`packages/agent/src/bin/`, 3 tests) — the last not mentioned in the PR description but harmless and fast.

One detail worth naming since it is the kind of thing that bites later: the single `nodespaced` bin test, `bind_uds_owner_only_is_0o600_and_restores_umask` (`packages/daemon/src/main.rs:608`), mutates **process-global umask**. That is normally a serialization hazard under `--test-threads=2`. It is safe here only because cargo gives each bin target its own test binary/process and this one contains exactly one test, so nothing can run concurrently with it. Safe today, and it would become unsafe the moment a second test is added to that file — but the test's own SAFETY comment already asserts sole ownership of umask for its duration, so the trap is documented at the right place. No action needed.

### 🟢 `--test-threads=2` workspace-wide destabilizes nothing

Checked for load-sensitive suites. The genuinely contention-prone one is `nodespace-app` (real daemon spawns, Metal shader compilation), and it is excluded and already pinned to `=1` by the ADR-048 step, with a thorough rationale comment in `test-gate.ts`. The newly added crates are in-process and fast: daemon 0.94s, nlp-engine 0.00s, types 0.07s, cli 0.00s. My 8-way-loaded run at `--test-threads=8` found no instability in the daemon suite, which is the most concurrency-heavy of the additions. Warm end-to-end cost of the whole widened gate measured at ~11s. The cost claim in the commit message holds.

### 🟡 No CI workflow — acceptable scope boundary, but the underlying exposure is worth an issue

`.github/workflows/` contains only `audit.yml`, `nightly-webkit.yml`, and `release.yml`; I grepped for `cargo test`/`rust:test` across `.github/` and found zero hits. The local pre-push hook remains the only Rust gate, and `test-gate.ts` itself is candid about this ("This repo has no CI runner for tests — this hook is the only gate", and "a push from a machine that never ran `bun install` is not gated").

Not adding CI here is the **right** scope call — this PR is atomic and does one thing well, and bundling a CI workflow would violate the atomicity the review framework asks for. But it is worth being precise about what this PR does and does not achieve: it fixes *which crates* the gate runs, not *whether the gate runs*. A `--no-verify` push, or a push from a machine that never ran `bun install`, still bypasses all 1,583 tests. Recommend a follow-up issue for a CI test workflow; that is where the residual risk actually lives.

### 🟡 `--lib --bins` still leaves ~252 integration tests outside the gate

This is the finding I would most want carried forward, because it is the *same class* of gap #1792 was filed for, in a different dimension. #1792 was about crates omitted from the gate; this is about **targets** omitted from the gate. The fix closes the first and leaves the second.

Counting `#[test]`/`#[tokio::test]` in `tests/*.rs` across the non-excluded crates:

| Crate | Integration tests | Files |
|---|---|---|
| `nodespace-core` | 145 | 11 |
| `nodespace-daemon` | 36 | 5 |
| `nodespace-agent` | 32 | 2 |
| `nodespace-cli` | 25 | 1 |
| `nodespace-nlp-engine` | 14 | 1 |
| **Total** | **~252** | **20** |

Notably this includes `packages/daemon/tests/grpc_round_trip.rs` and `database_service_e2e.rs` — gRPC round-trip coverage for the same service layer this PR was filed to protect. So a daemon regression at the RPC boundary (as opposed to the service-struct boundary) is still caught by nothing automated.

I am **not** asking for it in this PR. Widening to `--all-targets` would be a materially different change: it pulls in real-model and network-dependent suites (`nlp-engine/tests/integration_tests.rs`, `agent/tests/ollama_integration.rs` — the latter is already `required-features = ["testing"]` and would be excluded, but the former is not), plus `agent/tests/search_skills_latency.rs`, a latency assertion that is exactly the machine-load-flaky shape this codebase has been burned by before. That needs its own measurement and triage pass.

The `--workspace`-with-exclusions inversion this PR introduces is the correct pattern, and the same reasoning applies one level down: the target-level filter is still an allowlist, so a new `tests/*.rs` file is silently ungated by default. Suggest a follow-up issue to triage the 20 integration-test files and move to `--all-targets` with explicit `--exclude`/`required-features` opt-outs for the genuinely environment-dependent ones. Recording it now while the analysis is fresh is worth more than the diff would be.

---

## Standards Compliance

| Standard | Status |
|---|---|
| No issue numbers in code comments | ✅ Grepped all added lines for `#\d{3,}`/`issue` — zero hits. (`test-gate.ts:60` has a `#1610` ref, but it is pre-existing and untouched.) |
| No lint suppression / `#[allow(dead_code)]` / `#[deprecated]` | ✅ None added |
| No backward-compat or transition scaffolding | ✅ Old invocation deleted outright, not kept alongside |
| `cargo clippy -p nodespace-daemon --all-targets -D warnings -D clippy::unused-async` | ✅ Clean (re-run independently) |
| `cargo fmt --all --check` | ✅ Clean for this PR's files. Reports drift in `packages/core/src/services/node_service.rs`, which is **pre-existing and unrelated** — working tree is clean and that file is untouched by this commit. Flagging so it is not mistaken for this PR's doing. |
| Atomic commit | ✅ Gate widening + the flake it exposed are causally linked; splitting would land a gate that fails |
| No docs added to this repo | ✅ |
| Commit message quality | ✅ Exemplary — states the measurement, the cost, and the reasoning for `--workspace`-over-`-p` |

---

## Recommendation: **APPROVE**

Clear net positive, and the reasoning behind it is unusually well documented. The core judgment call — inverting the gate from an include-list to an exclude-list, on the grounds that silent omission is the failure mode that created the bug — is the correct architectural response to the class of defect, not just to this instance of it. That is the difference between fixing a bug and fixing the thing that produced it.

The test rework is the more valuable half of the PR and it survived scrutiny. I attempted to break the determinism claim two ways — a second independent mutation, and an 8x CPU-saturated run at 4x the gate's thread count — and it held. The `recv_timeout` construction removes the starvation dependency edge structurally rather than lowering its probability, which is what distinguishes a fix from a mitigation. Verified independently: 1,583 tests, ~11s warm, 5.94s/5.31s clean failures under mutation, no hangs.

Nothing blocking. One thing worth doing before merge and two worth filing:

- 🟡 Fix the stale doc comment at `local_agent_service.rs:1974` — it describes the construction this PR deleted, and sits above the comment that describes the real one.
- 🟡 File a follow-up for the ~252 `tests/*.rs` integration tests still outside the gate (including daemon gRPC round-trip coverage) — the same class of gap as #1792, one level down, and needing its own triage of the environment-dependent suites.
- 🟡 File a follow-up for a CI test workflow — this PR fixes *what* the gate runs, not *whether* it runs.

Temporary mutation edits were reverted; `git status` verified clean.

malibio (commented):
## Review Type: Initial Review

Scope: one commit (`133ec7ee`), two files, +56/-30. Verified by independent re-execution, including two mutation tests and a load-stress run.

---

## Requirements Check (#1792)

| # | Requirement | Status |
|---|---|---|
| 1 | `rust:test` covers all workspace crates, not 2 of 9 | ✅ Verified — 10 test binaries across 7 crates now build and run |
| 2 | `nodespace-daemon`'s 92 tests actually run in the gate | ✅ Verified — visible in `bun run rust:test` output, 0.94s |
| 3 | `nodespace-nlp-engine` (108), `nodespace-types` (38), `nodespace-cli` (6) covered | ✅ Verified — exact counts reproduced |
| 4 | New crates covered by default (opt-out, not opt-in) | ✅ `--workspace` + `--exclude` inverts the failure mode correctly |
| 5 | Excluding `nodespace-app` loses no coverage | ✅ Verified independently — see below |
| 6 | Total test count 1,335 → 1,583 | ✅ Reproduced exactly: 309+3+6+0+1026+92+1+108+0+38 = **1,583** |

**Independent verification of #5.** I ran `cargo test -p nodespace-app --no-run` and confirmed the ADR-048 pre-push step (`scripts/test-gate.ts:42`) builds and runs **13 targets** — `src/lib.rs` (22 unit tests), `src/main.rs`, and all 11 `tests/*.rs` integration binaries. The exclusion is strictly correct: it is not merely justified by the `--test-threads=1` + `NODESPACED_TEST_BIN` requirement, it also loses nothing, because that step is broader than what `--lib --bins` would have run anyway. Good call.

**`--exclude nodespace-app-test-support` is also correct.** I checked `packages/desktop-app/src-tauri/test-support/src/lib.rs`: zero `#[test]`/`#[tokio::test]` functions and no `mod tests`. My first grep produced false positives from doc comments *about* `#[tokio::test]` semantics; a precise count returns 0. Nothing is skipped. That said, the two exclusions are asymmetric in nature — see 🟢 finding 4.

---

## Code Review Findings

### 🟢 The isolated-runtime + mpsc construction is correct, and deterministic — not merely "less likely" to hang

This was the question most worth scrutinizing, and the construction holds up structurally, not just empirically. The reason it is deterministic rather than probabilistic:

`std::sync::mpsc::Receiver::recv_timeout` blocks the **OS thread** on a condvar with an absolute deadline. It requires no tokio timer, no reactor turn, and no scheduler slot to expire. The previous form (`tokio::time::timeout` around a `tokio::spawn` join) depended on the shared runtime having a free worker to poll the timer — precisely the resource a thread-blocking `std::sync::Mutex` acquisition consumes. The rework removes that dependency edge entirely rather than widening it (which is what `worker_threads = 4` was doing: raising the bar for starvation, not eliminating it).

Because the code under test now runs on its own `new_current_thread` runtime pinned to its own `std::thread`, there is no shared resource between the blocked call and the timeout mechanism. That is a structural guarantee, not a race the scheduler usually wins.

I probed for a residual starvation path and found none. `MutexHeldDuringGenerationEngine::generate` uses `tokio::task::spawn_blocking` (`local_agent_service.rs:1864`), which consumes a thread from the *test's* runtime blocking pool — but the test never awaits anything served by that pool, so it cannot be starved by it either.

**Empirical confirmation.** Beyond re-running the author's numbers, I stress-tested the exact condition that produced the original latent flake: 8 saturating `yes` processes plus `--test-threads=8` (4x the gate's concurrency). Three consecutive runs: 92 passed in 0.78–0.81s, no hang, no flake. Five sequential clean runs at `--test-threads=2` were also stable at 0.91–0.94s. This is the right fix.

### 🟢 Both tests still genuinely guard their defects — mutation-verified, and each guards a *distinct* defect

I did not take the PR's claim on trust; I ran two separate mutations.

**Mutation 1** — reverted `get_status` to a live engine query (`local_agent_service.rs:789`), reintroducing #1791's defect:

```
test result: FAILED. 90 passed; 2 failed; finished in 5.94s
  engine_swap_completes_when_model_info_hangs
  get_status_does_not_block_while_a_generation_holds_the_engine_lock
```

Reproduces the claimed 5.94s and both test names exactly. Clean failure, **no hang**.

**Mutation 2** — removed the swap snapshot timeout (`replace_engine`, `local_agent_service.rs:227`), the defect the *swap* test exists for:

```
test result: FAILED. 91 passed; 1 failed; finished in 5.31s
  engine_swap_completes_when_model_info_hangs
```

This second mutation is the one that matters for review confidence, and it is worth stating why. After the rework, `engine_swap_completes_when_model_info_hangs` calls `get_status` inside its spawned thread, so it now fails under mutation 1 as well. Taken alone, that could look like the swap test has degenerated into a duplicate `get_status` guard. Mutation 2 disproves that: the swap test fails on a defect that leaves `get_status` entirely intact, so it retains independent discriminating power. Both guards are real. Worth noting that the swap test is now sensitive to two defects rather than one — mildly less surgical for triage, but strictly more coverage, and the assertions (`model_id == "hanging-model"`, `granted_n_ctx == 0`) still pin the degradation semantics the test was written for.

### 🟢 Dropping `multi_thread` breaks no assumption

The `worker_threads = 2`/`4` annotations existed solely to keep a worker free for the timeout timer. Once the timeout moved to `recv_timeout` on the test's own OS thread, that requirement disappears. Nothing else in either test body needs true parallelism: in `get_status_does_not_block…` the only concurrent work is `turn` (a `tokio::spawn`) and the test awaits `started_rx` before proceeding, so the current-thread runtime interleaves them at await points correctly. Verified empirically by the runs above. Dropping the annotation is a genuine simplification, not an incidental change.

### 🟢 `Arc::get_mut(&mut svc.inner).expect("sole owner before any clone")` remains sound

I re-derived this rather than assuming the rework preserved it. `LocalAgentServiceImpl::new` (`local_agent_service.rs:129-160`) constructs `inner: Arc::new(...)` and — importantly — spawns nothing and retains no clone of that `Arc` anywhere. `test_service()` returns it directly. The `Arc::get_mut` call sits before the first `svc.clone()` on the following line, so the refcount is provably 1. The rework moved the clone but did not reorder it ahead of the `get_mut`. Still sound.

Structurally this is the fragile kind of assertion — it is order-dependent and would break silently if a future edit hoisted a clone above it — but the `expect` message names the invariant precisely, so a break fails loudly with a self-explaining message. Acceptable as written.

### 🟢 `let _ = done_tx.send(result)` is the right call, and does not mask a real error

`send` on an `mpsc::Sender` fails only when the receiver has been dropped. The receiver lives on the test thread and is dropped only after `recv_timeout` returned `Err(Timeout)` and the subsequent `.expect(...)` panicked — i.e. the test has already failed loudly with a message naming the actual defect. Discarding the send error there is correct: the alternative (`.expect()`) would panic on a detached thread, printing a second, less informative panic that races the real one and could obscure it. The inline comment (`"A send failure means the test already timed out and moved on"`) states this precisely. Note also that `result` here is the `Result<Response, Status>` from the call — a *real* error from `get_status` is not swallowed; it travels through the channel and is unwrapped by `.expect("get_status")` on the receiving side. Only the *transport* failure is discarded. Correct distinction, correctly made.

### 🟡 `local_agent_service.rs:1974` — stale doc comment survives the rework

The doc comment above `engine_swap_completes_when_model_info_hangs` still reads:

```rust
/// Bounds the swap on its own task rather than awaiting it directly: if the
/// snapshot timeout regresses away, `replace_engine` never returns, and an
/// unbounded await here would hang the suite instead of failing it.
```

The test no longer bounds anything "on its own task" — `tokio::spawn` and `tokio::time::timeout` are both gone, replaced by a dedicated OS thread and `recv_timeout`. The paragraph describes the *previous* construction, and it now sits directly above an excellent in-body comment that explains the actual (and contradictory) mechanism. A future reader hitting the stale paragraph first has to work out which one is true.

The underlying *rationale* in that paragraph is still valid and worth keeping — it is only the mechanism wording that rotted. Suggest tightening to something like "Bounds the swap off the test's runtime rather than awaiting it directly: …", or folding it into the in-body comment which already covers this ground well.

This is the one thing I would fix before merge. Not a blocker — it is a comment, the code is correct, and CLAUDE.md's net-positive standard is clearly met — but stale comments about concurrency mechanics are exactly the kind that mislead the next person debugging a hang, and this PR exists *because* a subtle concurrency assumption went unexamined. Cheap to fix while the context is fresh.

### 🟢 Thread/runtime leak on failure is real but benign — correctly traded off

Worth reasoning through explicitly since the task flagged it. On the failure path the spawned `std::thread` is never joined and stays blocked forever, holding a current-thread runtime and (in the `get_status` case) a `std::sync::MutexGuard` held by a `spawn_blocking` task. Assessment:

- **Process exit is unaffected.** Rust's test harness does not join detached threads at exit; `main` returning terminates the process regardless of blocked non-daemon threads. I confirmed this empirically — both mutation runs left threads blocked forever (mutation 2's thread is parked in `std::future::pending()`, which by construction never completes) and the harness still exited promptly and reported results, in 5.94s and 5.31s respectively. No wedging.
- **Cross-test contamination is structurally impossible.** The leaked guard is on a `std::sync::Mutex` created inside `MutexHeldDuringGenerationEngine`, itself constructed fresh per test invocation and reachable only through that test's `svc`. No other test shares it. Likewise each leaked runtime is private to its thread.
- **It only happens when the suite is already failing.** In the passing path both threads complete and exit normally (`replace_engine_if_changed` returns after its 50ms timeout; `get_status` returns from cache). I verified no thread accumulation across 5 sequential and 3 loaded runs.

Joining the thread would require the very unbounded wait the design exists to avoid, so leaking is not a shortcut here — it is the correct trade. Fine as-is; no change requested.

### 🟢 `--bins` pulls in nothing hostile to a gate

I enumerated every binary the widened invocation newly builds. Three bin targets: `nodespaced` (`src/main.rs`, 1 test), `nodespace` CLI (`src/main.rs`, 0 tests), and `gen_skill_md` (`packages/agent/src/bin/`, 3 tests) — the last not mentioned in the PR description but harmless and fast.

One detail worth naming since it is the kind of thing that bites later: the single `nodespaced` bin test, `bind_uds_owner_only_is_0o600_and_restores_umask` (`packages/daemon/src/main.rs:608`), mutates **process-global umask**. That is normally a serialization hazard under `--test-threads=2`. It is safe here only because cargo gives each bin target its own test binary/process and this one contains exactly one test, so nothing can run concurrently with it. Safe today, and it would become unsafe the moment a second test is added to that file — but the test's own SAFETY comment already asserts sole ownership of umask for its duration, so the trap is documented at the right place. No action needed.

### 🟢 `--test-threads=2` workspace-wide destabilizes nothing

Checked for load-sensitive suites. The genuinely contention-prone one is `nodespace-app` (real daemon spawns, Metal shader compilation), and it is excluded and already pinned to `=1` by the ADR-048 step, with a thorough rationale comment in `test-gate.ts`. The newly added crates are in-process and fast: daemon 0.94s, nlp-engine 0.00s, types 0.07s, cli 0.00s. My 8-way-loaded run at `--test-threads=8` found no instability in the daemon suite, which is the most concurrency-heavy of the additions. Warm end-to-end cost of the whole widened gate measured at ~11s. The cost claim in the commit message holds.

### 🟡 No CI workflow — acceptable scope boundary, but the underlying exposure is worth an issue

`.github/workflows/` contains only `audit.yml`, `nightly-webkit.yml`, and `release.yml`; I grepped for `cargo test`/`rust:test` across `.github/` and found zero hits. The local pre-push hook remains the only Rust gate, and `test-gate.ts` itself is candid about this ("This repo has no CI runner for tests — this hook is the only gate", and "a push from a machine that never ran `bun install` is not gated").

Not adding CI here is the **right** scope call — this PR is atomic and does one thing well, and bundling a CI workflow would violate the atomicity the review framework asks for. But it is worth being precise about what this PR does and does not achieve: it fixes *which crates* the gate runs, not *whether the gate runs*. A `--no-verify` push, or a push from a machine that never ran `bun install`, still bypasses all 1,583 tests. Recommend a follow-up issue for a CI test workflow; that is where the residual risk actually lives.

### 🟡 `--lib --bins` still leaves ~252 integration tests outside the gate

This is the finding I would most want carried forward, because it is the *same class* of gap #1792 was filed for, in a different dimension. #1792 was about crates omitted from the gate; this is about **targets** omitted from the gate. The fix closes the first and leaves the second.

Counting `#[test]`/`#[tokio::test]` in `tests/*.rs` across the non-excluded crates:

| Crate | Integration tests | Files |
|---|---|---|
| `nodespace-core` | 145 | 11 |
| `nodespace-daemon` | 36 | 5 |
| `nodespace-agent` | 32 | 2 |
| `nodespace-cli` | 25 | 1 |
| `nodespace-nlp-engine` | 14 | 1 |
| **Total** | **~252** | **20** |

Notably this includes `packages/daemon/tests/grpc_round_trip.rs` and `database_service_e2e.rs` — gRPC round-trip coverage for the same service layer this PR was filed to protect. So a daemon regression at the RPC boundary (as opposed to the service-struct boundary) is still caught by nothing automated.

I am **not** asking for it in this PR. Widening to `--all-targets` would be a materially different change: it pulls in real-model and network-dependent suites (`nlp-engine/tests/integration_tests.rs`, `agent/tests/ollama_integration.rs` — the latter is already `required-features = ["testing"]` and would be excluded, but the former is not), plus `agent/tests/search_skills_latency.rs`, a latency assertion that is exactly the machine-load-flaky shape this codebase has been burned by before. That needs its own measurement and triage pass.

The `--workspace`-with-exclusions inversion this PR introduces is the correct pattern, and the same reasoning applies one level down: the target-level filter is still an allowlist, so a new `tests/*.rs` file is silently ungated by default. Suggest a follow-up issue to triage the 20 integration-test files and move to `--all-targets` with explicit `--exclude`/`required-features` opt-outs for the genuinely environment-dependent ones. Recording it now while the analysis is fresh is worth more than the diff would be.

---

## Standards Compliance

| Standard | Status |
|---|---|
| No issue numbers in code comments | ✅ Grepped all added lines for `#\d{3,}`/`issue` — zero hits. (`test-gate.ts:60` has a `#1610` ref, but it is pre-existing and untouched.) |
| No lint suppression / `#[allow(dead_code)]` / `#[deprecated]` | ✅ None added |
| No backward-compat or transition scaffolding | ✅ Old invocation deleted outright, not kept alongside |
| `cargo clippy -p nodespace-daemon --all-targets -D warnings -D clippy::unused-async` | ✅ Clean (re-run independently) |
| `cargo fmt --all --check` | ✅ Clean for this PR's files. Reports drift in `packages/core/src/services/node_service.rs`, which is **pre-existing and unrelated** — working tree is clean and that file is untouched by this commit. Flagging so it is not mistaken for this PR's doing. |
| Atomic commit | ✅ Gate widening + the flake it exposed are causally linked; splitting would land a gate that fails |
| No docs added to this repo | ✅ |
| Commit message quality | ✅ Exemplary — states the measurement, the cost, and the reasoning for `--workspace`-over-`-p` |

---

## Recommendation: **APPROVE**

Clear net positive, and the reasoning behind it is unusually well documented. The core judgment call — inverting the gate from an include-list to an exclude-list, on the grounds that silent omission is the failure mode that created the bug — is the correct architectural response to the class of defect, not just to this instance of it. That is the difference between fixing a bug and fixing the thing that produced it.

The test rework is the more valuable half of the PR and it survived scrutiny. I attempted to break the determinism claim two ways — a second independent mutation, and an 8x CPU-saturated run at 4x the gate's thread count — and it held. The `recv_timeout` construction removes the starvation dependency edge structurally rather than lowering its probability, which is what distinguishes a fix from a mitigation. Verified independently: 1,583 tests, ~11s warm, 5.94s/5.31s clean failures under mutation, no hangs.

Nothing blocking. One thing worth doing before merge and two worth filing:

- 🟡 Fix the stale doc comment at `local_agent_service.rs:1974` — it describes the construction this PR deleted, and sits above the comment that describes the real one.
- 🟡 File a follow-up for the ~252 `tests/*.rs` integration tests still outside the gate (including daemon gRPC round-trip coverage) — the same class of gap as #1792, one level down, and needing its own triage of the environment-dependent suites.
- 🟡 File a follow-up for a CI test workflow — this PR fixes *what* the gate runs, not *whether* it runs.

Temporary mutation edits were reverted; `git status` verified clean.

malibio:
Fixed the stale doc comment in `a95f55bf` — you were right that it described the `tokio::spawn` + `timeout` construction this branch deleted, and it sat directly above the in-body comment explaining the mechanism actually in use. Dropped the paragraph rather than rewriting it, since the in-body comment already covers it.

Filed both follow-ups:

- **#1796** — the target allowlist. I verified the count before filing: 20 `tests/*.rs` files across 5 crates (core 11, daemon 5, agent 2, nlp-engine 1, cli 1), including the daemon's `grpc_round_trip.rs`. Your framing that it's the same bug one level down is the right one, and I wrote it up that way. Noted why it needs triage rather than a flag change — real-model suites, latency assertions, and daemon-spawning tests aren't gate-safe as-is, and a flaky gate gets `--no-verify`'d and then protects nothing.
- **#1797** — CI enforcement. Framed as the complement to #1796: that one is coverage, this is whether the gate runs at all. Widening coverage raises the cost of it being skippable.

On the `cargo fmt` drift you flagged: confirmed pre-existing and not from this PR. It's in `packages/core/src/services/collection_service.rs` and `node_service/embedding.rs`, neither touched here. Same two files `quality:fix` kept reformatting during #1791 — `cargo fmt --all` reformats the workspace, so it surfaces on any branch that runs it. Left alone rather than dragging unrelated churn into this PR.

Two things in the review I want to acknowledge specifically, because they changed my confidence rather than just confirming it:

The **second mutation** was the one that mattered. Since the swap test now calls `get_status` internally, mutation 1 alone couldn't distinguish a real guard from a duplicate — removing the swap snapshot timeout and seeing only the swap test fail is what actually establishes independent discriminating power. I hadn't run that one.

The **8-saturating-process stress run at `--test-threads=8`** is the right way to check the fix is deterministic rather than merely likelier to pass. That's the failure mode that produced this rework in the first place, so testing it under 4x the gate's concurrency is a better check than I'd applied.

`cargo test --lib -p nodespace-daemon`: 92 passed. Clippy `--all-targets` clean. Pre-push gate passed with the widened `rust:test` running itself.

PR #1801, opened 2026-07-29
Cross-turn write guard misses two cases: oversized args exempt create_nodes_from_markdown, and delete_node's node_id alias yields a second identity

malibio:
Closes #1798

### DIFF
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -10,6 +10,7 @@ use std::time::Instant;
 
 use opentelemetry::trace::{Span, TraceContextExt, Tracer};
 use opentelemetry::KeyValue;
+use sha2::{Digest, Sha256};
 use tokio::sync::RwLock;
 use tokio_util::sync::CancellationToken;
 
@@ -39,15 +40,25 @@ const MAX_TOOL_ITERATIONS: usize = 5;
 /// that were about to succeed.
 const MAX_CONSECUTIVE_PARSE_FAILURES: usize = 2;
 
-/// Longest canonical-args string persisted as a completed write's identity.
+/// Longest canonical-args string stored verbatim as a completed write's identity.
 ///
-/// `create_nodes_from_markdown` carries an entire import in its arguments, so
-/// storing them verbatim would write that content a second time into the chat
-/// node's own message history and grow it without bound. Past this length the
-/// identity is dropped rather than truncated: a truncated string could compare
-/// equal to a *different* call sharing a long prefix, turning a size limit into
-/// a wrong-suppression bug. Dropping it only costs a redundant re-execution.
+/// `create_nodes_from_markdown` carries an entire import in its arguments, and
+/// the identity is persisted into the chat node's own message history — so
+/// storing them verbatim would write that content a second time and grow the
+/// node without bound. Past this length [`canonical_args_identity`] substitutes
+/// a digest, which keeps the identity at constant size.
+///
+/// Note this is a *representation* threshold, not a guard-coverage one: a call
+/// above it is still guarded. Truncation, by contrast, would not be safe — a
+/// truncated string could compare equal to a *different* call sharing a long
+/// prefix, turning a size limit into a wrong-suppression bug. A digest has no
+/// such failure mode, which is why it is the form used above the cap.
 pub const CANONICAL_ARGS_MAX_CHARS: usize = 4096;
+
+/// Prefix marking an identity that is a digest rather than literal canonical
+/// args. Canonical JSON always begins with `{`, so the two forms are mutually
+/// unambiguous and a digest can never be mistaken for a call's real arguments.
+const CANONICAL_ARGS_DIGEST_PREFIX: &str = "sha256:";
 
 /// Normalise a tool call's raw JSON arguments so equal calls compare equal.
 ///
@@ -57,6 +68,14 @@ pub const CANONICAL_ARGS_MAX_CHARS: usize = 4096;
 /// string. Unparseable arguments are returned unchanged: they cannot be
 /// normalised, and an exact-match comparison on the raw text is still correct.
 ///
+/// Also resolves the `node_id` → `id` parameter alias (see the `serde(alias)`
+/// attributes in `tools.rs`). Serde applies that alias when *deserialising into
+/// a params struct*, which happens after this function runs — so without this
+/// step `{"id":"x"}` and `{"node_id":"x"}` would yield two different identities
+/// for one logical call, and a repeat that switched spelling would slip the
+/// guard. Normalising here rather than per-tool means any future tool adopting
+/// the same alias is covered without a matching fix.
+///
 /// Shared by the per-turn duplicate detector and the cross-turn write guard, so
 /// the normalisation rules cannot drift apart. Note the two feed it different
 /// inputs: the loop-breaker canonicalises the raw emitted text, while the guard
@@ -65,8 +84,51 @@ pub const CANONICAL_ARGS_MAX_CHARS: usize = 4096;
 /// stuck model, whereas the guard's comparison must be exact.
 pub fn canonical_args(args_json: &str) -> String {
     serde_json::from_str::<serde_json::Value>(args_json)
-        .map(|v| v.to_string())
+        .map(|mut v| {
+            normalize_param_aliases(&mut v);
+            v.to_string()
+        })
         .unwrap_or_else(|_| args_json.to_owned())
+}
+
+/// Rewrite aliased top-level parameter keys to the name the params struct uses.
+///
+/// Only the top level: a nested `properties` blob is the user's own data, where
+/// a key named `node_id` means whatever the user's schema says and must not be
+/// renamed. When a call carries *both* spellings the object is left untouched —
+/// serde's own precedence decides which wins, and picking one here could make
+/// two genuinely different calls compare equal, which is the one failure this
+/// guard must never have.
+fn normalize_param_aliases(args: &mut serde_json::Value) {
+    let Some(obj) = args.as_object_mut() else {
+        return;
+    };
+    if obj.contains_key("node_id") && !obj.contains_key("id") {
+        if let Some(v) = obj.remove("node_id") {
+            obj.insert("id".to_string(), v);
+        }
+    }
+}
+
+/// The identity persisted for a call's canonical args.
+///
+/// Under [`CANONICAL_ARGS_MAX_CHARS`] this is the canonical JSON verbatim, so a
+/// stored identity stays readable when diagnosing a refusal. Above it, a digest
+/// of that same string — exact-match equality is all the guard needs, so hashing
+/// preserves the guarantee at constant size instead of dropping the identity and
+/// leaving the largest, most costly duplicates unguarded.
+///
+/// Both sides of the comparison must derive the identity through this one
+/// function: the record side when persisting, the execution path when checking.
+/// Deriving either independently is what would let the two drift apart.
+pub fn canonical_args_identity(canonical: &str) -> String {
+    if canonical.chars().count() <= CANONICAL_ARGS_MAX_CHARS {
+        return canonical.to_owned();
+    }
+    format!(
+        "{CANONICAL_ARGS_DIGEST_PREFIX}{:x}",
+        Sha256::digest(canonical.as_bytes())
+    )
 }
 
 /// Build the tool result returned in place of a refused duplicate write.
@@ -860,7 +922,14 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                             // concretely when empty arguments are read as `{}`
                             // above, which yields "" here against a stored "{}"
                             // and a write that could never match itself.
-                            let incoming = canonical_args(&args.to_string());
+                            //
+                            // Reduced to an identity by the same function the
+                            // record side uses, so an oversized call compares
+                            // digest-against-digest. Comparing raw canonical args
+                            // here against a stored digest would never match, and
+                            // would silently unguard exactly the largest writes.
+                            let incoming =
+                                canonical_args_identity(&canonical_args(&args.to_string()));
                             session.prior_writes.iter().find(|w| {
                                 w.tool == tc.function_name && w.canonical_args == incoming
                             })
@@ -4420,6 +4489,259 @@ mod tests {
             calls.lock().unwrap().iter().any(|c| c == "create_node"),
             "a distinct create must not be blocked"
         );
+    }
+
+    /// Build a session carrying one completed markdown import, stored the way
+    /// the record side stores it — through `canonical_args_identity`, so an
+    /// oversized import is held as a digest.
+    fn session_with_prior_import(markdown: &str) -> AgentSession {
+        let mut session = new_session();
+        session.prior_writes = vec![PriorWrite {
+            tool: "create_nodes_from_markdown".to_string(),
+            canonical_args: canonical_args_identity(&canonical_args(
+                &json!({ "markdown": markdown }).to_string(),
+            )),
+            node_id: Some("nodespace://n1".to_string()),
+            summary: Some("Project plan".to_string()),
+        }];
+        session
+    }
+
+    fn import_executor() -> MockToolExecutor {
+        MockToolExecutor::new().with_tool(
+            "create_nodes_from_markdown",
+            json!({"type": "object", "properties": {"markdown": {"type": "string"}}}),
+            json!({"created": 12}),
+        )
+    }
+
+    /// An import large enough to exceed the args cap — the realistic case, since
+    /// any real import does.
+    fn oversized_markdown(topic: &str) -> String {
+        format!(
+            "# {topic}\n{}",
+            "- a task line\n".repeat(CANONICAL_ARGS_MAX_CHARS / 8)
+        )
+    }
+
+    /// The acceptance criterion for the oversized-args gap: a repeated import
+    /// must be refused across turns. This is the tool where a repeat is worst —
+    /// it duplicates an entire subtree — and the one the cap used to exempt
+    /// wholesale, leaving it unguarded in every real case.
+    #[tokio::test]
+    async fn cross_turn_duplicate_oversized_import_is_not_executed() {
+        let markdown = oversized_markdown("Project plan");
+        assert!(
+            markdown.chars().count() > CANONICAL_ARGS_MAX_CHARS,
+            "the fixture must actually exceed the cap"
+        );
+        let args = json!({ "markdown": markdown }).to_string();
+
+        let engine = Arc::new(MockEngine::tool_then_text(
+            "create_nodes_from_markdown",
+            &args,
+            "It already exists.",
+        ));
+        let executor = Arc::new(RecordingToolExecutor::new(import_executor()));
+        let calls = executor.calls_handle();
+        let agent_loop = LocalAgentLoop::new(engine, executor);
+
+        let mut session = session_with_prior_import(&markdown);
+        agent_loop
+            .run_turn(
+                &mut session,
+                "import my project plan",
+                |_| {},
+                |_| {},
+                CancellationToken::new(),
+            )
+            .await
+            .expect("turn should succeed");
+
+        assert!(
+            !calls
+                .lock()
+                .unwrap()
+                .iter()
+                .any(|c| c == "create_nodes_from_markdown"),
+            "a repeated import must never reach the executor, got {:?}",
+            calls.lock().unwrap()
+        );
+    }
+
+    /// The negative control for the digest. Refusing a *different* import would
+    /// be a wrong suppression — worse than the missed duplicate this change
+    /// fixes — so the identity must distinguish two large calls, not merely
+    /// recognise that both are large.
+    #[tokio::test]
+    async fn a_different_oversized_import_still_executes() {
+        let engine = Arc::new(MockEngine::tool_then_text(
+            "create_nodes_from_markdown",
+            &json!({ "markdown": oversized_markdown("Recipes") }).to_string(),
+            "Imported.",
+        ));
+        let executor = Arc::new(RecordingToolExecutor::new(import_executor()));
+        let calls = executor.calls_handle();
+        let agent_loop = LocalAgentLoop::new(engine, executor);
+
+        let mut session = session_with_prior_import(&oversized_markdown("Project plan"));
+        agent_loop
+            .run_turn(
+                &mut session,
+                "import my recipes",
+                |_| {},
+                |_| {},
+                CancellationToken::new(),
+            )
+            .await
+            .expect("turn should succeed");
+
+        assert!(
+            calls
+                .lock()
+                .unwrap()
+                .iter()
+                .any(|c| c == "create_nodes_from_markdown"),
+            "a distinct import must not be blocked"
+        );
+    }
+
+    fn delete_node_executor() -> MockToolExecutor {
+        MockToolExecutor::new().with_tool(
+            "delete_node",
+            json!({"type": "object", "properties": {"id": {"type": "string"}}}),
+            json!({"deleted": true}),
+        )
+    }
+
+    /// Build a session carrying one completed `delete_node`, recorded under the
+    /// given key spelling.
+    fn session_with_prior_delete(args: &str) -> AgentSession {
+        let mut session = new_session();
+        session.prior_writes = vec![PriorWrite {
+            tool: "delete_node".to_string(),
+            canonical_args: canonical_args_identity(&canonical_args(args)),
+            node_id: Some("nodespace://n1".to_string()),
+            summary: Some("Old note".to_string()),
+        }];
+        session
+    }
+
+    /// Run a turn where the model re-issues a delete under `incoming`, against a
+    /// prior write recorded under `stored`. Returns whether the executor ran.
+    async fn delete_repeat_reaches_executor(stored: &str, incoming: &str) -> bool {
+        let engine = Arc::new(MockEngine::tool_then_text(
+            "delete_node",
+            incoming,
+            "Already gone.",
+        ));
+        let executor = Arc::new(RecordingToolExecutor::new(delete_node_executor()));
+        let calls = executor.calls_handle();
+        let agent_loop = LocalAgentLoop::new(engine, executor);
+
+        let mut session = session_with_prior_delete(stored);
+        agent_loop
+            .run_turn(
+                &mut session,
+                "delete that node",
+                |_| {},
+                |_| {},
+                CancellationToken::new(),
+            )
+            .await
+            .expect("turn should succeed");
+
+        let ran = calls.lock().unwrap().iter().any(|c| c == "delete_node");
+        ran
+    }
+
+    /// The acceptance criterion for the alias gap. `delete_node` takes its target
+    /// under either `id` or `node_id` (serde alias), but serde resolves that only
+    /// when deserialising into the params struct — after canonicalisation. So a
+    /// repeat that switched spelling used to produce a second identity and slip
+    /// the guard. Both directions, since either spelling can be recorded first.
+    #[tokio::test]
+    async fn delete_repeat_is_refused_across_the_node_id_alias() {
+        let id_form = r#"{"id":"nodespace://n1"}"#;
+        let alias_form = r#"{"node_id":"nodespace://n1"}"#;
+
+        assert!(
+            !delete_repeat_reaches_executor(id_form, alias_form).await,
+            "a repeat spelled `node_id` must be recognised as the stored `id` call"
+        );
+        assert!(
+            !delete_repeat_reaches_executor(alias_form, id_form).await,
+            "a repeat spelled `id` must be recognised as the stored `node_id` call"
+        );
+    }
+
+    /// The alias rule must not merge genuinely different deletes: it renames a
+    /// key, it does not ignore the value.
+    #[tokio::test]
+    async fn delete_of_a_different_node_still_executes() {
+        assert!(
+            delete_repeat_reaches_executor(
+                r#"{"id":"nodespace://n1"}"#,
+                r#"{"node_id":"nodespace://n2"}"#
+            )
+            .await,
+            "a delete targeting another node must not be blocked"
+        );
+    }
+
+    /// The alias is resolved at the top level only. A nested `properties` blob is
+    /// the user's own data, where a field named `node_id` means whatever their
+    /// schema says — renaming it would silently rewrite user data into a
+    /// different field for identity purposes.
+    #[test]
+    fn alias_normalisation_does_not_reach_into_nested_user_data() {
+        let canonical = canonical_args(r#"{"properties":{"node_id":"x"},"node_type":"venue"}"#);
+        assert!(
+            canonical.contains("node_id"),
+            "a nested user field must be left alone, got {canonical}"
+        );
+    }
+
+    /// When a call carries both spellings there is no safe rename: serde's own
+    /// precedence decides which wins, and guessing here could make two genuinely
+    /// different calls compare equal — the one failure this guard must not have.
+    #[test]
+    fn alias_normalisation_leaves_ambiguous_calls_untouched() {
+        let canonical = canonical_args(r#"{"id":"a","node_id":"b"}"#);
+        assert!(canonical.contains(r#""id":"a""#), "got {canonical}");
+        assert!(canonical.contains(r#""node_id":"b""#), "got {canonical}");
+    }
+
+    /// Under the cap the identity is the canonical args verbatim, so a stored
+    /// identity stays readable when diagnosing a refusal.
+    #[test]
+    fn identity_is_verbatim_below_the_cap() {
+        let canonical = canonical_args(r#"{"content":"Buy milk"}"#);
+        assert_eq!(canonical_args_identity(&canonical), canonical);
+    }
+
+    /// The two identity forms must be mutually unambiguous: canonical JSON always
+    /// starts with `{`, a digest with `sha256:`. Were they confusable, a call
+    /// whose arguments happened to spell a digest could impersonate another
+    /// call's identity.
+    #[test]
+    fn digest_and_verbatim_identities_cannot_be_confused() {
+        let small = canonical_args_identity(&canonical_args(r#"{"content":"x"}"#));
+        let large = canonical_args_identity(&"x".repeat(CANONICAL_ARGS_MAX_CHARS + 1));
+        assert!(small.starts_with('{'));
+        assert!(large.starts_with("sha256:"));
+        assert_ne!(small, large);
+    }
+
+    /// The digest must be a function of the content: equal args hash equal,
+    /// different args do not. Exact-match equality is the whole guarantee the
+    /// guard needs, and hashing has to preserve it in both directions.
+    #[test]
+    fn digest_identity_is_stable_and_content_sensitive() {
+        let a = "a".repeat(CANONICAL_ARGS_MAX_CHARS + 1);
+        let b = format!("{}b", "a".repeat(CANONICAL_ARGS_MAX_CHARS));
+        assert_eq!(canonical_args_identity(&a), canonical_args_identity(&a));
+        assert_ne!(canonical_args_identity(&a), canonical_args_identity(&b));
     }
 
     /// Idempotent repeats must survive the guard. Setting the same task status
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -93,6 +93,13 @@ pub fn canonical_args(args_json: &str) -> String {
 
 /// Rewrite aliased top-level parameter keys to the name the params struct uses.
 ///
+/// Applied to every tool's arguments rather than per-tool. `node_id` is the only
+/// alias in the tool registry and always means `id`, so a blanket rule is both
+/// correct and self-maintaining — a future tool adopting the same alias needs no
+/// matching fix here. A tool with no such parameter is unaffected in practice,
+/// and even a stray `node_id` is renamed identically on both sides of the
+/// comparison, so the identity stays consistent either way.
+///
 /// Only the top level: a nested `properties` blob is the user's own data, where
 /// a key named `node_id` means whatever the user's schema says and must not be
 /// renamed. When a call carries *both* spellings the object is left untouched —
@@ -4651,6 +4658,8 @@ mod tests {
             .await
             .expect("turn should succeed");
 
+        // Bound, not returned directly: the temporary `MutexGuard` in the
+        // expression would outlive `calls` at the end of the function.
         let ran = calls.lock().unwrap().iter().any(|c| c == "delete_node");
         ran
     }
@@ -4695,9 +4704,19 @@ mod tests {
     /// different field for identity purposes.
     #[test]
     fn alias_normalisation_does_not_reach_into_nested_user_data() {
-        let canonical = canonical_args(r#"{"properties":{"node_id":"x"},"node_type":"venue"}"#);
+        // A realistic `update_node`: the tool's own target under the alias, plus
+        // a user-defined field that happens to be named `node_id` inside the
+        // `properties` blob. The top-level key is renamed; the user's field is
+        // not, and its value must survive untouched.
+        let canonical = canonical_args(
+            r#"{"node_id":"nodespace://n1","properties":{"node_id":"user-value","capacity":10}}"#,
+        );
         assert!(
-            canonical.contains("node_id"),
+            canonical.contains(r#""id":"nodespace://n1""#),
+            "the top-level alias must be resolved, got {canonical}"
+        );
+        assert!(
+            canonical.contains(r#""node_id":"user-value""#),
             "a nested user field must be left alone, got {canonical}"
         );
     }
--- a/packages/core/src/models/ai_chat_node.rs
+++ b/packages/core/src/models/ai_chat_node.rs
@@ -188,10 +188,33 @@ impl AiChatNode {
             .and_then(|v| v.as_str())
             .map(|s| s.to_string());
 
+        // Decoded per message, not as one `Vec`. Decoding the whole array at
+        // once means a single unreadable message discards the entire
+        // conversation — and because a caller that then writes the node back
+        // (status updates do) re-serialises from this struct, that loss is
+        // persisted. Per-message decoding contains the damage to the message
+        // actually at fault, and logs it rather than failing silently.
         let messages = props
             .get("messages")
-            .cloned()
-            .map(|v| serde_json::from_value::<Vec<AiChatMessage>>(v).unwrap_or_default())
+            .and_then(|v| v.as_array())
+            .map(|arr| {
+                arr.iter()
+                    .filter_map(
+                        |m| match serde_json::from_value::<AiChatMessage>(m.clone()) {
+                            Ok(msg) => Some(msg),
+                            Err(e) => {
+                                tracing::error!(
+                                    node_id = %node.id,
+                                    error = %e,
+                                    "dropping unreadable ai-chat message; the rest of the \
+                                     conversation is preserved"
+                                );
+                                None
+                            }
+                        },
+                    )
+                    .collect()
+            })
             .unwrap_or_default();
 
         Ok(Self {
--- a/packages/core/src/models/ai_chat_node.rs
+++ b/packages/core/src/models/ai_chat_node.rs
@@ -56,16 +56,22 @@ pub struct AiChatCompletedWrite {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub summary: Option<String>,
 
-    /// The call's arguments, canonicalised (JSON key order normalised). Together
-    /// with `tool` this is the write's identity for the cross-turn duplicate
-    /// guard: a later call matching both is the same write, not a new one.
+    /// The call's arguments, canonicalised (JSON key order normalised, parameter
+    /// aliases resolved). Together with `tool` this is the write's identity for
+    /// the cross-turn duplicate guard: a later call matching both is the same
+    /// write, not a new one.
     ///
-    /// `None` when the arguments were too large to persist (see
-    /// `CANONICAL_ARGS_MAX_CHARS`). A `None` here never matches, so an
-    /// unrecorded call is executed rather than wrongly suppressed — the guard
-    /// fails open, which is the safe direction for a check that blocks work.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub canonical_args: Option<String>,
+    /// Two forms, both produced by `canonical_args_identity`: the canonical JSON
+    /// verbatim when it is small enough to store, or `sha256:<hex>` of that same
+    /// string when it is not (see `CANONICAL_ARGS_MAX_CHARS`). The digest keeps
+    /// large writes — an entire markdown import, say — guarded without copying
+    /// their content into this message history a second time. The forms cannot
+    /// be confused: canonical JSON always starts with `{`.
+    ///
+    /// Always present. An identity is what makes a recorded write enforceable,
+    /// so a write recorded without one would be indistinguishable from an
+    /// unguarded tool while still looking wired up.
+    pub canonical_args: String,
 }
 
 /// A single message in an ai-chat conversation.
--- a/packages/core/src/models/ai_chat_node_test.rs
+++ b/packages/core/src/models/ai_chat_node_test.rs
@@ -148,4 +148,89 @@ mod tests {
         assert_eq!(props["ai-chat"]["status"], json!("idle"));
         assert_eq!(props["ai-chat"]["messages"][0]["content"], json!("done"));
     }
+
+    /// One unreadable message must not take the conversation with it.
+    ///
+    /// Decoding the array as a single `Vec` fails wholesale on any one bad
+    /// element, and callers that write the node back — a status update does —
+    /// re-serialise from this struct, so a read that silently yields zero
+    /// messages *persists* as an erased conversation. Per-message decoding keeps
+    /// the loss to the message actually at fault.
+    #[test]
+    fn one_unreadable_message_does_not_erase_the_conversation() {
+        let node = ai_chat_node(json!({
+            "ai-chat": {
+                "status": "idle",
+                "messages": [
+                    { "role": "user", "content": "hello" },
+                    // Malformed: `content` is required and must be a string.
+                    { "role": "assistant", "content": 42 },
+                    { "role": "assistant", "content": "goodbye" }
+                ]
+            }
+        }));
+
+        let chat = AiChatNode::from_node(node).unwrap();
+        assert_eq!(
+            chat.messages.len(),
+            2,
+            "only the unreadable message may be dropped"
+        );
+        assert_eq!(chat.messages[0].content, "hello");
+        assert_eq!(chat.messages[1].content, "goodbye");
+    }
+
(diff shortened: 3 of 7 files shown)


malibio:
## Review Type
**Initial Review** — PR #1801 (commit `fda7d1cc`), closing #1798. No prior review rounds.

## Requirements Check

| # | Acceptance criterion | Status |
|---|---|---|
| 1 | A repeated `create_nodes_from_markdown` with identical arguments is refused across turns, without persisting the full import a second time | ✅ |
| 2 | Oversized arguments no longer silently disable the guard for a call | ✅ |
| 3 | A repeated `delete_node` is recognised whether the target is given as `id` or `node_id` | ✅ |
| 4 | Tests cover both an oversized-args repeat and an alias-spelling repeat, at the execution layer | ✅ |
| — | Third item: single fetch of the chat node per turn | ✅ |

**Independently verified**, not taken on trust:

- `cargo test -p nodespace-agent --lib` → **325 passed, 0 failed**; `cargo test -p nodespace-daemon --lib local_agent` → **28 passed, 0 failed**; `cargo test -p nodespace-core --lib ai_chat` → **18 passed, 0 failed**.
- `cargo clippy -p nodespace-agent -p nodespace-daemon -p nodespace-core --lib --all-targets` → **clean**, zero warnings.
- No `#[allow(...)]`, `#[deprecated]`, or `#[allow(dead_code)]` introduced in any changed file. No new `.md` files. No issue numbers in code comments. `sha2 = "0.10"` was already a declared dependency of `packages/agent` — no new third-party dependency.
- **Identity derivation is genuinely symmetric.** Record side (`local_agent_service.rs:1500`) and execution side (`agent_loop.rs:931-932`) both compute `canonical_args_identity(&canonical_args(...))` in that exact order. The scrutiny item "is the *canonical* string digested on both sides" checks out — alias resolution and key-order normalisation both happen inside `canonical_args`, before the digest, on both paths. `oversized_identity_still_ignores_key_order` pins this.
- **No wrong-refusal path in the identity scheme.** I probed the rename for merge collisions directly: a rename only fires when `id` is absent, so it can never collapse two calls that differ in the `id` value. `{"id":"a"}` vs `{"node_id":"b"}` stay distinct; the both-spellings form is its own third identity distinct from either. The two identity forms are unambiguous — canonical JSON always begins with `{`, digests with `sha256:` — so a digest cannot impersonate literal args.
- **Single-fetch refactor preserves semantics.** `prior_writes_from_history(&messages)` borrows before `node_history_from_messages(messages)` consumes; the early-return-on-empty-history path is unchanged, and logging is now emitted exactly once from `load_chat_messages` rather than depending on the old ordering coupling. No log line was lost or duplicated.

---

## Code Review Findings

### 🔴 Critical

#### 1. Making `canonical_args` mandatory silently wipes any existing conversation — and then persists the wipe

📁 `packages/core/src/models/ai_chat_node.rs:74` (field), consumed at `packages/core/src/models/ai_chat_node.rs:190-194`

This is **not** a backward-compatibility objection. Per CLAUDE.md this repo is greenfield with zero users, and the `Option<String>` → `String` change is the right call — I am not asking for a migration, a compat shim, or a `#[serde(default)]`. The problem is the *failure handling*, which the review brief explicitly flagged as the thing to check, and which is bad independently of whether any legacy data exists today.

`AiChatNode::from_node` decodes history with:

```rust
let messages = props
    .get("messages")
    .cloned()
    .map(|v| serde_json::from_value::<Vec<AiChatMessage>>(v).unwrap_or_default())
    .unwrap_or_default();
```

`canonical_args` is now a required field with no `default`. Serde fails the **entire `Vec<AiChatMessage>`** if any single nested `completedWrites` entry lacks `canonicalArgs`, and `unwrap_or_default()` converts that total failure into an empty history with **no log line at all**.

I verified this end-to-end rather than reasoning about it: decoding a two-message history whose assistant turn carries one write missing `canonicalArgs` yields `0` messages, not 1, not an error.

The blast radius is worse than a skipped turn, because the empty read is then written back:

1. `run_ai_chat_turn` (`local_agent_service.rs:413-418`) reads `history.is_empty()`, logs the *misleading* `"ai-chat history empty — skipping turn"`, and calls `write_ai_chat_status(&node_id, "idle", None)`.
2. `write_ai_chat_status` (`local_agent_service.rs:642-650`) does `AiChatNode::from_node(node)` — decoding to empty again — then `props["ai-chat"] = ai_chat.to_properties_value()`.
3. `to_properties_value` rebuilds `messages` wholesale from the struct: `map.insert("messages", serde_json::to_value(&self.messages)...)`.

So one undecodable write does not merely make the turn a no-op — it **overwrites the user's stored conversation with `[]`**. The user sees their entire chat history vanish, and the only diagnostic is a warning claiming the history was already empty.

The engineering principle is *fail loudly at the boundary, never silently to a default that is also a valid state*. `unwrap_or_default()` is only defensible when "absent" and "corrupt" mean the same thing to the caller; here they emphatically do not — one is a new chat, the other is data loss. This pre-existed the PR, but this PR is what makes it *reachable*, by introducing the first non-defaulted field in that struct. Tightening the type without tightening the decode is what turns a latent sharp edge into a live one.

**Suggested fix** (either is fine, both are small and neither is a compat shim):
- Make the decode failure explicit — propagate the `serde_json` error out of `from_node` as a `ValidationError`, so `load_chat_messages` logs it with the same `tracing::warn!(node_id, error = %e, ...)` treatment it already gives `"node is not an ai-chat node"`, and `write_ai_chat_status` returns `Err` instead of round-tripping an empty history back to storage. This is the option I would pick: it makes the tightened type actually enforceable rather than quietly absorbing violations.
- At minimum, log the decode error before defaulting, and make `write_ai_chat_status` refuse to persist a `messages` array that is empty when the raw stored value was not.

A regression test pinning "a chat node whose `messages` blob fails to decode is never written back as empty" would lock this down.

---

#### 2. Third mirror of `AiChatCompletedWrite` was not updated — the TypeScript type now misdescribes the wire shape

📁 `packages/nodespace-types/src/ai_chat.rs:17-18` vs 📁 `packages/desktop-app/src/lib/types/ai-chat-node.ts:36`

The PR updates two of the **three** declarations of this type. `packages/nodespace-types/src/ai_chat.rs` still reads:

```rust
#[serde(default, skip_serializing_if = "Option::is_none")]
pub canonical_args: Option<String>,
```

This is not a dead duplicate. It is the struct that actually produces the frontend payload: `node_to_typed_value` → `ai_chat_node_to_value` (`packages/nodespace-types/src/convert.rs:156-197`) builds `messages` through *this* `AiChatMessage`/`AiChatCompletedWrite`, and `packages/desktop-app/src-tauri/src/types.rs:11` re-exports it as the canonical wire type for the Tauri commands.

Consequences:

- **The TS type is now wrong.** `ai-chat-node.ts:36` declares `canonicalArgs: string` (required, "Always present"), but the emitting struct still carries `skip_serializing_if = "Option::is_none"`, so the field is **omitted** from the JSON whenever it is `None`. The TS contract asserts a guarantee the wire does not make. Answering the brief's question directly: the TypeScript mirror is consistent with `packages/core`'s type but **not** with the type that actually serialises to the frontend.
- **It re-opens Gap 1 on this path.** `convert.rs:178` uses the same `unwrap_or_default()` pattern, so this lenient mirror silently absorbs the mismatch instead of surfacing it — which is exactly why the test suite is green despite the divergence.

The principle is *single source of truth for a wire contract*. Three hand-synced declarations of one shape is the same class of accidental complexity the guard itself was built to remove, and this PR demonstrates the failure mode: a field was tightened in two places and drifted in the third. The doc comment on line 8 — `Mirrors nodespace_core::models::AiChatCompletedWrite` — is now false.

**Suggested fix:** change `packages/nodespace-types/src/ai_chat.rs:17-18` to `pub canonical_args: String` (dropping the `serde` attributes), matching `packages/core`. That makes the TS declaration true. Longer term, these two Rust structs should be one type re-exported, not two kept in sync by comment — but that is out of scope here and worth its own issue.

---

### 🟡 Important

#### 3. `normalize_param_aliases` is applied to every tool, including those with no such alias

📁 `packages/agent/src/local_agent/agent_loop.rs:100-109`

The rename is unconditional on tool name, so a stray top-level `node_id` is rewritten to `id` even for tools whose params struct has no `node_id` alias. I confirmed the behaviour directly:

- `{"from_id":"a","node_id":"b"}` → `{"from_id":"a","id":"b"}` (`create_relationship` has no such alias)
- `{"content":"x","node_type":"task","node_id":"stray"}` → `{"content":"x","id":"stray","node_type":"task"}` (`create_node` has neither param)

This is **not** a correctness bug for the guard — the rename is applied identically on record and execution sides, and the value is preserved, so no two distinct calls can merge. I want to be clear I am not claiming a wrong refusal here; I probed for one and could not construct it.

But the doc comment says "the name the params struct uses", which overstates what the code does: it applies one tool's alias rule globally. The function is correct-by-accident rather than correct-by-construction, and the comment would mislead the next reader who adds a tool with a semantically different top-level `node_id`. The issue's own suggested direction — "normalise argument keys through the tool's own params type" — would be correct by construction, though it is meaningfully more work and I do not think it is required to land this.

**Suggestion:** keep the implementation, but correct the doc comment to state what it does — *"rewrites the `node_id` → `id` alias at the top level for all tools, since it is the only alias in the tool surface and applying it uniformly keeps record and execution sides in lockstep without a per-tool table"* — and note the invariant that makes it safe (the rename never fires when `id` is present, so it cannot merge two distinct calls). Accuracy in a comment guarding a safety-critical identity function is worth the two lines.

#### 4. Verify the nested-`properties` exemption holds for `update_node`, not just in the unit test

📁 `packages/agent/src/local_agent/agent_loop.rs:100-109`, test at `:4696-4703`

Restricting to top-level keys is the right call, and I agree with the reasoning: a nested `properties.node_id` is user schema data and renaming it would corrupt the identity of a legitimate call. `alias_normalisation_does_not_reach_into_nested_user_data` pins it, and `search_nodes`' nested `filters[].node_id` (`tools.rs:315`) is correctly untouched — I checked.

The gap is that the test exercises a bare JSON literal rather than a shape `update_node` actually emits, and `update_node` is the tool that carries both a top-level aliased `id` *and* a user-controlled `properties` blob. A test using a realistic `{"id": "...", "properties": {"node_id": "..."}}` payload would pin the interaction that actually matters — that the top-level key is renamed while the nested one is not, in the same call. Cheap to add, and it is the case a future refactor is most likely to break.

*(Note: `update_node` is `IdempotentWrite` and therefore not in the guarded set today, so this is about locking in the invariant before it matters, not fixing a live bug.)*

---

### 🟢 Suggestions

#### 5. `assert!(identity.len() < 128)` is a loose proxy for a fixed-size property

📁 `packages/daemon/src/services/local_agent_service.rs:2657`

A SHA-256 hex digest plus the `sha256:` prefix is *exactly* 71 characters. Asserting `< 128` would still pass if the digest were accidentally truncated to 8 hex chars — which is precisely the collision-inducing regression the bound is meant to catch. `assert_eq!(identity.len(), 71)` states the actual invariant and fails on the change you care about.

#### 6. `delete_repeat_reaches_executor` has a redundant binding

📁 `packages/agent/src/local_agent/agent_loop.rs:4658-4659`

```rust
let ran = calls.lock().unwrap().iter().any(|c| c == "delete_node");
ran
```

**Nit:** collapses to a single tail expression.

#### 7. Test-module `load_node_history` helper is the right call

📁 `packages/daemon/src/services/local_agent_service.rs:1684-1690`

Answering the brief's question directly: this does **not** hide a gap. It composes the two production functions in the same order and with the same data flow that `run_ai_chat_turn` uses, so the four pre-existing tests still exercise the real code path; it only spares them the two-line composition. It is a test-local convenience over production functions, not a parallel implementation. No change needed.

---

## Assessment of test quality

The tests are substantive, not decorative. Specifically:

- The negative controls genuinely guard the wrong-suppression direction. `a_different_oversized_import_still_executes` uses two imports that are *both* oversized and differ only in their heading — so it fails if the digest collapsed "all large calls" into one identity, which is the realistic way this could go wrong. `delete_of_a_different_node_still_executes` varies the node ID across the alias boundary (`{"id":"n1"}` stored, `{"node_id":"n2"}` incoming), catching a rename that ignored the value. These are the right adversaries.
- `delete_repeat_is_refused_across_the_node_id_alias` correctly tests **both** directions, since either spelling can be recorded first.
- `oversized_identity_still_ignores_key_order` is the test I would most have wanted, and it is present — it pins that normalisation precedes hashing, which is the ordering mistake that would unguard large writes while leaving small ones guarded (i.e. invisible to every other test).
- The execution-layer tests drive `run_turn` through a recording executor and assert on whether the tool was *reached*, which is the behaviour the acceptance criteria describe, not a proxy for it.

I did not independently re-run the implementer's mutation testing, but the tests are constructed such that the claimed sensitivity is plausible — each new assertion targets a distinct failure mode rather than restating the implementation.

---

## Recommendation

### 🔄 REQUEST CHANGES

The core of this change is well-designed and I want to be clear about that: the hybrid verbatim/digest identity is the right resolution of the debuggability-vs-coverage tension the issue laid out, the two forms are provably unambiguous, the alias fix is placed correctly (in shared canonicalisation, so `seen_calls` and the cross-turn guard cannot drift), the single-fetch refactor is a clean simplification that preserves the early-return and logging semantics, and the documentation genuinely explains *why* rather than *what*. All four acceptance criteria are met and independently verified. This is a clear net improvement to code health.

Two things should be fixed before merge, neither large:

1. **Finding #2 is a near-mechanical fix** — one field in `packages/nodespace-types/src/ai_chat.rs` — and without it the shipped TypeScript type asserts a guarantee the wire does not honour. Landing a knowingly-false type declaration is the kind of thing that costs someone an afternoon later.
2. **Finding #1 is the blocker.** Not because a breaking change is wrong — it is correct and expected here — but because tightening the type without tightening the decode means a violation is absorbed by `unwrap_or_default()` into an empty history, which `write_ai_chat_status` then **persists**, destroying the stored conversation while logging a message that says it was already empty. Silent data loss with a misleading diagnostic is worth the small fix now.

Findings #3–#7 are non-blocking; #3 (comment accuracy on a safety-critical function) is the one I would still encourage taking in this PR.

malibio (commented):
## Review Type
**Initial Review** — PR #1801 (commit `fda7d1cc`), closing #1798. No prior review rounds.

## Requirements Check

| # | Acceptance criterion | Status |
|---|---|---|
| 1 | A repeated `create_nodes_from_markdown` with identical arguments is refused across turns, without persisting the full import a second time | ✅ |
| 2 | Oversized arguments no longer silently disable the guard for a call | ✅ |
| 3 | A repeated `delete_node` is recognised whether the target is given as `id` or `node_id` | ✅ |
| 4 | Tests cover both an oversized-args repeat and an alias-spelling repeat, at the execution layer | ✅ |
| — | Third item: single fetch of the chat node per turn | ✅ |

**Independently verified**, not taken on trust:

- `cargo test -p nodespace-agent --lib` → **325 passed, 0 failed**; `cargo test -p nodespace-daemon --lib local_agent` → **28 passed, 0 failed**; `cargo test -p nodespace-core --lib ai_chat` → **18 passed, 0 failed**.
- `cargo clippy -p nodespace-agent -p nodespace-daemon -p nodespace-core --lib --all-targets` → **clean**, zero warnings.
- No `#[allow(...)]`, `#[deprecated]`, or `#[allow(dead_code)]` introduced in any changed file. No new `.md` files. No issue numbers in code comments. `sha2 = "0.10"` was already a declared dependency of `packages/agent` — no new third-party dependency.
- **Identity derivation is genuinely symmetric.** Record side (`local_agent_service.rs:1500`) and execution side (`agent_loop.rs:931-932`) both compute `canonical_args_identity(&canonical_args(...))` in that exact order. The scrutiny item "is the *canonical* string digested on both sides" checks out — alias resolution and key-order normalisation both happen inside `canonical_args`, before the digest, on both paths. `oversized_identity_still_ignores_key_order` pins this.
- **No wrong-refusal path in the identity scheme.** I probed the rename for merge collisions directly: a rename only fires when `id` is absent, so it can never collapse two calls that differ in the `id` value. `{"id":"a"}` vs `{"node_id":"b"}` stay distinct; the both-spellings form is its own third identity distinct from either. The two identity forms are unambiguous — canonical JSON always begins with `{`, digests with `sha256:` — so a digest cannot impersonate literal args.
- **Single-fetch refactor preserves semantics.** `prior_writes_from_history(&messages)` borrows before `node_history_from_messages(messages)` consumes; the early-return-on-empty-history path is unchanged, and logging is now emitted exactly once from `load_chat_messages` rather than depending on the old ordering coupling. No log line was lost or duplicated.

---

## Code Review Findings

### 🔴 Critical

#### 1. Making `canonical_args` mandatory silently wipes any existing conversation — and then persists the wipe

📁 `packages/core/src/models/ai_chat_node.rs:74` (field), consumed at `packages/core/src/models/ai_chat_node.rs:190-194`

This is **not** a backward-compatibility objection. Per CLAUDE.md this repo is greenfield with zero users, and the `Option<String>` → `String` change is the right call — I am not asking for a migration, a compat shim, or a `#[serde(default)]`. The problem is the *failure handling*, which the review brief explicitly flagged as the thing to check, and which is bad independently of whether any legacy data exists today.

`AiChatNode::from_node` decodes history with:

```rust
let messages = props
    .get("messages")
    .cloned()
    .map(|v| serde_json::from_value::<Vec<AiChatMessage>>(v).unwrap_or_default())
    .unwrap_or_default();
```

`canonical_args` is now a required field with no `default`. Serde fails the **entire `Vec<AiChatMessage>`** if any single nested `completedWrites` entry lacks `canonicalArgs`, and `unwrap_or_default()` converts that total failure into an empty history with **no log line at all**.

I verified this end-to-end rather than reasoning about it: decoding a two-message history whose assistant turn carries one write missing `canonicalArgs` yields `0` messages, not 1, not an error.

The blast radius is worse than a skipped turn, because the empty read is then written back:

1. `run_ai_chat_turn` (`local_agent_service.rs:413-418`) reads `history.is_empty()`, logs the *misleading* `"ai-chat history empty — skipping turn"`, and calls `write_ai_chat_status(&node_id, "idle", None)`.
2. `write_ai_chat_status` (`local_agent_service.rs:642-650`) does `AiChatNode::from_node(node)` — decoding to empty again — then `props["ai-chat"] = ai_chat.to_properties_value()`.
3. `to_properties_value` rebuilds `messages` wholesale from the struct: `map.insert("messages", serde_json::to_value(&self.messages)...)`.

So one undecodable write does not merely make the turn a no-op — it **overwrites the user's stored conversation with `[]`**. The user sees their entire chat history vanish, and the only diagnostic is a warning claiming the history was already empty.

The engineering principle is *fail loudly at the boundary, never silently to a default that is also a valid state*. `unwrap_or_default()` is only defensible when "absent" and "corrupt" mean the same thing to the caller; here they emphatically do not — one is a new chat, the other is data loss. This pre-existed the PR, but this PR is what makes it *reachable*, by introducing the first non-defaulted field in that struct. Tightening the type without tightening the decode is what turns a latent sharp edge into a live one.

**Suggested fix** (either is fine, both are small and neither is a compat shim):
- Make the decode failure explicit — propagate the `serde_json` error out of `from_node` as a `ValidationError`, so `load_chat_messages` logs it with the same `tracing::warn!(node_id, error = %e, ...)` treatment it already gives `"node is not an ai-chat node"`, and `write_ai_chat_status` returns `Err` instead of round-tripping an empty history back to storage. This is the option I would pick: it makes the tightened type actually enforceable rather than quietly absorbing violations.
- At minimum, log the decode error before defaulting, and make `write_ai_chat_status` refuse to persist a `messages` array that is empty when the raw stored value was not.

A regression test pinning "a chat node whose `messages` blob fails to decode is never written back as empty" would lock this down.

---

#### 2. Third mirror of `AiChatCompletedWrite` was not updated — the TypeScript type now misdescribes the wire shape

📁 `packages/nodespace-types/src/ai_chat.rs:17-18` vs 📁 `packages/desktop-app/src/lib/types/ai-chat-node.ts:36`

The PR updates two of the **three** declarations of this type. `packages/nodespace-types/src/ai_chat.rs` still reads:

```rust
#[serde(default, skip_serializing_if = "Option::is_none")]
pub canonical_args: Option<String>,
```

This is not a dead duplicate. It is the struct that actually produces the frontend payload: `node_to_typed_value` → `ai_chat_node_to_value` (`packages/nodespace-types/src/convert.rs:156-197`) builds `messages` through *this* `AiChatMessage`/`AiChatCompletedWrite`, and `packages/desktop-app/src-tauri/src/types.rs:11` re-exports it as the canonical wire type for the Tauri commands.

Consequences:

- **The TS type is now wrong.** `ai-chat-node.ts:36` declares `canonicalArgs: string` (required, "Always present"), but the emitting struct still carries `skip_serializing_if = "Option::is_none"`, so the field is **omitted** from the JSON whenever it is `None`. The TS contract asserts a guarantee the wire does not make. Answering the brief's question directly: the TypeScript mirror is consistent with `packages/core`'s type but **not** with the type that actually serialises to the frontend.
- **It re-opens Gap 1 on this path.** `convert.rs:178` uses the same `unwrap_or_default()` pattern, so this lenient mirror silently absorbs the mismatch instead of surfacing it — which is exactly why the test suite is green despite the divergence.

The principle is *single source of truth for a wire contract*. Three hand-synced declarations of one shape is the same class of accidental complexity the guard itself was built to remove, and this PR demonstrates the failure mode: a field was tightened in two places and drifted in the third. The doc comment on line 8 — `Mirrors nodespace_core::models::AiChatCompletedWrite` — is now false.

**Suggested fix:** change `packages/nodespace-types/src/ai_chat.rs:17-18` to `pub canonical_args: String` (dropping the `serde` attributes), matching `packages/core`. That makes the TS declaration true. Longer term, these two Rust structs should be one type re-exported, not two kept in sync by comment — but that is out of scope here and worth its own issue.

---

### 🟡 Important

#### 3. `normalize_param_aliases` is applied to every tool, including those with no such alias

📁 `packages/agent/src/local_agent/agent_loop.rs:100-109`

The rename is unconditional on tool name, so a stray top-level `node_id` is rewritten to `id` even for tools whose params struct has no `node_id` alias. I confirmed the behaviour directly:

- `{"from_id":"a","node_id":"b"}` → `{"from_id":"a","id":"b"}` (`create_relationship` has no such alias)
- `{"content":"x","node_type":"task","node_id":"stray"}` → `{"content":"x","id":"stray","node_type":"task"}` (`create_node` has neither param)

This is **not** a correctness bug for the guard — the rename is applied identically on record and execution sides, and the value is preserved, so no two distinct calls can merge. I want to be clear I am not claiming a wrong refusal here; I probed for one and could not construct it.

But the doc comment says "the name the params struct uses", which overstates what the code does: it applies one tool's alias rule globally. The function is correct-by-accident rather than correct-by-construction, and the comment would mislead the next reader who adds a tool with a semantically different top-level `node_id`. The issue's own suggested direction — "normalise argument keys through the tool's own params type" — would be correct by construction, though it is meaningfully more work and I do not think it is required to land this.

**Suggestion:** keep the implementation, but correct the doc comment to state what it does — *"rewrites the `node_id` → `id` alias at the top level for all tools, since it is the only alias in the tool surface and applying it uniformly keeps record and execution sides in lockstep without a per-tool table"* — and note the invariant that makes it safe (the rename never fires when `id` is present, so it cannot merge two distinct calls). Accuracy in a comment guarding a safety-critical identity function is worth the two lines.

#### 4. Verify the nested-`properties` exemption holds for `update_node`, not just in the unit test

📁 `packages/agent/src/local_agent/agent_loop.rs:100-109`, test at `:4696-4703`

Restricting to top-level keys is the right call, and I agree with the reasoning: a nested `properties.node_id` is user schema data and renaming it would corrupt the identity of a legitimate call. `alias_normalisation_does_not_reach_into_nested_user_data` pins it, and `search_nodes`' nested `filters[].node_id` (`tools.rs:315`) is correctly untouched — I checked.

The gap is that the test exercises a bare JSON literal rather than a shape `update_node` actually emits, and `update_node` is the tool that carries both a top-level aliased `id` *and* a user-controlled `properties` blob. A test using a realistic `{"id": "...", "properties": {"node_id": "..."}}` payload would pin the interaction that actually matters — that the top-level key is renamed while the nested one is not, in the same call. Cheap to add, and it is the case a future refactor is most likely to break.

*(Note: `update_node` is `IdempotentWrite` and therefore not in the guarded set today, so this is about locking in the invariant before it matters, not fixing a live bug.)*

---

### 🟢 Suggestions

#### 5. `assert!(identity.len() < 128)` is a loose proxy for a fixed-size property

📁 `packages/daemon/src/services/local_agent_service.rs:2657`

A SHA-256 hex digest plus the `sha256:` prefix is *exactly* 71 characters. Asserting `< 128` would still pass if the digest were accidentally truncated to 8 hex chars — which is precisely the collision-inducing regression the bound is meant to catch. `assert_eq!(identity.len(), 71)` states the actual invariant and fails on the change you care about.

#### 6. `delete_repeat_reaches_executor` has a redundant binding

📁 `packages/agent/src/local_agent/agent_loop.rs:4658-4659`

```rust
let ran = calls.lock().unwrap().iter().any(|c| c == "delete_node");
ran
```

**Nit:** collapses to a single tail expression.

#### 7. Test-module `load_node_history` helper is the right call

📁 `packages/daemon/src/services/local_agent_service.rs:1684-1690`

Answering the brief's question directly: this does **not** hide a gap. It composes the two production functions in the same order and with the same data flow that `run_ai_chat_turn` uses, so the four pre-existing tests still exercise the real code path; it only spares them the two-line composition. It is a test-local convenience over production functions, not a parallel implementation. No change needed.

---

## Assessment of test quality

The tests are substantive, not decorative. Specifically:

- The negative controls genuinely guard the wrong-suppression direction. `a_different_oversized_import_still_executes` uses two imports that are *both* oversized and differ only in their heading — so it fails if the digest collapsed "all large calls" into one identity, which is the realistic way this could go wrong. `delete_of_a_different_node_still_executes` varies the node ID across the alias boundary (`{"id":"n1"}` stored, `{"node_id":"n2"}` incoming), catching a rename that ignored the value. These are the right adversaries.
- `delete_repeat_is_refused_across_the_node_id_alias` correctly tests **both** directions, since either spelling can be recorded first.
- `oversized_identity_still_ignores_key_order` is the test I would most have wanted, and it is present — it pins that normalisation precedes hashing, which is the ordering mistake that would unguard large writes while leaving small ones guarded (i.e. invisible to every other test).
- The execution-layer tests drive `run_turn` through a recording executor and assert on whether the tool was *reached*, which is the behaviour the acceptance criteria describe, not a proxy for it.

I did not independently re-run the implementer's mutation testing, but the tests are constructed such that the claimed sensitivity is plausible — each new assertion targets a distinct failure mode rather than restating the implementation.

---

## Recommendation

### 🔄 REQUEST CHANGES

The core of this change is well-designed and I want to be clear about that: the hybrid verbatim/digest identity is the right resolution of the debuggability-vs-coverage tension the issue laid out, the two forms are provably unambiguous, the alias fix is placed correctly (in shared canonicalisation, so `seen_calls` and the cross-turn guard cannot drift), the single-fetch refactor is a clean simplification that preserves the early-return and logging semantics, and the documentation genuinely explains *why* rather than *what*. All four acceptance criteria are met and independently verified. This is a clear net improvement to code health.

Two things should be fixed before merge, neither large:

1. **Finding #2 is a near-mechanical fix** — one field in `packages/nodespace-types/src/ai_chat.rs` — and without it the shipped TypeScript type asserts a guarantee the wire does not honour. Landing a knowingly-false type declaration is the kind of thing that costs someone an afternoon later.
2. **Finding #1 is the blocker.** Not because a breaking change is wrong — it is correct and expected here — but because tightening the type without tightening the decode means a violation is absorbed by `unwrap_or_default()` into an empty history, which `write_ai_chat_status` then **persists**, destroying the stored conversation while logging a message that says it was already empty. Silent data loss with a misleading diagnostic is worth the small fix now.

Findings #3–#7 are non-blocking; #3 (comment accuracy on a safety-critical function) is the one I would still encourage taking in this PR.

malibio:
## Review Type
**Re-Review** — PR #1801, commit `a15aedfa` (previously reviewed at `fda7d1cc`). Scope: `git diff fda7d1cc..a15aedfa` only.

## Previous Review Summary

| | Count |
|---|---|
| Previous recommendations | 7 |
| Addressed | 6 |
| No action needed (agreed) | 1 (#7) |
| **New issues found this round** | **1 🟡** |

| # | Previous finding | Status |
|---|---|---|
| 1 | 🔴 Mandatory `canonical_args` silently wipes + persists erased conversations | ✅ **Fixed** in `packages/core` — see caveat under Focus Area 3 |
| 2 | 🔴 Third mirror (`nodespace-types`) not updated | ✅ **Fixed** |
| 3 | 🟡 `normalize_param_aliases` doc overstated scope | ✅ **Fixed** |
| 4 | 🟡 Unrealistic nested-`properties` test shape | ✅ **Fixed** |
| 5 | 🟢 Loose `assert!(identity.len() < 128)` | ✅ **Fixed** |
| 6 | 🟢 Redundant `let ran = …; ran` binding | ✅ **Correctly kept** — reasoning independently verified |
| 7 | 🟢 `load_node_history` test helper | ✅ No action, agreed |

## Independent Verification (not taken on trust)

- `cargo test -p nodespace-core --lib ai_chat` → **21 passed, 0 failed**, including all three new tests.
- Full gate suite `cargo test --lib --bins --workspace --exclude nodespace-app --exclude nodespace-app-test-support` → **1617 passed across 10 crate suites, 0 failed.**
- `cargo clippy -p nodespace-core -p nodespace-types -p nodespace-agent --lib --all-targets` → **clean, zero warnings.**
- **Finding 6's E0597 claim is correct.** I reproduced it in an isolated crate rather than assuming: `calls.lock().unwrap().iter().any(…)` as a tail expression genuinely fails with `` `calls` does not live long enough ``, and rustc's own `help:` output literally suggests the `let x = …; x` form the author used. The comment is accurate and the binding must stay.
- **Ollama characterisation confirmed.** `rust:test` is `cargo test --lib --bins`, so `packages/agent/tests/ollama_integration.rs` — an integration target — is never compiled or run by `test:all` or the pre-push gate. It self-skips unless a live server answers at `localhost:11434`. Outside this PR's blast radius; "pre-existing, not a regression" is a fair characterisation.
- Standards: no `#[allow(...)]`, no `#[deprecated]`, no `#[allow(dead_code)]`, no new `.md`, no issue numbers in code comments anywhere in the diff.
- **Mutation-test claim: NOT independently re-run.** Reproducing it required editing source, which my review constraints forbid. I verified it *analytically* instead, which I consider conclusive: `AiChatMessage.content` is a non-`Option` `String`, and `AiChatCompletedWrite.canonical_args` is now a non-`Option` `String` with no `#[serde(default)]`. Both fixture arrays therefore contain exactly one element that fails `from_value::<AiChatMessage>`. Under the old whole-`Vec` decode that error propagates to `unwrap_or_default()` → `vec![]`, and both `assert_eq!(len, 2)` assertions fail. The claim holds.

## Findings

### Focus Area 1 — Is Critical 1 fixed, and at the right layer?

📁 `packages/core/src/models/ai_chat_node.rs:191-218` — ✅ **Fixed, and I judge the layer correct.**

The per-message `filter_map` is the right call. My concrete opinion on the open design question: **containment plus loud logging is sufficient — do not add a write-path guard.** The counter-proposal is superficially attractive, so the reasoning is worth stating:

1. **A "refuse to write back a lossy history" guard would convert a degraded chat into a bricked one.** `write_ai_chat_status` is the *only* path that returns the node to `idle`. If it refuses to write on a lossy decode, the node stays pinned at `status: processing` forever — the spinner never stops, and every subsequent turn is skipped at the `history.is_empty()` check. Trading "one message lost" for "this conversation is permanently unusable and cannot self-heal" is a strictly worse failure mode. That is the KISS/YAGNI answer, not a compromise.
2. **The drop is now bounded, attributable, and observable.** `tracing::error!` carries `node_id` and the serde error. The prior failure was pernicious precisely because it was *silent* and *total*; it is now noisy and *partial*.
3. **Greenfield.** Per CLAUDE.md there are zero users and a resettable DB, so no persisted message can currently carry a legacy write missing `canonicalArgs`. The new tests defend the invariant going forward rather than rescuing real data.

One point of intellectual honesty the code already handles acceptably: a dropped message *does* drop its `completedWrites`, which could in principle let the duplicate guard re-execute a write it already performed. That is the guard degrading to its pre-guard behaviour for one message — not a new hazard — and the alternative is worse per (1).

📁 `packages/core/src/models/ai_chat_node.rs:191` — 🟢 **Nit:** the `.and_then(|v| v.as_array())` rewrite also avoids cloning the whole array. The per-element `m.clone()` remains, which is fine — bounded by message size and off the hot path.

### Focus Area 2 — Does `tracing::error!` in `packages/core` follow convention?

📁 `packages/core/src/models/ai_chat_node.rs:206` — ✅ **Legitimate, with one honest caveat.**

`tracing = { workspace = true }` is a declared dependency of `packages/core` (Cargo.toml:45), and the crate uses `tracing::` **152 times across ~15 modules** (`node_batch.rs`, `behaviors/mod.rs`, `db/sqlite_store/*`, the whole `playbook/` tree). This is unambiguously the crate's logging idiom, and the CLAUDE.md `console.log` prohibition is a frontend rule that does not apply here.

The caveat, stated plainly because the task asked me to check it: this **is the only `tracing::` call in all of `packages/core/src/models/`**. So it is a first for that specific directory. I do not think that blocks anything — `models/` is thin data-shape code that previously had nothing worth logging, and "no precedent in this directory" is not the same as "against convention" when the crate at large logs heavily. `error!` rather than `warn!` is also the right level: silent data loss is exactly what the previous round flagged.

### Focus Area 3 — Is the `nodespace-types` change complete?

📁 `packages/nodespace-types/src/ai_chat.rs:17-21` — ✅ **The declaration is fixed.** All four sites now agree the field is a required `String`:

| Site | Declaration |
|---|---|
| `packages/core/src/models/ai_chat_node.rs:74` | `pub canonical_args: String` |
| `packages/nodespace-types/src/ai_chat.rs:21` | `pub canonical_args: String` |
| `packages/agent/src/agent_types.rs:457` | `pub canonical_args: String` |
| `packages/desktop-app/src/lib/types/ai-chat-node.ts:36` | `canonicalArgs: string` |

No `skip_serializing_if` or `Option` remains on any of them; there is no `src-tauri/src/types.rs` mirror. The TS side is non-optional and typed `string` (no `any`).

**However — 🟡 the conversion layer was not checked, and it carries the same defect Critical 1 was about.**

📁 `packages/nodespace-types/src/convert.rs:175-179` — 🟡 **Improvement.** *Principle: DRY / consistency of invariants — the same rule enforced in two places should fail the same way in both.*

```rust
let messages = props
    .get("messages")
    .cloned()
    .map(|v| serde_json::from_value::<Vec<AiChatMessage>>(v).unwrap_or_default())
    .unwrap_or_default();
```

This is the **exact whole-`Vec` pattern** that `ai_chat_node.rs` just moved away from, still present in `ai_chat_node_to_value` — decoding into the very struct this commit just made stricter. Making `canonical_args` mandatory in `nodespace-types` therefore *created a new instance* of the original bug inside the same crate as the fix.

Blast radius is genuinely smaller than Critical 1, which is why this is 🟡 and not 🔴:
- `node_to_typed_value` is the **read/serialise path to the frontend** (`convert.rs:34` — "the single canonical implementation used by all entry points: Tauri commands, MCP, HTTP"). It does not write back to the DB, so a lossy decode here **displays** an empty conversation rather than **persisting** one. It is recoverable on reload; Critical 1 was not.
- It is also entirely silent — no `tracing` — so the same class of misleading-symptom debugging the previous round called out would recur, just as "the chat UI went blank" rather than "the history was erased."

Since the daemon flattens through this function on every ai-chat node read, one message failing to decode blanks the whole visible conversation in the UI. I'd apply the same `filter_map` + log treatment for symmetry. Not a merge blocker — no data is destroyed — but it is the one loose thread this commit left.

### Focus Area 4 — New issues introduced by `a15aedfa`

📁 `packages/core/src/models/ai_chat_node_test.rs:152-235` — ✅ **The three new tests are good.** Names state the invariant rather than the mechanism (`one_unreadable_message_does_not_erase_the_conversation`); the fixtures assert *which* messages survive and in what order, not merely the count; the inline comment marks precisely why element 2 is malformed. `completed_write_identity_round_trips` goes beyond what was asked and pins the positive case, so the mandatory field is covered in both directions.

📁 `packages/agent/src/local_agent/agent_loop.rs:4704-4721` — ✅ **Finding 4 properly addressed.** The test now exercises a real `update_node` shape and asserts *both* halves of the invariant: the top-level alias resolves to `"id"`, and the nested user field keeps both its key **and its value** (`"node_id":"user-value"`). The earlier `canonical.contains("node_id")` could have passed on the top-level key alone — the new form cannot.

📁 `packages/daemon/src/services/local_agent_service.rs:2661` — ✅ **`assert_eq!(identity.len(), 71)` is correct.** `"sha256:"` is 7 chars + 64 hex = 71. The comment explaining *why* exact-not-bounded is the right assertion is exactly the "comments explain why, not what" standard.

📁 `packages/agent/src/local_agent/agent_loop.rs:96-101` — ✅ **Doc comment now accurate.** It correctly states the blanket application, and the "renamed identically on both sides of the comparison, so the identity stays consistent either way" justification is the load-bearing argument — that is the real reason the blanket rule is safe.

No new lint suppressions, no dead code, no new dependencies, no backward-compat shims. Nothing in `a15aedfa` regresses anything.

## Recommendation

**COMMENT** — effectively an approve. *(GitHub rejects `--approve` here as author and reviewer are the same account.)*

Both Criticals from the previous round are genuinely resolved, and the fixes are reasoned rather than mechanical — particularly finding 6, where the author investigated, found the collapse doesn't compile, and documented *why* instead of silently leaving it or forcing a worse rewrite. That is the correct response to a review nit. All 1617 workspace Rust tests pass and clippy is clean.

The one new item (`convert.rs:175`) is a 🟡 that this commit surfaced by tightening the type — a read-path display bug, not data loss. **This is a net-positive change and I would not block merge on it.**

**Another review round is NOT needed. The PR is ready for the author to merge.** My suggestion: fix `convert.rs:175` in this PR if you want the invariant airtight in one commit — it is a ~10-line change mirroring one you already wrote — but a follow-up issue is an equally legitimate call given it cannot destroy data.

malibio (commented):
## Review Type
**Re-Review** — PR #1801, commit `a15aedfa` (previously reviewed at `fda7d1cc`). Scope: `git diff fda7d1cc..a15aedfa` only.

## Previous Review Summary

| | Count |
|---|---|
| Previous recommendations | 7 |
| Addressed | 6 |
| No action needed (agreed) | 1 (#7) |
| **New issues found this round** | **1 🟡** |

| # | Previous finding | Status |
|---|---|---|
| 1 | 🔴 Mandatory `canonical_args` silently wipes + persists erased conversations | ✅ **Fixed** in `packages/core` — see caveat under Focus Area 3 |
| 2 | 🔴 Third mirror (`nodespace-types`) not updated | ✅ **Fixed** |
| 3 | 🟡 `normalize_param_aliases` doc overstated scope | ✅ **Fixed** |
| 4 | 🟡 Unrealistic nested-`properties` test shape | ✅ **Fixed** |
| 5 | 🟢 Loose `assert!(identity.len() < 128)` | ✅ **Fixed** |
| 6 | 🟢 Redundant `let ran = …; ran` binding | ✅ **Correctly kept** — reasoning independently verified |
| 7 | 🟢 `load_node_history` test helper | ✅ No action, agreed |

## Independent Verification (not taken on trust)

- `cargo test -p nodespace-core --lib ai_chat` → **21 passed, 0 failed**, including all three new tests.
- Full gate suite `cargo test --lib --bins --workspace --exclude nodespace-app --exclude nodespace-app-test-support` → **1617 passed across 10 crate suites, 0 failed.**
- `cargo clippy -p nodespace-core -p nodespace-types -p nodespace-agent --lib --all-targets` → **clean, zero warnings.**
- **Finding 6's E0597 claim is correct.** I reproduced it in an isolated crate rather than assuming: `calls.lock().unwrap().iter().any(…)` as a tail expression genuinely fails with `` `calls` does not live long enough ``, and rustc's own `help:` output literally suggests the `let x = …; x` form the author used. The comment is accurate and the binding must stay.
- **Ollama characterisation confirmed.** `rust:test` is `cargo test --lib --bins`, so `packages/agent/tests/ollama_integration.rs` — an integration target — is never compiled or run by `test:all` or the pre-push gate. It self-skips unless a live server answers at `localhost:11434`. Outside this PR's blast radius; "pre-existing, not a regression" is a fair characterisation.
- Standards: no `#[allow(...)]`, no `#[deprecated]`, no `#[allow(dead_code)]`, no new `.md`, no issue numbers in code comments anywhere in the diff.
- **Mutation-test claim: NOT independently re-run.** Reproducing it required editing source, which my review constraints forbid. I verified it *analytically* instead, which I consider conclusive: `AiChatMessage.content` is a non-`Option` `String`, and `AiChatCompletedWrite.canonical_args` is now a non-`Option` `String` with no `#[serde(default)]`. Both fixture arrays therefore contain exactly one element that fails `from_value::<AiChatMessage>`. Under the old whole-`Vec` decode that error propagates to `unwrap_or_default()` → `vec![]`, and both `assert_eq!(len, 2)` assertions fail. The claim holds.

## Findings

### Focus Area 1 — Is Critical 1 fixed, and at the right layer?

📁 `packages/core/src/models/ai_chat_node.rs:191-218` — ✅ **Fixed, and I judge the layer correct.**

The per-message `filter_map` is the right call. My concrete opinion on the open design question: **containment plus loud logging is sufficient — do not add a write-path guard.** The counter-proposal is superficially attractive, so the reasoning is worth stating:

1. **A "refuse to write back a lossy history" guard would convert a degraded chat into a bricked one.** `write_ai_chat_status` is the *only* path that returns the node to `idle`. If it refuses to write on a lossy decode, the node stays pinned at `status: processing` forever — the spinner never stops, and every subsequent turn is skipped at the `history.is_empty()` check. Trading "one message lost" for "this conversation is permanently unusable and cannot self-heal" is a strictly worse failure mode. That is the KISS/YAGNI answer, not a compromise.
2. **The drop is now bounded, attributable, and observable.** `tracing::error!` carries `node_id` and the serde error. The prior failure was pernicious precisely because it was *silent* and *total*; it is now noisy and *partial*.
3. **Greenfield.** Per CLAUDE.md there are zero users and a resettable DB, so no persisted message can currently carry a legacy write missing `canonicalArgs`. The new tests defend the invariant going forward rather than rescuing real data.

One point of intellectual honesty the code already handles acceptably: a dropped message *does* drop its `completedWrites`, which could in principle let the duplicate guard re-execute a write it already performed. That is the guard degrading to its pre-guard behaviour for one message — not a new hazard — and the alternative is worse per (1).

📁 `packages/core/src/models/ai_chat_node.rs:191` — 🟢 **Nit:** the `.and_then(|v| v.as_array())` rewrite also avoids cloning the whole array. The per-element `m.clone()` remains, which is fine — bounded by message size and off the hot path.

### Focus Area 2 — Does `tracing::error!` in `packages/core` follow convention?

📁 `packages/core/src/models/ai_chat_node.rs:206` — ✅ **Legitimate, with one honest caveat.**

`tracing = { workspace = true }` is a declared dependency of `packages/core` (Cargo.toml:45), and the crate uses `tracing::` **152 times across ~15 modules** (`node_batch.rs`, `behaviors/mod.rs`, `db/sqlite_store/*`, the whole `playbook/` tree). This is unambiguously the crate's logging idiom, and the CLAUDE.md `console.log` prohibition is a frontend rule that does not apply here.

The caveat, stated plainly because the task asked me to check it: this **is the only `tracing::` call in all of `packages/core/src/models/`**. So it is a first for that specific directory. I do not think that blocks anything — `models/` is thin data-shape code that previously had nothing worth logging, and "no precedent in this directory" is not the same as "against convention" when the crate at large logs heavily. `error!` rather than `warn!` is also the right level: silent data loss is exactly what the previous round flagged.

### Focus Area 3 — Is the `nodespace-types` change complete?

📁 `packages/nodespace-types/src/ai_chat.rs:17-21` — ✅ **The declaration is fixed.** All four sites now agree the field is a required `String`:

| Site | Declaration |
|---|---|
| `packages/core/src/models/ai_chat_node.rs:74` | `pub canonical_args: String` |
| `packages/nodespace-types/src/ai_chat.rs:21` | `pub canonical_args: String` |
| `packages/agent/src/agent_types.rs:457` | `pub canonical_args: String` |
| `packages/desktop-app/src/lib/types/ai-chat-node.ts:36` | `canonicalArgs: string` |

No `skip_serializing_if` or `Option` remains on any of them; there is no `src-tauri/src/types.rs` mirror. The TS side is non-optional and typed `string` (no `any`).

**However — 🟡 the conversion layer was not checked, and it carries the same defect Critical 1 was about.**

📁 `packages/nodespace-types/src/convert.rs:175-179` — 🟡 **Improvement.** *Principle: DRY / consistency of invariants — the same rule enforced in two places should fail the same way in both.*

```rust
let messages = props
    .get("messages")
    .cloned()
    .map(|v| serde_json::from_value::<Vec<AiChatMessage>>(v).unwrap_or_default())
    .unwrap_or_default();
```

This is the **exact whole-`Vec` pattern** that `ai_chat_node.rs` just moved away from, still present in `ai_chat_node_to_value` — decoding into the very struct this commit just made stricter. Making `canonical_args` mandatory in `nodespace-types` therefore *created a new instance* of the original bug inside the same crate as the fix.

Blast radius is genuinely smaller than Critical 1, which is why this is 🟡 and not 🔴:
- `node_to_typed_value` is the **read/serialise path to the frontend** (`convert.rs:34` — "the single canonical implementation used by all entry points: Tauri commands, MCP, HTTP"). It does not write back to the DB, so a lossy decode here **displays** an empty conversation rather than **persisting** one. It is recoverable on reload; Critical 1 was not.
- It is also entirely silent — no `tracing` — so the same class of misleading-symptom debugging the previous round called out would recur, just as "the chat UI went blank" rather than "the history was erased."

Since the daemon flattens through this function on every ai-chat node read, one message failing to decode blanks the whole visible conversation in the UI. I'd apply the same `filter_map` + log treatment for symmetry. Not a merge blocker — no data is destroyed — but it is the one loose thread this commit left.

### Focus Area 4 — New issues introduced by `a15aedfa`

📁 `packages/core/src/models/ai_chat_node_test.rs:152-235` — ✅ **The three new tests are good.** Names state the invariant rather than the mechanism (`one_unreadable_message_does_not_erase_the_conversation`); the fixtures assert *which* messages survive and in what order, not merely the count; the inline comment marks precisely why element 2 is malformed. `completed_write_identity_round_trips` goes beyond what was asked and pins the positive case, so the mandatory field is covered in both directions.

📁 `packages/agent/src/local_agent/agent_loop.rs:4704-4721` — ✅ **Finding 4 properly addressed.** The test now exercises a real `update_node` shape and asserts *both* halves of the invariant: the top-level alias resolves to `"id"`, and the nested user field keeps both its key **and its value** (`"node_id":"user-value"`). The earlier `canonical.contains("node_id")` could have passed on the top-level key alone — the new form cannot.

📁 `packages/daemon/src/services/local_agent_service.rs:2661` — ✅ **`assert_eq!(identity.len(), 71)` is correct.** `"sha256:"` is 7 chars + 64 hex = 71. The comment explaining *why* exact-not-bounded is the right assertion is exactly the "comments explain why, not what" standard.

📁 `packages/agent/src/local_agent/agent_loop.rs:96-101` — ✅ **Doc comment now accurate.** It correctly states the blanket application, and the "renamed identically on both sides of the comparison, so the identity stays consistent either way" justification is the load-bearing argument — that is the real reason the blanket rule is safe.

No new lint suppressions, no dead code, no new dependencies, no backward-compat shims. Nothing in `a15aedfa` regresses anything.

## Recommendation

**COMMENT** — effectively an approve. *(GitHub rejects `--approve` here as author and reviewer are the same account.)*

Both Criticals from the previous round are genuinely resolved, and the fixes are reasoned rather than mechanical — particularly finding 6, where the author investigated, found the collapse doesn't compile, and documented *why* instead of silently leaving it or forcing a worse rewrite. That is the correct response to a review nit. All 1617 workspace Rust tests pass and clippy is clean.

The one new item (`convert.rs:175`) is a 🟡 that this commit surfaced by tightening the type — a read-path display bug, not data loss. **This is a net-positive change and I would not block merge on it.**

**Another review round is NOT needed. The PR is ready for the author to merge.** My suggestion: fix `convert.rs:175` in this PR if you want the invariant airtight in one commit — it is a ~10-line change mirroring one you already wrote — but a follow-up issue is an equally legitimate call given it cannot destroy data.

malibio:
## `/address-review` — verification pass

Run after the fact: the review findings were addressed manually in `a15aedfa` and `dbd0e8af` before this command was invoked. This pass re-derives the finding list from the PR comments and verifies each against the current tree, rather than trusting the earlier commits' claims.

**No new code changes were required — every finding was already resolved.**

### Round 1 (7 findings)

| # | Severity | Finding | State in tree | Evidence |
|---|---|---|---|---|
| 1 | 🔴 | Mandatory `canonicalArgs` wipes + persists erased conversations | ✅ Addressed | Per-message decode at `packages/core/src/models/ai_chat_node.rs:199` |
| 2 | 🔴 | Third mirror (`nodespace-types`) not updated | ✅ Addressed | `canonical_args: String` at `packages/nodespace-types/src/ai_chat.rs:21` |
| 3 | 🟡 | `normalize_param_aliases` doc overstated scope | ✅ Addressed | `agent_loop.rs:96` |
| 4 | 🟡 | Unrealistic nested-`properties` test shape | ✅ Addressed | `agent_loop.rs:4712,4719` — asserts both halves of the invariant |
| 5 | 🟢 | Loose `assert!(identity.len() < 128)` | ✅ Addressed | `assert_eq!(identity.len(), 71)` at `local_agent_service.rs:2664` |
| 6 | 🟢 | Redundant `let ran = …; ran` binding | ⏭️ **Skipped, deliberately** | Collapsing it fails to compile (E0597 — the temporary `MutexGuard` outlives `calls`). Binding kept with an explanatory comment at `agent_loop.rs:4661`. The re-review independently reproduced the error and agreed. |
| 7 | 🟢 | `load_node_history` test helper | ⏭️ No action — reviewer explicitly concluded no change needed |

### Round 2 (1 new finding)

| # | Severity | Finding | State in tree | Evidence |
|---|---|---|---|---|
| 8 | 🟡 | Same whole-`Vec` decode in the conversion layer | ✅ Addressed in `dbd0e8af` | Per-message decode at `packages/nodespace-types/src/convert.rs:182`, plus a regression test |

The round-2 🟢 nit on the residual per-element `m.clone()` was not actioned — the reviewer's own text says it "is fine — bounded by message size and off the hot path."

### Verification

- `bun run quality:fix` — exits 0. (Reformatted two unrelated files, `collection_service.rs` and `node_service/embedding.rs`, via `cargo fmt --all`; reverted, as they are outside this PR's scope.)
- `bun run test:all` — **exits 0**: 4106 frontend, 20 browser, 10 Rust crate suites, zero failures.
- Every behavioural fix in this PR was mutation-tested — each fix reverted individually, confirming the corresponding test fails. No test passes vacuously.

### Two pre-existing failures, explicitly not regressions

- **`ollama_integration` (5 tests)** — fail identically on unmodified `main` (`beb56627`); they require a live Ollama server. `rust:test` is `cargo test --lib --bins`, so this integration target is never run by `test:all` or the pre-push gate.
- **`playbook::validation::…::test_wrong_field_count_cron_rejected`** — failed once during this pass with `Failed to set journal_mode` (SQLite contention under a 4-crate parallel invocation). Did not reproduce in 3 subsequent worktree runs or 3 runs on clean `main`. Transient, load-dependent, unrelated to this change.

## Re-Review Decision

**Rationale:** This pass produced zero code changes — it verified work already reviewed. The re-review at `dbd0e8af`'s predecessor already stated another round was not needed, and the one item it surfaced (`convert.rs`) was then fixed with a mutation-tested regression test. The PR is ready for the author to merge.

PR #1803, opened 2026-07-29
Fix dev-proxy gRPC-js reconnect backoff so it recovers when the daemon socket appears

mstomar125:
Closes #1551.

## Problem
The Bun dev-proxy builds its gRPC-js client once at startup. When the proxy starts before the daemon binds its socket (which `DaemonTestHarness.startDeferred()` does deliberately), the channel's subchannel enters an ever-growing reconnect backoff (grpc-js default 1s→120s) that is independent of the socket actually becoming reachable moments later — so the next RPC fails silently with UNAVAILABLE. A real gap in the ADR-048 e2e harness that a raw socket probe can't detect. The production Tauri/tonic path has no equivalent cached-backoff state and is untouched.

## Fix
- Extract the gRPC client setup into `packages/dev-tools/src/grpc-client.ts` so the not-ready→recovered behavior is regression-testable in isolation.
- **Bound the reconnect backoff** to a fixed ~100ms via `grpc.initial_reconnect_backoff_ms` / `grpc.max_reconnect_backoff_ms` — the channel re-probes ~10×/second and recovers within ~100ms of the socket appearing (verified: ~127ms with the fix vs ~2234ms without).
- **Gate every RPC** (and the WatchNodes bridge) on `client.waitForReady(deadline)` so a call issued right after the daemon binds waits for READY instead of failing instantly.
- Add `dev-proxy-recovery.e2e.ts`: one RPC issued before a late-binding stub server, asserted to resolve promptly with no real-RPC polling. It fails without the fix and passes with it.
- Simplify `DaemonTestHarness.waitUntilProxyReady()` to socket reachability now that the proxy self-heals at the source.

## Notes
- I read grpc-js 1.14.4's source to confirm the mechanism: the backoff lives on the subchannel; pick-first's `resetBackoff` is a no-op and `startConnecting()` doesn't restart the timer — so bounding the backoff via channel options is the only public lever that works. Documented in the module header.
- `@grpc/grpc-js@1.14.4` + `@grpc/proto-loader` added as desktop-app **devDependencies** for the test only; the community/production bundle is unaffected (production uses tonic). Root already pins grpc-js to 1.14.4.

## Verification
Full pre-push gate green (test:all + daemon build + e2e). New regression test + the `startDeferred()` recovery test both pass against a real daemon.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -42,6 +42,8 @@
       },
       "devDependencies": {
         "@eslint/js": "^9.32.0",
+        "@grpc/grpc-js": "1.14.4",
+        "@grpc/proto-loader": "^0.7.0",
         "@internationalized/date": "^3.8.1",
         "@lucide/svelte": "^0.515.0",
         "@sveltejs/adapter-static": "^3.0.6",
--- a/packages/desktop-app/package.json
+++ b/packages/desktop-app/package.json
@@ -60,6 +60,8 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.32.0",
+    "@grpc/grpc-js": "1.14.4",
+    "@grpc/proto-loader": "^0.7.0",
     "@internationalized/date": "^3.8.1",
     "@lucide/svelte": "^0.515.0",
     "@sveltejs/adapter-static": "^3.0.6",
--- a/packages/desktop-app/src/tests/e2e/daemon-harness.ts
+++ b/packages/desktop-app/src/tests/e2e/daemon-harness.ts
@@ -312,37 +312,28 @@ export class DaemonTestHarness {
   }
 
   /**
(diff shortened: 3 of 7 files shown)


mstomar125:
**APPROVE** — Independent review + adversarial re-review both CONFIRMED. Empirically validated the fix (channel recovers ~127ms after socket bind vs ~2234ms unpatched; the regression test genuinely fails without the fix). Refactor is behavior-preserving (same proto paths/loader options), `waitForReady` guarantees no unbounded hang (rejects at deadline), simplified `waitUntilProxyReady()` verified safe for its sole caller, and the added grpc-js deps are dev-only (community/production tonic path untouched). Full pre-push gate green.

PR #1805, opened 2026-07-29
Make the create_schema eval assertions able to see a failed schema creation (#1799)

malibio:
Closes #1799

## What this changes

The `noExtraTypes` assertion counted `create_schema` calls and never checked what was persisted, so a call that was rejected outright scored as a pass. Scenario 3 had been a recorded false pass in every prior run.

The evidence needed to catch it was already produced and already printed — `aichat.ts` emits an `[ERROR]` marker per tool call — but `runner.ts` kept only the tool name and discarded the rest of the line. This carries the outcome through.

- **`agent_loop.rs`** logs `result_field_count`, taken from the tool RESULT rather than its arguments. Args are the model's report of what it asked for; the result is what the executor persisted. A bare integer because both previews truncate at 300 chars and a realistic `create_schema` payload exceeds that — a parser reading them would fail on exactly the well-formed calls it should pass.
- **`types.ts`** gains `toolCalls` alongside `toolsCalled` rather than replacing it, so `routing.ts` needs no edits and recorded baselines still join in `compareToBaseline`.
- **`agent-matrix.ts`** checks schema soundness on **both** the `noExtraTypes` and `toolOnce` branches. Scenarios 8a/8b reach `create_schema` through `toolOnce` and had the identical count-only hole; fixing only `noExtraTypes` (what the issue named) would have left most of the measurement defect in place.

## Two failure modes, not one

A rejected call persists nothing. But a call carrying neither `fields` nor `description` **succeeds** and persists a type with no properties (`packages/core/src/schema/mod.rs:275-277`) — invisible to any check that only asks whether the call errored. The issue's framing would have missed this; 8a/8b are exactly this case.

Absent `fieldCount` is treated as unknown, so baselines recorded before this do not read as fresh failures.

## Verification

Both branches mutation-tested: reverting either one fails exactly its own two unit tests and no others.

Run against **unmodified** guidance — so the assertion is demonstrated failing on current code, per the issue's first acceptance criterion:

- **3** — `create_schema was called but REJECTED — no schema persisted`
- **8a** — `create_schema succeeded but persisted a type with NO fields`
- **8b** — same

`bun run test:all` passes (4,106 frontend / 20 browser / 29 scripts / 10 Rust suites), no new failures.

## Scope: assertion only

The issue's checklist items 3-5 (routing `create_schema` through `search_skills`) are **not** included. That change was implemented and measured, and it makes behavior worse: 6/12 → 3/12 under identical conditions, with the model calling `search_skills` and then stopping. Its premise also does not hold — `SCHEMA_CREATION_RULES` is resident every turn and already carries the rule the model broke. Detail in the issue comments; the underlying guidance defect is now the spike #1802.

**No baseline is recorded here.** `scripts/results-e4b.json` is deliberately untouched: every run it holds was measured through an architecture #1802 is questioning, and at n=12 with only 3 schema scenarios the matrix cannot clear its own ~2-scenario noise floor. The assertion fix stands on mutation testing, not on any eval score.

## Note on the branch name

Created via the worktree tool, which names branches `worktree-<name>`; the CLAUDE.md convention is `issue-<number>-brief-desc`. Cosmetic only.

https://claude.ai/code/session_01C2fwCrhS3Bfq4uDGM61iCr

### DIFF
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -933,10 +933,22 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                     ),
                 };
 
+                // Field count from the tool RESULT, not its arguments: the result is the
+                // executor's report of what it persisted, while args are only the model's
+                // report of what it asked for. Logged as a bare integer because both
+                // previews truncate at 300 chars — a realistic create_schema payload
+                // exceeds that, so a parser reading them would fail on exactly the
+                // well-formed calls it is meant to pass.
+                let result_field_count = result_value
+                    .get("fields")
+                    .and_then(|f| f.as_array())
+                    .map(|a| a.len());
+
                 tracing::info!(
                     tool = %tc.function_name,
                     is_error,
                     duration_ms,
+                    result_field_count,
                     args_preview = %args.to_string().chars().take(300).collect::<String>(),
                     result_preview = %result_value.to_string().chars().take(300).collect::<String>(),
                     "Tool executed"
--- a/scripts/aichat.ts
+++ b/scripts/aichat.ts
@@ -157,11 +157,13 @@ function reportTurnLog(sinceByte: number): void {
     const tool = l.match(/tool="?([a-z_]+)"?/)?.[1] ?? "?";
     const args = l.match(/args_preview="?([^"]*?)"? result_preview/)?.[1] ?? "";
     const err = /is_error=true/.test(l) ? " [ERROR]" : "";
-    // Field count of the persisted result. tracing omits the field entirely when
-    // it is None, so "absent" (not a schema-creating tool) stays distinguishable
-    // from "=0" (a schema persisted with no properties) — the latter is a real
-    // failure that looks identical to success by tool name alone. Emitted before
-    // the args, which are free-form and truncated and so must stay last.
+    // Field count of the persisted result, emitted by any tool whose result
+    // carries a top-level `fields` array. tracing omits the field entirely when
+    // it is None, so "absent" (the result reports no fields) stays
+    // distinguishable from "=0" (a schema persisted with no properties) — the
+    // latter is a real failure that looks identical to success by tool name
+    // alone. Emitted before the args, which are free-form and truncated at the
+    // source and so must stay last on the line.
     const fields = l.match(/result_field_count=(\d+)/)?.[1];
     const fieldPart = fields === undefined ? "" : ` [fields=${fields}]`;
     console.log(`[tool] ${tool}${err}${fieldPart} ${args}`);
--- a/scripts/aichat.ts
+++ b/scripts/aichat.ts
@@ -30,7 +30,8 @@ import { fileURLToPath } from "node:url";
 
 const WORKTREE = resolve(dirname(fileURLToPath(import.meta.url)), "..");
 const NS_BIN = process.env.NS_BIN ?? join(WORKTREE, "target/release/nodespace");
-const SOCKET = process.env.NODESPACED_SOCKET ?? "/tmp/nodespaced-test/daemon.sock";
+const SOCKET =
+  process.env.NODESPACED_SOCKET ?? "/tmp/nodespaced-test/daemon.sock";
 const NS_LOG = process.env.NS_LOG ?? "/tmp/nodespaced-test/daemon.log";
 const NS_MODEL = process.env.NS_MODEL ?? "gemma-4-e4b-q4km";
 const TIMEOUT_MS = Number(process.env.NS_TIMEOUT_MS ?? 180_000);
@@ -44,10 +45,13 @@ interface AiChat {
 
 /** Run the nodespace CLI with --json and parse stdout. Throws on non-zero exit. */
 function ns(args: string[]): unknown {
-  const result = Bun.spawnSync([NS_BIN, "--socket", SOCKET, "--json", ...args], {
-    stdout: "pipe",
-    stderr: "pipe",
-  });
+  const result = Bun.spawnSync(
+    [NS_BIN, "--socket", SOCKET, "--json", ...args],
+    {
+      stdout: "pipe",
+      stderr: "pipe",
+    },
+  );
   const stdout = result.stdout.toString();
   if (result.exitCode !== 0) {
     throw new Error(
@@ -84,14 +88,25 @@ function defaultAiChat(): AiChat {
   return { provider: "native", model: NS_MODEL, status: "idle", messages: [] };
 }
 
-function batchUpdateProps(id: string, version: number | null, props: Record<string, unknown>): void {
+function batchUpdateProps(
+  id: string,
+  version: number | null,
+  props: Record<string, unknown>,
+): void {
   const item: Record<string, unknown> = { node_id: id, properties: props };
   if (version !== null) item.version = version;
   nsRaw(["node", "batch-update", "--updates", JSON.stringify([item])]);
 }
 
 function cmdNew(): string {
-  const created = ns(["node", "create", "--type", "ai-chat", "--content", "CLI test chat"]) as {
+  const created = ns([
+    "node",
+    "create",
+    "--type",
+    "ai-chat",
+    "--content",
+    "CLI test chat",
+  ]) as {
     id: string;
   };
   if (!created?.id) throw new Error("create returned no id");
@@ -113,7 +128,14 @@ function reportTurnLog(sinceByte: number): void {
   try {
     const buf = Bun.file(NS_LOG);
     // Read only the bytes appended during this turn.
-    slice = stripAnsi(Bun.spawnSync(["tail", "-c", `+${sinceByte + 1}`, NS_LOG]).stdout.toString());
+    slice = stripAnsi(
+      Bun.spawnSync([
+        "tail",
+        "-c",
+        `+${sinceByte + 1}`,
+        NS_LOG,
+      ]).stdout.toString(),
+    );
     if (!slice && buf) slice = "";
   } catch {
     return;
@@ -124,7 +146,9 @@ function reportTurnLog(sinceByte: number): void {
     const m = offered.match(/selected_tools="?([^"]*)"?/);
     if (m) console.log(`[tools offered] ${m[1]}`);
   }
-  const prepared = lines.filter((l) => l.includes("system prompt and tools prepared")).pop();
+  const prepared = lines
+    .filter((l) => l.includes("system prompt and tools prepared"))
+    .pop();
   if (prepared) {
     const m = prepared.match(/tool_names="?([^"]*?)"? system_prompt_len/);
     if (m) console.log(`[tools available] ${m[1]}`);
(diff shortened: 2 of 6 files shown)


malibio:
## Pragmatic Code Review — PR #1805

**Review Type**: Initial Review (0 prior reviews/comments)

**Verdict up front**: this is a measurement fix that makes a previously-invisible failure visible, and it is correct where it matters most. I verified the central risk myself — that the check might not actually catch its target bug — and it does. Two findings, both about comments/redundancy rather than behavior. Net positive; approving.

---

### Requirements Check (issue #1799)

Acceptance criteria:

- ✅ **`noExtraTypes` (or companion) fails when a schema is created with no fields, demonstrated failing on current code before the guidance fix.** `schemaCallsAreSound` (`scripts/eval/fixtures/agent-matrix.ts:81-102`) fails both on rejection and on zero-field success. Run against unmodified guidance, scenarios 3 / 8a / 8b now fail honestly.
- ✅ **Scenario 3 passes on an assertion that could have failed.** Inverted correctly: scenario 3 now *fails*, which is the honest outcome. The baseline file's own `caveats` field independently documents it as "a FALSE PASS in all three runs" — this PR closes exactly that hole.
- ✅ **Per-scenario tool traces and failure reasons recorded, not just a headline count.** `extra()` now records `toolCalls` (`agent-matrix.ts:299-310`), and failure strings name the specific mode.
- ✅ **No worked example added to guidance reuses eval prompt wording.** No guidance text touched at all.

Work checklist items 3-5 (routing `create_schema` through `search_skills`) — **deliberately not implemented.**

📝 **On the scope decision — I think it is correct, and I want to be explicit about why rather than just deferring to the author.** The change was implemented and measured, not hand-waved: 6/12 → 3/12 under controlled conditions (purged DB, rebuilt daemon, `skipped=0` verified). More importantly the item's *premise* is falsifiable and was falsified — `SCHEMA_CREATION_RULES` is resident every turn and already carries the rule the model broke, so "the model never sees the guidance" was not the actual defect. Shipping a change that is measurably worse in order to tick a checkbox would be process theater. Splitting the real problem out to spike #1802 is the right call.

📝 **On not recording a baseline (item 2) — also correct, and this is the more interesting judgment.** The issue itself documents that three runs of *identical* code scored 10/12, 9/12, 8/12 with scenario 5 flipping pass/fail/pass. At n=12 with only 3 schema-exercising scenarios, a recorded number could not clear its own noise floor, and committing one would manufacture false precision that later work would over-read. The issue also explicitly de-targets the headline score. Critically, **the correctness claim here does not rest on an eval score at all** — it rests on mutation testing, which is a deterministic argument. Recording a noisy number would have added no evidence and some risk.

---

### Verification I performed

I did not take the PR's claims on trust. Specifically:

**Mutation testing — claim confirmed exactly.** Reverting the `toolOnce` branch alone → exactly 2 failures, both `toolOnce` tests. Reverting `noExtraTypes` alone → exactly 2 failures, both `noExtraTypes` tests. No cross-talk, no test passing for the wrong reason. Working tree restored and verified clean.

**The decisive correctness question: does a fieldless success actually produce `fields: []` rather than an absent key?** This is the one thing that could have silently defeated the whole PR — since absent is treated as *pass*, a fieldless `create_schema` returning no `fields` key would mean the check never fires on its target bug. It holds:
- `CreateSchemaOutput.fields` (`packages/core/src/schema/mod.rs:193`) has **no `skip_serializing_if`**, unlike its `relationships` and `warnings` siblings which do.
- The fieldless path (`mod.rs:275-278`) stores `Vec::new()`, serializing to `"fields": []` → `Some(0)`, not `None`.
- `ok_result` passes the value through unwrapped, so the tool result *is* `CreateSchemaOutput`.

**The absent-vs-zero distinction holds at the tracing layer.** `impl<T: Value> Value for Option<T>` in tracing-core records nothing when `None`, so the field is genuinely omitted rather than rendered as `None`. The three-way distinction (absent / 0 / N) is real.

**Regex parity — `toolsCalled` is byte-identical.** Tested against 8 line shapes including `[ERROR]`, `[fields=N]`, both, trailing-space, the `[tool] ?` fallback, and args containing an injected `[tool] ... [fields=9]` string. Group 1 is unchanged in every case: the optional groups sit *after* `([a-z_]+)` and neither can begin with `[a-z_]`, so no backtracking can shorten the name capture. All 14 tools in `Tool::ALL` are lowercase+underscore — no digits, no `search_v2`-style name to truncate.

**Marker-position contract is sound.** `result_field_count` is declared *before* `args_preview`/`result_preview` at the tracing site (`agent_loop.rs:942-951`), so tracing emits it first. Free-form truncated content cannot precede the structured marker, which is what makes "structured markers before args" enforceable rather than merely conventional.

**Error paths cannot masquerade as zero-field successes.** All three error shapes (`agent_loop.rs:922`, `:923-933`, `error_result` at `tools.rs:214-221`) emit `{"error": ...}` only, so `result_field_count` is `None` and `isError` is `true`. The two assertion branches are cleanly disjoint.

`cargo clippy -p nodespace-agent --lib` clean. `bun run test:scripts` 29/29.

---

### Findings

#### 🟡 Important — the comment justifying `toolsCalled`'s retention states a mechanism that does not exist
📁 `scripts/eval/types.ts:38-45`

The comment says:

> `compareToBaseline` joins recorded runs against new ones. Changing it would orphan the recorded baselines

`compareToBaseline` (`scripts/eval/runner.ts:153-209`) reads exactly three things from the recorded JSON: `baseline.provenance` (:170), and `results[].id` / `results[].passed` (:178-193). It never reads `toolsCalled`, `toolCalls`, `turns`, or `extra`. Removing or renaming `toolsCalled` would orphan nothing.

**The decision to keep both fields is still right** — but for a different and stronger reason the comment does not give: `routing.ts` consumes `t.toolsCalled` from **live** turns at 8 sites (:240, :256, :260, :285, :318, :327, :328, :359) and has its own scoring that never touches `assertExpectation`. That is a genuine code-level consumer, which also means this is **not** a backward-compatibility artifact under CLAUDE.md's greenfield rule — it is a field with a current reader. The conclusion survives; the stated justification should be corrected to the real one.

**Engineering principle**: a comment asserting a specific mechanism is a load-bearing claim. When it is wrong, the next person either propagates the error or wastes a cycle disproving it — and here it would misdirect anyone later evaluating whether the field can be removed.

#### 🟢 Suggestion — `toolCalls` optionality guards a deserialization path nothing exercises
📁 `scripts/eval/types.ts:47-53`, `scripts/eval/runner.ts:83-89`

`toolCalls?` is optional so old results files deserialize. But nothing deserializes recorded `TurnRecord`s (see above). Repo-wide, `TurnRecord` is constructed in exactly two places, both in `runner.ts`: the success path (:107) which sets `toolCalls`, and the send-failure path (:83) which does not. So making it required would break exactly **one** in-repo site, not the eight implied.

Not worth changing now — the `?? []` at the two read sites is harmless and the send-failure path genuinely has no calls to report. Worth knowing that the meaningful "absence means unknown" handling lives on `fieldCount`, not on `toolCalls`; that one is doing real work and is correctly tested (`agent-matrix.test.ts:245-263`).

#### 🟢 Suggestion — the `aichat.ts` comment slightly overstates which tools emit the marker
📁 `scripts/aichat.ts:160-164`

The comment says absence distinguishes "not a schema-creating tool" from `=0`. `get_node` on a **schema** node also emits a top-level `fields` array (`SchemaNode.fields`, `packages/nodespace-types/src/schema.rs:145`, no `skip_serializing_if`), so `[tool] get_node [fields=4]` can appear.

This is **harmless and arguably correct** — for `get_node` on a schema, `fields` is also the persisted field list, so the number means the same thing, and `schemaCallsAreSound` filters on `c.name !== "create_schema"` before reading `fieldCount` (`agent-matrix.ts:83`), so no spurious failure is reachable. I checked all 14 tools in `Tool::ALL`: `create_schema` and schema-shaped `get_node` are the only two, and notably **`update_schema` is not** one — `SchemaUpdateOutput` is counts-only (`fieldsAdded`, a `usize`), so `.as_array()` rejects it. Comment accuracy only.

#### 🟢 Nit — `extra.toolCalls` duplicates `turns[].toolCalls`
📁 `scripts/eval/fixtures/agent-matrix.ts:299-310`

Both are serialized into the results file, so each call is recorded twice. Justified by the stated intent (human-readable evidence colocated with the verdict) and the volume is trivial. Noting only so it is a known duplication rather than a later surprise.

---

### What I checked and found no issue with

- **Security**: no user input, no auth surface, no secrets. The `fields` read is a total function over untrusted JSON — `.get()`/`.as_array()`/`.map()` cannot panic on a hostile shape.
- **Standards**: no lint suppression, no `any`, no `#[allow(dead_code)]`, no `#[deprecated]`, no raw `console.log` in production code (`aichat.ts` and `runner.ts` are tooling scripts, exempt), no issue numbers in code comments. `[a-z_]+` truncating a hypothetical `search_v2` is **pre-existing on both old and new regexes** — not a regression.
- **Test quality**: the tests do not assert the implementation back to itself. They encode the two distinct real-world failure modes separately, assert on failure-message content, cover the stale-baseline absence case, and include a negative control that an unrelated failing tool is ignored. The `toolOnce still reports a wrong call count before outcome` test correctly pins ordering so the new check cannot mask the pre-existing count assertion.

---

The core defect is fixed, the fix is verified by deterministic mutation testing rather than a noisy score, and I independently confirmed the one condition that could have silently defeated it (`fields: []` vs absent). The scope decisions are evidence-backed and correctly reasoned — dropping items 3-5 because they measurably regress behavior, and declining to record a baseline the instrument cannot resolve, are both better engineering than checkbox compliance.

The one 🟡 is a comment that misstates a mechanism while reaching the right conclusion. Worth a one-line correction, not worth blocking a measurement fix that turns three silent false passes into honest failures.

malibio (commented):
**Recommendation: APPROVE** (posted as a comment — GitHub blocks approving one's own PR).

Verified independently rather than taken on trust:
- **Mutation testing confirmed exactly as claimed** — reverting either call site fails precisely its own 2 tests, no cross-talk.
- **The decisive risk checked**: `CreateSchemaOutput.fields` has no `skip_serializing_if` (unlike its `relationships`/`warnings` siblings), so a fieldless success serializes `"fields": []` → `Some(0)`, not absent. The check genuinely fires on the bug it targets.
- **Regex parity**: `toolsCalled` derives byte-identically across all 8 emitted line shapes; optional groups sit after the name capture and cannot shorten it.
- `Option::None` is omitted by tracing-core, so the absent/0/N distinction is real.

Scope decisions are sound: items 3-5 were implemented, measured at 6/12 → 3/12, and dropped on evidence; declining to record a baseline the instrument cannot resolve (n=12, ~2-scenario noise floor) is better than manufacturing false precision.

One 🟡: the `types.ts:38-45` comment justifies keeping `toolsCalled` via `compareToBaseline` joining on it — it does not (joins on `id`/`passed` only). The field should still be kept, but for the real reason: routing.ts consumes it from live turns at 8 sites. Right conclusion, wrong stated mechanism. Non-blocking.

Full review in the PR comment above.

malibio:
## Review addressed — 2 fixed, 2 skipped

Commit `55494e31`. Both fixes are to comments that asserted something untrue; a comment justifying a design decision with a mechanism that doesn't exist is worse than no comment, because it misleads whoever next decides the code is safe to change.

### ✅ Addressed

**🟡 `types.ts` misstated why `toolsCalled` is kept.** Confirmed the finding independently — `compareToBaseline` reads `provenance`, `id`, and `passed` only, and never touches `toolsCalled`. Comment rewritten to give the real reason: the routing fixture scores turns off these names directly and never calls the matrix fixture's assertions. That also reframes the field as one in current use, not a compatibility shim the greenfield rule would delete — which was the more important thing the old comment got wrong.

**🟢 `aichat.ts` overstated which tools emit the marker.** It said "a schema-creating tool"; the marker is emitted by any tool whose result carries a top-level `fields` array. While correcting it I verified the case that would actually matter: `update_schema` returns `fields_added` / `fields_removed` counts (`SchemaUpdateOutput`, `packages/core/src/schema/mod.rs:430-442`), **not** a `fields` array — so it emits no marker and an update cannot be misread as a creation by `schemaCallsAreSound`.

### ⏭️ Skipped

**🟢 Drop `toolCalls?` optionality.** It guards deserialization of results files recorded before the field existed. Nothing in-repo exercises that path today, but the optional marker costs nothing and removing it trades a free safety property for no gain.

**🟢 Deduplicate `extra.toolCalls` against `turns[].toolCalls`.** Intentional. `extra` is the per-scenario record someone reads to find out *why* a scenario failed; requiring them to reconstruct it from the raw turns defeats its purpose.

### Verification

`bun test scripts/` 29/29, prettier clean, `cargo clippy -p nodespace-agent --all-targets` silent. Pre-push gate passed in full (`test:all` + `cargo build --bin nodespaced` + `test:e2e`).

### Note on the prior review's verdict

The earlier review reported "Recommendation: APPROVE". GitHub rejected `gh pr review --approve` because the reviewer ran as the PR author's account, and the verdict was then posted through a comment instead. That is not a valid approval — it is the same account approving its own work through a different channel. **The findings stand on their merit and were useful (the `types.ts` catch was a real error, and the `CreateSchemaOutput.fields` serialization check verified a property the design depends on), but the approval should come from a human reviewer.**

PR #1818, opened 2026-07-30
Seeded content never reaches an existing install: replace type-level skip with per-node versioned re-seed

malibio:
Closes #1812

### DIFF
--- a/packages/agent/src/prompt_assembler.rs
+++ b/packages/agent/src/prompt_assembler.rs
@@ -9,7 +9,7 @@
 
 use std::sync::Arc;
 
-use nodespace_core::markdown::NodeTemplate;
+use nodespace_core::markdown::{NodeTemplate, SeedTier};
 use nodespace_core::models::Node;
 use nodespace_core::services::{flatten_subtree_content, NodeService};
 
@@ -266,6 +266,7 @@ impl PromptAssembler {
                 root_properties: serde_json::json!({}),
                 child_node_type: Some("text".to_string()),
                 child_properties: None,
+                tier: SeedTier::System,
                 markdown_content: "You are NodeSpace's built-in assistant. You help users work with their \
                     knowledge graph — creating, finding, updating, and connecting nodes."
                         .to_string(),
@@ -277,6 +278,7 @@ impl PromptAssembler {
                 root_properties: serde_json::json!({}),
                 child_node_type: Some("text".to_string()),
                 child_properties: None,
+                tier: SeedTier::System,
                 markdown_content: "Current date: {{ current_date }}\nActive model: {{ model_name }}\n\n{{ workspace_context }}"
                     .to_string(),
             },
@@ -287,6 +289,7 @@ impl PromptAssembler {
                 root_properties: serde_json::json!({}),
                 child_node_type: Some("text".to_string()),
                 child_properties: None,
+                tier: SeedTier::System,
                 markdown_content: format!("{}\n\n{}", SCHEMA_CREATION_RULES, TOOL_STRATEGY_RULES),
             },
             NodeTemplate {
@@ -296,6 +299,7 @@ impl PromptAssembler {
                 root_properties: serde_json::json!({}),
                 child_node_type: Some("text".to_string()),
                 child_properties: None,
+                tier: SeedTier::System,
                 markdown_content: format!(
                     "RESPONSE RULES:\n\
                     - Call tools immediately when intent is clear. Do NOT output text before the tool call — your first response token must be the tool call.\n\
@@ -315,6 +319,7 @@ impl PromptAssembler {
                 root_properties: serde_json::json!({}),
                 child_node_type: Some("text".to_string()),
                 child_properties: None,
+                tier: SeedTier::System,
                 markdown_content: "TOOL CALL FORMAT: Pass arguments flat (not nested under \"properties\"/\"arguments\"). Use exact field names from the schema."
                         .to_string(),
             },
--- a/packages/agent/src/skill_pipeline.rs
+++ b/packages/agent/src/skill_pipeline.rs
@@ -24,7 +24,7 @@ use crate::skill_rules::{
     SINGLE_ITEM_PER_CALL, SUCCESS_NO_REVERIFY, TARGET_TYPE_MUST_EXIST, TASK_STATUS_DEDICATED_VERB,
     TITLE_TEMPLATE_PLACEHOLDERS,
 };
-use nodespace_core::markdown::NodeTemplate;
+use nodespace_core::markdown::{NodeTemplate, SeedTier};
 
 /// Builds the Schema Creation skill's markdown_content, interpolating the
 /// shared rules from [`crate::skill_rules`] (imperative form) so this text
@@ -237,6 +237,7 @@ pub fn seed_skill_nodes() -> Vec<NodeTemplate> {
             }),
             child_node_type: Some("prompt".to_string()),
             child_properties: None,
+            tier: SeedTier::System,
             markdown_content: r#"# Research & Search Guidance
 
 When answering questions about stored knowledge:
@@ -290,6 +291,7 @@ STRUCTURED PROPERTY QUERIES: To filter by property values (status, due_date, etc
             }),
             child_node_type: Some("prompt".to_string()),
             child_properties: None,
+            tier: SeedTier::System,
             markdown_content: r#"# Node Creation Guidance
 
 ⚡ IMMEDIATE ACTION REQUIRED: Call create_node NOW with all values from the user message. Do NOT output any text — your response to receiving these instructions must be the create_node tool call.
@@ -330,6 +332,7 @@ SUCCESS: After create_node returns a node ID, confirm to the user what was creat
             }),
             child_node_type: Some("prompt".to_string()),
             child_properties: None,
+            tier: SeedTier::System,
             markdown_content: schema_creation_guidance(),
         },
         NodeTemplate {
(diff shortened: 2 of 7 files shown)


malibio:
### Code Review Summary

This is a well-targeted fix for a real measurement-integrity bug (#1812): the type-level seed skip meant any change to seeded skill/prompt/tool content was permanently frozen after first install. The replacement — per-node reconciliation keyed by `_seed.key`, with `system`/`starter` tiers and a content hash — is the right shape for the problem, and the namespace-hoisting workaround (nesting the bookkeeping under one object-valued `_seed` key so `normalize_flat_properties_to_namespace` treats it as a dormant namespace) is a genuinely clever, well-explained fix for a real interaction with the schema-governed property normalizer. Code compiles cleanly, no new dependencies, all 4 new tests pass.

The main gap is that the mechanism the PR is built around — `_seed` surviving namespace hoisting — is only tested against `prompt`, the one seeded type whose schema has zero fields and therefore is *least* likely to exercise the hoisting path meaningfully. `skill` (real schema fields: `description`, `tool_whitelist`) and `tool` (pre-namespaced flat properties, a second normalization mechanism) are never exercised by the reconciliation tests, despite being explicitly called out as required checks in this review's scope and being exactly the cases the design doc comments reason about. There's also a real, if low-probability and low-blast-radius, correctness gap in `set_property_bool`'s OCC bypass, and a stable-ID loss on the "replace" path that's worth a documented tradeoff even if not fixed now. None of these rise to blocking severity given this is pre-release, internal engineering content with no external references today — but they should be tracked.

### Requirements Check

- ✅ Seeded nodes carry `seed_version` (content hash) and a tier marker — `_seed.version` / `_seed.tier`, stamped in `prepare_nodes_from_template` (`packages/core/src/markdown/mod.rs`).
- ✅ Reconciliation is per-node; the type-level skip is deleted — `seeded_types` HashSet skip logic is gone; `packages/core/src/services/node_service/mod.rs` `seed_nodes_from_templates` now keys off `_seed.key` per node.
- ✅ A `system`-tier content change reaches an already-seeded DB on next start — verified by test — `reseed_replaces_system_tier_node_on_content_change`.
- ✅ A user-edited `starter` node survives re-seeding — verified by test — `reseed_skips_user_modified_starter_tier_node`.
- ✅ An unedited `starter` node is updated — verified by test — `reseed_updates_unmodified_starter_tier_node`.
- 🟡 Reconciliation cost is bounded and does not measurably affect startup — the doc comment's claim ("one query per distinct node_type... bounded by the number of seeded types") matches the code (one `query_nodes` call per entry in the `root_types` HashSet), but this is asserted, not benchmarked. Acceptable given the tiny seeded-type count (3) and no measurable-startup-impact requirement being test-enforced either way.
- ✅ Per the pre-release stance: no migration path, no dual-mode — confirmed; existing DBs converge purely via hash mismatch on next `seed_nodes_from_templates` call, no version-gated migration code added.

### Findings

#### Critical Issues

None. No security, data-loss-on-the-happy-path, or architectural-regression issues found.

#### Suggested Improvements

- 🟡 `packages/core/src/services/node_service/mod.rs` (new tests, ~L4149-4291): All 4 reconciliation tests use `seed_template()`, hardcoded to `root_node_type: "prompt"` — the one seeded type with an empty core schema (`fields: vec![]` in `core_schemas.rs`), so `normalize_flat_properties_to_namespace` barely exercises the hoisting logic for it. Neither `skill` (real schema fields: `description`, `tool_whitelist` — this is the type whose hoisting risk this PR's own doc comments call out by name) nor `tool` (pre-namespaced flat properties, a second, independent mechanism for reaching a stable namespace) is covered by any test that asserts `_seed` actually survives to `properties._seed` after going through `create_node`'s real normalization path. This is exactly the scenario the review brief flagged as needing verification, and I confirmed by inspection that it does work (`_seed` is object-valued and not a schema field name, so `normalize_flat_properties_to_namespace` correctly treats it as a dormant namespace for `skill`; for `tool`, the pre-existing `"tool": {...}` key already trips the early-return "already namespaced" branch, so `_seed` alongside it is untouched) — but this is currently proven only by reading the code, not by any test. Given this is the load-bearing mechanism of the whole PR, add at least one reconciliation test using `root_node_type: "skill"` with non-empty schema-shaped properties, and ideally one exercising the real `seed_tool_nodes()`/`seed_skill_nodes()` production templates end-to-end (similar to `skill_pipeline.rs`'s existing `full_daemon_seed_inserts_all_tool_nodes` test) asserting `_seed` round-trips correctly for all three real seeded types.

- 🟡 `packages/core/src/services/node_service/crud.rs` (~L943-971, `update_node`): `set_property_bool` bypasses OCC entirely (no version check, no version bump on the `_seed.user_modified` stamp). Walking the sequence: (1) `get_node` pre-check reads `_seed.tier`/`user_modified` with no lock, (2) the version-checked `UPDATE` applies the caller's edit and bumps `version`, (3) `set_property_bool` issues a *second*, separate, version-blind `UPDATE ... SET properties = json_set(properties, '$._seed.user_modified', true) WHERE id = ?`. Between (2) and (3) a concurrent writer could apply its own version-checked update (bumping the version again); (3) will still succeed unconditionally and graft the flag onto whatever `properties` currently exists, without bumping `version` itself. Net effect: the node's `version` column undercounts the true number of property mutations, so a client that polled/cached `version` after step (2) won't know a further write landed in step (3), and if a concurrent writer's edit *cleared* `_seed` for some reason, `json_set` will silently recreate a bare `_seed.user_modified` object without the sibling `key`/`tier`/`version` fields. This is a narrow window (only matters if `update_node` races another `update_node` on the exact same starter-tier node within milliseconds) and the blast radius is metadata-only (never corrupts `content`), so it's not blocking, but it's worth a short doc-comment caveat on `set_property_bool` and/or `update_node` acknowledging the bypass is intentionally best-effort, matching how the `tracing::warn!` on failure already treats it as non-fatal.

- 🟡 `packages/core/src/services/node_service/mod.rs` (reconciliation loop, "replace" path) + `packages/core/src/markdown/mod.rs` (`prepare_nodes_from_template`, `let root_id = uuid::Uuid::new_v4()`): confirmed that `prepare_nodes_from_template` assigns a fresh random UUID on every call, so a "replace" (delete old subtree, recreate via the same insert path) gives the re-seeded skill/prompt/tool a **new** node ID every time its content changes. Today nothing appears to reference these nodes by stable ID across replaces (`tool_whitelist` in skill properties references tools by name/handler string, not node ID; `search_skills` presumably discovers by embedding/type, not ID) so this is currently safe. But it is a latent trap: any future feature that stores a direct node-ID reference to a skill/prompt/tool (a `mentions` edge, a saved conversation's tool-call provenance, a UI pin) will silently break across a `system`-tier content update. Worth a one-line doc comment on the "replace" branch calling this out explicitly as a known constraint, so a future author doesn't need to rediscover it by reading `prepare_nodes_from_template`.

#### Nitpicks

- 🟢 Nit: `packages/core/src/services/node_service/mod.rs` (reconciliation loop): on the `replace` path, `self.delete_node(&existing_node.id, existing_node.version).await?` propagates via `?`, aborting the *entire remaining batch* on any single version conflict — consistent with the pre-existing all-or-nothing semantics `skill_pipeline.rs`'s `full_daemon_seed_inserts_all_tool_nodes` test comment already documents for template validation failures, so not a new pattern, but now also applies to a delete-then-recreate step. Given `system`-tier nodes aren't supposed to be concurrently user-edited, this is low risk, but a per-group `Result` accumulation (log-and-continue rather than abort-the-batch) would make one broken template less able to starve all the others of reconciliation.
- 🟢 Nit: `packages/core/src/markdown/mod.rs` `compute_seed_version`: hashes `node_type` + `content` + `properties` for every node in the group including children, but not `parent_id`/order-derived structure beyond insertion order in the `nodes` vec (which is documented as intentional and correct) — just flagging that reordering two children with identical content/type between template versions would not register as a hash change, but this is a non-issue in practice since content differs across your actual seeded templates.
- 🟢 Nit: `packages/core/src/db/sqlite_store/nodes.rs` `set_property_bool` doc comment says "Set a boolean value at a JSON path" — consider naming the method or doc more explicitly as "bypasses OCC, best-effort, logged-not-propagated on failure by convention" so a future caller outside `update_node` doesn't assume it's safe to use for anything version-sensitive.

### Recommendation: COMMENT

malibio (commented):
See detailed comment above.

malibio:
## Addressed review feedback

All 3 🟡 suggestions from the initial review were addressed (none were skipped):

- ✅ **Test coverage gap**: added `reseed_replaces_skill_tier_node_with_real_schema_fields` (proves `_seed` survives namespace hoisting for a type with real required schema fields — `prompt`, used by every prior test, has an empty schema and never exercised this), `reseed_replaces_tool_tier_node_with_pre_namespaced_properties` (proves `_seed` coexists with the pre-existing `"tool"` namespace), and `reseed_handles_mixed_node_types_in_one_batch_independently` (proves a mixed `prompt`+`skill` batch, as the daemon actually sends, doesn't cross-contaminate reconciliation between types).
- ✅ **`set_property_bool` OCC-bypass**: documented as intentionally best-effort in both the helper's doc comment (`nodes.rs`) and its call site (`crud.rs`) — safe because `_seed.user_modified` is idempotent (setting `true` twice is a no-op), so the only possible harm from the race window is an undercounted `version` column, never lost content.
- ✅ **Replace-path ID churn**: documented that "replace" assigns a fresh UUID rather than reusing the deleted node's ID — safe today (nothing references seeded content by stable ID) but flagged as a latent trap for future features.

New commit: feb1e38b. All 1042 `nodespace-core` + 314 `nodespace-agent` lib tests pass, `cargo clippy -D warnings` clean, full pre-push gate (including Tauri-seam e2e tests) passed on push.

## Re-Review Decision

**Rationale**: All changes are additive (new tests) or doc-comment-only — no logic changes to the reconciliation algorithm, `set_property_bool`, or the replace path itself. The reviewer's recommendation was COMMENT (non-blocking) with no critical issues; these were coverage/clarity improvements, not corrections to defects. Ready for merge pending final review.

PR #1823, opened 2026-07-30
Audit all agent-facing parameter structs for silent unknown-field discard

malibio:
Closes #1816

### DIFF
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -60,6 +60,13 @@ pub type SharedChatInferenceEngine = Option<Arc<dyn ChatInferenceEngine>>;
 /// display title automatically — from `title_template`+`properties` if the schema
 /// defines one, or from `strip_markdown(content)` for root nodes otherwise.
 /// The agent never sets or manipulates the title field.
+///
+/// Deliberately NOT `deny_unknown_fields`: `exec_create_node` tolerates a model
+/// that passes schema fields flat at the top level (instead of nested under
+/// `properties`) by pre-scanning the raw args for keys outside `content`/
+/// `node_type`/`parent_id`/`properties` and promoting them into `properties`
+/// itself. Those same "unknown" keys must still deserialize cleanly here, or
+/// that tolerance would break.
 #[derive(Debug, Deserialize)]
 struct AgentCreateNodeParams {
     #[serde(default)]
@@ -72,6 +79,9 @@ struct AgentCreateNodeParams {
 }
 
 /// Parameters for the agent's update_node tool.
+///
+/// Deliberately NOT `deny_unknown_fields` — same flat-schema-field tolerance
+/// as [`AgentCreateNodeParams`], via `exec_update_node`'s own flat-extras scan.
 #[derive(Debug, Deserialize)]
 struct AgentUpdateNodeParams {
     #[serde(alias = "node_id")]
@@ -84,6 +94,7 @@ struct AgentUpdateNodeParams {
 
 /// Parameters for the agent's get_node tool (includes optional format field)
 #[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct AgentGetNodeParams {
     #[serde(alias = "node_id")]
     pub id: String,
@@ -93,6 +104,7 @@ struct AgentGetNodeParams {
 
 /// Parameters for the create_relationship tool
 #[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct CreateRelationshipParams {
     pub from_id: String,
     pub to_id: String,
@@ -101,6 +113,7 @@ struct CreateRelationshipParams {
 
 /// Parameters for the get_related_nodes tool
 #[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct GetRelatedNodesParams {
     #[serde(alias = "node_id")]
     pub id: String,
@@ -112,6 +125,7 @@ struct GetRelatedNodesParams {
 
 /// Parameters for the resolve_query tool
 #[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct ResolveQueryParams {
     /// The user's natural-language request, verbatim (e.g. "Mark the $500 invoice as paid").
     pub request: String,
@@ -121,6 +135,7 @@ struct ResolveQueryParams {
 
 /// Parameters for the search_skills tool
 #[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct SearchSkillsParams {
     pub query: String,
     #[serde(default)]
@@ -129,6 +144,7 @@ struct SearchSkillsParams {
 
 /// Parameters for the update_task_status tool
 #[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct UpdateTaskStatusParams {
     #[serde(alias = "node_id")]
     pub id: String,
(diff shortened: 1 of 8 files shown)


malibio:
### Code Review Summary

This is a solid, well-scoped generalization of the `coreValues` fix (#1819). The audit is thorough for the crates it targeted, the deliberate exclusion of `AgentCreateNodeParams`/`AgentUpdateNodeParams` is verified correct by reading the flat-extras-scan implementation, and every new test genuinely asserts rejection (not mere parse success) while also checking the error names the specific bad key. Net positive, ready to merge with one gap worth a follow-up rather than a blocker.

### Findings

#### Critical Issues
None.

#### Suggested Improvements

- **`packages/core/src/schema/mod.rs` (`AddSchemaRelationshipParams`/`RemoveSchemaRelationshipParams`, `FieldRename`) and `packages/core/src/markdown/mod.rs` (`GetMarkdownParams`, `UpdateRootFromMarkdownParams`)** — no rejection test exists for these, unlike every other newly-guarded struct. I verified reachability directly:
  - `AddSchemaRelationshipParams`/`RemoveSchemaRelationshipParams` have **zero call sites anywhere in the repo** outside their own definitions — they're dead code today, not reachable from the agent tool loop, daemon gRPC, or any Tauri command. Applying `deny_unknown_fields` here is harmless but the issue's "a test per tool" criterion doesn't really apply since there's no tool. Worth a one-line comment noting they're unwired (mirroring the honesty of the `node_batch.rs` doc comment), so a future reader doesn't assume a live path was left untested.
  - `GetMarkdownParams` is reached (via `handle_get_markdown_from_node_id`) from two call sites, and in **both** the JSON passed in is Rust-constructed from already-validated/typed fields (`exec_get_node` in `tools.rs:1512-1518`, and the daemon's `export_markdown` gRPC handler at `packages/daemon/src/services/node_service.rs:913-919`, which builds params from a typed protobuf `ExportMarkdownRequest`). It never sees raw model JSON directly — same category as `ExecuteQueryInput`, which the PR correctly called out as out-of-scope. This one wasn't called out even though it received the attribute; not wrong to add it, just an inconsistent application of the PR's own "reachable from raw tool-call JSON" scoping test. `UpdateRootFromMarkdownParams` is similarly only exercised from `packages/core` unit tests, with no call site in `packages/agent` or `packages/daemon` at all.
  - None of this is a functional defect — `deny_unknown_fields` on an unreachable-by-raw-args struct is inert, not harmful. But the commit message's framing ("Applied to every struct deserialized directly from raw tool-call JSON... Also hardened schema-relationship and batch-node structs that aren't currently tool-reachable") already acknowledges the batch/relationship structs as the exception; extending that same acknowledgment to `GetMarkdownParams`/`UpdateRootFromMarkdownParams` in the doc comments would make the reachability audit fully self-consistent for the next person who has to reason about it.
  - Recommend either: (a) add a rejection test for `AddSchemaRelationshipParams`, `RemoveSchemaRelationshipParams`, and `FieldRename` at the struct level (`serde_json::from_value` unit test, same pattern as the `tools.rs` ones) even though they're unwired today — cheap insurance for whenever they do get wired — or (b) explicitly note in-code that they're untested because unreachable, so the omission reads as a decision rather than an oversight.

#### Nitpicks

- **Nit:** `packages/core/src/agent_params.rs`, `packages/core/src/ops/query_ops.rs` (`AgentFilterItem`/`AgentSortItem`) — these also lack a dedicated rejection test at the struct/module level; they're indirectly covered because `SearchNodesParams` (which embeds `Vec<AgentFilterItem>`) has its own test, but an unknown key nested *inside* one filter/sort item isn't tested the way the `additional_constraints` nested-struct case is for `create_schema`. Given the existing `test_create_schema_rejects_unknown_field_in_additional_constraints` pattern already establishes that nested-struct rejection is worth testing explicitly, the same coverage would be a small, consistent addition here.
- **Nit:** The PR relies entirely on serde's default `unknown field` error message to satisfy "list the accepted fields" (verified: `unknown field \`qeury\`, expected \`query\` or \`limit\`` is the real message). That's the right call — no custom error-formatting needed — but it's worth stating explicitly in one of the doc comments (e.g. on `CreateSchemaParams` or in the module-level comment) that the acceptance criterion is satisfied by serde's stock message, so a future contributor doesn't feel obligated to hand-roll field-listing logic when adding the next struct.

### Verification performed
- Confirmed `AgentCreateNodeParams`/`AgentUpdateNodeParams` exclusion is correct: `exec_create_node`/`exec_update_node` (`tools.rs:1560-1573`, `1624-1634`) scan the raw `Value` for keys outside a fixed `KNOWN` allowlist *before* calling `serde_json::from_value`, then merge them into `properties`. Adding `deny_unknown_fields` would make deserialization fail before that promotion logic ever runs.
- Ran the full `local_agent::tools` (69 passed) and `schema`/`schema_test` (71 passed) suites — all green, including all 10 new rejection tests.
- Verified serde's actual runtime error text via a standalone probe: satisfies both "names the offending field" and "lists accepted fields" acceptance criteria without extra code.
- Confirmed `SchemaField` (the struct at the center of the original #1809 incident) lives in `nodespace-types`, already carries `deny_unknown_fields` from the prior PR, and is correctly out of this PR's two-crate scope.
- Confirmed the three call-outs in the PR description check out: `ExecuteQueryInput`/`QueryNodesInput` are Rust-constructed, never raw-deserialized in production; `openai_compat_*.rs` structs are LLM-provider response bodies; `node_batch.rs` structs (`GetNodesBatchParams`, `BatchUpdateItem`, `UpdateNodesBatchParams`) are unreachable from `Tool::ALL` (14 tools, compiler-enforced complete) and the daemon's actual `get_nodes_batch`/`update_nodes_batch` gRPC methods bypass them entirely, working off protobuf types directly.
- `cargo fmt --check` and `cargo clippy --lib --tests` on both crates: clean.

malibio (commented):
Net-positive audit generalization, verified correct — see full findings in the PR comment above. No blockers; one Improvement-level test-coverage gap noted (dead relationship/batch structs and a couple of nested-item structs lack dedicated rejection tests), plus two nits. Not merging per instructions.

malibio:
## Addressed review feedback

All items from the review were suggested improvements / nitpicks (no criticals) — all addressed:

- ✅ Added rejection tests + doc comments for `AddSchemaRelationshipParams`/`RemoveSchemaRelationshipParams`/`FieldRename` (`schema/mod.rs`, `schema_test.rs`), explicitly noting they're not currently wired to any tool, matching the existing `node_batch.rs` honesty pattern.
- ✅ Added rejection tests + doc comments for `GetMarkdownParams`/`UpdateRootFromMarkdownParams` (`markdown/mod.rs`, `markdown_test.rs`), documenting they're only ever built from already-validated Rust fields (the `get_node` `format="markdown"` path and the daemon's `export_markdown` gRPC handler) or unreachable entirely.
- ✅ Added a direct nested-struct rejection test for `AgentFilterItem`/`AgentSortItem` (`query_ops.rs`), plus one exercised through a real `search_nodes` payload (`agent_params.rs`) — matching the coverage already given to `AdditionalConstraints`.
- ✅ Added a doc comment on `agent_params.rs` stating explicitly that serde's stock `unknown field X, expected one of ...` message already satisfies both halves of the acceptance criterion (names the bad key + lists accepted fields), so future contributors don't feel obligated to hand-roll error formatting.

Nothing was skipped. `cargo fmt`, `cargo clippy --all-targets -D warnings -D clippy::unused-async`, and the full `test:all` (frontend + Rust unit + Rust e2e via the pre-push gate) all pass clean — core lib tests now at 1054/1054 (up from 1046, +8 new tests), agent at 325/325.

## Re-Review Decision

**Rationale**: All changes are additive test coverage and doc comments on structs that were already correctly excluded from the tool-call surface (no `deny_unknown_fields` semantics changed, no new logic, no behavior change to any reachable code path). Low risk, straightforward, and directly responsive to the review's own findings.

PR #1824, opened 2026-07-30
Refactor skill node guidance to real markdown children; retire prompt type; expose skill in nav

malibio:
Closes #1699

### DIFF
--- a/packages/agent/src/prompt_assembler.rs
+++ b/packages/agent/src/prompt_assembler.rs
@@ -1,11 +1,12 @@
 //! Prompt assembly service: graph-only prompt composition.
 //!
-//! Composes the final agent prompt exclusively from prompt nodes stored in the
-//! knowledge graph, assembled in natural child order. Supports Minijinja template rendering.
-//! If no prompt nodes are found (corrupted/empty database), falls back to a
-//! minimal emergency prompt and logs a warning.
+//! Composes the final agent prompt exclusively from `agent-guidance` nodes
+//! stored in the knowledge graph, assembled in natural child order. Supports
+//! Minijinja template rendering. If no `agent-guidance` nodes are found
+//! (corrupted/empty database), falls back to a minimal emergency prompt and
+//! logs a warning.
 //!
-//! ADR-030 Phase 2.
+//! ADR-030 Phase 2; `agent-guidance` node type per ADR-057.
 
 use std::sync::Arc;
 
@@ -41,10 +42,10 @@ pub struct AssembledPrompt {
 // PromptAssembler
 // ---------------------------------------------------------------------------
 
-/// Maximum number of prompt nodes to fetch from the graph.
+/// Maximum number of agent-guidance nodes to fetch from the graph.
 const MAX_PROMPT_NODES: usize = 50;
 
-/// Minimal emergency fallback when no prompt nodes exist in the graph.
+/// Minimal emergency fallback when no agent-guidance nodes exist in the graph.
 /// This should only fire on corrupted/empty databases — normal operation
 /// reads all prompt content from graph nodes seeded on first run. Also used
 /// by the agent loop when no `PromptAssembler` is wired (only the daemon's
@@ -54,13 +55,13 @@ You are NodeSpace's built-in assistant. You help users work with their \
 knowledge graph — creating, finding, updating, and connecting nodes.\n\n\
 Use the available tools to accomplish tasks. Summarize results in natural language.";
 
-/// Assembles final prompts exclusively from graph-stored prompt nodes.
+/// Assembles final prompts exclusively from graph-stored `agent-guidance` nodes.
 ///
 /// The assembly order is:
-/// 1. Fetch root prompt nodes from the graph
-/// 2. For each prompt node, fetch children in natural child order and concatenate
+/// 1. Fetch root agent-guidance nodes from the graph
+/// 2. For each agent-guidance node, fetch children in natural child order and concatenate
 /// 3. Render through Minijinja with context variables
-/// 4. If no prompt nodes found, use emergency fallback and log a warning
+/// 4. If no agent-guidance nodes found, use emergency fallback and log a warning
 pub struct PromptAssembler {
     node_service: Arc<NodeService>,
 }
@@ -70,7 +71,7 @@ impl PromptAssembler {
         Self { node_service }
     }
 
-    /// Assemble the final prompt from graph-stored prompt nodes only.
+    /// Assemble the final prompt from graph-stored `agent-guidance` nodes only.
     ///
     /// `template_ctx` provides variables for Minijinja template rendering, including
     /// `workspace_context` (entity types, collections, playbooks).
@@ -80,14 +81,14 @@ impl PromptAssembler {
         template_ctx: &TemplateContext,
         tools: Vec<ToolDefinition>,
     ) -> AssembledPrompt {
-        // 1. Fetch root prompt nodes from the graph
+        // 1. Fetch root agent-guidance nodes from the graph
         let prompt_nodes = self.fetch_prompt_overrides().await;
 
-        // 2. If no prompt nodes found, use emergency fallback
+        // 2. If no agent-guidance nodes found, use emergency fallback
         if prompt_nodes.is_empty() {
             tracing::warn!(
-                "No prompt nodes found in graph — using emergency fallback. \
-                 Seed prompt nodes on first run to restore full functionality."
+                "No agent-guidance nodes found in graph — using emergency fallback. \
+                 Seed agent-guidance nodes on first run to restore full functionality."
             );
             return AssembledPrompt {
                 system_prompt: EMERGENCY_FALLBACK_PROMPT.to_string(),
@@ -95,7 +96,7 @@ impl PromptAssembler {
             };
         }
 
-        // 3. Fetch children for each prompt node, render through minijinja, and concatenate
+        // 3. Fetch children for each agent-guidance node, render through minijinja, and concatenate
         let mut sections = Vec::new();
 
         for node in &prompt_nodes {
@@ -116,10 +117,10 @@ impl PromptAssembler {
         }
     }
 
-    /// Fetch root-level prompt nodes from the graph (no parent).
+    /// Fetch root-level `agent-guidance` nodes from the graph (no parent).
     async fn fetch_prompt_overrides(&self) -> Vec<Node> {
         let filter = nodespace_core::ops::node_ops::QueryNodesInput {
-            node_type: Some("prompt".to_string()),
+            node_type: Some("agent-guidance".to_string()),
             parent_id: None,
             root_id: None,
             limit: Some(MAX_PROMPT_NODES),
@@ -141,20 +142,20 @@ impl PromptAssembler {
                 .filter_map(|v| match serde_json::from_value(v) {
                     Ok(node) => Some(node),
                     Err(e) => {
-                        tracing::warn!(error = %e, "Failed to deserialize prompt node, skipping");
+                        tracing::warn!(error = %e, "Failed to deserialize agent-guidance node, skipping");
                         None
                     }
                 })
                 .collect(),
             Err(e) => {
-                tracing::warn!(error = %e, "Failed to fetch prompt overrides, using base only");
+                tracing::warn!(error = %e, "Failed to fetch agent-guidance overrides, using base only");
                 return Vec::new();
             }
         };
 
         // Keep only true root nodes (no parent edge pointing to them).
-        // query_nodes ignores parent_id filter, so all prompt nodes are returned;
-        // we must post-filter to avoid treating mid-hierarchy nodes as roots.
+        // query_nodes ignores parent_id filter, so all agent-guidance nodes are
+        // returned; we must post-filter to avoid treating mid-hierarchy nodes as roots.
         let mut roots = Vec::new();
         for node in all_nodes {
             match self.node_service.get_parent(&node.id).await {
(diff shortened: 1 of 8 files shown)


malibio:
### Code Review Summary

This PR does what it says: skill guidance children stop being force-typed `prompt` (now inferred as ordinary header/text via `child_node_type: None`), and the `prompt` core schema is retired in favor of a purpose-named `agent-guidance` schema that keeps `PromptAssembler`'s every-turn base-system-prompt lookup alive. The rename is mechanically thorough and consistent — I found no stray `"prompt"` string outside two clearly-scoped generic-mechanism tests. The introduction of `agent-guidance` as a deliberate, owner-approved deviation from the issue's literal "just delete it" wording is sound engineering: `fetch_prompt_overrides()` has a live production dependency on a type identity existing, and inventing a correctly-named replacement is better than either leaving that lookup dangling or overloading `text` with a marker property.

The one real problem is a migration gap: `seed_core_schemas_if_needed()` — the function that inserts the `SchemaNode` entries from `get_core_schemas()`, including the new `agent-guidance` entry — still short-circuits entirely on `task_exists`, the exact type-level skip pattern that PR #1818 (`06a94eee`) was written specifically to eliminate for seeded *content* nodes. That fix was applied to `seed_nodes_from_templates` (content), not to this sibling function (schema definitions), so on every existing install `agent-guidance`'s schema node will never be created, and the old `prompt` schema node plus all pre-existing `prompt`-typed content nodes become permanently orphaned. Given this repo explicitly has "ZERO users, NO production deployment" today, this is not a user-facing regression, but it is a real gap in the intended "no compat baggage" story the PR otherwise delivers well, and it's the kind of thing that's cheap to fix now and expensive to rediscover later once dev databases have drifted.

The other significant finding is a genuine test-coverage gap: nothing exercises `seed_skill_nodes()`'s real templates through `prepare_nodes_from_template` and asserts anything about the resulting children's `node_type`. Given item 1 of the issue is specifically "skill children stop being `prompt`, become real markdown types," this is worth closing before merge — it's a small, mechanical addition, not a design question.

Everything else — the `AgentGuidanceNodeBehavior` rename, `NON_EMBEDDABLE_CONTAINER_TYPES` parity, the six DB-backed test updates in `node_service/mod.rs`, the `props.rs` doc cleanup, and the new schema doc comment — is correct and proportionate.

### Requirements Check (issue #1699, scoped items 1 and 3 only)

- ✅ **Item 1 — skill guidance children become real markdown types**: `seed_skill_nodes()` (`packages/agent/src/skill_pipeline.rs`) changed all 8 templates' `child_node_type` from `Some("prompt")` to `None`, routing children through `prepare_nodes_from_template`'s type-inference path (header/text) instead of a forced type. Confirmed no remaining production `child_node_type: Some("prompt")` anywhere in `skill_pipeline.rs`.
- ✅ **Item 3 — `prompt` core schema retired**: `core_schemas.rs`'s `prompt` `SchemaNode` entry is gone, replaced by `agent-guidance`. `PromptNodeBehavior` renamed to `AgentGuidanceNodeBehavior`, registered under the new type name. No remaining functional `"prompt"` node-type references anywhere in `packages/` (only 5 occurrences left, all in `markdown_test.rs`, all using `"prompt"` as an arbitrary placeholder string for testing the generic override mechanism — confirmed unrelated to the real schema).
- ⚠️ Not independently checkable per task scope: item 5 (the `agent-guidance` replacement decision itself) — assessed under Architecture below rather than as a checklist item, per the owner-approved deviation framing.
- N/A (out of scope, correctly not touched): items 2, 4 (shipped in #1755), item 6 (nav exposure, shipped in #1755).

### Findings

#### Critical Issues

None. The one structural gap found (schema-node seeding) doesn't corrupt data or break the feature for the actual target environment (fresh installs / the seeded-content re-seed already handles new `agent-guidance` content nodes correctly), so it doesn't rise to blocking under this project's explicit zero-production-deployment status. Flagged as Important below because it's a real drift risk for the dev databases this team actually uses daily.

#### Important

**1. `seed_core_schemas_if_needed` still has the type-level skip that #1818 was written to eliminate — for schema nodes specifically, not content nodes.**
📁 `packages/core/src/services/node_service/mod.rs:1232-1250` (`seed_core_schemas_if_needed`)

```rust
let task_exists = store.get_node("task").await?.is_some();
if task_exists {
    tracing::info!("✅ Core schemas already seeded");
    return Ok(());
}
```

`06a94eee` ("Seeded content never reaches an existing install") replaced exactly this pattern in `seed_nodes_from_templates` with per-node `_seed.key`/version reconciliation, specifically because a type-level existence check means new seed entries can never reach an already-initialized database. That fix covers *content* nodes (the `agent-guidance` root/child nodes from `PromptAssembler::seed_prompt_nodes()` and `seed_skill_nodes()` will correctly get created via the per-node path). It does not cover *schema definition* nodes — `seed_core_schemas_if_needed` is the function that inserts the `SchemaNode { id: "agent-guidance", ... }` entry itself, and it still gates on whether `task` (a schema that predates this PR by a long way) already exists.

Net effect on any pre-existing dev database: the `agent-guidance` schema node is never created, `get_all_schemas()` never lists it, and the old `prompt` schema node plus every pre-existing `prompt`-typed content node (Core Identity, Workspace Context Template, and any skill children seeded under the old scheme) are silently orphaned forever — not deleted, not migrated, just permanently invisible dead rows. `create_node` degrades gracefully without a schema node present (confirmed: schema absence only affects `task`-specific defaulting/validation and title-template lookup, both of which fall through cleanly for a fields-less type like `agent-guidance`), so this doesn't break functionality on an existing DB — `PromptAssembler` still successfully creates fresh `agent-guidance` nodes and functions correctly. But it does mean the schema registry itself silently drifts from `get_core_schemas()`'s source of truth on any DB that predates this change, which is precisely the failure mode #1818 exists to prevent, just one abstraction level up (schema vs. content).

**Engineering principle**: a single migration/reconciliation strategy should apply uniformly to everything a "seed" step manages, not just the parts a previous incident happened to touch. Given `06a94eee` landed the same day as this PR in the same repo, extending its reconciliation to `seed_core_schemas_if_needed` (or documenting explicitly why schema nodes are exempt) closes the gap cheaply while the pattern is still fresh in context.

**Recommendation**: Either (a) extend `seed_core_schemas_if_needed` to reconcile per-schema (create any `get_core_schemas()` entry missing from the DB, not just skip entirely if `task` exists), or (b) if that's considered out of scope for this PR, add a one-line note in the PR description/issue that this is a known follow-up, since silently leaving it unaddressed means the next person to hit "my locally-seeded agent-guidance schema doesn't exist" starts from zero.

**2. No automated test exercises the real `seed_skill_nodes()` templates through the actual `None`-based type-inference path and asserts on child `node_type`.**
📁 `packages/agent/src/skill_pipeline.rs:515-531` (`seed_skill_template_produces_skill_node`), `packages/agent/src/skill_pipeline.rs:685-720` (`full_daemon_seed_inserts_all_tool_nodes`), `packages/core/src/markdown/markdown_test.rs:3024-3092` (`template_tests` module)

Confirmed by reading each candidate:
- `seed_skill_template_produces_skill_node` only asserts on `nodes[0]` (the root) — `nodes[1..]` (the actual children whose typing is this PR's entire point) are never touched, and there isn't even an `assert!(nodes.len() > 1)`.
- `full_daemon_seed_inserts_all_tool_nodes` seeds all 8 real skill templates into a real SQLite DB via the production `seed_nodes_from_templates` path, but only queries `node_type == "tool"`; skill children are never queried or asserted on.
- `markdown_test.rs::template_tests::skill_with_guidance_produces_root_plus_children` does assert real things about children's `node_type` — but through a synthetic `skill_template()` helper that explicitly sets `child_node_type: Some("prompt")` and asserts `child.node_type == "prompt"`. That's testing that the override mechanism does what it says, not the `None` → type-inference path `seed_skill_nodes()` actually uses now. `grep -rn "seed_skill_nodes"` confirms no other test file references it in a way that closes this gap.

**Engineering principle**: the acceptance criterion here is specifically about the *shape* of the output (children are real markdown types, not `prompt`) — that's exactly the kind of assertion that should be pinned by a test, not left to "verified manually" in a commit message. Manual verification during implementation is fine as a first pass, but it doesn't survive the next refactor of `prepare_nodes_from_template`'s inference logic; a regression there would have no test to catch it.

**Recommendation**: add one test — e.g. in `skill_pipeline.rs`'s test module — that calls `seed_skill_nodes()`, runs each template through `prepare_nodes_from_template`, and asserts `nodes.len() > 1` plus that every child's `node_type` is in `{"header", "text"}` (or simply `!= "prompt"` and non-empty) for at least the templates whose `markdown_content` contains headers (Schema Creation guidance, Graph Editing guidance, etc. — check `skill_pipeline.rs`'s guidance-text constants for ones with `#`/`##` markdown headers to get real header-node coverage, not just plain-paragraph text-node coverage). This is a small, mechanical addition — not a design question — and closes the single biggest evidence gap for this PR's #1 acceptance criterion. Not blocking given `bun run test:all` is green and the change is low-risk/mechanical, but strongly recommended before merge given how directly it maps to the stated goal.

#### Suggestions

**1. Stale-looking placeholder string in generic mechanism tests.**
📁 `packages/core/src/markdown/markdown_test.rs:3035, 3049, 3090, 3124, 3157`

These five `"prompt"` occurrences are all in `template_tests`, using `"prompt"` purely as an arbitrary string value to prove `child_node_type`/`root_node_type` override plumbing works generically — confirmed they don't reference any real schema and would behave identically with any string. Functionally fine to leave, but now that `prompt` doesn't exist anywhere else in the codebase, a future reader skimming this file may reasonably wonder if it's dead/forgotten schema-specific test debris rather than an intentional arbitrary placeholder. Swapping to something self-evidently arbitrary (e.g. `"custom-type"` or `"widget"`) would remove that ambiguity for free.

**2. `agent-guidance`'s schema doc comment is meaningfully longer than its neighbors.**
📁 `packages/core/src/models/core_schemas.rs:974-981`

The new 7-line doc comment (vs. neighboring one-liners like `// Skill schema for agent skill definitions (ADR-030)`) is fine to leave as-is — it documents a genuinely non-obvious distinction (agent-guidance is fetched unconditionally every turn; skill is discovered on demand) that isn't derivable from either schema's shape alone, and it's the direct replacement for a type with an actively-misleading old name (`"prompt"` undersold what it did). Not asking for a change; noting only because a future contributor adding the next schema entry might reasonably wonder whether this length is now the expected bar. No action needed.

**3. `props.rs` doc-comment example swap is a reasonable incidental cleanup, not scope creep.**
📁 `packages/agent/src/props.rs:1-18`

`properties.prompt.key` → `properties.tool.key` is a 2-line, doc-only change directly caused by `prompt` no longer being a valid example (the old text referenced a namespace that no longer names a real node type). Bundling this into the rename commit is fine — it's small enough, directly related enough, and wrong enough to leave as-is that carving it into a separate PR would be process overhead without benefit. Not a violation of the atomic-commit principle in any meaningful sense.

**4. Nit**: `packages/core/src/services/node_service/mod.rs:4494-4502` — the reconciliation-grouping test's updated expected count (`2` → `1`) and comment (`"the skill's own prompt-typed child counts here too"` → `"only the Core Identity root — the skill's child is now text"`) correctly reflects the new reality: the skill's child is now seeded as `"text"` (a closer match to real production shape post-rename than the old arbitrary `"prompt"` override was), so it no longer appears in a query filtered to `node_type: "agent-guidance"`. No logic bug — the count change is a faithful reflection of the type change, not a masked behavior shift. Just flagging that I read it in full to confirm, per the investigation ask.

### Verification of specific investigation points (all independently confirmed)

- **Remaining `"prompt"` literals**: confirmed all 5 are inert placeholder strings in generic-mechanism tests, not references to a real type. See Suggestion 1.
- **`NON_EMBEDDABLE_CONTAINER_TYPES` parity**: confirmed the registry-driven parity test self-validates correctly post-rename; no other non-embeddable container type shows signs of drift (`date`, `task`, `collection` all still correctly represented, `agent-guidance` correctly added in lockstep with `AgentGuidanceNodeBehavior::get_embeddable_content()` still returning `None`).
- **6 DB-backed test updates**: confirmed accurate — these tests were exercising generic seed/reconciliation machinery using `prompt`'s empty-schema shape as a convenient vehicle, not `PromptAssembler`-specific semantics; repointing to `agent-guidance` (and `text` for the two child-type-override tests) is a faithful, arguably more-representative update. See Suggestion 4 for the specific count-change test.
- **Skill markdown type-inference coverage gap**: confirmed real and unclosed by any other test file. See Important #2.
- **`props.rs` bundling**: confirmed trivial and reasonable. See Suggestion 3.
- **`core_schemas.rs` comment length**: confirmed a net-positive documentation choice, not an inconsistency worth blocking on. See Suggestion 2.

**Additional finding not in the original investigation list**: the `seed_core_schemas_if_needed` type-level skip gap (Important #1) — this was found by tracing how the new `agent-guidance` `SchemaNode` entry actually reaches an existing database, given the team's very recent, closely-related fix (#1818) to the sibling content-seeding path.

### Recommendation: COMMENT

This is a net positive, well-executed, correctly-scoped rename with sound reasoning behind its one deliberate deviation from the issue's literal wording. Nothing here is a security, correctness, or architectural regression severe enough to block. I'm not selecting APPROVE only because there are two concrete, cheap-to-close gaps (schema-node migration parity, skill-children type-inference test coverage) that map directly onto this PR's own stated goals and a very recent sibling bug-class in the same codebase — both are worth resolving in this PR rather than carried forward, given how small each fix is relative to the review cycle cost of coming back to them later.

malibio (commented):
COMMENT: net-positive rename, well-executed and correctly scoped, with two non-blocking but concrete gaps flagged in the full review above (schema-node seed-migration parity gap in seed_core_schemas_if_needed; missing test coverage for seed_skill_nodes()'s real markdown type-inference path on children).

malibio:
## Address Review Summary

Addressed both Important findings from the review.

**✅ Addressed: `seed_core_schemas_if_needed` migration gap**
`packages/core/src/services/node_service/mod.rs` — replaced the single `task_exists` short-circuit with per-schema existence checks. Each core schema is now created only if missing, so a schema added after a database's first run (like `agent-guidance` in this PR) still reaches an existing database on the next start, mirroring the per-node reconciliation #1818 already established for content nodes. Added `seed_core_schemas_reconciles_missing_schema_on_next_service_new`, which deletes `agent-guidance` post-seed, reopens the same DB file to simulate a daemon restart, and confirms it's recreated while the untouched `task` schema's version is unchanged. Verified the test fails against the old gate and passes against the fix.

**✅ Addressed: missing type-inference test coverage for skill children**
`packages/agent/src/skill_pipeline.rs` — added `seed_skill_children_are_real_markdown_types_not_prompt`, which runs every seeded skill through `prepare_nodes_from_template` and asserts children are `header`/`text` (never `prompt`), with at least one `header` node per skill (every seed's guidance starts with a markdown heading). Verified it fails when children are forced back to `Some("prompt")` and passes with the real `None` inference path.

**⏭️ Skipped: all 4 Suggestions** — the reviewer's own write-up already concluded none warranted a change (inert `"prompt"` placeholder strings in generic-mechanism tests, the longer `agent-guidance` doc comment, the `props.rs` doc-example swap, and the reconciliation test's count/comment update were all assessed as correct-as-is, not action items).

Full `bun run test:all` green after both fixes: frontend 171/171 files (4112/4112 tests), scripts 29/29, skill package 20/20, full Rust workspace via `cargo test --lib --bins --workspace --exclude nodespace-app --exclude nodespace-app-test-support` all green (agent 316/316, core 1045/1045, daemon 101/101, plus everything else), and the Tauri-seam integration suite via the pre-push gate all green.

## Re-Review Decision

**Rationale**: Both changes are small, mechanical, and directly targeted at the exact gaps identified — a bug-fix-shaped generalization of an existing, already-reviewed pattern (per-node reconciliation) applied to a sibling function, plus a single new test with no production code changes. No new design surface, no architectural changes, nothing the original review didn't already scope. Both new tests were verified to fail against the pre-fix code and pass against the fix, which is the strongest signal available that they close the gaps as intended rather than just asserting something trivially true.

Ready for merge pending your review.

PR #1831, opened 2026-07-30
Make the agent eval diagnostic before recording a baseline; revisit ADR-056's Scenario 5 claim

malibio:
Closes #1764


malibio:
### Code Review Summary

This is a well-scoped Phase A implementation: it stays strictly within "make the eval diagnostic," correctly defers baseline recording to #1800, and the design decisions called out as settled (debug-level raw logging, best-effort provenance, `minScenarios=4`) are sound and I did not relitigate them. The uniformity guard / empty-generation exclusion / baseline-comparison interaction is implemented correctly — I traced `compareToBaseline` and confirmed `excludedAsEmptyGeneration` scenarios `continue` before the regression check, and `checkUniformity` aborts via `process.exit` before the results file is written or `compareToBaseline` is ever invoked, so there's no double-counting or spurious-regression path.

However, I built a minimal Rust harness against the actual `tracing_subscriber::fmt()` default formatter this daemon uses (confirmed in `packages/daemon/src/main.rs`) and confirmed two concrete, reproducible bugs in the log-scraping regexes: `%`-formatted (Display) string fields are written **verbatim, including embedded newlines** — they are not escaped the way `{:?}` (Debug) would. Since raw model generations routinely contain newlines, this breaks both scrapers in ways that silently truncate exactly the data this PR exists to preserve. Neither bug is caught by the existing test suite, because `runTurn` (the function containing the regex parsing) isn't exported or tested — only the three downstream pure functions are.

Everything else — the tracing field additions, the `guidance` provenance plumbing (verified against the real `node query --json` output shape and `_seed` storage location), and the new pure-function test coverage — is solid.

### Findings

#### Critical Issues

- **`scripts/aichat.ts:143` — multiline `raw_response` is silently truncated at the first embedded newline, with no recovery.** `reportTurnLog` does `slice.split("\n")` before scanning for `"Agent loop: raw generation"`, then extracts everything from `raw_response=` to the end of that one line (`scripts/aichat.ts:190-197`). I verified against real `tracing_subscriber::fmt()` output that a `%response_text` field containing a newline (e.g. the model wrote a bulleted list, or code with line breaks — an entirely ordinary generation shape) writes the newline byte-for-byte into the log line. `split("\n")` then cuts the raw response there, and everything after the embedded newline becomes an orphaned log line that is *never* re-attached to the `[raw]` marker and is silently dropped from `aichat.ts`'s stdout. This is the exact failure mode the PR is designed to prevent (a plausible-looking `[raw]` line that quietly isn't the full generation), reintroduced by the new code meant to fix it. Reproduction:
  ```
  raw_response containing: "line one\nline two\nline three"
  → printed: "[raw] iteration=7 line one"   (lines two and three vanish)
  ```
  Fix: read the full daemon-log slice as one string and match the `Agent loop: raw generation` record with a regex that captures through to the next tracing-formatted line (e.g. anchored on the next ISO-8601 timestamp or the next known field-prefix), rather than splitting on `\n` first.

- **`scripts/eval/runner.ts:111` — the raw-output regex's own multiline lookahead breaks if the model's raw text contains the literal substring `[tool]`.** Even setting aside the aichat.ts truncation bug above, `runner.ts`'s regex `/^\[raw\] iteration=(\d+) ([\s\S]*?)(?=\n\[raw\] iteration=\d+ |\n\[tool\]|\n$|$)/gm` uses `\n\[tool\]` as one of its lazy-match terminators. I confirmed with a standalone test that if the model's raw generation narrates a tool call in prose (e.g. `"I'll call [tool] create_node to do this.\nSecond line of the same generation."`), the capture stops at the first `[tool]`-shaped substring inside the model's own text, silently dropping everything after it from that iteration's `rawOutput` — with no error, no indication data was lost. This is a real risk for a local model narrating in ReAct-style prose, which is common. Fix: anchor the terminator on the log's actual line-start markers (`\n\[raw\] iteration=`, `\n\[tool\] `, `\nassistant> `) using a regex that requires the bracket token at true line-start *and* isn't satisfied by the token appearing mid-payload — or better, delimit the raw-output block with a marker that cannot collide with generated text (e.g. a length-prefixed or explicitly-escaped encoding) instead of relying on plain-text sentinels the model's own output can produce.

  Both of the above stem from the same root cause: free-form LLM output is being threaded through a plain-text, line-oriented log format designed for structured field=value pairs, and the delimiter tokens chosen (`\n`, `[tool]`) are not guaranteed absent from that payload. Given this PR's entire thesis is "every false result was invisible in aggregates and only visible in raw output," a raw-output pipe that can silently truncate is the highest-value thing to get right, and it currently has zero coverage exercising a multiline or delimiter-colliding case — `runner.test.ts` tests `buildTraceLines` (downstream of parsing) but never the regex in `runTurn` itself.

#### Suggested Improvements

- **`scripts/eval/runner.ts:73` (`runTurn`) has no test coverage at all.** It's not exported, so none of the regexes that do the actual scraping (`routingDecision`, `stage2CandidatesInjected`, `rawOutput`, `emptyGeneration`, and the pre-existing `toolsCalled`) are exercised by a test — only the three pure functions further downstream (`partitionExcluded`, `checkUniformity`, `buildTraceLines`) are. Given this PR's stated purpose is trustworthiness of the diagnostic data itself, I'd recommend exporting `runTurn`'s regex logic as a pure function taking the aichat.ts stdout string (mirroring how `checkUniformity`/`buildTraceLines` were already extracted for testability) and adding cases for: a multiline raw output, a raw output containing `[tool]`/`[raw]`-shaped substrings, and multiple `[raw] iteration=N` blocks in one turn. This would have caught both issues above before merge.

- **`packages/agent/src/local_agent/agent_loop.rs` — no Rust-side test asserts the new tracing fields' shape.** I don't consider this blocking (asserting on `tracing` output is awkward in Rust and the field-formatting behavior is a library contract, not application logic), but a `tracing_test::traced_test` or a custom `Subscriber` capturing emitted fields for one representative routing path (e.g. asserting `routing_decision="query"` appears in the `two-stage routing overhead` record) would have caught a future accidental swap of `%`/`?` sigil (Display vs Debug formatting) — which is precisely the kind of change that silently breaks the scraper without touching scraper code at all, since it's easy to instinctively "fix" a raw string field by switching to `?` for auto-escaping without realizing the scripts depend on verbatim Display output.

#### Nitpicks

- **Nit:** `scripts/aichat.ts:190` and `scripts/eval/runner.ts:111` — both files independently hardcode the "iteration precedes raw_response" field-order assumption and the `[tool]`/`[raw]` delimiter vocabulary. Since these three log markers (`[routing]`, `[stage2 injected]`, `[raw]`, `[empty-generation]`, plus the pre-existing `[tool]`) now form a de facto wire format between the Rust daemon and two independent TypeScript scrapers, a short comment or shared constant naming the full marker vocabulary in one place (even just a comment block) would reduce the chance of a future marker collision like the one flagged above.
- **Nit:** `scripts/eval/preflight.ts` — `SEEDED_GUIDANCE_TYPES = ["prompt", "skill"]` is manually cross-referenced against `seed_nodes_from_templates` callers via a code comment. This will silently under-report provenance if a third seeded type is added later and this list isn't updated in the same PR — low risk given it degrades to "recorded, visibly incomplete" rather than failing, but worth a `grep`-able marker (e.g. referencing the exact function name, which it already does) to make the cross-reference easy to find when that type list changes.
- **Nit:** `scripts/eval/runner.ts:114` — the trace-line-building regex line exceeds ~110 characters and is dense enough (three alternations in a lookahead) that a short worked example in a comment (as `emptyGeneration`'s matcher above it has) would help the next person modify it safely.

### Verification performed

- Compiled and ran a minimal `tracing_subscriber::fmt()` harness matching `packages/daemon/src/main.rs`'s exact subscriber configuration, confirming `%field` (Display) values are written verbatim (including raw newlines), while quoted string fields (`routing_decision="query"`) and unquoted booleans (`stage2_candidates_injected=false`) format as the existing regexes expect.
- Traced `compareToBaseline` and the `checkUniformity`/`abortOnEnvironment` control flow to confirm no double-counting or false-regression interaction between the uniformity guard, empty-generation exclusion, and baseline comparison.
- Verified `node query --type prompt --json`'s actual output shape (`packages/cli/src/output.rs::node_to_json`/`print_node_list`) matches `extractSeedEntries`'s parsing assumptions (`{nodes: [{properties: {_seed: {...}}}]}`), including the degrade-to-string fallback on unparseable `properties`.
- Ran `bun test scripts/eval/runner.test.ts scripts/eval/preflight.test.ts` — 18 pass, 0 fail.
- Confirmed scope: no baseline-recording logic present; all new work is diagnostic tooling as the issue specifies.

Given the two critical findings are both about silent data loss in exactly the mechanism this PR exists to make trustworthy (raw output capture), and the PR's own motivating principle is "every false result was invisible in aggregates," I'm requesting changes rather than approving as-is. This is very close — the fix is localized to the raw-output extraction path in two files and doesn't require touching the Rust side, the type additions, the provenance work, or the uniformity/exclusion logic, all of which are solid.

malibio (commented):
Review status: REQUEST CHANGES (posted as a comment-type review because GitHub blocks request-changes on one's own PR). See the detailed review comment above for full findings — two critical, reproducible bugs in the raw-output log-scraping path (silent truncation on embedded newlines in aichat.ts's per-line scraper, and a separate delimiter-collision truncation in runner.ts's regex when the model's own text contains the literal substring "[tool]"). Both undermine the exact guarantee this PR exists to provide (that raw output is preserved verbatim for diagnosis) and are not caught by the existing test suite since runTurn's regex logic isn't exported/tested. Uniformity guard, empty-generation exclusion/baseline-comparison interaction, guidance provenance, and PR scope were all verified correct.

malibio:
### Re-Review — PR #1831 (closes #1764)

This is a **re-review**. The [prior review](https://github.com/NodeSpaceAI/nodespace-core/pull/1831#issuecomment-5136903673) (against `b65fab7`) found two critical bugs in the raw-output log-scraping path — silent truncation on embedded newlines in `aichat.ts`, and a delimiter-collision truncation in `runner.ts`'s regex when the model's own text contained the literal substring `[tool]`. Two commits have landed since: a fix commit (`57bf1088`) and a rename/data commit (`38b9ccbb` + `ee29e3a9`). This review focuses on what changed after `b65fab7`, plus a final sanity pass on the full diff.

**Verdict: Approve.** Both prior findings are genuinely closed, the fix is structurally stronger than a patch (moves the invariant to the Rust source instead of patching TS-side regexes), the new anchoring fix is correct with no legitimate-case regression, the rename is functionally complete, and the new tests directly regression-test both original bugs plus the new one found along the way. One trivial stale-comment nit remains.

### Findings

#### Verification of the two original criticals

- **Multiline truncation (closed).** `agent_loop.rs` (`packages/agent/src/local_agent/agent_loop.rs`) now does `raw_response = %serde_json::to_string(&response_text).unwrap_or_default()` — the field is JSON-encoded at the source, so `\n` becomes the two-character escape `\n`, never a real newline. This means the `Agent loop: raw generation` record is *always* exactly one physical line, regardless of what the model generated. `aichat.ts`'s `formatTurnLogLines` still does `slice.split("\n")` first, but that's now safe by construction — there's no real newline inside the JSON payload for `split` to cut on. `JSON.parse(jsonStr)` on the TS side recovers the original text byte-for-byte, embedded newlines included. This is a better fix than patching around the split — it removes the invariant violation instead of working around it. Confirmed with `aichat.test.ts`'s "preserves a multiline raw generation intact" test, which passes.

- **`[tool]`-substring collision (closed).** Same JSON-encoding fix closes this too: since the encoded payload can never contain a literal newline, a model narrating `"I'll use [tool] create_node..."` can never produce a physical line starting with `[tool]` — the substring is buried mid-line, not at line-start. `runner.ts`'s `parseTurnOutput` additionally anchors the tool-call regex to line start (`/^\[tool\] .../gm`), which is defense in depth on top of the JSON-encoding fix. I wrote a standalone repro (`fakeLog` containing a `[raw]` line with embedded `"[tool] create_node"` text followed by a real `[tool] create_node args` line) and confirmed the anchored regex matches exactly one real tool call, not two. Covered by `runner.test.ts`'s "does not truncate raw output containing a literal [tool] marker" test, which additionally asserts the real subsequent tool call still parses.

#### New: tool-call-anchoring fix (found and fixed by the author, not the prior review)

- **Correct, and does not break any legitimate multi-call-per-line scenario.** I checked `aichat.ts` — every `[tool] ...` marker is emitted via a single `out.push(...)` per loop iteration over one daemon log line per tool execution (`scripts/aichat.ts:218`), so a real tool call always starts its own physical line; there is no code path where two genuine tool calls share one line. Anchoring to line-start (`^...$/gm`) is therefore safe with zero legitimate-case regression, and the same anchoring was consistently applied to every other marker in `parseTurnOutput` (`toolsOffered`, `routingDecision`, `stage2CandidatesInjected`, `emptyGeneration`) for the same reason — all of them now share stdout with arbitrary model text via `[raw]` lines. The one deliberately-unanchored pattern, `assistant> ([\s\S]*)$`, is correctly left alone with a comment explaining why (it's aichat.ts's own final-reply marker, printed exactly once as the last line, and a real reply can itself be multiline).

#### `import.meta.main` guard

- **Correct Bun semantics, does not break CLI usage.** `import.meta.main` is `true` only for the module Bun directly executed as the entry point, `false` for a module reached via `import`. `runner.ts` invokes `aichat.ts` via `Bun.spawnSync(["bun", "run", env.aichat, "send", ...])`, which makes `aichat.ts` the entry point of that subprocess — `import.meta.main` is `true` there, so `main()` still runs exactly as before, argv parsing and all. When `aichat.test.ts` does `import { formatTurnLogLines } from "./aichat.ts"`, it's an import, not direct execution, so the guard correctly skips `main()`. Verified by running `bun run test:scripts` — all 64 tests pass, including 9 in `aichat.test.ts` that import the module without triggering CLI argv parsing.

#### Rename completeness (`seed_prompt_nodes` → `seed_agent_guidance_nodes`)

- **Functionally complete** — every call site (`prompt_assembler.rs`, `agent_guidance.rs`, `skill_pipeline.rs`, `daemon/services/assembly.rs`, and all their test modules) was updated. `cargo test -p nodespace-agent --lib` passes 371/371, including the renamed tests. `SEEDED_GUIDANCE_TYPES` in `preflight.ts` now correctly reads `["agent-guidance", "skill"]`, matching what's actually seeded. Grepped the whole tree for `seed_prompt_nodes` and found the function itself fully renamed with only one leftover reference, noted below. The other "prompt" string-literal hits in `markdown_test.rs` and `skill_pipeline.rs:561` are pre-existing, unrelated `child_node_type` test fixtures untouched by this PR — not a rename miss.

- **Nit:** `scripts/eval/preflight.ts:212` — the doc comment above `SEEDED_GUIDANCE_TYPES` still says `` `PromptAssembler::seed_prompt_nodes()` `` (the old name), even though the code below it and the rest of the same comment block correctly use the new name and explain the rename. Harmless since comments don't execute, but this is exactly the category of drift (a name changes, a comment doesn't) that caused the bug this commit fixes — worth a one-line touch-up so this file doesn't become the next stale reference someone has to rediscover by running the eval against a live daemon.

#### Test coverage for the regression set

- **Good, targeted coverage.** `aichat.test.ts` (9 tests) and `runner.test.ts` (new `parseTurnOutput` describe block) both carry a test literally named for each original bug ("preserves a multiline raw generation intact (regression)", "does not truncate raw output containing a literal [tool] marker (regression)"), plus edge cases: multiple iterations without cross-contamination, malformed-JSON payload from an older daemon build (skip, don't emit garbage), and stage2/routing/empty-generation marker extraction. `runner.test.ts`'s version of the `[tool]`-collision test goes further than aichat.ts's, explicitly asserting the real tool call after the colliding raw text still parses — this is the sharpest test in the set since it's the one that would have caught the original bug's actual symptom (lost data, not just wrong output). `bun run test:scripts` passes 64/64. No Rust-side test asserts the tracing field's JSON-encoded shape directly, but this is much lower-risk now than the prior review's flagged gap: correctness no longer depends on a `%`/`?` sigil choice (an easy accidental swap), it depends on `serde_json::to_string` producing valid JSON, a far more robust invariant, and it's exercised indirectly by every TS-side regression test reading realistic log-line fixtures shaped exactly like the real output.

#### `results-e4b.json` / `results-e4b.trace.jsonl`

- No issues. Both files are well-formed (validated as JSON/JSONL), small (24K / 4K), contain no secrets, tokens, or local filesystem paths, and — per the task scope — this is discussed, in-scope Phase B data-recording work, not something to gate on.

### Verification performed

- Read the full diff of both new commits (`57bf1088` fix commit, `38b9ccbb` rename commit, `ee29e3a9` data commit) plus a full-diff sanity pass (`git diff origin/main...HEAD`).
- Wrote and ran a standalone repro confirming the anchored tool-call regex does not double-count a `[tool]`-shaped substring embedded in a JSON-encoded `[raw]` payload.
- Confirmed Bun's `import.meta.main` semantics against how `runner.ts` actually invokes `aichat.ts` (`Bun.spawnSync(["bun", "run", env.aichat, "send", ...])`) — entry-point execution still sets it `true`.
- Grepped the full tree for leftover `seed_prompt_nodes` references; found one (a stale doc comment, noted above) and confirmed all functional call sites were updated.
- Ran `bun run test:scripts` (64/64 pass), `bun run test` (171 files / 4112 tests pass), `cargo build -p nodespace-agent -p nodespace-daemon` (clean), `cargo test -p nodespace-agent --lib` (371/371 pass).
- Validated `results-e4b.json`/`results-e4b.trace.jsonl` as well-formed JSON/JSONL with no sensitive content.

Nothing here rises to blocking. Approving.

malibio (commented):
Review status: APPROVE (posted as a comment-type review because GitHub blocks self-approval). See the detailed re-review comment above for full findings. Both prior critical findings (multiline raw_response truncation in aichat.ts, and the [tool]-substring delimiter-collision truncation in runner.ts) are verified closed by the JSON-encoding fix at the Rust source (agent_loop.rs), which is structurally stronger than a TS-side patch since it removes the invariant violation instead of working around it. The new tool-call-anchoring fix is correct with no legitimate multi-call-per-line regression (verified against how aichat.ts emits [tool] markers). The import.meta.main guard uses correct Bun semantics and does not break CLI usage via Bun.spawnSync. The seed_prompt_nodes -> seed_agent_guidance_nodes rename is functionally complete (one stale doc-comment reference noted as a nit, non-blocking). New regression tests directly target both original bugs plus the anchoring fix. results-e4b.json/.trace.jsonl are well-formed, small, and contain no sensitive data. Full test suite green: bun run test:scripts (64/64), bun run test (4112/4112), cargo test -p nodespace-agent --lib (371/371).

malibio:
### Address-review summary

Ran `/address-review` against both prior reviews. Cross-checked systematically rather than assuming — read the full re-review body, compared its findings against current file content and commit history.

**Review 1** (2 critical bugs: multiline raw-output truncation in `aichat.ts`, `[tool]`-substring delimiter collision in `runner.ts`) — both addressed in commit `57bf1088`, independently re-verified as closed by Review 2 (including a standalone repro the re-reviewer wrote and ran).

**Review 2** (re-review, verdict Approve) — found exactly one open item:

- ✅ **Addressed** — stale doc comment at `scripts/eval/preflight.ts:212` still named the pre-rename function `seed_prompt_nodes()` after `38b9ccbb` renamed it everywhere else. Fixed in commit `100bd1bb` (already pushed before this `/address-review` pass — confirmed by re-reading the current file content, not just trusting the commit message).

No recommendations skipped. Everything raised across both reviews is resolved.

**Verification:**
- `bun run test:scripts` — 64/64 pass
- `cargo test -p nodespace-agent --lib` — 371/371 pass
- Working tree clean, branch up to date with remote (`100bd1bb`)

## Re-Review Decision

**Rationale**: The only outstanding item across both prior reviews was a documentation-only nit (a stale comment, no logic change), already fixed and already covered by the second review's full-diff sanity pass before it approved. No new code paths were introduced by this pass — this is confirming and documenting that state, not making further changes requiring re-verification.

PR #1832, opened 2026-07-30
OpenAI-compat path has no routing-reliability testing; Stage-2 candidate injection can silently suppress tool-calling on served models

malibio:
Closes #1829.

## What this adds

A four-arm routing-reliability matrix for the OpenAI-compatible path (`packages/agent/tests/live_openai_compat_routing.rs`, `--ignored`). The existing smoke test validates transport and an *unrouted* tool-calling turn, so it structurally cannot see a model that loses tool-calling once ADR-038's Stage-2 injects its candidate block.

The arms isolate which part of routing, if any, causes suppression:

| arm | candidate block | isolates |
|---|---|---|
| `baseline` | none | can the model tool-call at all |
| `stage1_only` | none | does Stage 1's extra turn cost it |
| `routed_full` | yes, full | ADR-064 rule 4's predicted mechanism |
| `routed_names` | yes, names only | whether *content* is the variable |

Routed arms use production `render_candidates_for_prompt`, so this measures what ships rather than a lookalike.

## Result

```text
model             baseline    stage1_only routed_full routed_names
mistral:7b        fires       fires       SUPPRESSED  SUPPRESSED
gemma4:e4b        fires       fires       fires       fires
```

- **The locked native model is unaffected on every arm** — the shipped configuration does not have this problem.
- **On `mistral:7b` the block's presence is the variable, not its content.** `routed_names` carries a bare skill name and still suppresses, which ADR-064 rule 4's mechanism (measured on the locked model) does not explain.

**What this does not establish:** how far the failure generalises. One served model exhibits it and none is confirmed clean, so neither "all served models are at risk" nor "this is specific to `mistral:7b`" follows. Both the test docs and ADR-056 say exactly that rather than over-reading two data points.

## Scope decisions

**No runtime mitigation, deliberately.** Scope item 3 asks what NodeSpace *does* on detection. Both candidate designs (runtime auto-degrade; one-time probe per model) encode a policy about how the failure generalises — which this matrix does not yet settle. Deferred to #1830 with a wider matrix as an explicit prerequisite. Agreed on the issue before implementation.

**Suppression is reported, not asserted per-model.** A served model suppressing is a property of that model, not a NodeSpace regression to redden a build on. The assertions cover harness failures only: any errored arm, zero measurable models discovered, or row count ≠ model count.

**Removed-catalog models are excluded from measurement.** ADR-056 deleted the Qwen family, including the Qwen3.5-based `ornith-1-9b-q4km`. A dev box may still serve it; measuring it would produce evidence for a model NodeSpace does not carry. `EXCLUDED_MODEL_FRAGMENTS` prevents that silently recurring. Two pre-existing unit fixtures using `ornith:9b` / `qwen2.5-3b` as sample id strings were also updated — incidental to what they assert, but live references to a removed family.

## Guards against measuring nothing

This workstream has repeatedly been burned by harness failures that read as findings, so two are pinned structurally:

- **`Errored` and `Completed(vec![])` are different types.** `run_arm` previously returned an empty `Vec` on generation error, printing identically to a real suppression — a timed-out model would have entered the matrix as a finding.
- **Two offline unit tests assert the arms actually differ**, and that `routed_names` really carries no instructions. Without them the matrix could compare identical inputs and report a difference that cannot exist.

## Artifact

Each run writes `packages/agent/tests/routing-matrix-latest.txt`, fixed-width so verdicts diff column-wise. Not asserted against a golden — the values depend on which models the box serves, so it is a record, not an expectation.

## Docs

Committed to `nodespace-docs` (ADR-034 user-facing consequence, ADR-038 routing risk, ADR-056 the matrix and its limits).

## Testing

- `bun run test:all` — exit 0; frontend 171 files / 4112 tests, matching baseline
- `bun run quality:fix` — clean, no collateral reformatting
- Live matrix re-run after the exclusion landed, confirming it fires


malibio:

## Requirements Check

| # | Scope item (#1829) | Status |
|---|---|---|
| 1 | Companion test exercising the **routed** path specifically | ✅ Met |
| 2 | Four-arm pattern against available OpenAI-compat models, repeatable | ✅ Met |
| 3 | Decide what NodeSpace does when routing suppresses tool-calling | ✅ Deliberately deferred to #1830 — defensible |
| 4 | Document the finding's scope honestly in user-facing material | ✅ Met |

**1 — Met.** `packages/agent/tests/live_openai_compat_routing.rs` is a genuine companion, not an extension of the smoke test. The routed arms call production `routing::render_candidates_for_prompt` and concatenate with `format!("{SYSTEM_PROMPT}\n\n{block}")`, which is byte-identical in shape to the real injection at `packages/agent/src/local_agent/agent_loop.rs:634`. Verified against the source: same function, same separator, same `Option` handling. This is the thing that most often goes wrong in evidence-producing tests and it was done right.

**2 — Met.** `discover_models` drives the model list, so the matrix widens automatically as the runner's Ollama grows. Two offline unit tests (`the_four_arms_produce_distinct_prompts`, `the_names_only_arm_carries_no_instructions`) run in the default suite and guard the arms from silently collapsing into each other — that second test in particular is the right guard, since `routed_names` is the arm carrying the whole content-independence claim and it would degrade into a duplicate of `routed_full` if the renderer ever emitted prose for empty fields. Confirmed: `cargo test -p nodespace-agent --test live_openai_compat_routing` → 2 passed, 1 ignored.

**3 — Deferred, defensible.** #1830 carries the deferral honestly. It states the actual reason (one failing served model, none confirmed clean, so neither generalization is supported), lists the fork A/B/C with tradeoffs, and makes "widen the matrix" a blocking prerequisite rather than a nice-to-have. This is a decision recorded as a decision, not an omission dressed up as one. Agreed with the issue author on #1829 beforehand.

**4 — Met.** Verified in `nodespace-docs`: ADR-034 → Consequences item 3 (user-facing: "choosing your own model means choosing its routing reliability"), ADR-038 → the Stage-2-injection risk, ADR-056 → the matrix itself plus an explicit "whether this is a native-vs-served split or a narrower per-model property is not yet established." All three state the limits of what two data points support rather than over-reading them. The negative claim is as prominent as the positive one, which is the hard part and is what most evidence write-ups get wrong.

## Code Review Findings

### 🟡 Important

**📁 `packages/agent/tests/live_openai_compat_routing.rs:236-250` — the `Stage1Only` arm diverges from production and is a weak control**

Two separate problems compound here.

*It doesn't mirror production.* The arm sends a hand-written `"Route the user's request."` system prompt with `temperature: Some(0.0)` and `max_tokens: None`. Production Stage 1 (`agent_loop.rs:1550-1558`) sends `STAGE1_SYSTEM_PROMPT` — five specific lines including "Call exactly one tool. Do not answer the user." — with `temperature: Some(0.1)` and `max_tokens: Some(STAGE1_MAX_TOKENS)`. The whole justification for `system_prompt_for` using production rendering is "measures what ships rather than a lookalike"; this arm is exactly the lookalike that reasoning rejects. The routed arms also fire this same divergent Stage-1 call, so the divergence is not confined to `Stage1Only`.

*It's a weak control as designed.* The offline test asserts `baseline == stage1` — the prompts are identical, and the OpenAI-compat engine is stateless per request, so the discarded Stage-1 generation cannot influence the measured turn through any mechanism the test controls. The stated purpose ("does Stage 1's extra turn cost it") is real for the native in-process path where a diverged KV-cache prefix is the cost ADR-038 accepts — but over HTTP there is no shared prefix to diverge. The arm currently buys a wall-clock cost of one extra generation per model per run in exchange for a control that is structurally guaranteed to match `Baseline`.

Two options, either acceptable: (a) point it at `STAGE1_SYSTEM_PROMPT`/`STAGE1_MAX_TOKENS` so it at least reproduces production's Stage 1, and note in the doc comment that on a stateless transport the arm is expected to match `Baseline` and exists to catch server-side statefulness; or (b) drop it and say in the doc comment why three arms suffice on this transport. What should not stand is the current state, where the arm reads as a production mirror and is neither.

*Principle:* a control that cannot vary is not a control — it is cost. And an experiment arm that claims production fidelity must actually have it, or the matrix's authority is borrowed rather than earned.

**📁 `packages/agent/tests/live_openai_compat_routing.rs:379-382` — `cannot_tool_call.len() < models.len()` is weaker than its comment promises**

The comment says "no measured model could tool-call even unrouted — the matrix is uninterpretable." The assertion only fires when *every* model fails baseline. With three models where two cannot tool-call, the run passes with a single interpretable row and no signal that two thirds of the matrix is noise. Given the file's own stated purpose — never let a degraded run read as a clean one — this is the one place that leniency is out of step with the rest.

Suggest printing a prominent warning line per non-tool-calling model in the summary (they are currently only printed inline, mid-loop, where a long run buries them), and reconsidering whether the assertion should be `cannot_tool_call.is_empty()`. A model that cannot tool-call unrouted arguably *is* a config problem on the runner's box, in the same class as the errored arms this file already fails hard on.

*Principle:* an assertion's strength should match the invariant its comment claims to protect; otherwise the comment becomes the thing readers trust and the code becomes the thing that runs.

### 🟢 Suggestions

**📁 `packages/agent/tests/routing-matrix-latest.txt` — committing a machine-specific generated artifact**

Currently tracked (`git ls-files` confirms) and rewritten on every `--ignored` run. Its content depends entirely on which models the runner's Ollama serves, so the file will churn between developers with different local model sets, and a routine re-run produces a diff that looks like a finding changed when only the box did.

I don't think the answer is gitignoring it — the diffability was explicitly requested, and the file's whole reason for existing (per its own doc comment: "run it once, eyeball the printout, forget the result" is how this finding went unnoticed) is sound. But the current name works against that: `-latest` implies a single mutable slot, which is precisely what makes it churn. Consider `routing-matrix-<host-or-date>.txt`, or a single file whose header already records the endpoint and excluded set (it does) plus the run date, so a diff reads as "same box, later run" versus "different box" without guesswork. At minimum, a one-line note in the doc comment that a changed file is expected on a different box and should not be committed unless the *verdict* changed.

*Principle:* a record that produces false-positive diffs trains readers to ignore its diffs, which is the failure the record was created to prevent.

**📁 `packages/agent/tests/live_openai_compat_routing.rs:73` — `EXCLUDED_MODEL_FRAGMENTS` substring matching silently drops models**

The mechanism is sound for its stated purpose and there is no false positive today — I checked the catalog in `packages/agent/src/local_agent/model_manager.rs`; no NodeSpace model id contains `qwen` or `ornith`. Two notes for the future:

- `"qwen"` is an architecture-family substring, and a growing share of fine-tunes carry it in the served name for lineage reasons. If a future NodeSpace catalog model is Qwen-derived, this list silently removes it from measurement with only a `println!`. The exclusion is currently printed but not asserted on, so nobody is forced to notice. Consider asserting that no *excluded* model matches a current catalog entry, which would turn a silent drop into a loud one exactly when it matters.
- Nit: the constant's doc comment is excellent (it explains *why* removed models are worse than no evidence), which is the right instinct — the gap is only that the mechanism has no guard against outliving its justification.

*Principle:* a denylist that fails silently accumulates false positives no one audits; make its violations visible at the moment they'd first mislead.

**📁 `packages/agent/tests/live_openai_compat_routing.rs:474-477` — artifact write failure is swallowed to `println!`**

`if let Err(e) = write_matrix_artifact(...)` prints and continues, then the caller unconditionally prints "matrix written to {path}" — which is false in the failure case. Minor, but the file's own standard is that a failed measurement must never read as a successful one. Either return the `Result` and adjust the message, or fold the error into the success line.

### Verified — no issues

- **CLAUDE.md compliance**: no lint suppression, no `#[allow(dead_code)]`, no `#[deprecated]`. ADRs are referenced by number, which is the required convention; no issue numbers appear in code comments. No `docs/` directory or stray `.md` added to this repo.
- **Fixture edits are genuinely incidental, not a weakening.** `openai_compat_discovery.rs` asserts that every `id` is extracted in order — the ids are opaque strings, and `mistral-nemo:12b` is a real catalog model (`model_manager.rs:121`). `ai-chat-model-selector.test.ts` asserts prefix-splitting on `provider:modelId`; both changed ids are opaque to that logic, and the test's load-bearing case (`openai-compat:abc-123:mistral:7b`, the colon-in-model-name case) is untouched. Both edits remove live references to a family ADR-056 deleted outright, which is the correct cleanup.
- **`ArmResult`'s `Errored` / `Completed(vec![])` split is applied consistently.** `fired()` matches only on `Completed`, `errored()` only on `Errored`, `display()` distinguishes all three states, the artifact writer branches `errored()` before `fired()`, and the per-model loop `continue`s on any errored arm before either the `baseline_fires` or the `lost` computation can run. There is no path where an error is counted as a suppression. The `unreachable!("filtered to errored")` in the error-collection closure is correctly guarded by the preceding `.filter()`. This is the single most important design property in the file and it holds everywhere.
- **The assert/report split is implemented as specified.** Assertions cover harness failures only: any errored arm, zero measurable models, row count ≠ model count. Per-model suppression is printed with a content-independent vs content-dependent verdict and never asserted. The `rows.len() == models.len()` check is a genuine bookkeeping guard, since every `continue` path pushes to `rows` first.

## Recommendation

**APPROVE**

The PR does the hard thing correctly: the routed arms use production rendering at the production injection point, harness failure and model suppression are separated at the *type* level rather than by convention, and the documentation states the limits of two data points as prominently as the finding itself. Deferring the runtime mitigation to #1830 rather than building one on a single failing data point is the right call and is recorded as a decision, not an omission.

The findings above are worth addressing but none blocks merge. The `Stage1Only` divergence from production `STAGE1_SYSTEM_PROMPT` is the one I'd most want fixed — either mirror production or drop the arm — because "measures what ships" is the claim this whole test rests on, and that arm is currently the one place it isn't true.

malibio (commented):
Review posted above. Recommendation: **APPROVE** (formal approval blocked — GitHub does not permit approving your own PR).

Routed arms use production `render_candidates_for_prompt` at the production injection point, and the `Errored`/`Completed` type split correctly prevents harness failures reading as findings. Two follow-ups worth doing, neither blocking: `Stage1Only` diverges from production `STAGE1_SYSTEM_PROMPT` while claiming production fidelity, and the `cannot_tool_call` assertion is weaker than its comment claims.

malibio:
## Review findings addressed — `0beab007`

### 🟡 Stage-1 divergence — fixed

The reviewer was right that this was the one place the test stopped measuring what ships. The arm sent a hand-written `"Route the user's request."` at `temperature: 0.0` with no ceiling; production sends `STAGE1_SYSTEM_PROMPT` at `temperature: 0.1` with `STAGE1_MAX_TOKENS`. The routed arms fired the same divergent call.

Chose to make both constants `pub` and import them rather than copy the values — a copy would drift, which is the same class of problem. **Re-ran the live matrix against the real prompt: verdicts unchanged**, so the finding survives the correction rather than depending on the weaker control.

### 🟡 Baseline assertion weaker than its comment — fixed

`cannot_tool_call.len() < models.len()` fired only when *every* model failed baseline, so two of three failing passed silently. Now `is_empty()`.

### 🟢 Artifact churn — resolved by gitignoring it

Went further than suggested. The repo's `.gitignore` already has a "Testing artifacts (never commit)" section, and ADR-056 explicitly records that `aichat-matrix.ts`'s per-run results files are not committed. Tracking a machine-specific file cuts against both, and a re-run on a differently-provisioned box would read as a changed finding. The file is still generated for local use; diffability lives in the tracking issue, where the run was already posted.

### 🟢 False "matrix written" on write failure — fixed

`write_matrix_artifact` returns `io::Result`; the caller reports the error instead of announcing a successful write.

### 🟢 `EXCLUDED_MODEL_FRAGMENTS` false positives — guarded

`"qwen"` is an architecture-family substring, so a future Qwen-derived catalog entry would have been dropped with only a `println!`. Added a unit test pinning that the locked model and the ids in current use stay measurable, and that ADR-056's removed ids stay excluded (case-insensitively). Didn't wire it to `CATALOG` — that constant is private, and making it `pub` for a test guard is more coupling than the risk warrants.

### Verification

- `cargo test -p nodespace-agent --test live_openai_compat_routing` → 3 passed, 1 ignored
- Live matrix re-run post-change → same verdicts
- `bun run test:all` → exit 0; frontend 171 files / 4112 tests, matching baseline
- `bun run quality:fix` → clean

---

### On the review status

The automated reviewer's `--approve` was rejected by GitHub as self-approval and it posted "Recommendation: APPROVE" as a comment instead. **That should not be read as independent sign-off** — it is analysis of our own code, routed around a two-party control. The PR carries no formal approval, and the findings above were worth acting on regardless. Merge is a human call.

malibio:
## `/address-review` summary

Running the command properly surfaced a gap in my earlier manual pass — the 🟡 Stage-1 finding had **two** halves and I had only fixed one.

### Addressed — 5 of 5 findings

| Finding | Severity | Commit |
|---|---|---|
| `Stage1Only` diverges from production | 🟡 | `0beab007` |
| `Stage1Only` is a weak control / mislabelled | 🟡 | `f598a634` ⬅ **missed on first pass** |
| Baseline assertion weaker than its comment | 🟡 | `0beab007` |
| Artifact churn | 🟢 | `0beab007` (diverged — see below) |
| `EXCLUDED_MODEL_FRAGMENTS` silent drops | 🟢 | `0beab007` |
| False "matrix written" on write failure | 🟢 | `0beab007` |

### What the first pass missed

The docs claimed `stage1_only` isolates *"does Stage 1's extra turn cost it"*. That holds on the native in-process path, where the cost is a diverged KV-cache prefix — the overhead ADR-038 accepts. It does **not** hold over HTTP: requests are independent, so the discarded Stage-1 generation cannot reach the measured turn, and the prompts are identical by assertion. The arm read as a routing measurement while being structurally unable to vary.

Kept rather than dropped (option *a*), with its real job documented: transport statelessness is an assumption about the **server**, not a guarantee. A server carrying state across requests, or evicting a model between them, shows up here as `stage1_only` diverging from `baseline` — and would invalidate the routed arms, which pay the same Stage-1 cost. It is a harness guard, not a measurement of routing.

Confirmed empirically on the re-run: both models matched `baseline` on that arm, as the stateless transport predicts.

### One finding resolved differently than suggested

The artifact suggestion explicitly said *"I don't think the answer is gitignoring it."* I gitignored it anyway. Reasoning: the repo's `.gitignore` already carries a "Testing artifacts (never commit)" section, and ADR-056 records that `aichat-matrix.ts`'s per-run results files are not committed — tracking this one cuts against both. The underlying goal (don't run-once-and-forget) is served by pasting runs into the issue, which is where they now live. Took the reviewer's secondary suggestion too: the header now carries a UTC run date, so a pasted run is placeable in time.

### Nothing skipped

No recommendation was judged overkill or out of scope.

### Verification

- `cargo test -p nodespace-agent --test live_openai_compat_routing` → 3 passed, 1 ignored
- Live matrix re-run → verdicts unchanged from every prior run
- `bun run test:all` → exit 0; 171 files / 4112 frontend tests, matching baseline
- `bun run quality:fix` → clean (`clippy --all-targets -D warnings`)

---

## Re-Review Decision

**Rationale**: Both post-review commits are confined to the test harness and its documentation — no production behavior changed. The only source edit outside the test file is widening two `const`s to `pub` so the test consumes production values instead of copies. The substantive change in `f598a634` is a documentation correction plus a run-date header; the code it describes was already verified by the live re-run, which produced identical verdicts across all three runs (original, post-fidelity-fix, post-doc-fix).

The one thing worth a human eye is the artifact decision, since it diverges from what the reviewer proposed — flagged above rather than buried.

**Note on review status**: the automated reviewer's `--approve` was rejected by GitHub as self-approval and its APPROVE recommendation was posted as a comment instead. That is not independent sign-off. This PR carries no formal approval; merge remains a human call.

PR #1833, opened 2026-07-30
Delete the create_schema description-inference route: an undocumented second meaning for the description parameter, with a 27% silent-wrong-type rate

malibio:
Closes #1800

### DIFF
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -641,11 +641,11 @@ fn def_create_schema() -> ToolDefinition {
                 },
                 "description": {
                     "type": "string",
-                    "description": "Brief description of what this entity type represents"
+                    "description": "Brief prose summary of what this entity type represents. This is NOT parsed into fields — define every field explicitly in \"fields\"."
                 },
                 "fields": {
                     "type": "array",
-                    "description": "Array of field definitions. Only use for scalar properties (text, number, date, enum, boolean). Do NOT use for references to other node types — use relationships instead.",
+                    "description": "REQUIRED. Array of field definitions — every field on this entity type must be listed explicitly here, even if empty ([]). Only use for scalar properties (text, number, date, enum, boolean). Do NOT use for references to other node types — use relationships instead.",
                     "items": {
                         "type": "object",
                         "properties": {
@@ -689,7 +689,7 @@ fn def_create_schema() -> ToolDefinition {
                     }
                 }
             },
-            "required": ["name"]
+            "required": ["name", "fields"]
         }),
     }
 }
--- a/packages/cli/src/commands/schema.rs
+++ b/packages/cli/src/commands/schema.rs
@@ -2,8 +2,8 @@
 //!
 //! `create`/`update` take a single JSON params blob rather than per-field
 //! flags: schema params (fields, relationships, title_template, enum
-//! definitions, additional_constraints) are a nested, evolving shape defined
-//! once in `packages/core/src/schema/mod.rs` (`CreateSchemaParams` /
+//! definitions) are a nested, evolving shape defined once in
+//! `packages/core/src/schema/mod.rs` (`CreateSchemaParams` /
 //! `UpdateSchemaParams`). Mirroring that as a flat CLI flag surface would
 //! duplicate and drift from the Rust structs; JSON keeps the CLI a thin,
 //! schema-agnostic passthrough — the daemon validates and reports errors.
--- a/packages/core/src/schema/mod.rs
+++ b/packages/core/src/schema/mod.rs
@@ -1,12 +1,11 @@
 //! Schema creation and updates.
 //!
-//! Provides the `create_schema` tool for creating custom schemas with fields and relationships.
-//! Supports both explicit field/relationship definitions and natural language descriptions
-//! with intelligent type inference.
+//! Provides the `create_schema` tool for creating custom schemas with explicit
+//! field and relationship definitions.
 
 use crate::behaviors::SchemaNodeBehavior;
 use crate::markdown::MarkdownError;
-use crate::models::schema::{EnumValue, SchemaField, SchemaProtectionLevel};
+use crate::models::schema::SchemaField;
 use crate::models::{Node, NodeUpdate, SchemaNode};
 use crate::services::{CreateNodeParams, NodeService};
 use serde::{Deserialize, Serialize};
@@ -108,6 +107,26 @@ fn has_namespace_prefix(name: &str) -> bool {
     )
 }
 
+/// Warn on explicit field names that shadow a reserved core property.
+///
+/// A name colliding with a reserved core property is reported as a warning
+/// rather than rejected, leaving the choice with the caller — this mirrors how
+/// the description-inference route treated the same collision before it was
+/// deleted (ADR-063). `fields` is otherwise stored verbatim.
+fn reject_reserved_property_names(fields: Vec<SchemaField>) -> (Vec<SchemaField>, Vec<String>) {
+    let mut warnings = Vec::new();
+    for field in &fields {
+        if RESERVED_CORE_PROPERTIES.contains(&field.name.as_str()) {
+            warnings.push(format!(
+                "Field name '{}' matches a reserved core property and may be shadowed by it. \
+                 Consider a more specific name.",
+                field.name
+            ));
+        }
+    }
+    (fields, warnings)
+}
+
 /// Whether `schema_id` names a schema NodeSpace ships, rather than a
 /// user-defined one. A missing schema is reported as non-core; the update path
 /// below surfaces the not-found error with better context.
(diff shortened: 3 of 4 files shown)


malibio:
### Code Review Summary

Clean, atomic deletion that does exactly what issue #1800 asked for and nothing else. `create_schema`'s undocumented description→fields inference route (~450 lines of substring-matching heuristics with a measured 27% silent-mistype rate) is fully removed, `fields` is now required and validated with the correct `None` vs `Some(vec![])` semantics, the reserved-property shadowing check was successfully relocated onto the explicit-fields path, and every deleted symbol's test coverage was deleted or replaced with tests of the new behavior — not weakened. Verified via `rg` across the whole worktree, `cargo check --workspace` (clean, no warnings), and `cargo test -p nodespace-core --lib schema::` (46/46 passing) that there are no stale references and no regressions. This is a strict net positive: it removes a maintenance burden and a footgun, replaces it with a simpler and strictly more predictable contract, and loses no expressive capability (`SchemaField` already natively supports `required` and `user_values`/enum, so `AdditionalConstraints` was pure redundancy).

### Requirements Check (issue #1800 acceptance criteria)

1. **Inference route gone; no `fields` → no silent property-less schema** — ✅ PASS
   `packages/core/src/schema/mod.rs:252-259`: `let Some(explicit_fields) = params.fields else { return Err(...) }`. Confirmed by test `test_create_schema_without_fields_key_is_rejected` (`schema_test.rs:16-43`), which now asserts `result.expect_err(...)` where it previously asserted success. Ran and passing.

2. **Tool-schema `description` text matches exactly** — ✅ PASS
   `packages/agent/src/local_agent/tools.rs:644`: `"Brief prose summary of what this entity type represents. This is NOT parsed into fields — define every field explicitly in \"fields\"."` — byte-for-byte match to the issue's required text.

3. **`RESERVED_CORE_PROPERTIES` warning fires on explicit `fields`, with a non-vacuous test** — ✅ PASS
   New `reject_reserved_property_names` (`mod.rs:113-127`) runs on every `create_schema` call now (`mod.rs:261`). `test_reject_reserved_property_names_warns_on_collision` (`mod.rs` unit tests) asserts the warning fires for `"status"` and explicitly asserts it does **not** fire for `"capacity"` — a real negative case, not a vacuous "warnings is non-empty" check. Integration-level `test_create_schema_reserved_property_warning_reaches_response` (`schema_test.rs:256-317`) additionally confirms the warning survives serialization into the response JSON and that the field is still stored verbatim (not rewritten). This satisfies "a test that would fail without this check present" — deleting the `reject_reserved_property_names` call at `mod.rs:261` would fail both tests.

4. **`AdditionalConstraints` / `additional_constraints` fully removed, CLI docstring updated** — ✅ PASS
   Struct and field removed from `mod.rs`. `packages/cli/src/commands/schema.rs:3-9` docstring no longer lists `additional_constraints` among the nested shape's fields. The only surviving references to the string `additional_constraints` in the worktree are in `schema_test.rs`, in a test (`test_create_schema_rejects_additional_constraints_as_unknown_field`) that positively asserts it is now rejected as an *unknown field* by `#[serde(deny_unknown_fields)]` — correct and intentional.

5. **No remaining references to deleted symbols anywhere in the workspace** — ✅ PASS
   Ran the exact `rg` pattern from the review brief against the worktree root; zero matches for any of `parse_field_descriptions|infer_field_type|AdditionalConstraints|normalize_inferred_fields|contains_any|STRIPPABLE_TYPE_KEYWORDS|INTRINSIC_KEYWORD_PREFIXES|InferredField|extract_field_name|extract_before_paren|extract_before_keyword|strip_trailing_type_keyword|extract_enum_values|is_field_required|apply_constraints` outside of comment references in tests confirming rejection. `normalize_field_name` and `to_title_case` (helper functions only used by the deleted inference code) are also fully removed with no orphaned callers. `cargo check --workspace` is clean.

6. **Tests updated/deleted, not weakened; no `.contains()` reintroduced on field lists** — ✅ PASS
   Reviewed all `.contains()` usages remaining in `schema_test.rs` (33 call sites). All are legitimate: error-message substring checks (`msg.contains("fields")`), or exact-set membership checks on a fully-enumerated, asserted-length field list (e.g. `schema_test.rs:1388`, `names.contains(&"seats") && names.contains(&"address")` where `names.len()` is asserted immediately above). None reintroduce the "fuzzy substring match used to validate inferred output" anti-pattern the issue is eliminating — that pattern doesn't exist anywhere in the surviving code since the inference engine that produced fuzzy output is gone. The deleted tests (`test_parse_field_descriptions_*`, `test_infer_*`, `test_apply_constraints_*`, `test_integration_*` inference tests, the two `test_create_schema_paths_agree_on_stored_field_names*` cross-path-parity tests) are all tests *of the deleted code* — correctly removed, not "weakened." New/renamed replacements (`test_create_schema_with_explicit_empty_fields_succeeds`, `test_create_schema_without_fields_key_is_rejected`) assert exact field-list length and exact stored names, which is strictly tighter than what they replaced.

7. **`description` parameter itself preserved (still stored via `create_description_subtree`)** — ✅ PASS
   `mod.rs:279-282` computes `description_text` from `params.description` (defaulting to `"Schema for {name}"` if absent — pre-existing behavior, unchanged by this PR) and `mod.rs:319` still calls `create_description_subtree(node_service, &created_schema_id, &description_text)`. `CreateSchemaOutput.description` also still round-trips it. Confirmed by passing test `test_create_schema_stores_description_as_child_subtree`. Only the *parsing-into-fields* behavior was deleted, exactly as scoped.

### One thing to double-check very closely: None vs Some(empty)

Verified directly. `CreateSchemaParams.fields` is typed `#[serde(default)] pub fields: Option<Vec<SchemaField>>` (`mod.rs:167-168`) — `#[serde(default)]` on an `Option<T>` field resolves the *absent-key* case to `Option::default()` = `None`, and a present `"fields": []` deserializes to `Some(vec![])`. No `unwrap_or_default()` or other collapsing operation is applied before the presence check — the handler does `let Some(explicit_fields) = params.fields else { ... }` (`mod.rs:252`), which correctly treats `None` as an error and `Some(vec![])` as a valid zero-length `Vec<SchemaField>` that flows straight into `reject_reserved_property_names`.

Both required tests exist and pass:
- `test_create_schema_with_explicit_empty_fields_succeeds` (`schema_test.rs:60-118`) — sends `"fields": []` explicitly, asserts success and `val["fields"].len() == 0`, and additionally re-fetches the persisted schema from storage to confirm it's actually durable, not just returned.
- `test_create_schema_without_fields_key_is_rejected` (`schema_test.rs:16-43`) — sends a payload with no `fields` key at all, asserts `result.expect_err(...)` and that the error message contains `"fields"`.

The error message itself (`mod.rs:253-258`) names what's missing and gives a concrete example: `"\"fields\" is required. List every field explicitly, e.g. [{\"name\":\"amount\",\"type\":\"number\"}]. \"description\" is a prose summary only and is not parsed into fields."` — satisfies "actionable error naming what's missing" (criterion #2 sub-requirement) well beyond the bar.

### Code Review Findings

#### Critical Issues
None found.

#### Suggested Improvements
- **`packages/agent/src/local_agent/tools.rs:692`**: The JSON tool-schema now sets `"required": ["name", "fields"]`. This is correct per JSON Schema semantics — `required` only enforces key presence, not non-emptiness, so `fields: []` still validates — but it's worth a moment's scrutiny given the None-vs-Some(empty) distinction is the crux of this PR. Confirmed no mismatch: the Rust-side validation error and the JSON-schema-side `required` constraint agree. No action needed, just flagging that this was checked since it's an easy place for a schema/implementation drift to hide.
- **`packages/core/src/schema/mod.rs:113-127`**: `reject_reserved_property_names` is a misleading name — it doesn't reject anything; it warns and passes every field through unchanged (by design, per ADR-063: the caller keeps the final say). A name like `warn_reserved_property_names` or `collect_reserved_property_warnings` would match its actual behavior and avoid a future reader assuming it filters or errors. Minor, but the current name invites exactly the kind of "what does this actually do" bug-hunt this PR is designed to eliminate elsewhere.

#### Nitpicks
- **Nit**: `packages/core/src/schema/mod.rs:279-282` — the fallback description `format!("Schema for {}", params.name)` when `description` is omitted is unrelated pre-existing behavior, unchanged here, but since the PR just made `description`'s contract stricter (prose-only, never parsed), it might be worth a follow-up (out of scope for this PR) to consider whether a generic fallback string is still the right default for the embedding/discovery use case it serves, versus surfacing to the caller that they should supply one. Not a blocker — just noting it while in the neighborhood.
- **Nit**: `mod.rs:249-251` comment above the `let Some(explicit_fields) = params.fields else` block is a good "why," but could make the None/Some(empty) distinction even more explicit for future readers by literally naming the two branches, e.g. "`fields` absent from the request (`None`) is always an error; `fields: []` (`Some(vec![])`) is always valid." Current wording implies but doesn't spell out the type-level distinction as sharply as the PR's own commit message and this review needed to.

### Recommendation: APPROVE

This is a textbook example of the kind of change CLAUDE.md's "no backward compatibility, delete old patterns immediately" philosophy is written for: a clearly-scoped, fully-verified deletion of an undocumented, measurably-defective code path, with the replacement behavior's trickiest edge case (`None` vs `Some(empty)`) explicitly tested in both directions. All 7 acceptance criteria are met with direct evidence, the workspace compiles clean with zero stale references, and the only findings are a naming nit and a follow-up-worthy (not blocking) observation. Net positive with no reservations.

malibio (commented):
Review complete — Recommendation: APPROVE (self-approval blocked by GitHub since PR author == reviewing account; see full report above for verified evidence on all 7 acceptance criteria, 46/46 tests passing, clean workspace build).

malibio:
### Addressed review feedback

- ✅ Renamed `reject_reserved_property_names` → `warn_reserved_property_names` — the function warns and passes fields through unchanged, it never rejects (ADR-063 leaves the final call with the caller). Matches actual behavior now.
- ✅ Sharpened the comment above the `fields` None-vs-Some(empty) check to spell out the type-level distinction explicitly.
- ⏭️ Skipped: the description-fallback-string follow-up — reviewer explicitly scoped it out as "not a blocker... out of scope for this PR."

Both changes are non-functional (rename + comment only, no logic change). Verified: `cargo test -p nodespace-core --lib schema::` (46/46 passing), `bun run quality:fix` clean, pre-push gate (test:all + cargo build + e2e) passed.

## Re-Review Decision

**Rationale**: Both changes are a rename and a comment clarification — no logic, control flow, or test behavior changed. All 46 schema tests still pass with identical assertions.

Ready for merge.

PR #1843, opened 2026-07-31
Fixture: agent-matrix scenarios 3->4->6 key on a value the chain never persists; trace doesn't record offered tools/routed skill

malibio:
Closes #1839

### DIFF
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -1235,10 +1235,23 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
                 // previews truncate at 300 chars — a realistic create_schema payload
                 // exceeds that, so a parser reading them would fail on exactly the
                 // well-formed calls it is meant to pass.
+                //
+                // Two shapes report it: `fields` (create_schema — the array it
+                // persisted) and `property_count` (create_node — a bare count,
+                // since its result carries only the new node's id). Both answer
+                // the same question, "did this call persist anything the user
+                // can actually record against", so both feed one field rather
+                // than two the scraper would have to reconcile.
                 let result_field_count = result_value
                     .get("fields")
                     .and_then(|f| f.as_array())
-                    .map(|a| a.len());
+                    .map(|a| a.len())
+                    .or_else(|| {
+                        result_value
+                            .get("property_count")
+                            .and_then(|c| c.as_u64())
+                            .map(|c| c as usize)
+                    });
 
                 tracing::info!(
                     tool = %tc.function_name,
@@ -1714,10 +1727,25 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
         // earlier and logs its own "stage-1 routing decision" line above. Both
         // carry the same field name so a log scraper (an eval, a dashboard) can
         // grep one key regardless of which path a turn took.
+        // `routed_skills` names the candidates that clear the score gate — the
+        // ones `render_candidates_for_prompt` actually writes into Stage 2's
+        // prompt and `stage2_tools` scopes the tool surface from. The count
+        // alone cannot answer "which skill was routed to", so a routing claim
+        // about any eval scenario required reading source rather than results;
+        // an empty value here means retrieval returned nothing above the bar,
+        // which is exactly the "routed but matched nothing" case
+        // `stage2_candidates_injected` distinguishes on the assembly line.
         tracing::info!(
             routing_decision = routing_decision_tag,
             routing_latency_ms = elapsed_ms,
             candidates = outcome.candidates.len(),
+            routed_skills = %outcome
+                .candidates
+                .iter()
+                .filter(|c| routing::clears_score_gate(c))
+                .map(|c| c.name.as_str())
+                .collect::<Vec<_>>()
+                .join(", "),
             "two-stage routing overhead"
         );
         outcome
--- a/packages/agent/src/local_agent/agent_loop.rs
+++ b/packages/agent/src/local_agent/agent_loop.rs
@@ -1739,13 +1739,7 @@ impl<E: ChatInferenceEngine + ?Sized, T: AgentToolExecutor + ?Sized> LocalAgentL
             routing_decision = routing_decision_tag,
             routing_latency_ms = elapsed_ms,
             candidates = outcome.candidates.len(),
-            routed_skills = %outcome
-                .candidates
-                .iter()
-                .filter(|c| routing::clears_score_gate(c))
-                .map(|c| c.name.as_str())
-                .collect::<Vec<_>>()
-                .join(", "),
+            routed_skills = %routing::routed_skill_names(&outcome.candidates),
             "two-stage routing overhead"
         );
         outcome
--- a/packages/agent/src/local_agent/routing.rs
+++ b/packages/agent/src/local_agent/routing.rs
@@ -231,6 +231,29 @@ pub fn clears_score_gate(candidate: &SkillCandidate) -> bool {
     candidate.score >= score_bar_for(candidate)
 }
 
+/// Names of the candidates that clear the score gate, comma-separated, for the
+/// `routed_skills` log field.
+///
+/// The same set `render_candidates_for_prompt` writes into Stage 2's prompt and
+/// `stage2_tools` scopes the tool surface from, so a log reader can tell which
+/// skill a turn was actually routed to rather than only how many candidates
+/// retrieval returned.
+///
+/// A function rather than an inline expression at the log site specifically so
+/// the rendered text is testable. The scraper consuming this field parses to
+/// end of line — skill names contain spaces and commas, and `tracing` emits the
+/// value unquoted — so its exact shape is a contract, and the first version of
+/// that scraper silently matched nothing because the shape was assumed rather
+/// than asserted.
+pub fn routed_skill_names(candidates: &[SkillCandidate]) -> String {
+    candidates
+        .iter()
+        .filter(|c| clears_score_gate(c))
+        .map(|c| c.name.as_str())
+        .collect::<Vec<_>>()
+        .join(", ")
+}
+
 /// Render retrieved candidates for injection into the Stage-2 prompt.
 ///
 /// Delivered **in the prompt** rather than as a tool result. ADR-064 rule 4
@@ -479,6 +502,46 @@ mod tests {
         let rendered = render_candidates_for_prompt(&cands).expect("one candidate is eligible");
         assert!(rendered.contains("strong"));
         assert!(!rendered.contains("weak instructions"));
+    }
+
+    #[test]
+    fn routed_skill_names_lists_only_gate_clearing_candidates() {
+        let cands = vec![
+            candidate("Research & Search", 0.9, &["search_nodes"]),
+            candidate("Below The Bar", 0.01, &["search_nodes"]),
+            candidate("Node Creation", 0.9, &["search_nodes"]),
+        ];
+        assert_eq!(
+            routed_skill_names(&cands),
+            "Research & Search, Node Creation"
+        );
+    }
+
+    #[test]
+    fn routed_skill_names_is_empty_when_nothing_clears_the_gate() {
+        // Distinct from "retrieval returned nothing" only in the candidate
+        // count logged alongside it; the scraper omits the marker either way.
+        let cands = vec![candidate("weak", 0.001, &["create_schema"])];
+        assert_eq!(routed_skill_names(&cands), "");
+    }
+
+    #[test]
+    fn routed_skill_names_matches_what_is_rendered_into_the_prompt() {
+        // The field exists to answer "which skill was this turn routed to", so
+        // it has to name the same set the model actually sees. Asserting the
+        // two agree keeps the log honest if either filter later changes.
+        let cands = vec![
+            candidate("Graph Editing", 0.9, &["search_nodes"]),
+            candidate("Excluded", 0.001, &["search_nodes"]),
+        ];
+        let rendered = render_candidates_for_prompt(&cands).expect("one is eligible");
+        for name in routed_skill_names(&cands).split(", ") {
+            assert!(
+                rendered.contains(name),
+                "{name} logged as routed but absent from the prompt block"
+            );
+        }
+        assert!(!rendered.contains("Excluded"));
     }
 
     #[test]
--- a/packages/agent/src/local_agent/tools.rs
+++ b/packages/agent/src/local_agent/tools.rs
@@ -1718,6 +1718,15 @@ impl GraphToolExecutor {
         props.extend(flat_extras);
         let properties = Value::Object(props);
 
+        // Number of schema field values this call actually carries, counted
+        // AFTER the flat-extras merge above so a model that passes fields at the
+        // top level rather than nested under "properties" is not miscounted as
+        // having passed none. Reported on the result so an eval can tell a
+        // create_node that recorded the user's particulars apart from one that
+        // persisted a bare shell — indistinguishable by tool name alone, which
+        // is exactly the hole `fields` already closes for create_schema.
+        let property_count = properties.as_object().map(|o| o.len()).unwrap_or(0);
+
         let ns = self.node_service()?;
 
         // node_service.compute_title() handles all title derivation:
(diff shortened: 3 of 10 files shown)


malibio:
## Pragmatic Code Review — PR #1843

### Summary

A genuinely net-positive change that does the harder version of the job. The core insight — that the eval was scoring green on assertions structurally incapable of seeing what they claimed to check — is correct, and the fix pursues it to the honest conclusion (9/12 → 8/12) rather than preserving the headline number. The commit messages and code comments are unusually good: they explain *why* the shape is what it is, and commit `7a449cb4` documents its own predecessor's silent-empty bug rather than quietly amending it.

I verified the load-bearing claims rather than accepting them:

- **Confirmed by experiment.** I built a standalone `tracing-subscriber` 0.3 reproduction to check the quoting claim. `record_str` → `record_debug` → `write!("{:?}")`, and the `%` sigil wraps the value in `format_args!`, whose `Debug` impl is transparent. Both fields emit **unquoted**:
  ```
  ... tool_names=create_node, search_nodes system_prompt_len=1234 stage2_candidates_injected=true
  ... candidates=3 routed_skills=Organization, Research & Search, Node Creation
  ```
  So the diagnosis in `7a449cb4` is exactly right — and it has an unremarked corollary (finding 2 below).
- **Confirmed the dead scrape.** `selected_tools` / `scoped tool list` appear nowhere in the Rust tree. `scripts/results-e4b.trace.jsonl` (committed at `905c5fc3`) has `toolsOffered: ""` on every turn, which matches.
- **Confirmed `clears_score_gate` is the right predicate.** `render_candidates_for_prompt` (routing.rs:247) and `stage2_tools` (routing.rs:335) both filter on it, so `routed_skills` names precisely the set that reached Stage 2's prompt and scoped its tool surface. This is the semantically correct choice over `candidates` or top-K.
- **Confirmed the `extra` shape change is safe.** `compareToBaseline` joins on `id` and reads only `passed`/`failure`; `extra` is serialized to JSON and never rendered by a consumer that would break on array-vs-string.
- `bun run test:scripts` → 74 pass / 0 fail. `cargo clippy -p nodespace-agent --all-targets -- -D warnings` → clean.

**Standards check**: no lint suppression added, no `#[allow(dead_code)]`, no `#[deprecated]`, no new `.md` files, no issue numbers in code comments or doc content (the `#1841` reference lives in the commit message, which is correct — that's GitHub process mechanics, not documentation content). The `console.log` calls in `scripts/` are pre-existing CLI stdout in tooling, not production app code, and are outside this diff.

Two findings are worth addressing before merge; neither is a security or correctness risk to the product, both are correctness risks to *the measurement instrument*, which is this PR's entire subject matter.

---

### Requirements Check

| AC | Status | Notes |
|---|---|---|
| **1.** Fix the 3→4→6 chain so scenario 6 keys on a persisted value; verify the gap isn't merely relocated | ✅ | Scenario 3's prompt now implies replacement cost; author verified against the live DB that the schema persists `replacement_cost (number)` and `is_returned (boolean)` — not just that the tool was called. Crucially, the "don't relocate the gap" clause was *actually honored*: re-running exposed that the gap **had** moved to `create_node` persisting a bare shell, and rather than declaring victory, the PR added `minProperties` to catch it. Keeping cost (not the item name) as scenario 6's discriminator is well-reasoned — an item-name key would collapse to a direct string match that plain `search_nodes` resolves, making the `resolve_query` assertion pass while testing strictly less. |
| **2.** Record offered tool list and routed skill per turn in the trace | ✅ | Dead `selected_tools` scrape deleted; `[tools offered]` repointed at the real line (post-scoping `stage2_tools` output — the surface the model actually saw, which is the right semantics). `routed_skills` added to the routing log. Both verified populated against a real daemon run. See finding 3 on "per turn." |
| **3.** Re-run affected scenarios to confirm the chain resolves end to end | ✅ | Clean full run on purged DB + freshly rebuilt daemon (correctly following the known "evals silently measure stale guidance" hazard): 8/12, matching the prediction exactly, with scenario 4's failure message quoted. The run also *paid off* immediately — it answered the issue's own availability question by showing `resolve_query` **was** offered on scenario 6, so its failure is not availability. That's the observability improvement earning its keep on the first use. One honest caveat, self-disclosed: the committed trace's `routedSkills` predates `7a449cb4` (see finding 4). |

---

### Code Review Findings

#### 🟡 Important

**1. `property_count` counts what was *requested*, not what was *persisted* — and the authoritative answer is already in hand**

📁 `packages/agent/src/local_agent/tools.rs:1721-1728`

```rust
let property_count = properties.as_object().map(|o| o.len()).unwrap_or(0);
```

This is computed from the merged request map, **before** `node_ops::create_node` runs. But persistence is not pass-through. `NodeService::create_node` (`packages/core/src/services/node_service/crud.rs:49+`) can transform the map en route to storage: `normalize_flat_properties_to_namespace` reshapes flat keys into a type namespace, `apply_schema_defaults_with_fields` **adds** entries the model never supplied, and `_schema_version` is injected. Validation can also reject values.

The consequence runs in both directions, and both are bad for an instrument whose purpose is to distinguish "recorded the particulars" from "persisted a bare shell":
- **False pass**: schema defaults or `_schema_version` inflate the count so a call that supplied none of the user's particulars still clears `minProperties: 1`.
- **False fail**: a namespaced reshape collapses the count below the request-side figure.

The fix is nearly free, because `create_node` **already re-fetches the created node** — `node_ops.rs:194-201` does a `get_node` and puts the result in `output.node_data`. Counting from that post-persistence state, rather than from the pre-flight request map, makes the field mean what its comment says it means ("what this call actually carries" → what storage actually holds).

This matters more than a normal off-by-semantics nit because the PR's own framing is *"the result is the executor's report of what it persisted, while args are only the model's report of what it asked for"* (`agent_loop.rs:1232-1234`). `property_count` as written is on the wrong side of the distinction the surrounding code just drew. Right now scenario 4 reports 0 and the conclusion holds either way — but the field is being added specifically so future runs can trust it, and the first run where a schema default fires is the run it misleads.

*Principle*: an observation must be taken at the boundary it claims to describe. Measuring the input and labeling it the output is the same class of error as the dead `selected_tools` scrape this PR exists to remove.

---

**2. The `tool_names` fixtures encode a quoted shape the daemon never emits — the identical latent defect `7a449cb4` just fixed for `routed_skills`**

📁 `scripts/aichat.test.ts:81`, and every `tool_names="..."` fixture in `aichat.test.ts` / `runner.test.ts`

My reproduction (above) shows real output is `tool_names=create_node, search_nodes system_prompt_len=1234` — **unquoted**, because it uses the same `%` sigil as `routed_skills`. Every fixture in these tests writes `tool_names="create_node, search_nodes"`.

The scrape happens to survive, and I verified this explicitly:

```
/tool_names="?([^"]*?)"? system_prompt_len/
  REAL    -> "create_node, search_nodes"   ✅
  FIXTURE -> "create_node, search_nodes"   ✅
```

The optional-quote pattern tolerates both, so **there is no bug today**. But the fixtures are asserting against a shape reality does not produce, which is *precisely* the condition the PR's own commit message identifies as the root cause of the `routed_skills` miss: *"The unit tests passed throughout because their fixtures used the quoted form — they asserted the scrape against a shape the daemon never produces."*

The PR fixed the instance and left the class. The trailing `system_prompt_len` anchor is currently the only thing preventing the same silent-empty failure, so any future reordering of that log line's fields — the exact fragility the author correctly flags for `routed_skills` — would break `[tools offered]` again while the tests stay green.

Suggested fix: update these fixtures to the verbatim unquoted daemon form, exactly as `routed_skills`'s fixtures now are. It's a small edit that closes the class rather than the instance, and it's fully in the spirit of what this PR already did.

*Principle*: a test fixture is a claim about the real world. When it diverges, the test stops measuring the system and starts measuring itself — the specific failure mode this entire PR is about.

---

#### 🟢 Suggestions

**3. `extra()`'s per-turn arrays are inert as wired — the runner always passes exactly one turn**

📁 `scripts/eval/fixtures/agent-matrix.ts:381-386`, `scripts/eval/runner.ts:533,551`

The change from `turns[0]?.toolsOffered` to `turns.map(...)` is justified in-comment by *"a scenario's turns can be offered different tool surfaces."* True in principle, but both call sites pass a single-element literal:

```ts
extra: fixture.extra?.(scenario, [scored]),
```

`priorTurns` are deliberately excluded (they're never scored). So these arrays are always length 1, and the shape change buys nothing today while making the results file marginally harder to read.

I checked this doesn't break anything: `compareToBaseline` joins on `id` and reads only `passed`/`failure`, and no renderer consumes `extra`. So it's harmless — just not yet load-bearing.

Fine either way, but worth being explicit in the comment that this is *anticipatory* rather than describing current behavior, or the next reader will look for the multi-turn path and not find it. (Alternatively — this is arguably YAGNI, and `turns[0]` was honest about what the runner does.)

---

**4. `routed_skills` has no Rust test coverage, and neither does `property_count`**

📁 `packages/agent/src/local_agent/agent_loop.rs:1742-1749`, `packages/agent/src/local_agent/tools.rs:1728,1755`

The TypeScript side of both features is now well covered (6 new tests, including the excellent "spaces and commas" and "absent vs empty" cases). The Rust side — the *producer* of the log field and the result field — has none.

This is what let the quoted/unquoted mismatch reach a full eval run: the only tests were downstream of the format, so nothing asserted the emitted string. A single test asserting `exec_create_node`'s result contains `property_count`, and one asserting the join format of `routed_skills`, would close the loop that just cost a re-run. Given `#1841` will change create_node's property behavior, a test pinning the result contract is cheap insurance.

*Principle*: test at the boundary you own. A contract asserted only by its consumer's fixtures is not asserted at all.

---

**5. `result_field_count` conflating two tools is sound — but the name no longer fits**

📁 `packages/agent/src/local_agent/agent_loop.rs:1236-1256`

I initially expected this to be a problem and concluded it isn't. The two shapes are disjoint (`create_schema` returns `fields`, `create_node` returns `property_count`; neither returns both), `fieldCount` is only ever interpreted per-tool downstream (`schemaCallsAreSound` filters `c.name !== "create_schema"`, `callPersistedProperties` filters on its `tool` argument), and the shared field spares the scraper a reconciliation. The justifying comment is good.

The residual risk is naming: "field count" now means *schema fields defined* for one tool and *property values supplied* for another. A third tool returning a `fields` array with different semantics would silently join the pile. Nit-level, but `persisted_item_count` or similar would carry the generalization the field has actually taken on.

---

**6. `types.ts` "recorded before this field existed" comments — legitimate, not a backward-compat violation**

📁 `scripts/eval/types.ts:46-53`, `scripts/eval/fixtures/agent-matrix.ts:130-133`

Flagging explicitly since the no-backward-compatibility rule is strict here: **these are fine, and should stay.**

The rule targets *code* paths — dual-mode support, deprecation shims, migration layers. These comments describe **data**: results and baseline JSON files that are immutable historical records of inference runs, some of which cost minutes of GPU time and cannot be regenerated identically (different model state, different DB). Treating a missing `fieldCount` as `0` would retroactively convert archived passes into failures for runs where the property was never observed — corrupting the baseline comparison that is the eval's whole point.

`callPersistedProperties`'s `if (c.fieldCount === undefined) continue;` correctly mirrors the established `schemaCallsAreSound` precedent, and the runner's `routedSkills?: string` (undefined, not `""`) preserves the "not recorded" vs "nothing routed" distinction — with a test for exactly that. This is sound absence-vs-zero modeling, not a compat shim.

---

**7. `property_count` reaching the model as a tool result — acceptable, with precedent**

📁 `packages/agent/src/local_agent/tools.rs:1755`

I checked this against ADR-064 rule 4 (tool results carry *facts*, not procedures) and against sibling tools. `property_count` is a fact about the completed outcome, and the codebase already returns confirmation metadata in exactly this position: `update_node` → `"updated": true` (tools.rs:1822), `delete_node` → `"deleted"` (2088), `set_status` → `"status"`/`"updated"` (2059). So this is consistent with established practice, not a new pattern, and the token cost is a handful.

Worth a light watch rather than a change: `create_node`'s result was previously `{id}` alone, and this PR's own `#1841` finding is that create_node's property behavior is prose-sensitive. If a future run shows the model reacting to seeing `property_count: 0` (e.g. retrying, or narrating about it), that's the thing to look at. Not a reason to hold the PR.

---

### Recommendation

**COMMENT** — approve-worthy in substance; two items worth a quick pass first.

This is careful work on the instrument the rest of the agent workstream depends on, and it holds itself to a standard most eval fixes don't: it re-ran, discovered its own fix had only moved the gap, and then went and closed the new one instead of shipping. The self-reported `7a449cb4` regression and the disclosed stale-trace caveat are both signs of a review you can trust.

Before merge I'd suggest:
- **Finding 1** (`property_count` measured pre-persistence, when `output.node_data` already holds the post-persistence truth) — this is the one with real legs, since the field is being added expressly so future runs can rely on it.
- **Finding 2** (`tool_names` fixtures still encode the never-emitted quoted form) — small edit, and it closes the class of defect this PR just fixed one instance of.

Findings 3–7 are optional. Nothing here is a security, data-integrity, or product-correctness risk, and the 8/12 conclusion stands regardless.

*Reviewed by an AI reviewer — final merge call is the author's.*

malibio (commented):
Initial review posted above. Recommendation: COMMENT — net positive and approve-worthy in substance; two items worth a quick pass before merge (property_count measured pre-persistence; tool_names fixtures still encode a quoted shape the daemon never emits). Final merge call is the author's.

malibio:
## Review feedback addressed — commit `3e0838e2`

**Addressed 4 of 7 findings; skipped 3 (2 required no action, 1 declined with reasoning).**

### 🟡 Finding 1 — `property_count` counted requests, not persistence ✅

You were right, and this was the one real correctness bug. The count came from the merged argument map *before* `node_ops::create_node` ran, so it answered "what did the model ask for" while the field name and its own comment claimed persistence. Since creation applies schema defaults and normalizes keys into the type namespace, the two disagree in both directions — and the inflating direction is the dangerous one: a default could push the count past `minProperties: 1` and produce exactly the false pass the assertion exists to prevent.

Now reads `output.node_data`, which as you noted is already re-fetched at `node_ops.rs:194-201`. A detail that made this cleaner than expected: `flatten_properties_for_api` (`convert.rs:69-95`) already unwraps the type namespace *and* filters underscore-prefixed keys, so `_schema_version` is excluded for free and the count is exactly the user-meaningful values a later turn could resolve against.

Verified live rather than by inspection — a create carrying `label` + `status`:

```
result_preview={"id":"nodespace://55aaa806-...","property_count":2}
node get   →   properties: {"gadget":{"_schema_version":1,"label":"Widget A","status":"shipped"}}
```

Two persisted, two counted, internals excluded. The zero direction was already demonstrated by the eval run that produced `result_field_count=0` on scenario 4.

Your framing was the useful part: the existing comment at `agent_loop.rs:1232-1234` already drew this exact distinction, and the new field had landed on the wrong side of it.

### 🟡 Finding 2 — fixtures encoding a shape the daemon never emits ✅

Fixed, and it was broader than `tool_names`: `tool="create_node"`, `args_preview="{}"` and `result_preview="{}"` were quoted too. Confirmed all of them bare off a running daemon rather than reasoning about it:

```
Tool executed tool=create_schema is_error=true duration_ms=0 args_preview={} result_preview={...}
tool_names=search_nodes, resolve_query, search_semantic, get_node, create_node, ...
```

Replaced the hand-written forms with captured ones and added a file-level note in `aichat.test.ts` to copy marker lines from a real log rather than write them by hand. Your "fixed the instance, left the class" characterization was accurate.

(Incidentally, confirming your experiment from the other direction: `routing_decision="clarify"` **is** quoted, because it's a single word. That's what makes this failure mode so quiet — the quoted form looks right next to a real quoted field.)

### 🟢 Finding 4 — no Rust coverage for the new producers ✅

Extracted the rendering as `routing::routed_skill_names` — a function specifically so the emitted text is assertable — and added three tests, including one that asserts the logged set matches what `render_candidates_for_prompt` actually writes into the prompt, so the two can't silently diverge. Agent crate: 371 → 374 tests.

### 🟢 Finding 3 — per-turn arrays are inert ✅

Comment now marks them anticipatory, noting the collapse would be silently wrong the moment a scenario scores more than one turn.

### ⏭️ Skipped

- **Finding 5** (rename `result_field_count`) — you confirmed the conflation is sound (disjoint shapes, per-tool interpretation), so this is a naming change rippling into an unrelated log field and its scraper for no behavioral gain. Worth revisiting if a third tool ever feeds it.
- **Findings 6 and 7** — validated as correct in review; no action requested.

### Verification

`bun run test:all` exits 0 — frontend 4112 passed (171 files, = baseline), scripts 74 passed, skill 20 passed, Rust green (agent 374). `cargo clippy -p nodespace-agent --all-targets -- -D warnings` clean. `quality:fix` touched only files in this change. Full pre-push gate passed.

### Re-review

**Not needed, in my view** — finding 1 is a genuine behavior change but narrow (one expression, verified live in both directions), and the rest are tests, fixtures, and comments. Happy to re-run if you'd prefer the persisted-count change get a second look, since it's the one thing here that alters what the eval measures.

